PR #8, opened 2017-11-17
 Generate a different controller when Twig is not installed

### LINKED ISSUE #6: Return a simple Response from make:controller is twig is not installed
We use `$this->render()` in `make:controller` to render a pretty page :). But we cannot do that if TwigBundle is not installed. In that case, we should generate a simple `new Response()`.

javiereguiluz:
It fixes #6 and replaces #7.

### DIFF
--- a/src/Resources/skeleton/controller/Controller.php.txt
+++ b/src/Resources/skeleton/controller/Controller.php.txt
@@ -13,6 +13,6 @@ class {{ controller_class_name }} extends Controller
      */
     public function index()
     {
-        return new Response('Hello World');
+        return new Response('Welcome to your new controller!');
     }
 }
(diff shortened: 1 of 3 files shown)


src/Resources/skeleton/controller/Controller.php.txt
```diff
@@ -13,7 +13,6 @@ class {{ controller_class_name }} extends Controller
      */
     public function index()
     {
-        // replace this line with your own code!
-        return $this->render('@Maker/demoPage.html.twig', [ 'path' => str_replace($this->getParameter('kernel.project_dir').'/', '', __FILE__) ]);
+        return new Response('Hello World');
```
weaverryan: How about:

> // replace this line with your own code!
> Welcome to your new controller!

fabpot:
Thank you @javiereguiluz.

PR #10, opened 2017-11-17
[DoctrineMigration] add the maker command for doctrine migration

Simperfit:
I will add the test (need to switch room) later today.

### DIFF
--- a/src/Command/MakeMigrationCommand.php
+++ b/src/Command/MakeMigrationCommand.php
@@ -1,0 +1,89 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Command;
+
+use Doctrine\Bundle\MigrationsBundle\Command\DoctrineCommand;
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Component\Console\Exception\CommandNotFoundException;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ */
+final class MakeMigrationCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:migration';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Creates a migration with the diff or an empty if empty is provided.')
+            ->addOption('empty', null, InputOption::VALUE_OPTIONAL, 'Generate an empty classes.', false)
+            ->addOption('db', null, InputOption::VALUE_REQUIRED, 'The database connection to use for this command.')
+            ->addOption('em', null, InputOption::VALUE_OPTIONAL, 'The entity manager to use for this command.')
+            ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeMigration.txt'))
+        ;
+    }
+
+    protected function writeNextStepsMessage(array $params, ConsoleStyle $io)
+    {
+        $io->text([
+            'Next: Edit Version(lastVersion) to customize the migration.',
+            'Find the documentation at <fg=yellow>https://symfony.com/doc/current/bundles/DoctrineMigrationsBundle/index.html</>'
+        ]);
+    }
+
+    protected function configureDependencies(DependencyBuilder $dependencies)
+    {
+        $dependencies->addClassDependency(
+            DoctrineCommand::class,
+            'migrations'
+        );
+    }
+
+    protected function getParameters(): array
+    {
+        return [];
+    }
+
+    protected function getFiles(array $params): array
+    {
+        return [];
+    }
+
+    protected function generateCode(OutputInterface $output)
+    {
+        $isEmpty = $this->input->getOption('empty');
+        $options = ['db' => $this->input->getOption('db'), 'em' => $this->input->getOption('em'), 'shard' => $this->input->getOption('shard')];
+
+        try {
+            $diffMigrationCommand = $this->getApplication()->find('doctrine:migrations:diff');
+            $generateMigrationCommand = $this->getApplication()->find('doctrine:migrations:generate');
+        } catch (CommandNotFoundException $e) {
+            throw new RuntimeCommandException(sprintf("Missing package %s: to use the migration, run: \n\ncomposer require %s", 'migrations', 'migrations'));
+        }
+
+        if (true === $isEmpty) {
+            $options['command'] = 'doctrine:migrations:diff';
+            $generateMigrationCommand->run(new ArgvInput($options), $output);
+        } else {
+            $options['command'] = 'doctrine:migrations:generate';
+            $diffMigrationCommand->run(new ArgvInput($options), $output);
+        }
+    }
+
+}
--- a/tests/Command/FunctionalTest.php
+++ b/tests/Command/FunctionalTest.php
@@ -3,12 +3,14 @@
 namespace Symfony\Bundle\MakerBundle\Tests\Command;
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Bundle\MakerBundle\Command\AbstractCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeCommandCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeControllerCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeEntityCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeFormCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeFunctionalTestCommand;
+use Symfony\Bundle\MakerBundle\Command\MakeMigrationCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeSubscriberCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeTwigExtensionCommand;
 use Symfony\Bundle\MakerBundle\Command\MakeUnitTestCommand;
@@ -17,6 +19,13 @@ use Symfony\Bundle\MakerBundle\Command\MakeVoterCommand;
 use Symfony\Bundle\MakerBundle\EventRegistry;
 use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\HelperSet;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputDefinition;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\ConsoleOutput;
 use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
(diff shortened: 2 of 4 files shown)


src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        if (true === $isEmpty) {
+            $options['command'] = 'doctrine:migrations:diff';
+            $generateMigrationCommand->run(new ArgvInput($options), $output);
+        } else {
+            $options['command'] = 'doctrine:migrations:migrate';
+            $diffMigrationCommand->run(new ArgvInput($options), $output);
+        }
+
+        $this->io->newLine();
+        $this->io->writeln(' <bg=green;fg=white>          </>');
```
OskarStark: why not?
```php
$this->io->success('Success!');
```

Simperfit: I'm using the same format as the other commands, but why not I need to look

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+ */
+final class MakeMigrationCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:migrations';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Creates a migration with the diff or an empty migration if there are no diffs')
+            ->addOption('empty', null, InputOption::VALUE_OPTIONAL, 'Generate an empty classes', 'false')
```
weaverryan: should be `false`, not a string, right?

tests/Command/MakeMigrationCommandTest.php
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Symfony\Bundle\MakerBundle\Tests\Command;
+
+use Symfony\Bundle\MakerBundle\Command\MakeMigrationCommand;
+use PHPUnit\Framework\TestCase;
+
+class MakeMigrationCommandTest extends TestCase
+{
+
+}
```
weaverryan: We should add a test here: https://github.com/symfony/maker-bundle/blob/master/tests/Command/FunctionalTest.php#L68

And then we might not need a separate test at all :)

src/Command/MakeMigrationCommand.php
weaverryan: iirc, this is just a way to have a little bit prettier format that `$io->success()`, but I don't remember 100% :)

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return [];
+    }
+
+    protected function getFiles(array $params): array
+    {
+        return [];
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
```
weaverryan: It's not really an intention for commands to override `execute()`. I totally see why you did it though - this generator is a little bit different :). What about this:

* Add a new `protected function generateCode()` in `AbstractCommand` and put the `$this->generator->generate()` code there by default.

* Override *that* method instead of `execute()` in this class.

Then we won't need to repeat things, like the success message at the bottom.

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+ * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ */
+final class MakeMigrationCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:migrations';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Creates a migration with the diff or an empty migration if there are no diffs')
```
weaverryan: After some more thinking, I think we should *not* automatically create an empty migration if the diff is empty. Instead, we should show a message:

> No changes were detected to your mapping metadata.
> If you want to generate an empty migration, re-run the command with --empty

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        } catch (CommandNotFoundException $e) {
+            throw new RuntimeCommandException(sprintf("Missing package%s: to use the migrations, run: \n\ncomposer require %s", 'migrations', 'migrations'));
+        }
+
+        if (true === $isEmpty) {
+            $options['command'] = 'doctrine:migrations:diff';
+            $generateMigrationCommand->run(new ArgvInput($options), $output);
+        } else {
+            $options['command'] = 'doctrine:migrations:migrate';
+            $diffMigrationCommand->run(new ArgvInput($options), $output);
```
weaverryan: Should this be `ArgvInput` or `ArrayInput`? I'm actually really not sure.

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        try {
+            $diffMigrationCommand = $this->getApplication()->find('doctrine:migrations:diff');
+            $generateMigrationCommand = $this->getApplication()->find('doctrine:migrations:migrate');
+        } catch (CommandNotFoundException $e) {
+            throw new RuntimeCommandException(sprintf("Missing package%s: to use the migrations, run: \n\ncomposer require %s", 'migrations', 'migrations'));
+        }
+
+        if (true === $isEmpty) {
+            $options['command'] = 'doctrine:migrations:diff';
+            $generateMigrationCommand->run(new ArgvInput($options), $output);
```
weaverryan: Is it possible to determine the new filename that was generated? I would love to repeat that in the "next steps" message.

src/Resources/help/MakeMigrations.txt
```diff
@@ -0,0 +1,5 @@
+The <info>%command.name%</info> command generates a new migrations:
+
+<info>php %command.full_name% doctrine:migrations:migrate</info>
```
weaverryan: Should be `make:migration`

src/Resources/help/MakeMigrations.txt
```diff
@@ -0,0 +1,5 @@
+The <info>%command.name%</info> command generates a new migrations:
```
weaverryan: migrations -> migration

src/Resources/help/MakeMigrations.txt
```diff
@@ -0,0 +1,5 @@
...
+
+<info>php %command.full_name% doctrine:migrations:migrate</info>
+
+If the argument is missing, the command will ask for the command name interactively.
```
weaverryan: not needed. But we should probably show one other example that shows the options being used.

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ */
+final class MakeMigrationCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:migrations';
```
weaverryan: `make:migration`

src/Command/MakeMigrationCommand.php
javiereguiluz: Do we really need all these options? Simplicity is the main feature of this bundle. Ideally commands have zero or one option. This looks too much to me ... but I don't use migrations, so please @weaverryan review the UX/DX of this command and see if we really need all these options. Thanks!

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Creates a migration with the diff or an empty migration if there are no diffs')
+            ->addOption('empty', null, InputOption::VALUE_OPTIONAL, 'Generate an empty classes', 'false')
+            ->addOption('db', null, InputOption::VALUE_REQUIRED, 'The database connection to use for this command.')
```
fabpot: Dots at the end of each description (same below) should be removed.

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+            ->addOption('em', null, InputOption::VALUE_OPTIONAL, 'The entity manager to use for this command.')
+            ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeMigrations.txt'))
+        ;
+    }
+
+    protected function writeNextStepsMessage(array $params, ConsoleStyle $io)
+    {
+        $io->text([
+            'Next: We just generated the migration',
```
fabpot: The `We just generated the migration` should probably be removed with something along the lines of "Edit FILENAME to customize the migration."

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+            'Find the documentation at <fg=yellow>http://symfony.com/doc/current/bundles/DoctrineMigrationsBundle/index.html</>'
```
fabpot: Use https instead of http

src/Command/MakeMigrationCommand.php
OskarStark: or sth like this: `make:migration --allow-empty`, `make:migration:diff` (`make::diff-migration`) or `make:migration:empty` (`make:empty-migration`)

javiereguiluz: That looks too much and too complicated. Simplicity is the main feature of this bundle. Thanks!

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        $params = $this->getParameters();
+
+        $isEmpty = $this->input->getOption('empty');
+        $options = ['db' => $this->input->getOption('db'), 'em' => $this->input->getOption('em'), 'shard' => $this->input->getOption('shard')];
+
+        try {
+            $diffMigrationCommand = $this->getApplication()->find('doctrine:migrations:diff');
+            $generateMigrationCommand = $this->getApplication()->find('doctrine:migrations:migrate');
+        } catch (CommandNotFoundException $e) {
+            throw new RuntimeCommandException(sprintf("Missing package%s: to use the migrations, run: \n\ncomposer require %s", 'migrations', 'migrations'));
```
OskarStark:
```diff
- Missing package%s: to use...
+ Missing package %s: to use... 
```

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        if (true === $isEmpty) {
```
OskarStark: is this needed?

```diff
- if (true === $isEmpty) {
+ if ($isEmpty) {
```

src/Command/MakeMigrationCommand.php
Simperfit: I think ArgvInput is correct here too, but i'm not sure too.

src/Command/MakeMigrationCommand.php
Simperfit: I don't want it to be another type and equal to true.

src/Command/MakeMigrationCommand.php
Simperfit: I can't have the filename since it just in the output, or is there something that i'm missing ?

Simperfit:
I think there is something wrong with the code, since I don't have any arguments and it expect me to have one

src/Command/MakeMigrationCommand.php
Simperfit: @alcaeus is it possible ? We may have talked about that but I can't remember

alcaeus: We discussed it. The command doesn't return anything and I don't think it would be good for the command to keep the name of the last generated file in its state.

The only other option that comes to mind is actively parsing the output to grab the generated filename from the output.

weaverryan:
If/when this is merged, the official docs will need to be updated.

src/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,89 @@
...
+            $options['command'] = 'doctrine:migrations:diff';
+            $generateMigrationCommand->run(new ArgvInput($options), $output);
+        } else {
+            $options['command'] = 'doctrine:migrations:generate';
+            $diffMigrationCommand->run(new ArgvInput($options), $output);
+        }
+    }
+
+}
```
OskarStark: EOF missing

tests/Command/FunctionalTest.php
```diff
@@ -67,6 +76,58 @@ public function testCommands(AbstractCommand $command, array $inputs)
...
+            ->willReturn($inputDefinition);
+
+        $command->setApplication($application);
+        $tester = new CommandTester($command);
+        $tester->setInputs(array());
+        $tester->execute(array());
+
+        $this->assertContains('Success', $tester->getDisplay());
+        $this->assertContains('Version', $tester->getDisplay());
+
```
OskarStark: please remove this blank line

OskarStark:
We should activate fabot.io for this repository. Could you do this @weaverryan ?

fabpot:
@OskarStark fabbot activated for this repo.

weaverryan:
@Simperfit Can you rebase (sorry, I changed everything!) and look at the comments from @OskarStark?

weaverryan:
Replacing this with #97 - original commit included!

PR #12, opened 2017-11-17
Add initialize and interact stubs to Command skeleton

JoeKre:
I often find myself at least needing the initialize method for my commands. 

This PR includes initialize and interact in the Command skeleton with the information from the docs about the time they are executed.

### DIFF
--- a/src/Resources/skeleton/command/Command.php.txt
+++ b/src/Resources/skeleton/command/Command.php.txt
@@ -22,6 +22,26 @@ class {{ command_class_name }} extends Command
         ;
     }
 
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        //Optional.
+        //If no initialization is required either leave this empty or remove it.
+
+        //This method is executed before the interact() and the execute() methods.
+        //Its main purpose is to initialize variables used in the rest of the command methods.
+    }
+
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        //Optional.
+        //If no interaction from the user is expected either leave this empty or remove it.
+
+        //This method is executed after initialize() and before execute().
+        //Its purpose is to check if some of the options/arguments are missing and interactively ask the user for those values.
+        //This is the last place where you can ask for missing options/arguments.
+        //After this command, missing options/arguments will result in an error.
+    }
+
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $io = new SymfonyStyle($input, $output);
--- a/src/Resources/skeleton/command/Command.php.txt
+++ b/src/Resources/skeleton/command/Command.php.txt
@@ -22,24 +22,27 @@ class {{ command_class_name }} extends Command
         ;
     }
 
+    /**
+     * Optional.
+     *
+     * This method is executed before the interact() and the execute() methods.
+     * Its main purpose is to initialize variables used in the rest of the command methods.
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/command/Command.php.txt
```diff
@@ -22,6 +22,26 @@ class {{ command_class_name }} extends Command
...
     }
 
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        //Optional.
+        //If no initialization is required either leave this empty or remove it.
+
+        //This method is executed before the interact() and the execute() methods.
+        //Its main purpose is to initialize variables used in the rest of the command methods.
+    }
```
weaverryan: Hmm. I'd rather *not* have this one. I've actually never used it before. And I think `__construct()` might now be a better place.

src/Resources/skeleton/command/Command.php.txt
```diff
@@ -22,6 +22,26 @@ class {{ command_class_name }} extends Command
...
+
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        //Optional.
+        //If no interaction from the user is expected either leave this empty or remove it.
+
+        //This method is executed after initialize() and before execute().
+        //Its purpose is to check if some of the options/arguments are missing and interactively ask the user for those values.
+        //This is the last place where you can ask for missing options/arguments.
+        //After this command, missing options/arguments will result in an error.
```
weaverryan: Can you move this to comments (`/**`) above the method?

And I think we can say less:

> This method is executed before execute() and can be used
> to interactively ask for information from the user
> https://symfony.com/doc/current/console.html#command-lifecycle

javiereguiluz:
At first sight I'm 👎 about this proposal. We're aggressively pushing for total simplicity in this bundle. That means this won't solve all problems, but in exchange, this bundle will always be simple and lightweight (we don't want to repeat the old GeneratorBundle issues).

src/Resources/skeleton/command/Command.php.txt
JoeKre: Depending on what needs to be initialized you are right `__construct` is the place to go, unless you need the user provided arguments for initialization. In this case often execute is used directly which clutters the actual task that's to be done a bit. But of course it's a matter of taste.

src/Resources/skeleton/command/Command.php.txt
JoeKre: I moved both comments outside of the methods and adjusted the contents like you suggested. 
Leaving both methods included for now. @javiereguiluz has a valid point concerning the overall simplicity of this bundle. Which could lead to complete removal of both additions anyways.

JoeKre:
@javiereguiluz that is a completely valid point of course. I understand both sides and don't think either side is wrong in any way. 
Once you are starting to make exceptions there will be more. 
I think the skeletons are a good way to show beginners some more options than the minimalistic basics too. 

Even i find it hard to think about how you could differentiate if an addition is helpful or useless for most of the users. 
So going for a pure "simplicity" strategy and only include the absolute minimal requirements isn't a bad option by any means. 

I'll leave that up for you guys to decide and won't take it personal in any way if the PR gets declined :)

weaverryan:
Yes... I agree with Javier. Thanks for the PR - I hope we see more :)

PR #13, opened 2017-11-17
Added the initial documentation

javiereguiluz:


README.md
```diff
@@ -3,5 +3,11 @@ SymfonyMakerBundle
...
-bundle replaces [SensioGeneratorBundle](https://github.com/sensiolabs/SensioGeneratorBundle)
-in modern Symfony applications.
+bundle replaces [SensioGeneratorBundle][1] in modern Symfony applications and
+requires using Symfony 3.4 or newer and [Symfony Flex][2].
+
+[Read the docs][3]
+
+[1]: https://github.com/sensiolabs/SensioGeneratorBundle
+[2]: https://symfony.com/doc/current/setup/flex.html
+[3]: Resources/doc/index.rst
```
xabbuh: I would link to the rendered documentation on symfony.com instead. Reading reStructuredText files may not be much fun.

src/Resources/doc/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+    -class ProductCommand extends Command
+    +class ProductCommand extends AbstractCommand
+    {
+        protected static $defaultName = 'make:report';
+
+        // ...
+    }
+
+Finally, implement the methods requires by the ``AbstractCommand`` class::
```
xabbuh: required

src/Resources/doc/index.rst
```diff
@@ -0,0 +1,121 @@
...
+        {
+            return [
+                __DIR__.'/../Resources/skeleton/report.txt' => 'reports/'.$params['filename'];
+            ];
+        }
+
+        // Optionally, display some message after the generation of code
+        protected function writeNextStepsMessage(array $params, ConsoleStyle $io)
+        {
+            $io->text(sprintf('A new report was generated in %s file.', $params['filename']));
```
xabbuh: [...] in the [...]

README.md
```diff
@@ -3,5 +3,11 @@ SymfonyMakerBundle
 
 Symfony Maker helps you creating empty commands, controllers, form classes,
 tests and more so you can forget about the required boilerplate code. This
-bundle replaces [SensioGeneratorBundle](https://github.com/sensiolabs/SensioGeneratorBundle)
-in modern Symfony applications.
+bundle replaces [SensioGeneratorBundle][1] in modern Symfony applications and
```
fabpot: I'm sure we can claim that it replaces the generator bundle. At least, not yet as the main feature is CRUD generation, which this bundle does not do yet.

README.md
javiereguiluz: Yes! We'll update this when/if docs are published on symfony.com

src/Resources/doc/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: terminal
+
+    $ php bin/console make:
+
+     make:command            Creates a new console command class
+     make:controller         Creates a new controller class
+     make:entity             Creates a new Doctrine entity class
+
+     [...]
```
maidmaid: What about to explicitly list all the commands here? We'd know exactly all that is possible to _make_ only by reading the readme.

src/Resources/doc/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+In case your applications need to generate custom boilerplate code, you can
+create your own ``make: ...`` command reusing the tools provided by this bundle.
+Imagine that you need to create a ``make:report`` command. First, create an
+empty command:
+
+.. code-block:: terminal
+
+    $ cd your-project/
+    $ php bin/console make:command 'make:report'
```
maidmaid: Why not create a `make:maker` command? :smiley:

src/Resources/doc/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+Usage
+-----
+
+This bundle provides several commands under the ``maker:`` namespace. List them
+all executing this command:
+
+.. code-block:: terminal
+
+    $ php bin/console make:
```
fabpot: The command to run is `php bin/console list make`. Yours does not work AFAIK.

src/Resources/doc/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: terminal
+
+    $ php bin/console make:controller --help
+
+Creating your Own Makers
+------------------------
+
+In case your applications need to generate custom boilerplate code, you can
+create your own ``make: ...`` command reusing the tools provided by this bundle.
```
fabpot: I would remove the space after`make:`

src/Resources/doc/index.rst
javiereguiluz: I prefer to not do that to avoid having to maintain that list forever. This bundle just started, so we'll update this list of commands often ... and the less work we put on ourselves (e.g. maintaining this doc) the better.

src/Resources/doc/index.rst
javiereguiluz: For the moment this is an "edge feature". If it gets popular, we would add it in the future.

src/Resources/doc/index.rst
javiereguiluz: It works ... but it's not pretty:

command-1

Your proposal is much better (although the output is a bit noisy):

command-2

fabpot:
Thank you @javiereguiluz.

PR #21, opened 2017-11-18
Validating empty arguments

jzawadzki:
Right now almost all commands are throwing errors if the `name` argument is empty.
This small PR adds validation to every command to check for empty values.

(#SymfonyConHackday2017)

### DIFF
--- a/src/Command/MakeCommandCommand.php
+++ b/src/Command/MakeCommandCommand.php
@@ -13,6 +13,7 @@ namespace Symfony\Bundle\MakerBundle\Command;
 
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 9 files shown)


javiereguiluz:
But the optional command arguments have a default value, right? It shouldn't fail when executing the commands without arguments.

jzawadzki:
@javiereguiluz Only `MakeTwigExtensionCommand` have default `AppExtension`

When I'm trying to not give the value, they raise the exception (but pretty much every command = different exception because of different logic in the commands)
bug

javiereguiluz:
I've just tested it and you are right. However, this must be some last minute change. This worked before and must work again. There's no way this argument is empty. We always provide a default value or ask for it ... so there should be no need for these checks.

weaverryan:
I think this happens when you actually answer the interactive question with zero content.

A better solution (I think) is to update AbstractCommand to re-ask the question if you leave it blank or use the default/suggested value (not sure which is better)

weaverryan:
I'm not sure when this was fixed, but I believe this is no longer an issue :). The base `MakerCommand` checks for blank content on input: https://github.com/symfony/maker-bundle/blob/master/src/Command/MakerCommand.php#L82

This gives a nice "This value should not be blank" error instead of the bad error from before. Thanks for the PR and reporting the issue!

stof:
@weaverryan this is https://github.com/symfony/maker-bundle/pull/52

PR #22, opened 2017-11-18
Add entity class name to Voter skeleton

yceruto:


weaverryan:
Would it be too magic to check to see if the entity class exists and only do this change if it does?

And yea... make:form has the same issue. Let’s make sure we solve it correctly. For the form l, I almost want to ask for the model/entity class name in the prompt, instead of the form class name (which we could generate from the first answer). But then, if you don’t want to bound to a class (which is rare), you would need to leave the prompt blank (weird) and then we would ask you for the form class name. But maybe that’s better - it covers most cases. We could in theory do the same here.

yceruto:
@weaverryan I like your idea, but probably it requires more advanced template engine to avoid many skeleton versions, conditional block codes would help (@javiereguiluz' 
 proposal points in this direction https://github.com/symfony/maker-bundle/issues/33)

By now, I think this is consistent with other skeletons.

weaverryan:
@yceruto I've been accomplishing these more complex things by doing more logic in the command. For example, you could set a `extraSupportsCode` variable either to an empty string or ` && $subject instanceof {{ entity_class_name }};` and then *always* render this in the template. It looks a little ugly, but it works :). We could even test both possibilities in the functional test.

yceruto:
@weaverryan :+1: done!

yceruto:
Closing here because this probably should be achieved in a different way after https://github.com/symfony/maker-bundle/pull/59

weaverryan:
@yceruto yep, but if you still like the idea, let's create another PR later! :)

PR #23, opened 2017-11-18
Invocation paramter type are not compatible with declared

yceruto:
> Expected string, got array.


javiereguiluz:
Nice catch. Could we add PHP 7.1 typehints in addClassDependency instead of the current phpdoc?

yceruto:
> Nice catch. Could we add PHP 7.1 typehints in addClassDependency instead of the current phpdoc?

See https://github.com/symfony/maker-bundle/pull/24

fabpot:
Thank you @yceruto.

Issue #17, opened 2017-11-18
Include less common makers


maidmaid:
The documentation covers a lof of other boilerplates, but theses one are less common. So, we can imagine the next commands:

- `make:security:guard` [doc](http://symfony.com/doc/current/security/guard_authentication.html#step-1-create-the-authenticator-class)
- `make:serializer:encoder` [doc](http://symfony.com/doc/current/serializer/custom_encoders.html)
- `make:profiler:data-collector` [doc](http://symfony.com/doc/current/profiler/data_collector.html)
- `make:security:form-login` [doc](http://symfony.com/doc/current/security/form_login_setup.html)
- `make:monolog:processor` [doc](http://symfony.com/doc/current/logging/processors.html)
- ...

This bundle is interested only by the most common _makers_ or these previous examples are potentially to include in this project?

nicolas-grekas:
we could even handle a tag so that makers could be provided by bundles

javiereguiluz:
Not sure it's needed because a maker is just a console command whose name starts with make:

weaverryan:
Short answer: yes, we want these commands :). I’m planning several varieties of guard auth make commands.

About extensibility, we do want to empower other bundles to add make commands. Probably they can/should just extend our base class. Or, a separate class with an interface might be even better.

Javier - it’s true that anyone can just create a make: command. But we need just enough extensibility so that those commands can have the same look & feel :)

nicolas-grekas:
Great :) Note that extending the base command might be a stronger coupling than providing a service ppl could use instead. Either way, we need to be careful to make the maker-bundle a mandatory requirement of all other bundles that want to hook into it.

luishdez:
May be `make:kernel`  to create separated kernels eg a front script that loads ApiKernel with another set of bundles … configs

Though the prefix for kernels in the new sf has been remove so not sure how will be the new or best practice approach to achieve this.

javiereguiluz:
@luishdez thanks for your proposal. The problem is that newer Symfony versions (specifically Symfony 4 with Flex) is discouraging the creation of multiple-kernel apps. Instead it's recommending the creation of multiple small apps.

ad3n:
Making bundle have/has a tag is better than create an command with make as prefix

yceruto:
> The problem is that newer Symfony versions (specifically Symfony 4 with Flex) is discouraging the creation of multiple-kernel apps. Instead it's recommending the creation of multiple small apps.

[...]

PR #27, opened 2017-11-18
Added Event Listener generation command.

piotrgradzinski:
May be useful to create an Event Listener using make command.

### DIFF
--- a/src/Command/MakeListenerCommand.php
+++ b/src/Command/MakeListenerCommand.php
@@ -1,0 +1,127 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Command;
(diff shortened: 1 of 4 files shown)


weaverryan:
Thanks for the PR! This seems logical, but I’m -1. Event listeners require service config to work, and I don’t want beginners to be confused and not know which maker command to use.

nicolas-grekas:
I agree with Ryan: making a subscriber is a better option because it is self describing, thus auto configurable.

piotrgradzinski:
@nicolas-grekas, @weaverryan - sure that making a subscriber is a better option mainly due to configuration option. I have to admit that having to copy/paste additional code to `services.yaml` is a pain. 

If you don't see the space for this PR please close it :)

weaverryan:
Closing now :). But I hope to see more PR’s in the future!

piotrgradzinski:
@weaverryan - sure, actually working on some commands from #17 issue :)

javiereguiluz:
@piotrgradzinski thanks for your willingness to contribute ... but let's slow down a bit. We're going to add more commands, but not immediately. Let' fix first all the current issues, problems and bad choices. Thanks!

PR #29, opened 2017-11-19
Fixing MakeCommandCommand dependency configuration.

piotrgradzinski:
In `src/Command/MakeCommandCommand.php` second argument for class dependency is passed as array what causes an error when dependency is missing in `src/Command/AbstractCommand.php:87`.

This PR should fix that in `src/Command/MakeCommandCommand.php`.


piotrgradzinski:
Looks to be fixed earlier :)

PR #30, opened 2017-11-19
Added MakeSerializerEncoder command.

piotrgradzinski:
@javiereguiluz - I know that this is an early stage of this repository and you've mentioned in #27 that the focus should be more on fixing current issues and problems. I would love to help on those :)

Nevertheless I hope you'll find this PR useful as a implementation of one of proposed commands from #17.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,12 +17,12 @@
         "symfony/dependency-injection": "^3.4|^4.0",
         "symfony/filesystem": "^3.4|^4.0",
         "symfony/framework-bundle": "^3.4|^4.0",
-        "symfony/http-kernel": "^3.4|^4.0",
-        "symfony/serializer": "^3.4|^4.0"
+        "symfony/http-kernel": "^3.4|^4.0"
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4|^4.0",
-        "symfony/process": "^3.4|^4.0"
+        "symfony/process": "^3.4|^4.0",
+        "symfony/serializer": "^3.4|^4.0"
     },
     "config": {
         "preferred-install": {
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
         "symfony/dependency-injection": "^3.4|^4.0",
         "symfony/filesystem": "^3.4|^4.0",
         "symfony/framework-bundle": "^3.4|^4.0",
-        "symfony/http-kernel": "^3.4|^4.0"
+        "symfony/http-kernel": "^3.4|^4.0",
+        "symfony/serializer": "^3.4|^4.0"
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4|^4.0",
--- a/composer.json
+++ b/composer.json
@@ -21,8 +21,7 @@
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4|^4.0",
-        "symfony/process": "^3.4|^4.0",
-        "symfony/serializer": "^3.4|^4.0"
+        "symfony/process": "^3.4|^4.0"
     },
     "config": {
         "preferred-install": {
--- a/src/Command/MakeSerializerEncoderCommand.php
+++ b/src/Command/MakeSerializerEncoderCommand.php
@@ -68,8 +68,8 @@ final class MakeSerializerEncoderCommand extends AbstractCommand
     protected function configureDependencies(DependencyBuilder $dependencies)
     {
         $dependencies->addClassDependency(
-          Serializer::class,
-          'serializer'
+            Serializer::class,
+            'serializer'
         );
     }
 }
--- a/src/Command/MakeSerializerEncoderCommand.php
+++ b/src/Command/MakeSerializerEncoderCommand.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Command;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+use Symfony\Component\Serializer\Serializer;
+
+/**
+ * @author Piotr Grabski-Gradzinski <piotr.gradzinski@gmail.com>
+ */
+final class MakeSerializerEncoderCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:serializer:encoder';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Creates a new custom encoder class')
+            ->addArgument('name', InputArgument::OPTIONAL, 'Choose a class name for your encoder (e.g. <fg=yellow>YamlEncoder</>).')
+            ->addArgument('format', InputArgument::OPTIONAL, 'Pick your format name (e.g. <fg=yellow>yaml</>)')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeSerializerEncoder.txt'))
+        ;
+    }
+
+    protected function getParameters(): array
+    {
+        $encoderClassName = Str::asClassName($this->input->getArgument('name'), 'Encoder');
+        Validator::validateClassName($encoderClassName);
+        $format = $this->input->getArgument('format');
+
+        return [
+            'encoder_class_name' => $encoderClassName,
+            'format' => $format,
+        ];
+    }
+
+    protected function getFiles(array $params): array
+    {
+        return [
+            __DIR__.'/../Resources/skeleton/serializer/Encoder.php.txt' => 'src/Serializer/'.$params['encoder_class_name'].'.php'
+        ];
+    }
+
+    protected function writeNextStepsMessage(array $params, ConsoleStyle $io)
+    {
+        $io->text([
+            'Next: Open your new encoder class and start customizing it.',
+            'Find the documentation at <fg=yellow>http://symfony.com/doc/current/serializer/custom_encoders.html</>'
+        ]);
+    }
+
+    protected function configureDependencies(DependencyBuilder $dependencies)
+    {
+        $dependencies->addClassDependency(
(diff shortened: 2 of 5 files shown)


src/Command/MakeSerializerEncoderCommand.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        return [
+            __DIR__.'/../Resources/skeleton/serializer/Encoder.php.txt' => 'src/Serializer/'.$params['encoder_class_name'].'.php'
+        ];
+    }
+
+    protected function writeNextStepsMessage(array $params, ConsoleStyle $io)
+    {
+        $io->text([
+            'Next: Open your new encoder class and start customizing it.',
```
javiereguiluz: This is not a question about your Pull Request but a general question: are these "next steps" instructions really useful?

* This doesn't provide much: *"Next: Open your new encoder class and start customizing it."*
* And this is more suited for a "demo app" or "newcomer learning resource": *"Find the documentation at ..."*

I propose to remove all these obvious "next steps" and only keep the ones really needed.

weaverryan: I think you may be right. But I still want a clear “success” message at the bottom. Maybe there’s a default, generic success message and we only override it when there really are additional steps

composer.json
```diff
@@ -17,7 +17,8 @@
         "symfony/dependency-injection": "^3.4|^4.0",
         "symfony/filesystem": "^3.4|^4.0",
         "symfony/framework-bundle": "^3.4|^4.0",
-        "symfony/http-kernel": "^3.4|^4.0"
+        "symfony/http-kernel": "^3.4|^4.0",
+        "symfony/serializer": "^3.4|^4.0"
```
weaverryan: This should be in require-dev - it’s only needed to test the app.

src/Command/MakeSerializerEncoderCommand.php
piotrgradzinski: The exact *next* information is not really helpful in this case (just followed the convention from other commands). 

Just to point that when you generate new command over SensioGeneratorBundle after confirmation you'll receive

```
You are going to generate a foo-bar command inside AppBundle bundle.
Do you confirm generation [yes]? 
  created ./src/AppBundle/Command/
  created ./src/AppBundle/Command/FooBarCommand.php
Generated the foo-bar command in AppBundle

                                         
  Everything is OK! Now get to work :).  
```

Not sure if and how close to this maker should be.

If agreed I can refactor existing commands for that.

src/Resources/skeleton/serializer/Encoder.php.txt
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function encode($data, $format, array $context = array())
+    {
+        // TODO: return your encoded data
+        return '';
+    }
+
+    public function supportsEncoding($format)
+    {
+        return '{{ format }}' === $format;
```
ogizanagi: The format could be in a `FORMAT` class const, as done in built-in encoders:

https://github.com/symfony/symfony/blob/8eb97f5c483b57d8cec2fe999d1437fadfeae575/src/Symfony/Component/Serializer/Encoder/JsonEncoder.php#L21

piotrgradzinski: @ogizanagi - makes sense. Fixed.

piotrgradzinski:
@weaverryan - fixed `composer.json`
@ogizanagi - changed format to const.

src/Command/MakeSerializerEncoderCommand.php
```diff
@@ -0,0 +1,75 @@
...
+ * @author Piotr Grabski-Gradzinski <piotr.gradzinski@gmail.com>
+ */
+final class MakeSerializerEncoderCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:serializer:encoder';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Creates a new custom encoder class')
```
javiereguiluz: What about:

```
Creates a new serializer encoder class
```

src/Resources/skeleton/serializer/Encoder.php.txt
```diff
@@ -0,0 +1,33 @@
...
+{
+    const FORMAT = '{{ format }}';
+    
+    public function encode($data, $format, array $context = array())
+    {
+        // TODO: return your encoded data
+        return '';
+    }
+
+    public function supportsEncoding($format)
```
javiereguiluz: Just asking: we can't do it right and add type hints (e.g. `supportsEncoding(string $format): bool`) because of the interfaces, right?

src/Command/MakeSerializerEncoderCommand.php
javiereguiluz: OK ... let's discuss about this in a separate PR and, if needed, let's make the change at once for all commands.

src/Command/MakeSerializerEncoderCommand.php
piotrgradzinski: Yup, fixed in command description, success message (as is :) ) and command help.

src/Resources/skeleton/serializer/Encoder.php.txt
piotrgradzinski: @javiereguiluz indeed. Since there is no type-hinting in interface trying to add some in implemented method will end up with 
```
Fatal error: Declaration of YamlEncoder::supportsEncoding(string $format) must be compatible with EncoderInterface::supportsEncoding($format)
```

ogizanagi: In fact you could at least add the return type: https://3v4l.org/GiabR.

piotrgradzinski: @ogizanagi right! Done!

composer.json
```diff
@@ -21,7 +21,8 @@
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4|^4.0",
-        "symfony/process": "^3.4|^4.0"
+        "symfony/process": "^3.4|^4.0",
+        "symfony/serializer": "^3.4|^4.0"
```
yceruto: It is really necessary? we don't do this for the current ones.

piotrgradzinski: @yceruto - it was from the beginning of the repo. Correct me if I'm wrong but I think removing this should go to a separate issue.

yceruto: `symfony/serializer` deps has been introduced in this PR, no? I think is not necessary for tests either, as tests don't check dependencies currently:
https://github.com/yceruto/maker-bundle/blob/c402dac16d4744bf0c1d159ec67bdaf0d3fbd619/tests/Command/FunctionalTest.php#L49

src/Resources/skeleton/serializer/Encoder.php.txt
```diff
@@ -0,0 +1,33 @@
...
+{
+    const FORMAT = '{{ format }}';
+    
+    public function encode($data, $format, array $context = array())
+    {
+        // TODO: return your encoded data
+        return '';
+    }
+
+    public function supportsEncoding($format): bool
```
dunglas: `?string $format` should work (not tested it)

src/Resources/skeleton/serializer/Encoder.php.txt
```diff
@@ -0,0 +1,33 @@
...
+namespace App\Serializer;
+
+use Symfony\Component\Serializer\Encoder\DecoderInterface;
+use Symfony\Component\Serializer\Encoder\EncoderInterface;
+
+class {{ encoder_class_name }} implements EncoderInterface, DecoderInterface
+{
+    const FORMAT = '{{ format }}';
+    
+    public function encode($data, $format, array $context = array())
```
dunglas: `$context = []` as we're on 7.1+?

src/Resources/skeleton/serializer/Encoder.php.txt
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace App\Serializer;
+
+use Symfony\Component\Serializer\Encoder\DecoderInterface;
+use Symfony\Component\Serializer\Encoder\EncoderInterface;
+
+class {{ encoder_class_name }} implements EncoderInterface, DecoderInterface
+{
+    const FORMAT = '{{ format }}';
```
dunglas: `public const`?

composer.json
piotrgradzinski: Your right. Removed from `composer.json`.

src/Resources/skeleton/serializer/Encoder.php.txt
piotrgradzinski: @dunglas - unfortunately the problem remains, tested on PHP 7.1.6

```
Fatal error: Declaration of YamlEncoder::supportsEncoding(?string $format) must be compatible with EncoderInterface::supportsEncoding($format)
```

src/Resources/skeleton/serializer/Encoder.php.txt
piotrgradzinski: Fixed. Thank you!

src/Resources/skeleton/serializer/Encoder.php.txt
piotrgradzinski: Fixed. Thank you!

src/Command/MakeSerializerEncoderCommand.php
```diff
@@ -0,0 +1,75 @@
...
+            'Next: Open your new serializer encoder class and start customizing it.',
+            'Find the documentation at <fg=yellow>http://symfony.com/doc/current/serializer/custom_encoders.html</>'
+        ]);
+    }
+
+    protected function configureDependencies(DependencyBuilder $dependencies)
+    {
+        $dependencies->addClassDependency(
+          Serializer::class,
+          'serializer'
```
yceruto: use 4 space for each level of indentation.

piotrgradzinski: Fixed. Thank you!

piotrgradzinski:
@weaverryan - are you happy with the recent changes?

piotrgradzinski:
Guys, I know that in the light of @javiereguiluz recent post about this bundle the approach is to discuss the potential new command over a ticket but I am wondering whether this PR can be merged? It's been discussed in #17 and I hope make:serializer:encode code is good enough to be merged. 

Thank you for your attention!

weaverryan:
Thanks Piotr for working on this feature, this is much appreciated.

weaverryan:
And thanks also for making the quick changes! Woo!

src/Resources/skeleton/serializer/Encoder.php.txt
stof: adding it relies on PHP 7.2 new features.

stof:
I think making custom normalizer is much more likely to happen than custom encoders though.

piotrgradzinski:
@stof - this is a good idea I think. If you are OK with that I could start new issue to discuss this further. If there be a will for that I would love to prepare such command.

PR #36, opened 2017-11-21
Adding a new make:auth:empty

weaverryan:
This is the first of a few authenticator commands that I want to add.

It's not SUPER helpful - just a blank authenticator. My hope is that the other `make:auth:` commands will be used more often anyways (e.g. `make:auth:form`).

Cheers!


src/Command/MakeAuthenticatorEmptyCommand.php
```diff
@@ -0,0 +1,69 @@
...
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+final class MakeAuthenticatorEmptyCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:auth:empty';
+
+    public function configure()
+    {
+        $this
+            ->setDescription('Creates a new Doctrine entity class')
```
yceruto: copy+paste :) should be updated

tests/Command/FunctionalTest.php
```diff
@@ -183,6 +184,14 @@ public function getCommandTests()
             ]
         ];
 
+        $commands['auth_empty'] = [
+            new MakeAuthenticatorEmptyCommand($generator),
+            [
+                // class name
+                'AppCustomVoter'
```
yceruto: same here :) `AppCustomAuthenticator`?

weaverryan:
Thanks @yceruto :)

javiereguiluz:
I like this, but I don't like the `:empty` part of the name. The rest of the makers also generate empty things and they are not called `make:contorller:empty`, `make:command:empty`, etc.  What if we rename this to just `make:auth` but keep using the 3-part names for the other auths (e.g. `make:auth:form`)

PR #38, opened 2017-11-22
[EntityGenerator] add the support of ApiPlatform

Simperfit:
I would like to add the support of ApiPlatform in maker, since api-platform is already in the recipes of flex.
dsc_0003
(PRception)

### DIFF
--- a/src/Command/MakeEntityCommand.php
+++ b/src/Command/MakeEntityCommand.php
@@ -17,6 +17,7 @@ use Symfony\Bundle\MakerBundle\DependencyBuilder;
 use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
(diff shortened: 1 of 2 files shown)


src/Command/MakeEntityCommand.php
```diff
@@ -46,13 +49,14 @@ protected function getParameters(): array
             'entity_class_name' => $entityClassName,
             'entity_alias' => $entityAlias,
             'repository_class_name' => $repositoryClassName,
+            'api_platform' => $this->input->hasOption('with-api-platform') ? 'Api' : ''
```
fabpot: I'd would use --with-api instead

Simperfit: Yeah feels better, you are right.

javiereguiluz:
Not sure about this. ApiPlatform is popular but not standard or ubiquitous. Besides, ApiPlatform is so nice that it doesn't require much boilerplate: this maker would only add this to the generated entity -> `@ApiResource`

src/Command/MakeEntityCommand.php
```diff
@@ -72,6 +75,12 @@ protected function writeNextStepsMessage(array $params, ConsoleStyle $io)
 
     protected function configureDependencies(DependencyBuilder $dependencies)
     {
+        if ($this->input->hasOption('with-api') && true === $this->input->getOption('with-api')) {
```
stof: `hasOption` is useless. This command always has this option in its definition

weaverryan:
Yep, I think I'm 👎 on this too. I don't think we should add another question to the generator. And since the annotation is pretty easy.... :)

PR #40, opened 2017-11-22
[Entity] Made repository generation optional (true by default) and add namespace changement

### LINKED ISSUE #31: Namespaces and repository classes
Some questions:

1.  Is that possible to create entities or controllers with namespaces?
2.  How to disable creating repository class while making doctrine entity?

Simperfit:
This fixes #31 
It is now possible to generate an entity without a repository and with a custom namespace

### DIFF
--- a/src/Command/MakeEntityCommand.php
+++ b/src/Command/MakeEntityCommand.php
@@ -17,6 +17,7 @@ use Symfony\Bundle\MakerBundle\DependencyBuilder;
 use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Not sure about this. Simplicity is the main (and the best) feature of this bundle. We want to include options for everything. They provide flexibility ... but they complicate everything.

Next time is better to wait until the related issue (#31 in this case) is approved or rejected. Thanks!

Simperfit:
I totally understand your point ! Thanks for looking at the PR ;).

fabpot:
I think it is indeed not worth it. If one does not want the repository, he can delete the file. Sounds easy enough.

weaverryan:
Closing for the reasons listed above. Thanks @Simperfit!

PR #41, opened 2017-11-22
[User - WIP] add the user generator.

Simperfit:
- [ ] Add a test

### DIFF
--- a/src/Command/MakeUserCommand.php
+++ b/src/Command/MakeUserCommand.php
@@ -1,0 +1,76 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Command;
+
+use Doctrine\ORM\Mapping\Column;
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Ryan Weaver <weaverryan@gmail.com>
+ * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ */
+final class MakeUserCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:user';
+
+    public function configure()
+    {
+        $this
+            ->setDescription('Creates a the user doctrine class, the user provider and the user repository')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeUser.txt'))
+        ;
+    }
+
+    protected function getParameters(): array
+    {
+        return [];
+    }
+
+    protected function getFiles(array $params): array
+    {
+        return [
+            __DIR__.'/../Resources/skeleton/user/User.php.txt' => 'src/Entity/User.php',
+            __DIR__.'/../Resources/skeleton/user/UserRepository.php.txt' => 'src/Repository/UserRepository.php',
+            __DIR__.'/../Resources/skeleton/user/UserProvider.php.txt' => 'src/Security/UserProvider.php',
+        ];
+    }
+
+    protected function getResultMessage(array $params): string
+    {
+        return 'User entity, User provider and user repository has been created successfully.';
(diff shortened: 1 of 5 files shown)


javiereguiluz:
Thanks again for your contributions ... but I'd like to ask you to please stop sending pull requests. As we explained yesterday in https://symfony.com/blog/introducing-the-symfony-maker-bundle when introducing this bundle:

new-features

We need issues to discuss about ideas first. Pull requests must be sent only for approved issues. Thanks for understanding it!

src/Resources/skeleton/user/User.php.txt
```diff
@@ -0,0 +1,72 @@
...
+        return null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function serialize()
+    {
+        return serialize([
+            $this->id,
```
stof: not serializing the username will cause issues (incompatible with the remember me)

src/Resources/skeleton/user/UserProvider.php.txt
```diff
@@ -0,0 +1,81 @@
...
+            throw new UsernameNotFoundException(sprintf('The user "%s" does not exist.', $username));
+        }
+
+        return $user;
+    }
+
+    /**
+     * @throws UsernameNotFoundException if the user is not found
+     */
+    public function loadUserByResourceOwner($resourceOwner, string $provider): UserInterface
```
stof: does not make any sense in core (this is specific to HWIOAuthBundle AFAICT)

src/Resources/skeleton/user/User.php.txt
Simperfit: As i'm not using this feature I was not thinking about it, I understand, I will add it.

weaverryan:
As I mentioned in #42, I'm +1 for this command (I think @Simperfit may even have gotten the idea from a list I have that I shared with him).

@Simperfit could you rebase / update things for the latest `MakerInterface` change I made?

src/Resources/skeleton/user/UserProvider.php.txt
```diff
@@ -0,0 +1,81 @@
...
+
+use App\Entity\User;
+use App\Repository\UserRepository;
+use Doctrine\ORM\NoResultException;
+use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
+use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+
+class UserProvider implements UserProviderInterface
```
weaverryan: If we know the user wants an "entity" provider, then does it make sense to generate this class? They could just use the built-in `entity` provider.

src/Command/MakeUserCommand.php
```diff
@@ -0,0 +1,76 @@
...
+ * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ */
+final class MakeUserCommand extends AbstractCommand
+{
+    protected static $defaultName = 'make:user';
+
+    public function configure()
+    {
+        $this
+            ->setDescription('Creates a the user doctrine class, the user provider and the user repository')
```
weaverryan: Obviously, we want to make these commands as simple, and *not* interactive as possible (i.e. don't ask the user 10 questions). But, I think one question might make sense here:

> Do you need to store User details in the database?

This will tell us whether or not they need a `User` entity, or instead, a `User` model class. This would also change whether or not we generate a `UserProvider` (or just tell them to use the built-in `entity` provider). And finally, it will change what recommendations we give at the end: i.e. the exact changes they'll need to make to their `security.yml` file.

weaverryan: We may also need to ask a second question... if the first question is "yes" to storing user information:

> Will you need to store an encoded password for each User in the database?

This would influence how we handle the `getPassword()` method and whether or not we have a `password` field. Sometimes, *even* when people have a local `User` entity, they validate their password in a different system (e.g. via a SSO API). This is very confusing to users, so it's a great opportunity to help generate the `User` class correctly based on this.

src/Resources/skeleton/user/User.php.txt
```diff
@@ -0,0 +1,72 @@
...
+    public function eraseCredentials()
+    {
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getUsername(): string
+    {
+        return $this->email;
```
OskarStark: where does email come from?

ro0NL:
Just saw this.. FYI i just released a full working `make:user`... :sweat_smile: 

https://medium.com/@ro0NL/adding-user-management-to-your-symfony-application-ceeefe2a2e9

let me know if we should talk about name conflicts. Either way i'd very sadd with  `class User implements UserInterface` as the recommended approach.

PR #43, opened 2017-11-22
Add @ORM\Table by default

weaverryan:
I think having it is better. When a user later needs to add indexes, or (for some reason) wants to control the table name, it's there waiting for them.

### DIFF
--- a/src/Resources/skeleton/doctrine/Entity.php.txt
+++ b/src/Resources/skeleton/doctrine/Entity.php.txt
@@ -6,6 +6,7 @@ use Doctrine\ORM\Mapping as ORM;
 
 /**
  * @ORM\Entity(repositoryClass="App\Repository\{{ repository_class_name }}")
+ * @ORM\Table()
  */
 class {{ entity_class_name }}
 {


javiereguiluz:
I'm not sure this is truly useful ... because it doesn't help you in any way with the right syntax to use. Example:

```
@ORM\Table('users')         <-- wrong
@ORM\Table(name = 'users')  <-- wrong
@ORM\Table(name = "users")  <-- correct
```

And it's even worse for index, because the syntax is nested and not simple:

```
 * @Table(
 *     uniqueConstraints={@UniqueConstraint(name="user_unique",columns={"username"})},
 *     indexes={@Index(name="user_idx", columns={"email"})}
 * )
```

weaverryan:
We can't fix the nested part. But at least when they google or read the docs, they can see the `@ORM\Table` and know where it goes. Or do you think it's still not worth it?

javiereguiluz:
Another argument against doing this change: I don't think we should promote developers defining the name of their tables by hand. There are of course valid use cases for doing that ... but should we promote it by default?

Doctrine recipe already solves this (`naming_strategy: doctrine.orm.naming_strategy.underscore`) so the `@ORM\Table` would only be really used for index ... but that requires to do more changes and syntax is obscure.

weaverryan:
That sounds fine with me :)

PR #44, opened 2017-11-22
make:entity: Removing overridden __construct in repository

weaverryan:
You don't need this. You will only need to override the constructor if you need to add dependencies. And while showing parent::__construct() is useful for that situation, we should actually recommend against doing this (you shouldn't normally need extra deps in your repository class)


weaverryan:
Nevermind, you totally do need this. I'm doing too many things right now :)

PR #46, opened 2017-11-23
Remove the demo page for new controllers

javiereguiluz:
This demo page is great after installing Symfony (and we already have that thanks to the brilliant trick contributed by @yceruto) but for scaffolded controllers, I think it's better to not generate it because it's something you need to clean/remove after the code generation and that's always cumbersome and easy to forget.


src/Resources/skeleton/controller/ControllerWithTwig.php.txt
```diff
@@ -13,7 +13,6 @@ class {{ controller_class_name }} extends Controller
      */
     public function index()
     {
-        // replace this line with your own code!
-        return $this->render('@Maker/demoPage.html.twig', [ 'path' => str_replace($this->getParameter('kernel.project_dir').'/', '', __FILE__) ]);
+        return $this->render('your_own_template.html.twig');
```
nicolas-grekas: actually, it would be great to add a second argument to hint you can pass vars to it,
and also, would be greate to generate the twig file, using a simple naming scheme (similar to the one used by the `@Template` annotation?.)
The content of the file could be trivial, eg:
```
{% extends base.html.twig %}
{% block body %}
Very simple welcome message, or next steps hint
{% block endblock %}
```

javiereguiluz: We "can't" do that because maybe this controller is used in a API app that returns JSON contents. We can't assume Twig, templates, etc.  I know that the `$this->render()` call assumes that ... but it's simple to remove, unlike having to look for and remove a separate Twig file, etc.

nicolas-grekas: It's trivial to remove, and as you said is already opinionated since "render" is called.
The currently generated code it broken, which is bad experience.
Just dropping an additional file and the experience becomes smooth.

fabpot: I agree with @nicolas-grekas here. Generating some code that triggers an error is not a good DX.

stof: the example given by @nicolas-grekas still assumes a `base.html.twig` template, which may not be there either.

And btw, if Twig is not installed, this would also stay broken.

yceruto: I agree with @nicolas-grekas and @stof here, should be a simple template without `extends` tag.

> And btw, if Twig is not installed, this would also stay broken.

this template is only loaded if Twig dep is available.

nicolas-grekas: *with* the extends tag, so that discovering the base template is made easier, and valid html pages are generated by default :)

Tobion: why not just `return new Reponse('...')`?

nicolas-grekas: Because this is the controller that is generated when twig is installed.

Tobion: I don't see why we should distinguish that in the first place. Just because twig is installed, you still don't know what the controller is meant for. So just do the MVP.

nicolas-grekas: We're doing all of this to help the 80% use case, and to help newcomers discover how this is easy to use (and discover)
I think it's not unreasonable (like >50% OK) to think that if you have twig, you might want to use it to create your pages. That's the reasoning, and I fully support it :)

PR #47, opened 2017-11-23
Improved some PHPdoc comments

javiereguiluz:


fabpot:
Thank you @javiereguiluz.

PR #48, opened 2017-11-23
Minor code improvements

javiereguiluz:
Spotted with PHP-CS-Fixer and Php Inspections (EA Extended).


fabpot:
Thank you @javiereguiluz.

PR #50, opened 2017-11-23
Add supports to override skeleton templates

### LINKED ISSUE #33: [RFC] Allow to override the skeleton files?
When creating this bundle we decided to not require Twig to keep things simple. That means that our skeleton files can't have complex logic and we need to interpolate the placeholder variables ourselves. Not a big deal ... but not using Twig and not defining skeleton files as Twig templates has another problem...

... you can't easily override the skeleton files in your project (see #32 and lots of related issues in the SensioGeneratorBundle). Should we...

... add Twig as dependency and turn skeletons into templates?
... try to make the skeleton files overridable using the kernel file locator ?
... not support this feature explicitly?

yceruto:
This allows us to load skeleton templates relative to:
* <del>`%kernel.project_dir%/config/maker/skeleton/`<del>
* `%kernel.root_dir%/Resources/MakerBundle/skeleton/` (overriding dir convention)
* `Resources/skeleton/` (from MakerBundle)

Thus we would have a simple mechanism to override all or one template without create/override the whole command and vice versa, this allows us to reuse the MakerBundle skeletons in our custom commands by passing the relative template name.

Fixes #33 

What do you think?

### DIFF
--- a/src/Command/MakeCommandCommand.php
+++ b/src/Command/MakeCommandCommand.php
@@ -50,7 +50,7 @@ final class MakeCommandCommand extends AbstractCommand
     protected function getFiles(array $params): array
     {
         return [
-            __DIR__.'/../Resources/skeleton/command/Command.php.txt' => 'src/Command/'.$params['command_class_name'].'.php',
+            'command/Command.php.txt' => 'src/Command/'.$params['command_class_name'].'.php',
         ];
     }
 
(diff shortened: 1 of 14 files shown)


javiereguiluz:
At first sight I don't like the `config/maker/...` convention. We should be careful with the `config/` file and its `config/packages/...` convention. Moreover, the skeleton files are not config files ... they are resources or templates, etc.

yceruto:
Also this allows us to reuse the MakerBundle skeletons in our own commands.

yceruto:
@javiereguiluz What about `%kernel.root_dir%/Resources/MakerBundle/skeleton/` instead?

yceruto:
I made the changes about overriding dir convention, how does it look now?

javiereguiluz:
I still don't know what to do about this. But here is another issue that could be simple to solve if skeleton files were easy to override: https://github.com/symfony/maker-bundle/issues/58#issuecomment-347209656

yceruto:
Closing here because this probably should be achieved in a different way after #59

PR #52, opened 2017-11-24
Add NotBlank constraint to interactive arguments

yceruto:
**before**
not-blank-constraint

**after**
not-blank-constraint-ok


fabpot:
Thank you @yceruto.

PR #53, opened 2017-11-24
Add BrowserKit component as dependency to functional test commands

yceruto:
Same as other commands, the generated file does not work if all dependencies are not available.

### DIFF
--- a/src/Maker/MakeFunctionalTest.php
+++ b/src/Maker/MakeFunctionalTest.php
@@ -20,6 +20,7 @@ use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\BrowserKit\Client;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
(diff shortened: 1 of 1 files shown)


fabpot:
👎 for this as it does not work when using a globally installed PHPUnit.

yceruto:
Agree, I made it optional, but I expected some comment messages for those who don't use global installation. Wdyt?

fabpot:
Even if this is optional, I don't want to make people think that this is what they need to do.

jzawadzki:
on same note, I think we should add BrowserKit as dependency for functional test

yceruto:
Agree again, but I'm placed in the broken side, I don't have phpunit globally installed, I've generated the unit test file and it does not work, what can I do ... :/ Shouldn't something like "You might require install phpunit through composer or globally" help people to solve the broken state? Not as a red comment but as a grey note comment. Wdyt?

fabpot:
You can't hint at installing PHPUnit globally as you have no idea if that's the case or not. I don't want a warning each time I generate a unit test file if I have PHPUnit installed globally.

yceruto:
@fabpot fair enough, thank you for all feedback.

@jzawadzki thanks, that's true, I've changed the goal of the PR.

weaverryan:
Thank you @yceruto!

PR #54, opened 2017-11-24
Add travis config

ck-developer:


.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+        - $HOME/.composer/cache/files
+        - vendor
+        - .phpunit
+
+env:
+    global:
+        - SYMFONY_PHPUNIT_DIR=.phpunit
+
+install:
+    - composer --prefer-source install
```
stof: `--prefer-source` must be replaced by `--prefer-dist` (otherwise persisting the composer cache is useless)

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+    depth: 1
+
+php:
+  - 7.1
+  - 7.2
+
+cache:
+    directories:
+        - $HOME/.composer/cache/files
+        - vendor
```
stof: I would not cache the vendor folder. This can make things really hard to debug if one of the build is corrupting it. Relying on the download cache is enough when using dist installations (which are not rate-limited anymore).

.travis.yml
stof: you should also add `./vendor/bin/simple-phpunit install` in the install section, so that the PHPUnit installation gets done in `install` rather than when running tests. This makes the build output less confusing.

ck-developer:
Hello @stof thank you for review I updated PR

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+    directories:
+        - $HOME/.composer/cache/files
+        - .phpunit
+
+env:
+    global:
+        - SYMFONY_PHPUNIT_DIR=.phpunit
+
+install:
+    - composer --prefer-dist install
```
yceruto: shouldn't be `composer --prefer-dist update` instead?

fabpot:
Thank you @ck-developer.

PR #55, opened 2017-11-25
[RFC] Generate UnitTest from file path instead of class name?

yceruto:
Most of the time (almost always I think) running `make:unit-test <class-name>` would require to move the generated test file to the right directory and change its namespace according to the best practice:
> By convention, the `tests/` directory should replicate the directory of your bundle for unit tests. So, if you're testing a class in the `src/Util/` directory, put the test in the `tests/Util/` directory.

My proposal (with shell auto-completation <3):
```bash
./bin/console make:unit-test src/Security/Voter/FooVoter.php
```
This would generate the file in `tests/Security/Voter/FooVoterTest.php` with namespace `App\Tests\Security\Voter`.

For now, it's optional and still one can to pass the class name, but probably the proposed one should be the only way IMHO.

What do you think?

### DIFF
--- a/src/Command/MakeUnitTestCommand.php
+++ b/src/Command/MakeUnitTestCommand.php
@@ -13,6 +13,7 @@ namespace Symfony\Bundle\MakerBundle\Command;
 
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Generator;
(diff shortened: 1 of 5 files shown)


nicolas-grekas:
I like the shell auto-completion this provides!

yceruto:
Closing here because this probably should be achieved in a different way after #59

yceruto:
See revamped version https://github.com/symfony/maker-bundle/pull/89 :)

PR #59, opened 2017-11-28
Adding a new MakerInterface, instead of extending AbstractCommand

weaverryan:
Hi guys!

This diff looks big, but it makes no visible changes to the bundle. This only changes the way that custom maker commands are created: instead of extending from a base class and overriding methods, you'll implement the new `MakerInterface`. The reason behind this is simple: we need to tag a stable release and having a concrete interface gives us more flexibility going forward (i.e. we could add other optional interfaces for more functionality, deprecate this interface in favor of a new one, etc).

There are 2 powerful functional tests that pretty much cover everything. One actually tries all of the Maker classes to make sure they work (that existed before) and another tests that all Makers are wired correctly in the container.

The most important thing to look at is the new `MakerInterface`.

Cheers!


README.md
```diff
@@ -1,13 +1,29 @@
...
-tests and more so you can forget about the required boilerplate code. This
-bundle is an alternative to [SensioGeneratorBundle][1] for modern Symfony
-applications and requires using Symfony 3.4 or newer and [Symfony Flex][2].
+The MakerBundle is the fastest way to generate the most common code you'll
+need in a Symfony app: commands, controllers, form classes, event susbcribers
+and more! This bundle replaces [SensioGeneratorBundle][1] for modern Symfony
+applications and requires Symfony 3.4 or newer and [Symfony Flex][2].
 
-[Read the docs][3]
+[![Build Status](https://travis-ci.org/knpuniversity/oauth2-client-bundle.svg)](http://travis-ci.org/symfony/maker-bundle)
```
fabpot: Looks like something is wrong here :)

README.md
```diff
@@ -1,13 +1,29 @@
...
+The Symfony MakerBundle
+=======================
 
-Symfony Maker helps you creating empty commands, controllers, form classes,
-tests and more so you can forget about the required boilerplate code. This
-bundle is an alternative to [SensioGeneratorBundle][1] for modern Symfony
-applications and requires using Symfony 3.4 or newer and [Symfony Flex][2].
+The MakerBundle is the fastest way to generate the most common code you'll
+need in a Symfony app: commands, controllers, form classes, event susbcribers
+and more! This bundle replaces [SensioGeneratorBundle][1] for modern Symfony
```
fabpot: Right now, it does not replace the generator bundle as there is no CRUD generation, which is the main feature of the old bundle.

README.md
fabpot: But anyway, I would remove this build status that does not give any useful information for people reading the README.

src/EventRegistry.php
```diff
@@ -61,7 +64,7 @@ public function __construct(EventDispatcherInterface $eventDispatcher)
      */
     public function getAllActiveEvents(): array
     {
-        $activeEvents = [];
+        $activeEvents = array();
```
fabpot: I think using `[]` is fine. We just need to create a `.php_cs` file to tweak the default fabbot config regarding CS.

weaverryan:
Changes made! Though, fabbot is not respecting my new `.php_cs.dist` file... did I bork it somehow? It works locally with php-cs-fixer.

src/Resources/config/makers.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+        <services>
+            <defaults public="false" />
+
+            <service id="maker.maker.make_authenticator" class="Symfony\Bundle\MakerBundle\Maker\MakeAuthenticator">
```
javiereguiluz: Unimportant proposal: I'd love to use autowiring ... but if it's not possible, could we use the class as the service ID? ... if that's not possible either, could we rename the IDs to `maker.command.*` (the `maker.maker.*` format looks too redundant).

weaverryan: So.... this is done on purpose :)

* If third-party bundles use autowiring, then it will slow the dev experience for users a little bit. I'd like to make it a standard to *not* use autowiring for this reason.

* If we use the class names as service id's, they will show up in `debug:autowiring`. We need to tell the bundle world to not do this - only use class names or class aliases *if* it is something that makes sense to autowire.

* The `maker.command.*` is a little misleading... as these aren't actually commands, but are... "makers"... whatever that means ;).

src/MakerInterface.php
```diff
@@ -0,0 +1,90 @@
...
+    public function configureDependencies(DependencyBuilder $dependencies): void;
+
+    /**
+     * If necessary, you can use this method to interactively ask the user for input.
+     *
+     * @param InputInterface $input
+     * @param ConsoleStyle   $io
+     * @param Command        $command
+     */
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void;
```
stof: I found only 1 maker needing to perform custom interactions to fill arguments. So should this really be part of the interface implemented by all makers ?

PR #60, opened 2017-11-28
Php templates

weaverryan:
This depends on #59. So sorry for the messy diff until that is merged:

tl;dr; This renders the templates as PHP, instead of simple `{{ foo }}` find and replace. We do not want to make the templates super dynamic and fancy, but we're already running into a few spots where we're hitting this limitation (case: `make:controller` which has *two* possible templates, because the current replace is too limiting).

The only ugly part is that the `<?php` in the templates needs to be written as `//PHP_OPEN` to avoid syntax errors (we then replace this with `<?php` after including). Here is an example: https://github.com/symfony/maker-bundle/compare/master...weaverryan:php-templates?expand=1#diff-3ecc33834a0f1ec4f8e5516d71e4ad35

Btw, if you're wondering "Why not use Twig!?". We don't want to create a dependency on Twig to use the maker :).

Cheers!


src/Resources/skeleton/authenticator/Empty.tpl.php
```diff
@@ -9,7 +9,7 @@
 use Symfony\Component\Security\Core\User\UserProviderInterface;
 use Symfony\Component\Security\Guard\GuardAuthenticator;
 
-class {{ class_name }} extends GuardAuthenticator
+class <?php echo $class_name; ?> extends GuardAuthenticator
```
javiereguiluz: Just asking: would it be acceptable to replace this:

```php
<?php echo $var; ?>
```

by this?

```php
<?= $var ?>
```

The `<?=` syntax is always available since PHP 5.4 and it cannot be disabled in any way.

weaverryan: Hmm, I didn't know that. I thought that syntax was somehow deemed "horrible" by someone... but I can't remember. I'm fine with it - we'll see if anyone has a good reason against it.

javiereguiluz: Maybe it's horrible. Not sure. The `<?php echo ...` looked too long to me ... but maybe it's better to use it because it's the normal syntax everybody know and likes.

nicolas-grekas: I like the short syntax :) This is PHP, let's use the language's features ;)

src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Bundle\MakerBundle\MakerInterface;
+use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Reference;
+
+class MakeCommandRegistrationPass implements CompilerPassInterface
+{
+    const MAKER_TAG = 'maker.command';
```
piotrgradzinski: Shouldn't we use explicit `public const`?

weaverryan: Why not? :) I've done it over on #59

weaverryan:
This is rebased and ready. I'm still getting false fabbot failures on the `[]` array syntax...

src/FileManager.php
```diff
@@ -40,13 +40,15 @@ public function setIO(SymfonyStyle $io): void
...
     {
-        $keys = array_keys($parameters);
-        $values = array_values($parameters);
-        $placeholders = array_map(function ($name) {
-            return "{{ $name }}";
-        }, $keys);
+        ob_start();
+        extract($parameters);
+        include($templatePath);
+        $contents = ob_get_contents();
```
nicolas-grekas: ob_get_clean()

src/FileManager.php
```diff
@@ -40,13 +40,15 @@ public function setIO(SymfonyStyle $io): void
...
+        ob_end_clean();
 
-        return str_replace($placeholders, $values, file_get_contents($templatePath));
+        $contents = str_replace('//PHP_OPEN', '<?php', $contents);
```
nicolas-grekas: return str_replace...?

src/Resources/skeleton/authenticator/Empty.tpl.php
```diff
@@ -1,4 +1,4 @@
-<?php
+//PHP_OPEN
```
nicolas-grekas: `<?php ?>` instead?

weaverryan: But then it would be evaluated by the PHP processor, and would disappear. I guess we could *always* assume we need to prepend the entire file with `<?php`. Is it worth it? They're bother weird honestly...

weaverryan:
Changes made... except for the one change...

src/Resources/skeleton/authenticator/Empty.tpl.php
nicolas-grekas: > we could always assume we need to prepend the entire file

looks like a good idea to me

src/Resources/skeleton/authenticator/Empty.tpl.php
fabpot: I like the short syntax as well.

src/Resources/skeleton/authenticator/Empty.tpl.php
stof: always prepending is bad IMO: a generator may have a case for generating another type of file.

another option might be this (if it works):

```php
<?= '<?php' ?>
```

src/FileManager.php
```diff
@@ -40,13 +40,12 @@ public function setIO(SymfonyStyle $io): void
 
     public function parseTemplate(string $templatePath, array $parameters): string
     {
-        $keys = array_keys($parameters);
-        $values = array_values($parameters);
-        $placeholders = array_map(function ($name) {
-            return "{{ $name }}";
-        }, $keys);
+        ob_start();
+        extract($parameters);
```
stof: should use `EXTR_SKIP` instead of the default `EXTR_OVERWRITE`

src/Resources/skeleton/event/Subscriber.tpl.php
```diff
@@ -0,0 +1,20 @@
+//PHP_OPEN
+
+namespace App\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+<?php echo $eventUseStatement; ?>
+class <?php echo $subscriber_class_name; ?> implements EventSubscriberInterface
```
stof: should have an empty line between the use statement and the class

src/Resources/skeleton/authenticator/Empty.tpl.php
stof: I agree about using the short echo syntax

src/Resources/skeleton/authenticator/Empty.tpl.php
weaverryan: What about `//<?php`?

weaverryan:
Changes made! fabbot still dislikes me, but it's unrelated :)

stof:
@fabpot any idea why fabbot would not use the config file in the repo ?

src/MakerInterface.php
```diff
@@ -71,9 +71,13 @@ public function getParameters(InputInterface $input): array;
      * For example:
      *
      *    return array(
-     *        __DIR__.'/../Resources/skeleton/command/Command.php.txt' => 'src/Command/'.$params['command_class_name'].'.php',
+     *        __DIR__.'/../Resources/skeleton/command/Command.tpl.php' => 'src/Command/'.$params['command_class_name'].'.php',
      *    );
      *
+     * These files are parsed as PHP. To avoid syntax errors, replace the
+     * <?php tag in your templates with `//PHP_OPEN`. After evaluating the
```
nicolas-grekas: leftover

src/Resources/skeleton/authenticator/Empty.tpl.php
```diff
@@ -1,4 +1,5 @@
-<?php
+<?= '<?php' ?>
```
nicolas-grekas: `<?= "<?php\n" ?>`?

src/Resources/skeleton/event/Subscriber.tpl.php
```diff
@@ -0,0 +1,22 @@
+<?= '<?php' ?>
+
+
+namespace App\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+<?= $eventUseStatement ?>
+
```
stof: I think you need a second empty line here: The first EOL char will be swallowed by `?>`, so this would not create an empty line between the use statement and the class at the end.

weaverryan:
@stof nice catch! I've just added php-cs-fixer to the tests, so that we can verify the generated files follow coding standards (especially important because that empty line stuff is very easy to mess up).

Thanks!

PR #61, opened 2017-11-29
Add missing import

codedmonkey:
Add a missing import after the revamp of the Command class by @weaverryan


fabpot:
Thank you @codedmonkey.

PR #64, opened 2017-11-29
Replace demo page by twig file

nicolas-grekas:
Follow up of #46

Seriously, this will make demoing Symfony a no brainer. Wow effect. Newbies <3
And for others, that's still >50% likelly OK.

image


src/Resources/skeleton/controller/ControllerWithTwig.twig.php
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html.twig' %}
```
yceruto: If `base.html.twig` template does not exists then the generated code is broken, which is bad experience :)

yceruto: I mean, after removing the default base template provided by Twig-Bundle recipe.

javiereguiluz:
I'm 👎 for some reasons:

1) Maybe we haven't explained well the purpose/target of this bundle. It's NOT for newcomers, it's for Symfony professionals (newcomers and experts). We have the Symfony Demo app for newcomers and to show the great Symfony features.

2) The SymfonyMaker must strive to be useful for everybody. Specifically, this means that we must minimize the number of changes between the generated code and the code which is useful for you. Otherwise, we'll repeat the same errors of the AcmeDemoBundle.

3) Generating only the controller class (and pointing the `render()` method to an internal template of the bundle) it's great because you just need to change the `render()` value. Maybe you change the template name, maybe you delete that line entirely. But that's all.

4) If we generate the controller class and the Twig template, users will need to make more changes. Maybe the Twig template is not called as they ant, maybe it's not in the directory where they want, maybe they are not using a Twig template for this controller (think JSON responses), etc.

5) We can't extend from the `base.html.twig` template because there's no guarantee that the template exists. Then we'd need to create a stand-alone Twig template adding some HTML code, etc.

nicolas-grekas:
1. If something is useful to newcomers, it IS useful to advanced ones. This has been proven so many times. I don't understand the purpose of the bundle then. For pros means absolutely nothing to me.
2. I agree. We're talking about one file, which replaces a demo that serves not technical purpose. Why do we generated a doctrine repository following your logic?
3. Why does the generated controller call render then?
4. That's easy to resolve.

javiereguiluz:
I disagree with this:

> If something is useful to newcomers, it IS useful to advanced ones.

For example, adding long help notes in the source code of the Symfony Demo app is great for newcomers, but awful for experts.

Again, the confusion may come for not having explained well enough the target/purpose of this bundle. This is not the AcmeDemoBundle or the Symfony Demo app. This is a tool that every Symfony developer should use. It's not to demo/showcase things, it's to generate fully useful code.

In my opinion, the Doctrine repository class example is not the same. That repository class is generated with the right file name, at the right location, with the right contents and all non-trivial apps will use it.

nicolas-grekas:
> If something is useful to newcomers, it IS useful to advanced ones.
> Symfony Demo app is great for newcomers, but awful for experts.

I was talking about *features* of course, not doc. Accessibility for example, autowiring also at the beginning, etc. Many great technological features that are for "new comers" *are* useful to the "expert" majority. That's what I meant.

> This is not the AcmeDemoBundle or the Symfony Demo app

Of course, I got that since the beginning, that's plain clear to me!
I was talking about demoing yourself, or someone teaching Symfony to someone (like live demos we did with Fabien): if, in a very few commands, we're able to very quickly make a working app, this is WOW. And the maker-bundle is exactly about this to me: get fast to something working (and the demo-effect is then just a nice consequence.)

So yes: in my opinion, generating this twig file IS like doctrine repos: if you have twig, we can estimate you're going to use "render()", and we can generate a twig file for you. The fact that is eases a lot demoing Symfony to new comers *is* the proof to it IMHO.

src/Maker/MakeController.php
```diff
@@ -53,26 +54,42 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
 
     public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
     {
+        $this->projectDir = $command->getApplication()->getKernel()->getContainer()->getParameter('kernel.project_dir');
```
nicolas-grekas: Any better/canonical way to get the target dir? this currently makes the test suite fail, so needs to be resolve one way or another. Hint welcomed.

src/Maker/MakeController.php
```diff
@@ -53,26 +54,42 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
     public function getParameters(InputInterface $input): array
     {
         $controllerClassName = Str::asClassName($input->getArgument('controller-class'), 'Controller');
         Validator::validateClassName($controllerClassName);
 
+        $title = "Hello {$controllerClassName}!";
+        if (file_exists($this->projectDir.'/templates/base.html.twig')) {
+            $twigFirstLine = "{% extends 'base.html.twig' %}\n{% block title %}{$title}{% endblock %}";
+        } else {
+            $twigFirstLine = "<!DOCTYPE html>\n<title>{$title}</title>";
```
nicolas-grekas: This is valid HTML5

src/Resources/skeleton/controller/ControllerWithTwig.tpl.php
```diff
@@ -13,7 +13,6 @@ class <?= $controller_class_name ?> extends Controller
      */
     public function index()
     {
-        // replace this line with your own code!
-        return $this->render('@Maker/demoPage.html.twig', [ 'path' => str_replace($this->getParameter('kernel.project_dir').'/', '', __FILE__) ]);
+        return $this->render('<?= $twig_file ?>');
```
nicolas-grekas: not calling getParameter anymore, this gives back the possibility to extend AbstractController

src/Str.php
```diff
@@ -97,6 +99,14 @@ public static function asEventMethod(string $eventName): string
         return sprintf('on%s', self::asClassName($eventName));
     }
 
+    public static function asFilePath(string $value): string
+    {
+        $value = Container::underscore(trim($value));
```
nicolas-grekas: perfect fit to generate a path from a class

src/Str.php
```diff
@@ -97,6 +99,14 @@ public static function asEventMethod(string $eventName): string
...
+        $value = str_replace('\\', '/', $value);
```
nicolas-grekas: this anticipates the possibility to generate controllers in subnamespace
(which is forbidden right now, but doesn't have to be really)

src/Maker/MakeController.php
yceruto: `__construct(..., string $projectDir)`?

src/Maker/MakeController.php
```diff
@@ -53,26 +54,42 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+        $this->projectDir = $command->getApplication()->getKernel()->getContainer()->getParameter('kernel.project_dir');
     }
 
     public function getParameters(InputInterface $input): array
     {
         $controllerClassName = Str::asClassName($input->getArgument('controller-class'), 'Controller');
         Validator::validateClassName($controllerClassName);
 
+        $title = "Hello {$controllerClassName}!";
+        if (file_exists($this->projectDir.'/templates/base.html.twig')) {
```
yceruto: However, I think what you need here is the `%twig.default_path%` value, later should be `$this->templateDir` instead?

src/Maker/MakeController.php
```diff
@@ -53,26 +54,42 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+        $classFile = 'src/Controller/'.$params['controller_class_name'].'.php';
+
+        if (!$this->isTwigInstalled()) {
+            return [$dir.'Controller.tpl.php' => $classFile];
+        }
 
         return [
-            __DIR__.'/../Resources/skeleton/controller/'.$skeletonFile => 'src/Controller/'.$params['controller_class_name'].'.php',
+            $dir.'ControllerWithTwig.tpl.php' => $classFile,
+            $dir.'ControllerWithTwig.twig.php' => 'templates/'.$params['twig_file'],
```
yceruto: same here `$this->templateDir`.

src/Maker/MakeController.php
nicolas-grekas: We cannot use `twig.default_path` because the parameter is not always defined.
But anyway, just targeting the default dir structure is fine IMHO.

nicolas-grekas:
@yceruto comments addressed, thanks.

src/Maker/MakeController.php
yceruto: The bundle is intended to work with Symfony 3.4 or newer, right? and the `twig.default_path` parameter was introduced since 3.4, so I guess this should work?

Anyway at least we can avoid to hardcode the `templates/` directory with something like this?:
```
<argument type="expression">container.hasParameter('twig.default_path') ? parameter('twig.default_path') : parameter('kernel.project_dir') ~ '/app/Resources/views'</argument>
```

nicolas-grekas: It's not a matter of version: the param exists only if TwigBundle is.
I think this is not worth pursuing: `templates/` is what the default recipe provides.
There are already many hardcoded, eg the `src/` folder. This is legitimate for the purpose.
The day this statement will become false, someone could try a PR, fixing all hardcoded things at one. Dealing with the required complexity is not needed now, if anytime ever.
My 2cts:)

weaverryan: Sorry, I thought I commented on this line, but I apparently did not! I believe we should use `FileManager::fileExists()` - it's aware of the root directory and I want to keep all the path stuff going through this

src/Maker/MakeController.php
```diff
@@ -60,19 +62,34 @@ public function getParameters(InputInterface $input): array
         $controllerClassName = Str::asClassName($input->getArgument('controller-class'), 'Controller');
         Validator::validateClassName($controllerClassName);
 
+        $title = "Hello {$controllerClassName}!";
+        if (file_exists($this->projectDir.'/templates/base.html.twig')) {
+            $twigFirstLine = "{% extends 'base.html.twig' %}\n{% block title %}{$title}{% endblock %}";
```
weaverryan: 2 empty lines - so that we get:

```
{% extends 'base.html.twig' %}

{% block title %}TITLE{% endblock %}
```

src/Resources/skeleton/controller/ControllerWithTwig.tpl.php
weaverryan: What do you think about passing some variable in as the second arg (to show how that works)?

src/Maker/MakeController.php
```diff
@@ -60,19 +62,34 @@ public function getParameters(InputInterface $input): array
...
             'route_path' => Str::asRoutePath(str_replace('Controller', '', $controllerClassName)),
             'route_name' => Str::asRouteName(str_replace('Controller', '', $controllerClassName)),
+            'twig_file' => Str::asFilePath(str_replace('Controller', '', $controllerClassName)).'.html.twig',
+            'twig_first_line' => $twigFirstLine,
         ];
     }
 
     public function getFiles(array $params): array
     {
-        $skeletonFile = $this->isTwigInstalled() ? 'ControllerWithTwig.tpl.php' : 'Controller.tpl.php';
```
weaverryan: Now that we have php support in the templates, could we unify these 2 files into 1? Unless they are VERY different, that seems less error-prone and the files will stay in sync with each other.

src/Maker/MakeController.php
```diff
@@ -60,19 +62,34 @@ public function getParameters(InputInterface $input): array
...
+        $dir = __DIR__.'/../Resources/skeleton/controller/';
+        $classFile = 'src/Controller/'.$params['controller_class_name'].'.php';
+
+        if (!$this->isTwigInstalled()) {
+            return [$dir.'Controller.tpl.php' => $classFile];
+        }
```
weaverryan: I'm being annoying, but I'd rather see this reversed: the array with one entry only created first (and set to some `$paths` variable. Then an if statement to append the `.twig.php` file `if ($this->isTwigInstalled())`

tests/Command/FunctionalTest.php
```diff
@@ -92,7 +92,7 @@ public function getCommandTests()
             ->method('getRouteCollection')
             ->willReturn(new RouteCollection());
         yield 'controller' => array(
-            new MakeController($router),
+            new MakeController($router, sys_get_temp_dir()),
```
weaverryan: you should be able to remove this if you start using `FileManager::fileExists()` I believe

yceruto: I afraid we don't have direct access to the `FileManager` from _makers_ right now. However, still injecting it the problem about the path will persist here, because this is a test provider function and `$this->targetDir` is null at this moment.

weaverryan: Bah, that's annoying. That's not a good reason to write the code wrong though, so we need to think of some clever way to fix the tests. Probably we should make `FileManager` a 3rd argument to `testCommands` and make each provider pass in its own `FileManager` (we could add a nice `createFileManager()` method to make that easy, it would do all the `sys_get_temp_dir()` stuff)

nicolas-grekas: I left things as is for now, maybe resolve this in another PR?

nicolas-grekas:
Generated controllers now extend AbstractController, fixes #88

src/Maker/MakeController.php
```diff
@@ -60,20 +62,34 @@ public function getParameters(InputInterface $input): array
...
 
+        if (file_exists($this->projectDir.'/templates/base.html.twig')) {
+            $twigFirstLine = "{% extends 'base.html.twig' %}\n\n{% block title %}Hello {{ controller_name }}!{% endblock %}\n";
+        } else {
+            $twigFirstLine = "<!DOCTYPE html>\n\n<title>Hello {{ controller_name }}!</title>\n";
+        }
+
         return [
             'controller_class_name' => $controllerClassName,
+            'controller_class_file' => 'src/Controller/'.str_replace('\\', '/', $params['controller_class_name']).'.php',
```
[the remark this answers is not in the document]
nicolas-grekas: fixes, thanks

src/Resources/skeleton/controller/twig_template.tpl.php
```diff
@@ -0,0 +1,26 @@
...
+<title>Hello {{ controller_name }}!</title>
+<?php endif; ?>
+
+{% block body %}
+<style>
+    #example-wrapper { margin: 1em auto; max-width: 800px; width: 95%; font: 18px/1.5 sans-serif; }
+    #example-wrapper code { background: #F5F5F5; padding: 2px 6px; }
+</style>
+
+<div id="example-wrapper">
```
nicolas-grekas: class instead of id would be better practice AFAIK :)

PR #65, opened 2017-11-29
Be compatible with PHP 7.0

### LINKED ISSUE #63: Make the bundle 7.0 compatible
There are too many people that have just 7.0 installed and we would be way more welcoming to them and every newcomers if things worked out of the box here with eg Ubuntu LTS.

nicolas-grekas:
Fixes #63

Let's not be elitist. We want adoption. We need adoption.


src/MakerInterface.php
```diff
@@ -53,7 +53,7 @@ public function configureDependencies(DependencyBuilder $dependencies): void;
      * @param ConsoleStyle   $io
      * @param Command        $command
      */
-    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void;
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command);
```
stof: This is too late. 1.0 is already released, and removing the return type of an interface is a BC break

nicolas-grekas: It's too late, unless we don't want it to be too late. We're not going to fork the bundle, neither to tag v2. Will we break anything if we bump to v1.0.1? Hell no. Let's be pragmatic.

stof: shipping a BC break in a patch release in an official SF bundle is not good: people would think it is OK to do such change in their own release, and this kind of defeat the BC promise (especially given that our BC documentation on http://symfony.com/doc/current/contributing/code/bc.html does not yet describe all changes regarding PHP 7 return types).

The *only* reason such BC break would have little impact here is because this interface was merged and released today, and so nobody had enough time to start using it. But this does not mean it is not a BC break.

stof: So answer about "will we break something" is yes: we will provide an official example that it is OK to do such change without considering it as a BC break, which impacts the usefulness of the BC rules for others (while we want them to follow them as well)

nicolas-grekas: OK, all things considered, I'm still :+1: (a HUGE ONE) to do the mode.
Now calling to others.

javiereguiluz:
Thanks @nicolas-grekas.

dunglas:
Too late, but 👍 too.

PR #66, opened 2017-11-29
Remove minimum-stability from Composer config

javiereguiluz:
Reported here: https://twitter.com/electricmaxxx/status/935944477998534658

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,6 @@
             "homepage": "https://symfony.com/contributors"
         }
     ],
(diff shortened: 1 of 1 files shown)


stof:
I agree about this change, but not actually for the reason given in the tweet

stof:
hmm, actually, as it requires Symfony 3.4+, such change has to wait until tomorrow (otherwise the CI cannot install dependencies when allowing only stable ones)

PR #75, opened 2017-11-30
[Voter] Add entity class only if it exists

yceruto:
This is not ideal yet, but at least will avoid generating a broken voter class.


yceruto:
This is considered as a bugfix?

yceruto:
This is considered as a bugfix?

javiereguiluz:
Quick question: if we don't do this change and `App\Entity\Blog` doesn't exist, does PHP fail when compiling the voter class or when using it? If it's when using it, I don't care about this because user MUST change the voter class before using it, so changing the entity class is just another thing they must do.

yceruto:
> if we don't do this change and App\Entity\Blog doesn't exist, does PHP fail when compiling the voter class or when using it?

No. so closing this.

PR #81, opened 2017-12-02
Fix extends on ControllerWithTwig template

welcoMattic:
`Symfony\Bundle\FrameworkBundle\Controller\AbstractController` doesn't implement `getParameter` method. I replaced it by `Symfony\Bundle\FrameworkBundle\Controller\Controller`

# How to test

- `bin/console make:controller HomeController`
- Open `/home`

### DIFF
--- a/src/Resources/skeleton/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/controller/Controller.tpl.php
@@ -3,15 +3,15 @@
 namespace App\Controller;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
-use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Response;
 
-class <?= $controller_class_name ?> extends AbstractController
+class <?= $controller_class_name ?> extends Controller
 {
     /**
      * @Route("<?= $route_path ?>", name="<?= $route_name ?>")
      */
-    public function index()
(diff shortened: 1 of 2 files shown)


weaverryan:
Also fixed in #82. Are the semi-colons needed? I would like to also update the other controller skeleton so that they are both consistent. You PR was first, so if you can update, we’ll merge your’s :)

welcoMattic:
Hi @weaverryan,

The PR is up to date ;)

Semi-colons are required by fabbot.io which is not happy on my first push for coding style issues. So, I fixed it.

javiereguiluz:
I'd say we should ignore fabbot.io about semicolons. Not adding the semicolon is find in PHP and IMO it looks much better.

About the Controller vs AbstractController debate, it's clear we have a problem. In theory, AbstractController is "better" and we want to promote it ... but in practice some users don't like it and it's been proposed to keep using Controller. My opinion: either we promote AbstractController for real ... or we should remove it.

welcoMattic:
Hi @javiereguiluz, I removed semi-colons. 

For the debate about AbstractController vs Controller, let me know if I have to switch again. IMHO, we should promote AbstractController. And may be move `getParameter` in AbstractController if there is no side effects.

weaverryan:
Thank you @welcoMattic!

We should discuss the `AbstractController` in a separate issue. I agree that we should promote it, the question is how *soon* we should promote it, has people are still accustomed to having access to the container and there is not real disadvantage to `Controller`. Javier and I discussed this a few weeks ago, and for the documentation, the plan is to start showing `AbstractController` for 4.1.

Issue #67, opened 2017-11-29
[BUG][Controller] getParameter is not available


sescandell:
Hi,

When we generate a controller from the `make:controller` command and Twig is available, the template generated is invalid: it uses the `getParameter` method which is not available on a `AbstractController`.

yceruto:
related to https://github.com/symfony/maker-bundle/pull/64

javiereguiluz:
Closing as fixed because we removed the `getParameter()` call in #64.

PR #82, opened 2017-12-02
Fixed extended class for both controller templates

vertisan:
It's little fix for issue #67 for both `make:controller` templates


weaverryan:
Closed - I merged in #81, which just barely was here first :) 

Thanks!

PR #85, opened 2017-12-04
Add maker command for form type extensions

### LINKED ISSUE #78: [RFC] Add a make:form:type-extension command
Hi guys!

This maker command would generates a new (form) type extension class under `src/Form/Extension/` and the argument is the extended type class (the class name is derived from it).

What do you think?

yceruto:
I've two versions of this command:
1. This asks only for the class name of the form type extension and the extended type is always the `FormType::class`. Then, the user is responsible for changing it if necessary (with IDE autocompletion this task is very fast).
2. (The current one) Ask also for the FQCN of the extended type + class exists validation, but it could be a boring task considering that most of the time this form type comes from a vendor with a long namespace. To improve it, I thought about injecting all available form types and allowing the input of the class name only, something similar to what was done with the command `debug:form <FormType>` argument.

form-type-extension

closes #78

### DIFF
--- a/src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
+++ b/src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
@@ -18,6 +18,7 @@ use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\Form\Extension\Core\CoreExtension;
 
 class MakeCommandRegistrationPass implements CompilerPassInterface
 {
@@ -25,6 +26,8 @@ class MakeCommandRegistrationPass implements CompilerPassInterface
 
     public function process(ContainerBuilder $container)
     {
+        $this->processFormTypeExtensionMaker($container);
+
         foreach ($container->findTaggedServiceIds(self::MAKER_TAG) as $id => $tags) {
             $def = $container->getDefinition($id);
             $class = $container->getParameterBag()->resolveValue($def->getClass());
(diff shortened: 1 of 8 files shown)


yceruto:
Anyone there!? :) very low review activity on this amazing new bundle by the way :speedboat: I'd like to put 2th gear

yceruto:
@javiereguiluz thanks for the minor reword ;) it look better now.

weaverryan:
Hey @yceruto!

Sorry for the delay - I wanted to get the new functional test system in place before working on any additional makers.

Imo, I *would* prefer option (2): where we allow the user to type in the extended type, but help them a bit (because yea, nobody wants to type the long class name, especially since `\` needs to be doubled to `\\`). We do something similar in `MakeSubscriber`: we list all the events, then allow them to free-type, but with auto-completion. I agree that we should only make the user type the *short* version of the class name (e.g. `FormType`) (if there is a conflict, I guess we'll need to somehow use a less-shortened version... but that's an edge case).

Could you update the maker for these changes? Also, the `FunctionalTest` class has changed and there has been a slight change in `MakerInterface` after #97. 

> very low review activity on this amazing new bundle by the way 🚤 I'd like to put 2th gear

I agree :D. I think it's because I've been busy with some "boring" low-level tasks, but those are done now!

yceruto:
Hi! happy new year to everyone :tada: 

@weaverryan yeah, i'll try to do it this weekend, cheers.

yceruto:
Latest changes:
* Added extended type argument.
* Show autocomplete values, populated with all "available" form types (core types + registered as services).
* Show ambiguous validation & choice question if there is more than one FQCN with the same CN.
* Added class exists validation to `Validator`.

Ready for review!

yceruto:
Updated description with preview!

yceruto:
this is ready from my side... ;)

yceruto:
I'm closing here because I can't fix this after rebase on master.

weaverryan:
@yceruto sorry about that - I should have communicated better. I wanted to get a few internal updates done (which are now done) before merging this. If you can’t rebate, I understand. But both of your PR’s are good imo :)

PR #89, opened 2017-12-06
(make:unit-test) Generate test class from file path

yceruto:
> By convention, the `tests/` directory should replicate the directory of your bundle for unit tests. So, if you're testing a class in the `src/Util/` directory, put the test in the `tests/Util/` directory.

Following this convention, I'd like to offer another way to generate unit tests to accomplish this statement.

I think by passing the source file path with the benefits of shell autocompletion, we can easily derive the test path, the namespace and the final class. e.g:
```bash
$ bin/console make:unit-test src/Security/Voter/FooVoter.php
```
Generating this:
```php
// tests/Security/Voter/FooVoterTest.php
namespace App\Tests\Security\Voter;

// ...
```
Wdyt?

### DIFF
--- a/src/Maker/MakeUnitTest.php
+++ b/src/Maker/MakeUnitTest.php
@@ -27,6 +27,13 @@ use Symfony\Component\Console\Input\InputInterface;
  */
 final class MakeUnitTest implements MakerInterface
 {
+    private $projectDir;
+
+    public function __construct(string $projectDir)
+    {
+        $this->projectDir = $projectDir;
+    }
+
     public static function getCommandName(): string
     {
         return 'make:unit-test';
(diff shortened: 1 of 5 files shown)


javiereguiluz:
Just an idea: could we make this command smart? In practice:

1.) You type: `bin/console make:unit-test FooVoter` instead of `bin/console make:unit-test src/Security/Voter/FooVoter.php`

2.) This command scans the PHP files defined in `src/` and assumes that 1 PHP file == 1 PHP class with the same name (FooVoter class is stored in FooVoter.php file) and:

2.1) If it finds just 1 class named like that, it uses it;

2.2) If it finds several classes named like that, it displays them and you pick the right one:
  |> Pick the PHP class you are creating the test for:
  |> 1) src/Security/Voter/FooVoter.php
  |> 2) src/Admin/Security/Voter/FooVoter.php
  |> 3) src/Security/Legacy/Admin/Voter/FooVoter.php

2.3) If it finds lots and lots of classes, don't display the list and just generate the test class as it's done now.

2.4) If no class is found, just generate the test class as it's done now.

So:

1) Do you like this idea?

2) If you like it:
2.1) It's technically viable? Can we scan the src/ dir fast enough even for large projects? Would it work on Windows, Linux, etc.?
2.2) Any other technical issue you can imagine?

yceruto:
Hi @javiereguiluz, thanks for the reply!

> Do you like this idea?

Well, the idea is not bad, but it implies a complexity + interaction question that I don't know if it will be worth it.

I'm using this maker actually in combination with PhpStorm (External Tool + keyboard shortcut) and the DX is really amazing:
tool
just one step, triggering the shortcut on the source file and my unit-test is ready in the right dir thanks to the relative path option and `$FilePathRelativeToProjectRoot$` macro. I wouldn't like type something or do something to achieve it. I hope share it with everyone if this PR is merged.

Cheers.

stof:
@yceruto for the usage in the IDE, things would still work fine IMO:

- for the interaction part, just add `-n` in your command to force running in non-interactive mode (the auto-detection might already handle it btw, depending of how PhpStorm calls the command)
- as you pass a path to an existing file already, there would be a single match (and this can be done fast by using `is_file` before the scanning).

This does not forbid making the command better for usage by humans.

yceruto:
I'm closing here because I can't fix this, again.

PR #91, opened 2017-12-07
If value fails validation, ask again!

### LINKED ISSUE #90: If value fails validation, ask again!
To repeat:

* `bin/console make:entity`
* Just hit enter when asked the question

screen shot 2017-12-07 at 2 47 14 pm

The validation is correct. But why not give the user another try? We all deserve second chances! :)

yceruto:
@weaverryan Look like throwing a `Symfony\Component\Console\Exception\RuntimeException` into validator function breaks the [loop](https://github.com/symfony/symfony/blob/784186b2498303343a684a6a6f71eca1f7fa2bf2/src/Symfony/Component/Console/Helper/QuestionHelper.php#L364-L365).

But still we need a nice error message without line trace, so implementing `ExceptionInterface` should be enough.

Fix #90.

### DIFF
--- a/src/Exception/RuntimeCommandException.php
+++ b/src/Exception/RuntimeCommandException.php
@@ -11,13 +11,13 @@
 
 namespace Symfony\Bundle\MakerBundle\Exception;
 
-use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Exception\ExceptionInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


yceruto:
@weaverryan could you check (please) if this solve the issue?

weaverryan:
Awesome job! Thank you @yceruto - and sorry for the slow merge :)

PR #94, opened 2017-12-12
Ability to use sub namespace for entity

pink6440:
I don't get what i'm missing with the coding convention ...if someone can explain me ...

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -46,14 +46,33 @@ final class MakeEntity implements MakerInterface
     {
     }
 
-    public function getParameters(InputInterface $input): array
+   public function getParameters(InputInterface $input): array
     {
-        $entityClassName = Str::asClassName($input->getArgument('entity-class'));
+
+        $classname = $input->getArgument('entity-class');
+
+        $parts = explode('\\',$classname);
+        $short_classname = $parts[count($parts)-1] ;
+        $namespace = implode('\\',array_slice($parts,0,-1));
+        $classname = $short_classname ;
+
+        $path = str_replace('\\','/',$namespace);
+
+        if ($namespace) {
+
+            $path = $path.'/' ;
+        }
+
+
+
+        $entityClassName = Str::asClassName($classname);
         Validator::validateClassName($entityClassName);
         $entityAlias = strtolower($entityClassName[0]);
         $repositoryClassName = Str::addSuffix($entityClassName, 'Repository');
 
         return [
+            'path' => $path,
+            'entity_namespace' => $namespace,
             'entity_class_name' => $entityClassName,
             'entity_alias' => $entityAlias,
             'repository_class_name' => $repositoryClassName,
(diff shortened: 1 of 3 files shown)


javiereguiluz:
Some context: someone suggested this in the Symfony Slack. I told him that I was going to think about this. I was going to submit an issue so we could discuss it, instead of sending a pull request. We always prefer issues over pull requests.

---

Why an issue? First to see if we approve or reject this feature. Second, to design it to work as greatly as we can.

For example:

1) Usability: people always use the wrong slash for namespaces ( `\` and `/`) ... a good bundle wouldn't say anything about this to the user and it'd accept everything, but silently change the wrong slash behind the scenes to use the correct slash.

2) Flexibility: should we allow to provide partial namespaces (`Something\User`) or full namespaces (`App\Entity\Something\User`) or both? I'd say both and silently remove the `App\Entity\` part if included. And what if the user provides a FQCN outside of App\Entity ? We don't care and always assume FQCN are relative to App\Entity\ ... that's one of the simplifications we can make in this bundle to keep things simple.

3) Code quality: we could simplify the templates if we store the full real FQCN in the  `entity_namespace` variable to avoid doing hacks and concatenations in the template, etc.

4) Bugs and problems: creating directories recursively is always problematic to avoid creating existing dirs, make it work on all OS, avoid race conditions, etc.

As you can see, even a small feature needs a lot of thinking to do it great. That's why I always prefer issues over pull requests. But let's not close this PR now that you have created. Let's just stop working on it until the feature is approved. Thanks.

pink6440:
Thanks for your time and explanations !

I'm more than agree with all the remarks.

weaverryan:
I think I'm +1 for this, as it doesn't add complexity for the *user*. Actually, it removes some complexity. Most users will just enter `FooBar` as their class name. Currently (I believe), if they type `FooBar\\Baz`, their generated code will have an error (or maybe we show a validator error?).

Anyways, if the user wants to type a sub-namespace, I suppose it's fine. Of course, it's kind of annoying, because they'll need the double `\\`, but it's really more for expert users.

zorn-v:
>> creating directories recursively is always problematic
RLY ? `mkdir -p` or `mkdir('problematic/dir/that/does/not/exists', 0755, true)`
With the rest I agree.

src/Maker/MakeEntity.php
```diff
@@ -48,12 +48,27 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+        $parts = explode('\\',$classname);
+        $short_classname = $parts[count($parts)-1] ;
+        $namespace = implode('\\',array_slice($parts,0,-1));
+        $classname = $short_classname ;
+
+        $path = str_replace('\\','/',$namespace);
+
+        if ($namespace) {
+            $path = $path.'/' ;
+        }
```
weaverryan: This should somehow be extracted into an external class, as we'll probably want to re-use this other places too.

enleur:
image

iSDP:
I'm rooting for this to be added ASAP, manually moving files around is a lot of added workload

weaverryan:
It’s high on my list - I’ll finish the updated make:entity this week. Then we should add this ability to all maker commands :)

iSDP:
@weaverryan you're the best 😁

weaverryan:
Thanks for starting this @pink6440! I've included your work and finished it at #120.

PR #95, opened 2017-12-13
make:doctrine:entities command

zorn-v:
It is port for `doctrine:generate:entities` command from DoctrineBundle for flex based apps.
Namespaced with "doctrine" intentionally to not conflict with "entity" so you don't have to write full command like in SensioGeneratorBundle

May be not so elegant, but this command really needed

### DIFF
--- a/src/Command/GenerateEntitiesCommand.php
+++ b/src/Command/GenerateEntitiesCommand.php
@@ -21,7 +21,7 @@ class GenerateEntitiesCommand extends ContainerAwareCommand
             ->setDescription('Generates entity classes and method stubs from your mapping information for flex based project')
             ->addArgument('name', InputArgument::OPTIONAL, 'An entity name')
             ->addOption('path', null, InputOption::VALUE_REQUIRED, 'The path where to generate entities when it cannot be guessed')
-            ->addOption('no-backup', null, InputOption::VALUE_NONE, 'Do not backup existing entities files.')
+            ->addOption('backup', null, InputOption::VALUE_NONE, 'Backup existing entities files.')
         ;
     }
 
@@ -46,7 +46,7 @@ class GenerateEntitiesCommand extends ContainerAwareCommand
 
         $generator = $this->getEntityGenerator();
 
-        $backupExisting = !$input->getOption('no-backup');
+        $backupExisting = $input->getOption('backup');
         $generator->setBackupExisting($backupExisting);
 
         $repoGenerator = new EntityRepositoryGenerator();
--- a/src/Command/GenerateEntitiesCommand.php
+++ b/src/Command/GenerateEntitiesCommand.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace Symfony\Bundle\MakerBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Doctrine\ORM\Tools\EntityRepositoryGenerator;
+use Symfony\Bundle\MakerBundle\Doctrine\EntityGenerator;
+use Symfony\Bundle\MakerBundle\Doctrine\DisconnectedMetadataFactory;
+
+class GenerateEntitiesCommand extends ContainerAwareCommand
+{
+    protected static $defaultName = 'make:doctrine:entities';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Generates entity classes and method stubs from your mapping information for flex based project')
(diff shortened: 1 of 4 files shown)


yceruto:
I guess this should be done by using `MakerInterface`, right?

zorn-v:
Actually this not "generic" maker command. It just need to execute. I use it for generating getters/setters etc. after adding some fields/relations. Consider it as "patch for legacy" :smile: 
Maybe remove `--no-backup` option BTW ? I never revert from backup after this command yet. Project must be under version control anyway.

Majkl578:
Please let's not add new commands based on deprecated tools (doctrine/doctrine2#6869). EntityGenerator encourages very bad practices (symfony/symfony-docs#8893) and makes the model anemic and mostly useless. That's why it should die in peace, not be used even more. 😞

Nightbr:
Like zoonr-v said, we need to consider this as "patch for legacy". I think this is a useful feature with no alternative right now. For example, if you use ApiPlatform, you need to generate all setter/getter to expose your field. I think we should merge this and add a warning for user. And when doctrine will remove this tool, we will have find a better solution to handle this problem :)

codedmonkey:
I'm personally 👎 on this one. While I very much agree that creating getters and setters is annoying and should be easier, EntityGenerator didn't solve that problem for me. Are you using a proper IDE? PhpStorm has built-in functionality to generate such code, and it's a pretty customizable too.

Nightbr:
I think you never use this command... I'm using Atom with getter/setter generator plugin and sometimes phpstorm but it's very different. This command generate proper getter/setter and also add and remove functions for relations (ManyToOne, OneToMany and ManyToMany) and your IDE can't do this... 

I don't understand why we are reluctant to add feature which are usefull for developers and propose no suitable alternative...

codedmonkey:
I have used the command, and again, I agree it's very useful to have but it wasn't the solution for me. 

I'm not a member of the Symfony team, but for now, your proposal sounds outside the scope of this bundle, we don't even have a proper interactive `make:entity` command yet. For now the best solution would be to use the code of your pull request as a separate bundle, it sounds like you're not the only one who wants this.

Nightbr:
agreed about interactive `make:entity`, sure we need this before all 👍  

But the framework have to resolve developer problems. And what is will create here is developer will turn back to the framework and create their own solution... Pretty sad that we can find a consensus here...

IMO, I think I will create my DevBundle and integrate what is missing... 
Don't miss understand my comment, **I really love the way Symfony Flex reorganises the framework** and the structure we can use. **I already find my way to create really organized project and loved it!** I understand that Flex is young and it will be improve but we need to conduct developer in the transition.

weaverryan:
Hi guys!

My plan is to first, enhance `make:entity` so that it is a bit more like `doctrine:generate:entity`... but better ;). Inside that command, we will generate getters & setters for the fields we add (if the user wants this). 

I think that covers *most* cases... but I think there are people who will still add new fields by hand, but want a generator to add getters & setters. That's the odd case: using the generator for only *half* of the work. I'm not sure about this yet... or if this will even be a use-case once the much-better `make:entity` (where I plan to also allow you to add fields to an existing entity) is finished.

codedmonkey:
Personally I never liked the idea of executing generate:entities and it generating all my class files. I’m in favor of make:entity working only on single classes. if the class already exists, append the new fields. And a separate option to disable the field generator, only generate the relevant methods in that case.

zorn-v:
@weaverryan totally agree
I urgently needed this command and I wrote this ugly "patch". Then I thought to share it. 
In particular to raise the issue of generating the entities.

pink6440:
@codedmonkey i think generate single entity at a time and all entities are both usefull ; especially at the beginning of the project when you just had wrote your 40+ entities mappings and you have to generate the getters/setters.

codedmonkey:
Yes I understand. I know the intention of  this bundle is to get rid of quirky namespaces, but the most semantic command names if we end up with 2 commands imo would be `make:orm-entity` and `make:orm-mappings`.

deivid11:
An enchaced make:entity command with getters and setters generator is fine when you add a new entity from scratch, the pros of doctrine:generate:entities is that you can add more fields later by hand and then generate getters and setters. IDEs are not smart enough yet to generate methods for relationships (add, remove) doing it by hand is prone to errors

weaverryan:
@deivid11 the idea of the enhanced `make:entity` is that it would ALSO allow you to add more fields to an existing entity. I also plan to have it handle relationships correctly (yep, adders, removers, synchronizing inverse/owning side of relation).

But yea, there still may be a use-case for people that prefer to add the fields by hand, then run one command to have all the getter/setter/adder/remover stuff generated automatically. We'll have to finish `make:entity` first and then see (at that point, it shouldn't be too much work to do it).

I'm going to close this issue, as we definitely won't use this approach. I'm working to get the `make:entity` stuff done as quickly as possible - thanks everyone for your input and patience :).

PR #96, opened 2017-12-13
add --dev option with composer

mikysan:
I suggest to add --dev option to the composer require command in the docs because the maker bundle shouldn't be needed in production environment, in fact flex already registers it automatically to be used only in dev...


weaverryan:
Definitely! Thanks!

PR #97, opened 2017-12-14
New make:migration command

weaverryan:
This replaces #10.

There are 2 cases:

(A) When there ARE schema changes:

screen shot 2017-12-29 at 11 07 47 pm

(B) When there are NOT schema changes:

screen shot 2017-12-29 at 11 07 14 pm

This includes a BC break on `MakerInterface`, but it will only affect people who have created custom makers (and they'll get a clear error).

### DIFF
--- a/src/Command/MakerCommand.php
+++ b/src/Command/MakerCommand.php
@@ -11,13 +11,16 @@
 
 namespace Symfony\Bundle\MakerBundle\Command;
 
+use Symfony\Bundle\MakerBundle\ApplicationAwareMakerInterface;
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
 use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\ExtraGenerationMakerInterface;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\MakerInterface;
 use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -92,13 +95,24 @@ final class MakerCommand extends Command
         $params = $this->maker->getParameters($input);
         $this->generator->generate($params, $this->maker->getFiles($params));
 
-        $this->io->newLine();
-        $this->io->writeln(' <bg=green;fg=white>          </>');
-        $this->io->writeln(' <bg=green;fg=white> Success! </>');
-        $this->io->writeln(' <bg=green;fg=white>          </>');
-        $this->io->newLine();
+        if ($this->maker instanceof ExtraGenerationMakerInterface) {
+            $this->maker->afterGenerate($this->io, $params);
+        }
 
-        $this->maker->writeNextStepsMessage($params, $this->io);
+        $this->maker->writeSuccessMessage($params, $this->io);
+    }
+
+    public function setApplication(Application $application = null)
+    {
+        parent::setApplication($application);
+
+        if ($this->maker instanceof ApplicationAwareMakerInterface) {
+            if (null === $application) {
+                throw new \RuntimeException('Application cannot be null.');
+            }
+
+            $this->maker->setApplication($application);
+        }
     }
 
     /**
--- a/src/ExtraGenerationMakerInterface.php
+++ b/src/ExtraGenerationMakerInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle;
+
+/**
+ * Interface to do extra code generation in a maker.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+interface ExtraGenerationMakerInterface
+{
+    /**
+     * Called after normal code generation.
+     */
+    public function afterGenerate(ConsoleStyle $io, array $params);
+}
--- a/src/Maker/MakeMigration.php
+++ b/src/Maker/MakeMigration.php
@@ -1,0 +1,142 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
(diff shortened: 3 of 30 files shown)


src/Command/MakerCommand.php
```diff
@@ -79,7 +82,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
                 continue;
             }
 
-            $value = $this->io->ask($argument->getDescription(), $argument->getDefault(), [Validator::class, 'notBlank']);
+            $value = $this->io->ask($argument->getDescription(), $argument->getDefault(), array(Validator::class, 'notBlank'));
```
javiereguiluz: Please, revert this change. We don't use the legacy `array()` notation in this bundle.

src/ExtraGenerationMakerInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * Interface to do extra code generation in a maker.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+interface ExtraGenerationMakerInterface
+{
+    /**
+     * Called after normal code generation.
+     *
+     * @param ConsoleStyle $io
```
javiereguiluz: The PHPdoc is not needed because we use type-hints.

src/Maker/MakeMigration.php
```diff
@@ -0,0 +1,102 @@
...
+    public function setApplication(Application $application)
+    {
+        $this->application = $application;
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConf)
+    {
+        $command
+            ->setDescription('Creates a new migration based on database changes.')
+            ->addOption('db', null, InputOption::VALUE_REQUIRED, 'The database connection to use for this command.')
```
javiereguiluz: I'd remove *"...to use for this command"* in all descriptions because it's redundant. Also, I'd like better descriptions. For example, this doesn't help me -> `db = The database connection.` I can't provide a database connection to a console command. Is this the name of the database connection? The name of the Doctrine database connection? etc.

javiereguiluz: Should all these options define default values for the common case where there's only 1 database, 1 em, etc.?

src/Maker/MakeMigration.php
```diff
@@ -0,0 +1,102 @@
...
+                'db' => $input->getOption('db'),
+                'em' => $input->getOption('em'),
+                'shard' => $input->getOption('shard'),
+            ),
+        );
+    }
+
+    public function getFiles(array $params): array
+    {
+        return array();
```
javiereguiluz: More `array()` notation to replace by `[]` Thanks!

src/Maker/MakeMigration.php
```diff
@@ -0,0 +1,102 @@
...
+        $options['command'] = 'doctrine:migrations:diff';
+        $generateMigrationCommand = $this->application->find('doctrine:migrations:diff');
+        $generateMigrationCommand->run(new ArgvInput($options), $io);
+    }
+
+    public function writeNextStepsMessage(array $params, ConsoleStyle $io)
+    {
+        $io->text(array(
+            'Next: Edit Version(lastVersion) to customize the migration.',
+            'Find the documentation at <fg=yellow>https://symfony.com/doc/current/bundles/DoctrineMigrationsBundle/index.html</>',
```
javiereguiluz: Could we standardize these doc links in all commands? `Find the documentation at...` looks too long to me. What about `See ...` ?

src/Command/MakerCommand.php
Pierstoval: And it's available since PHP 5.4. Anyone using this bundle will probably have at least 5.5 or 5.6

src/Maker/MakeMigration.php
Pierstoval: I think you can already get the default connection from Doctrine, as well as for the entity manager, especially if you have only one connection

src/Maker/MakeMigration.php
```diff
@@ -0,0 +1,102 @@
...
+            ->addOption('db', null, InputOption::VALUE_REQUIRED, 'The database connection to use for this command.')
+            ->addOption('em', null, InputOption::VALUE_OPTIONAL, 'The entity manager to use for this command.')
+            ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeMigration.txt'))
+        ;
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
+    {
+    }
```
Pierstoval: Why implement interact when it's empty?

tests/Command/FunctionalTest.php
```diff
@@ -207,6 +208,12 @@ public function getCommandTests()
                 'AppCustomAuthenticator',
             ),
         );
+
+        yield 'migration' => array(
+            new MakeMigration(),
+            array(
+            ),
+        );
```
Pierstoval: Again please use short-syntax for arrays 👍

weaverryan:
This is ready to be reviewed :) - screenshots back in the original description. This also uses the new test system, which means there are real functional tests that actually run the commands and assert that the 2 outputs above occur.

src/Maker/MakeMigration.php
```diff
@@ -0,0 +1,142 @@
...
+
+    public function writeSuccessMessage(array $params, ConsoleStyle $io)
+    {
+        if (false !== strpos($this->migrationOutput, 'No changes detected')) {
+            $io->warning([
+                'No database changes were detected.'
+            ]);
+            $io->text([
+                'The database schema was compared with your mapping information and they are already in sync: there is no generation to migrate.',
```
fabpot: I don't understand the "there is no generation to migrate." part :) I suppose it should be "there is no migration to generate."

weaverryan: Whoooops :)

weaverryan:
Comment addressed (and a few other things). Tests are passing, ready for a final review :).

And sorry for the noise in the diff: this command is a bit different, so we needed to expand the maker system a bit and also the tests, which affected some unrelated things.

src/Maker/MakeMigration.php
```diff
@@ -0,0 +1,142 @@
...
+        $generateMigrationCommand = $this->application->find('doctrine:migrations:diff');
+
+        $commandOutput = new BufferedOutput($io->getVerbosity());
+        $generateMigrationCommand->run(new ArgvInput($options), $commandOutput);
+        $this->migrationOutput = $commandOutput->fetch();
+    }
+
+    public function writeSuccessMessage(array $params, ConsoleStyle $io)
+    {
+        if (false !== strpos($this->migrationOutput, 'No changes detected')) {
```
javiereguiluz: I don't like this logic because it looks fragile ... but there's no other possible way to detect if a migration was generated or not. If MigrationsBundle supported JSON as one of its output formats ... but it doesn't.

src/Maker/MakeMigration.php
```diff
@@ -0,0 +1,142 @@
...
+            $io->warning([
+                'No database changes were detected.',
```
javiereguiluz: Just wondering if we could reword the messages in case no migration is needed:

Before:

```
[WARNING] No database changes were detected.

The database schema was compared with your mapping information and they are
already in sync: no migration was generated.
```

After:

```
[WARNING] No migration was generated.

The database schema and the application mapping information are already in sync.
```

src/Maker/MakeMigration.php
weaverryan: You're exactly right. So I'm depending on our tests to catch this :/

PR #99, opened 2017-12-21
Refactoring tests so that they are functionally tested in a real app

weaverryan:
This is more important than it looks at first :).

Currently, we generate some code and run `phpcs-fixer` on it. This is basically a smoke test. But, does the generated code work in a real app? Or did we overlook something (wrong base class, missing template, missing package).

With this change, inside the test, we create a new flex project and allow it to install its recipes like normal. Then, we run the generated on that outside project and execute functional tests against it. For example, after running `make:controller`, we run a functional test to make sure the new endpoint works. Pretty cool :), and it will give us better quality and the ability to iterate faster with confidence.

This is WIP only because I need to finish converting the remaining makers (I've only done command and controller so far). Some of the code is bizarre, so feedback is welcome!

Cheers to making things! ;)


weaverryan:
Ok, this is ready! There are a few crazy things I get the test system working, so I would MUCH love technical r view before I merge this :).

javiereguiluz:
Thanks for this. It's not easy to review 😓  but I tried to run tests first and I found an issue. Is this working for you?

```
1) Symfony\Bundle\MakerBundle\Tests\Maker\FunctionalTest::testCommands with data set "functional" ('MakeFunctionalTest', 'Symfony\Bundle\MakerBundle\Ma...alTest')
Exception: Running maker command failed: "
 [ERROR] Missing package: to use the make:functional-test command, run:

         composer require css-selector

" "2017-12-28T17:06:16+01:00 [debug] Command "make:functional-test" exited with code "1"
"

src/Test/MakerTestCase.php:71
tests/Maker/FunctionalTest.php:52
```

weaverryan:
@javiereguiluz hmm, no, that shouldn't happy... obviously :D. The test *should* `composer require` all the libraries that are defined in the `configureDependencies()` method of the maker being tested. So, in this case, `MakeFunctionalTest::configureDependencies()`, which (thanks to this PR) now (does) require this library.

So, it's possible that this command failed for some reason (though this should have been reported)! This is likely a legitimate bug with my setup. Could you help me debug? Could you look at your `tests/tmp/cache/MakeFunctionalTest/` directory (this is created automatically in the test) - does it's `composer.json` file have `symfony/css-selector` or `browser-kit` in it? If you, `rm -rf tests/tmp/*`, does it fail again?

Thanks!

weaverryan:
I just pushed some changes. Unrelated to your issue, but pull down fresh before trying again :)

weaverryan:
Merging :). We can improve over time - this is all a test refactoring :)

PR #102, opened 2018-01-02
Changing configureDependencies() to be static

weaverryan:
Hi guys!

This includes another minor BC break (I say minor, because it only affects people creating custom maker classes). This is necessary because of the following situation:

A) Suppose I create a maker (e.g. `make:entity`) that requires Doctrine's EntityManager in its `__construct()` method
B) If I call `make:entity` but don't have the ORM installed, I will get a dependency error from the container, instead of the nice `Please run composer require orm` error.

But, after this PR, if the user run `make:entity --help`, they will get the `Please run composer require orm` error instead of seeing the help information. That's because `MakerInterface::configureCommand()` is still non-static. But I think this is ok.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,3 +6,6 @@
   `AbstractMaker` instead of implementing the interface directly,
   and use `parent::writeSuccessMessage()` to get the normal success
   message after the command.
+
+* [BC BREAK] The MakerInterface changed: `configureDependencies()`
+  was changed to be static.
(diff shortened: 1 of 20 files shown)


weaverryan:
Actually, I'm going to close this PR. For this to work, `configureCommand()` would also need to be static. Instead, I think we'll document that if makers need a service that may not exist, they should inject those in a way where failing is ok (`on-invalid="null"` or via a service locator).

PR #104, opened 2018-01-07
make:entity improved: interactive field created & entity update support

weaverryan:
Hi guys!

The PR we (or at least me) have been waiting for :). This makes `make:entity` behave similar to `doctrine:generate:entity` from SensioGeneratorBundle: it interactively asks you for the field names and generates the getters/setters... but with some improved UX.

## Features:

* Ability to create new entities & fields
* Ability to update *existing* entities 🎆 
* Support for adding *relations* (double 🎆)
* A `--regenerate` flag to add getters/setters for existing properties (replaces `doctrine:generate:entities`)
* Generates code that passes phpcs

Please help test! It's easy: [composer require details to for testing](https://gist.github.com/weaverryan/a20fc281ccd1faea2ea04a5da7fdaa55)

## Blocker 😢 

We're using the newest version of `nikic/php-parser` heavily... but it only has a *beta* release so far :(. We need to wait until this is stable. OR, we need to remove it as a dependency temporarily, then throw an error to tell people to install it manually (at the beta version). :/ https://github.com/nikic/PHP-Parser/issues/474

## Important notes:

* I bumped the min php version to 7.1 (from 7.0). This was so that we could consistently generated code using nullable types.

* Some of the code generated for relationships is opinionated. For example, I automatically synchronize (i.e. by calling the setter) the *owning* side if you set data on the inverse side of the relation. I also make all relation return types nullable (even if the relation is required in the db), to be friendly in case they're not set yet.

* The setters are not fluent. We *could* make them fluent, or make that an option on the bundle (i.e. one in a config file, not asked interactively)

make:entity

If you want to see the available type list (relations are now in this list, but not shown):

screen shot 2018-01-08 at 3 09 32 pm

I would love feedback! I added many little "features" to make the user experience as excellent as possible. But, I need users to help validate that!

Cheers!


src/Validator.php
```diff
@@ -24,16 +22,76 @@ public static function validateClassName(string $className, string $errorMessage
         if (!preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $className)) {
             $errorMessage = $errorMessage ?: sprintf('"%s" is not valid as a PHP class name (it must start with a letter or underscore, followed by any number of letters, numbers, or underscores)', $className);
 
-            throw new RuntimeCommandException($errorMessage);
+            throw new \InvalidArgumentException($errorMessage);
```
yceruto: These changes come back to the "ugly" exception message, what about to use `Symfony\Component\Console\Exception\InvalidArgumentException` instead? or create your own `InvalidArgumentException` that extends from it?

src/Maker/MakeEntity.php
```diff
@@ -80,4 +171,117 @@ public function configureDependencies(DependencyBuilder $dependencies)
...
+        $question->setValidator([Validator::class, 'validateBoolean']);
+        $question->setAutocompleterValues(['true', 'false']);
+        if ($nullable = $io->askQuestion($question)) {
+            $data['nullable'] = $nullable;
+        }
+
+        return $data;
+    }
+
+    private function printAvailableTypes(ConsoleStyle $io)
```
javiereguiluz: We could simplify this method as follows:

```php
private function printAvailableTypes(ConsoleStyle $io)
{
    $io->write(' <info>Available types:</info> ');

    $types = array_keys(Type::getTypesMap());
    $io->write(wordwrap(implode(', ', $types), 50));
    $io->writeln('');
}
```

And if you want to keep the `<comment>` styling (adjust the word-wrap line-length accordingly):

```php
private function printAvailableTypes(ConsoleStyle $io)
{
    $io->write(' <info>Available types:</info> ');

    $types = array_map(function ($v) { return '<comment>'.$v.'</comment>'; }, array_keys(Type::getTypesMap()));
    $io->write(wordwrap(implode(', ', $types), 80));
    $io->writeln('');
}
```

weaverryan: Actually, I just rewrote this algorithm to be more complex, but make the output much fancier and easier for the user.

src/Validator.php
weaverryan: Good catch. It doesn't make sense MOST of the time (i.e. it doesn't make sense if these are callbacks to a Question validation, where the exception is caught), but in case these methods are ever called in a different situation, better to throw the "prettier" exception.

weaverryan:
Made a few tweaks and improvements! Ready for review again :)

src/Maker/MakeEntity.php
```diff
@@ -80,4 +170,248 @@ public function configureDependencies(DependencyBuilder $dependencies)
...
+    private function askForNextField(ConsoleStyle $io, array $fields, string $entityShortClass)
+    {
+        $io->writeln('');
+        $fieldName = $io->ask('New field name (press <return> to stop adding fields)', null, function ($name) use ($fields) {
+            if (in_array($name, $fields)) {
+                throw new \InvalidArgumentException(sprintf('The "%s" property already exists.', $name));
+            }
+            // check reserved words
+            if ($this->isReservedKeyword($name)) {
+                throw new \InvalidArgumentException(sprintf('Name "%s" is a reserved word.', $name));
```
yceruto: Could we allow it by quoting the column name? i.e.:
```php
@Column(name="`number`", ...)
```
> http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/basic-mapping.html#quoting-reserved-words
> Doctrine will then quote this column name in all SQL statements according to the used database platform.

weaverryan:
I've also *started* support for relations/associations. You can see the beginning of how this will work (the command now supports relation types). But nothing generates yet :).

codedmonkey:
Very nice work. I like the idea of the `ClassSourceManipulator`, as it could be a basis for structured customization of the output, but it would need to be injectable for that.

I also like the idea of a relation helper. Here's my idea of a verbose guide that puts less emphasis on the words one and many and more on the collection aspect:

     Class name of the entity to create or update (e.g. GentleChef):
     > Article

     New field name (press <return> to stop adding fields):
     > sections

     > relation

     What class should this entity be related to?:
     > Section

    ManyToOne  The sections field is a reference to a single Section object.
               The Section object can have a field with a collection of Article objects.

    OneToMany  The sections field is a collection of Section objects.
               The Section object has a field referencing a single Article object.

    ManyToMany The sections field is a collection of Section objects.
               The Section object has a field with a collection of Article objects.

    OneToOne   The sections field is a reference to a single Section object.
               The Section object can have a field referencing a single Article object.

Just an idea of course :) Keep up the good work.

weaverryan:
FYI = I've just set this back to WIP, and I'm going to finish the relation stuff in this PR.

@codedmonkey thanks for the quick review of things and UX suggestions! I actually already modified the text a bit further before reading this, but eventually, your version might be better. When I've fully finished the command (as I'm still messing around and doing all kinds of craziness), I would very much appreciate your review again! :)

tvogt:
This is excellent. Will it work on existing Entities? Because most of the time, while prototyping, I will start with a simple entity, and then add fields and relations to it alter.

TheKassaK:
Hello, i'm here because I'm looking for the old command : php bin/console doctrine:generate:entities for generate getters and setters... Any news ?

biji:
Adding setter and getter to existing entity would be great

weaverryan:
I’ll be pushing the final version during the next day I hope. Generating getters/setters to an existing entity IS supported - I think you’ll like it ;)

weaverryan:
Ok! Just added support for regenerating getters/setters (to replace `doctrine:generate:entities`) AND support for relationships! Wooo!

There is an external blocker for release (see above), but it's time to review and test. I'm quite confident in this feature, but (as you can see by the PR), it's quite complex.

src/Maker/MakeEntity.php
```diff
@@ -35,49 +60,706 @@ public static function getCommandName(): string
...
+                    Str::getShortClassName($owningClass),
+                    Str::getShortClassName($inverseClass)
+                ),
+            ]);
+
+            return $io->confirm(sprintf('Do you want to automatically delete orphaned <comment>%s</comment> objects (orphanRemoval)?', $owningClass), false);
+        };
+
+        $setMapInverseSide = function (EntityRelation $relation) {
+            $relation->setMapInverseRelation(!$this->isClassInVendor($relation->getInverseClass()));
```
stof: So, it is even impossible to disable the generation of the bidirectional relation here, as you don't ask 😢

src/Maker/MakeEntity.php
```diff
@@ -35,49 +60,706 @@ public static function getCommandName(): string
...
+                    $targetEntityClass
+                );
+                $relation->setOwningProperty($newFieldName);
+
+                $relation->setIsNullable($askIsNullable(
+                    $relation->getOwningProperty(),
+                    $relation->getOwningClass()
+                ));
+
+                $setMapInverseSide($relation);
```
stof: you should never generate a bidirectional one-to-one relation IMO. they are a performance killer (the inverse side cannot use a lazy-loading proxy, as it does not have the id in its own table, so it *always* creates an extra query each time)

src/Resources/skeleton/doctrine/Repository.tpl.php
```diff
@@ -13,15 +13,31 @@ public function __construct(RegistryInterface $registry)
...
     {
         return $this->createQueryBuilder('<?= $entity_alias ?>')
-            ->where('<?= $entity_alias ?>.something = :value')->setParameter('value', $value)
+            ->andWhere('<?= $entity_alias ?>.exampleField = :val')
+            ->setParameter('val', $value)
             ->orderBy('<?= $entity_alias ?>.id', 'ASC')
             ->setMaxResults(10)
             ->getQuery()
-            ->getResult()
+            ->execute()
```
stof: `getResult` is better than `execute` for a SELECT query builder IMO (especially when dealing with other hydration modes after that)

src/Resources/skeleton/doctrine/Repository.tpl.php
```diff
@@ -13,15 +13,31 @@ public function __construct(RegistryInterface $registry)
...
+        ;
+    }
+    */
+
+    /*
+    public function findOneBySomeField($value): <?= $entity_class_name."\n" ?>
```
stof: missing `?` for the return type. It MUST be a nullable one for a method return one or null result

src/Str.php
```diff
@@ -77,6 +79,16 @@ public static function asTwigVariable(string $value): string
...
     }
 
+    public static function asLowerCamelCase(string $str): string
+    {
+        return lcfirst(self::asCamelCase($str));
+    }
+
+    public static function asCamelCase(string $str): string
+    {
+        return strtr(ucwords(strtr($str, ['_' => ' ', '.' => '_ ', '\\' => '_ '])), [' ' => '']);
```
stof: why transforming dots and backslashes to underscores here ?

src/Str.php
```diff
@@ -87,6 +99,11 @@ public static function asRouteName(string $value): string
         return self::asTwigVariable($value);
     }
 
+    public static function snakeCase(string $value): string
```
stof: should be `asSnakeCase` for consistency

src/Util/ClassSourceManipulator.php
```diff
@@ -0,0 +1,992 @@
...
+    /**
+     * @param string $annotationClass The annotation: e.g. "@ORM\Column"
+     * @param array  $options         Key-value pair of options for the annotation
+     *
+     * @return string
+     */
+    private function buildAnnotationLine(string $annotationClass, array $options)
+    {
+        $formattedOptions = array_map(function ($option, $value) {
+            $quoteValue = !(is_bool($value) || is_int($value) || is_array($value));
```
stof: what about `null` ?

src/Util/ClassSourceManipulator.php
```diff
@@ -0,0 +1,992 @@
...
+            }
+
+            if (is_array($value)) {
+                if (!isset($value[0])) {
+                    // associative array: we'll add this if/when we need it
+                    throw new \Exception('Not currently supported');
+                }
+
+                $value = sprintf('{%s}', implode(', ', array_map(function ($val) {
+                    return sprintf('"%s"', $val);
```
stof: this only supports string values

src/Util/ClassSourceManipulator.php
```diff
@@ -0,0 +1,992 @@
...
+            // because even though these are required in the db, they may not be set
+            true
+        );
+
+        $setterNodeBuilder = $this->createSetterNodeBuilder(
+            $relation->getPropertyName(),
+            $typeHint,
+            // make the type-hint nullable always for ManyToOne to allow the owning
+            // side to be set to null, which is needed for orphanRemoval
+            // e.g. $userAvatarPhoto->setUser(null);
```
stof: that's wrong. OrphanRemoval does not require setting it to `null`. It only requires removing it from the inverse collection.

stof: or, the case of ManyToOne vs OnetoOne are swapped

src/Util/ClassSourceManipulator.php
```diff
@@ -0,0 +1,992 @@
...
+        switch ($doctrineType) {
+            case 'string':
+            case 'text':
+            case 'guid':
+                return 'string';
+
+            case 'array':
+            case 'simple_array':
+            case 'json_array':
+            case 'json':
```
stof: not good. You can store other things than arrays in a `json` field (even though it is the common need)

src/Doctrine/EntityRelation.php
```diff
@@ -0,0 +1,200 @@
...
+
+    private $owningProperty;
+
+    private $inverseProperty;
+
+    private $isNullable = false;
+
+    private $orphanRemoval = false;
+
+    private $mapInverseRelation = true;
```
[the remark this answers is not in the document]
weaverryan: This is one of those tough spots. What are the actual *downsides* of mapping the inverse side? The generator actually generates setter/adder/remover methods on the inverse side that *set* the owning side, which makes the "I set the inverse side and nothing happened" problem lessened or completely eliminated.

src/Str.php
weaverryan: great question - I can't remember anymore where this came from :/. But it doesn't make any sense

src/Maker/MakeEntity.php
```diff
@@ -35,49 +60,706 @@ public static function getCommandName(): string
...
+        if (!$fieldName) {
+            return;
+        }
+
+        $defaultType = 'string';
+        // try to guess the type by the field name prefix/suffix
+        // convert to camel case for simplicity
+        $camelCasedField = Str::snakeCase($fieldName);
+        if ('_at' == substr($camelCasedField, -3)) {
+            $defaultType = 'datetime';
```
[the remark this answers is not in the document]
Majkl578: Please do not use mutable datetime at all, use immutable only.

src/Doctrine/EntityRelation.php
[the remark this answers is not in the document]
Majkl578: Inverse 1:1 assocations could be a performance nightmare. Also synchronization is.

src/Maker/MakeEntity.php
```diff
@@ -35,49 +60,706 @@ public static function getCommandName(): string
...
+                $annotationOptions = $newField;
+                unset($annotationOptions['fieldName']);
+                $manipulator->addEntityField($newField['fieldName'], $annotationOptions);
+
+                $currentFields[] = $newField['fieldName'];
+            } elseif ($newField instanceof EntityRelation) {
+                // both overridden below for OneToMany
+                $newFieldName = $newField->getOwningProperty();
+                $otherManipulatorFilename = $this->getPathOfClass($newField->getInverseClass());
+                $otherManipulator = $this->createClassManipulator($otherManipulatorFilename, $io, $overwrite);
```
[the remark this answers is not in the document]
weaverryan: It *is* a bit wasteful, but it just gets ugly the other way - we use it a few places below, except for OneToMany, where we override it. I tried wrapping this in an if statement and it only seemed to make things look worse.

src/Maker/MakeEntity.php
```diff
@@ -35,49 +60,706 @@ public static function getCommandName(): string
...
+        });
+
+        if (!$fieldName) {
+            return;
+        }
+
+        $defaultType = 'string';
+        // try to guess the type by the field name prefix/suffix
+        // convert to camel case for simplicity
+        $camelCasedField = Str::snakeCase($fieldName);
```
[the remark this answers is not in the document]
weaverryan: Hahaha. This is what I get for copying some from SensioGeneratorBundle, then - *partially* - updating it :). Good catch

src/Maker/MakeEntity.php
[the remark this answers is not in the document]
weaverryan: Any downside to using `datetimetz`? Like, if I don't care about the timezone? It seems like it's just discarded for DB's that don't support it anyways... which means that it *would* be a safe default. But want to be sure.

src/Maker/MakeEntity.php
weaverryan: That's correct :). I need some convincing that there is a downside before we add more options

src/Maker/MakeEntity.php
weaverryan: I love having feedback at this deep of a level - awesome :).

I just tested, and it's not *quite* as you say: only *1* query is made, but it's a join query. But I think it's basically the same issue: if you query directly for the inverse entity, you query for 2x the data, even if you don't need it. So yea, I'm going to reverse this: no inverse relation for OneToOne

src/Util/ClassSourceManipulator.php
weaverryan: Yea, this is confusing. The issue is that the generated `remove*` method on the inverse side (and really, the mutators for all the inverse sides) synchronizes the owning side - e.g. it looks like this:

```php
// User.php
public function removeAvatarPhoto(UserAvatarPhoto $photo)
{
    $this->avatarPhotos->removeElement($photo);
    // This is the key: we sync the owning side, which is why it needs to be nullable
    $photo->setUser(null);
}
```

When orphanRemoval is set, yea, we don't need to sync the owning side, as the orphanRemoval will cause the deletion. But since we *are* sync'ing it (and I think that's generally a good idea - and I don't want to NOT sync it just for the case of `orphanRemoval=true` - want to not create so many variations), we need to allow the setter to be null. 

These are the tough balances

src/Util/ClassSourceManipulator.php
weaverryan: Actually, check out the JsonType class - it uses `$val = json_decode($value, true);` in `convertToPHPValue`. So it looks like it *does* always return an array.

However, that method could also return `null`. So probably having no return type is safer.

src/Doctrine/EntityRelation.php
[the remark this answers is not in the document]
weaverryan: I'm going to add a question to ask the user if they want an inverse side, instead of always adding it. Then, it's their fault ;). Plus, we only synchronize the owning side (from the inverse side) to minimize performance. If you're not going to set the inverse side, then you won't call these methods and have the penalty anyways. But if you *do* set the inverse side, then it's *really* nice to synchronize and I think the penalty is worth it.

tests/Doctrine/fixtures/expected_no_overwrite/src/Entity/UserAvatar.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function getUser(): ?User
+    {
+        return $this->user;
+    }
+
+    public function setUser(?User $user)
+    {
+        $this->user = $user;
```
Xymanek: What if I update this directly? Then `User#avatars` will not be updated

weaverryan: Do you mean: if you call $avatar->setUser($user), the problem is that the avatar will not all be added to User#avatars? i.e. the problem is that the relationship is not synced?

If so, it’s by design. If you set the inverse side of a relation, the generated code *does* sync the opening side. But not the other way around. The reason is that (a) it doesn’t need to be set for persistence and (b) more importantly, it could have major performance consequences. For example, if there are 1000 avatar, by setting User#avatars, Doctrine would first query for all 1000 and then set it. We gotta avoid that :)

Xymanek: Here's an example:

1. Create a new article, `setCategory($category)`.
2. Something later (eg. listener) grabs `$article->getCategory()` and passes it somewhere else
3. It ends up in code that does `foreach ($category->getArticles() as $article)`
4. Weird results

I understand what you are saying but relying on chance that the collection will not be used later during request.... not really sound strategy.

> Doctrine would first query for all 1000 and then set it

Not is it's set to extra lazy: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/extra-lazy-associations.html

weaverryan: So... this is tricky. If we generate code that has fetch="EXTRA_LAZY", then setting the inverse side, only costs one "fast" to check of the item already is related. If we do not set "EXTRA_LAZY", then it is far too costly.

This leaves us with 2 choices:
1) Generate all relationships as EXTRA_LAZY, sync the inverse side (just for edge-case convenience) and suffer a small perf impact

2) Do no generate relationships as EXTRA_LAZY, do not sync the inverse side (edge case problems could arise) and suffer no per impact.

I'm really not convinced either way. @stof @Majkl578 you usually have words of warning - what do you have here? Btw, we now ask the user IF they want to generate the inverse side. If they choose no, we don't have to worry about any of this.

I lean slightly towards option (1). EXTRA_LAZY seems better in 99% of the cases. And I don't like the small perf impact, but it really is small, and avoids WTF situations. I can't think of a situation where this causes a BIG perf impact.

Kudze:
I've tested it. Looks really nice, however I have some suggestions for make:entity.

When generating a string `generate:doctrine:entity` used to ask for length, and this one doesn't. In my opinion it should ask for it.

Also an option for setters that would return the object would be nice. (If you can't understand what i mean by `setters that would return the object`, heres an example)
```
class Entity {

    private $value;

    //...

    public funcition setValue(string $value) : Entity
    {
        $this->value = $value;

        return $this;
    }

    //...

}
```
This would allow to set a lot of variables at once like this:
```
$entity->setValue1("val1")->setValue2(420);
```

Xymanek:
> Also an option for setters that would return the object 
Usually called fluent setters

gabiudrescu:
I just (briefly) tested this and looks very nice. It is definitely an improvement from previous command. 

Two mentions, one already said:
 - the new command does not ask how long it should be a string field (something the old command did); but it's not a deal breaker, IMO
 - there seems to be an opinionated command that allows only annotations mapping; not a bad thing, but some prefer yaml or xml.

ok, I admit, I prefer yaml. I know that Doctrine maintainers are not big fans of this format and they plan to remove the support for it. but is this command going to support other mapping formats besides annotations?

weaverryan:
Thanks for the feedback guys! Awesome! Both suggestions - the length & fluent setters - were something I was thinking about. I’ll add those :).

About Yaml or XML, it’s sort of supported :). You cannot use the command to generate XML mapping code or YAML. But if you write some YAML, you can run make:entity -- regenerate to create al the getters and setters. I’d really like to *not* support anything beyond this... but we’ll see ;)

tests/Doctrine/fixtures/expected_no_overwrite/src/Entity/UserAvatar.php
stof: > EXTRA_LAZY seems better in 99% of the cases.

That's not actually true. If your business need means that the collection will never become huge, performing DB queries each time you check `contains` or `count` or similar APIs on the collection is actually much worse for performance than loading the collection on first usage (trading a bit of memory to win on time). I actually reverted a bunch of my extra lazy collections to be only lazy ones (i.e. the default) to solve performance issues.

For me, the right fix is to stop generating bidirectional relations by default. All this comes from the fact that the code expects a fully synchronized bi-directional relation. If you make the relation unidirectional, you make all these issues disappear (if you try to load the list of all articles before saving the change the DB, you might have a harder time writing your listener indeed as there is *no* place having the source of truth for the updated full list yet, but then it might push you to reconsider this need instead of killing performance and making the entity code very complex to have this list available).

src/Util/ClassSourceManipulator.php
stof: nope, it can also return any scalar or null. The `true` argument just means it will never return you a `stdClass`

src/Util/ClassSourceManipulator.php
stof: but this syncing forces you to make the setter accept null on the owning side, even if `null` would be an invalid value because of the field being non-nullable, allowing invalid usage of the class.

stof:
> About Yaml or XML, it’s sort of supported :). You cannot use the command to generate XML mapping code or YAML. But if you write some YAML, you can run make:entity -- regenerate to create al the getters and setters. I’d really like to not support anything beyond this... but we’ll see ;)

@weaverryan would this regeneration generate the mapping annotations on properties too or no ? If yes, your solution does not work (having both the XML and the annotation mapping will cause issues, as you would then have 2 source of truth for the mapping, with Doctrine probably using only one of them, but maybe not the one you modified)

gabiudrescu:
to make the discussion simpler: if someone has only XML annotations in the app, is that person going to be able to use this command or not?

or should the app be migrated to metadata written as annotation so this command can be used?

I'm asking this to clarify the behavior, not because I'm requesting any implementations. I feel that it should remain written so it can be indexed by search engines and answer future questions.

weaverryan:
Hey guys!

If you have xml or yaml, the regenerate option generates the getters and setters, but does NOT add annotations to your class. So, your original xml or yaml is still used.

So yes! This command IS still useful for someone uselse by XML or yaml. Well, *half* of the command is useful: the normal version that asks for new fields is not useful. In fact, that *would* generate new fields with annotations. I should throw an exception in this case to tell the user what’s going on :). But the regenerate option *does* work fine with xml or yaml.

gabiudrescu:
if I got this right, when starting a new project you have the following options for your entities mapping:
 - write yaml or xml by hand and use the command to generate PHP classes based on these files
 - use the command entirely and go with annotations

well, although personally I would have liked the idea of having yaml generated by the command to, given that Doctrine is planning to drop support for it, this sounds like a good compromise for switching to annotations.

and the new command also sweetens the deal of adopting a new format.

tests/Doctrine/fixtures/expected_no_overwrite/src/Entity/UserAvatar.php
weaverryan: I thought more about this too and had the same thought as you Stof - that EXTRA_LAZY could cause tons of count queries in certain situations.

> If you make the relation unidirectional, you make all these issues disappear

It's true. But you've also killed a lot of usability for many users. To strike a balance, the command now DOES ask if you want the inverse side to be mapped. So, it's up to you

weaverryan:
Updates! Added:

* Fluent setters/adders/removers
* For string field, `make:entity` now asks for the field length
* Error if you try to generate classes that use non-annotation mapping

The only current block is waiting for the stable 4.0 of nikic/PHP-Parser - see nikic/PHP-Parser#474

nckenn:
nikic/PHP-Parser#474 has now a stable version, hope you can make release version of this PR :)

weaverryan:
Ok! I've just updated to use `nikic/PHP-Parser` 4.0 *stable*!

There are no known, pending issues with this PR. So, now would be the time for anyone to do any last review or testing.

The tests on appveyor ARE randomly failing - different tests each time :/. I've started debugging them, but the circumstances are very odd. As long as no appveyor tests consistently fail, I'm not going to let that block this PR (i.e. we'll debug the phantom failures later).

Cheers!

MrNicodemuz:
@weaverryan regarding the randomly failing tests, maybe you can let your CI script echo out the User entity that seems to have the missing methods.

Maybe it's a file permission issue or the disk is full?

weaverryan:
Thanks for checking the tests :). I actually did that earlier from inside the test class that reports them missing. Ready for your mind to be blown? The methods were there! It’s as if the test class in the tmp project already have an old version of the class in memory, but I don’t know how that’s possible, or what on windows would cause it. I’m open to ideas :)

Gwynei:
This is the most wanted feature for me in Symfony 4.

I have tested it, and I have two suggestions:

- Is it possible to regenerate ALL entities at once? Not that all entities will be changed, but thing is that it speeds up developing. For example, I make changes in six entities (deleted ManyToOne, changed data type, added field, etc), now I have to run the --regenerate command six times and I have to find out the exact names of those six entities. In Symfony 3.4 I could run doctrine:generate:entities App and it would check/update every entity.

- Minor one: When generating an entity and you use the maker to generate fields, the // your fields here comment is at the bottom, after the $id field and the GetId() getter. Previously it was between those two (I usually have all my fields first, and then the getters/setters.

Thanks a bunch!

weaverryan:
Thanks for testing @Gwynei :). About your questions/comments:

> Is it possible to regenerate ALL entities at once? Not that all entities will be changed, but thing is that it speeds up developing. For example, I make changes in six entities (deleted ManyToOne, changed data type, added field, etc), now I have to run the --regenerate command six times and I have to find out the exact names of those six entities. In Symfony 3.4 I could run doctrine:generate:entities App and it would check/update every entity.

Yes, you should be able to just do `bin/console make:entity --regenerate` and use the default `App` value (i.e. the namespace, not the class name) for the one question. You can regenerate a single class, or *all* classes matching a namespace. Let me know if you're seeing otherwise.

> Minor one: When generating an entity and you use the maker to generate fields, the // your fields here comment is at the bottom, after the $id field and the GetId() getter. Previously it was between those two (I usually have all my fields first, and then the getters/setters.

Good catch! I'm going to remove the comment entirely.

> Found a bug (I think) fields with nullable=true give this error while regenerating:

> 2018-03-06T09:29:26+01:00 [error] Error thrown while running command "make:entity --regenerate". ` Message: "Notice: Undefined index: nullable"
> In EntityRegenerator.php line 127:
> Notice: Undefined index: nullable

Another good catch - I should have it fixed now :). But let me know if you see anything else like this.

Gwynei:
Thanks for the fixes @weaverryan - works fine now!

If I may do one more request? Can you make it so that I can give a parameter to the entity/entities to regenerate? Such as this:

`bin/console make:entity --regenerate App`
or
`bin/console make:entity --regenerate App\Entity\User`

Not mandatory but just a convenience. :)

src/Util/ClassSourceManipulator.php
```diff
@@ -0,0 +1,1011 @@
...
+     *
+     * @return string
+     */
+    private function buildAnnotationLine(string $annotationClass, array $options)
+    {
+        $formattedOptions = array_map(function ($option, $value) {
+            if (is_array($value)) {
+                if (!isset($value[0])) {
+                    // associative array: we'll add this if/when we need it
+                    throw new \Exception('Not currently supported');
```
petrvacha: ```$value = ["default" => "now()"]``` fires the exception

weaverryan: It *is* true that you could use the `ClassSourceManipulator` directly (specifically `addEntityField()`) and pass in some annotations where the value is an associative array. But, this situation is not actually possible in the bundle (at least at this time) - either `addEntityField()` is called by `MakeEntity`, and it passes in only scalar values, or it's called by `EntityRegenerator`, but in that case, the annotations aren't actually printed on the field.

So, it's just not a possible use-case right now :). If someone wants to PR this to fill it in, that's cool. I don't want to fill it in quickly and realize it's incorrect if we *do* in fact later use it.

weaverryan:
@Gwynei good catch - I've fixed the command so you can just run `bin/console make:entity --regenerate App` and *not* get the interactive question - that was *indeed* the intention :)

MrNicodemuz:
Can we add an option to skip annotations? I'm using YML.

MrNicodemuz:
Adding a comment on an field does not work:

    App\Entity\Testing:
        type: entity
        table: testing
        repositoryClass: App\Repository\TestingRepository
        id:
            id:
                type: integer
                id: true
                generator:
                    strategy: AUTO
        fields:
            created:
                type: datetime
                options:
                    comment: Date of creation.

It gives me the following error:

    In ClassSourceManipulator.php line 211:

      Not currently supported

weaverryan:
@MrNicodemuz 

> Can we add an option to skip generation of annotations? I'm using YML.

I don't understand what you mean. If you use YAML, you cannot use this command to add new fields: you can only use the `--regenerate` option. And when you use the `--regenerate` option, it only generates the properties, getters, setters, etc - it does *not* generate annotations above the properties. If you see something different, please let me know :).

> It gives me the following error:

This is legit - I just finished telling someone this situation wasn't possible. I was wrong :). I'll fix it.

mnavarrocarter:
This is amazing. Thanks @weaverryan!

src/Util/ClassSourceManipulator.php
```diff
@@ -0,0 +1,1015 @@
...
+    {
+        $formattedOptions = array_map(function ($option, $value) {
+            if (is_array($value)) {
+                if (!isset($value[0])) {
+                    return sprintf('%s={%s}', $option, implode(', ', array_map(function ($val, $key) {
+                        return sprintf('"%s" = %s', $key, $this->quoteAnnotationValue($val));
+                    }, $value, array_keys($value))));
+
+                    // associative array: we'll add this if/when we need it
+                    throw new \Exception('Not currently supported');
```
stof: this is dead code

deivid11:
Thanks! So wanted feature! @weaverryan 😄

PR #105, opened 2018-01-13
Added a maker to generate empty fixture classes

### LINKED ISSUE #84: [RFC] Add make:fixture command
Since doctrine/doctrine-fixtures-bundle has a very nice Flex support, why not including it in this bundle, too?

javiereguiluz:
This fixes #84 following the new best practices recommended for Doctrine fixtures (https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html).


src/Maker/MakeFixture.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+
+/**
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Ryan Weaver <weaverryan@gmail.com>
+ */
+final class MakeFixture extends AbstractMaker
+{
+    private $eventRegistry;
```
weaverryan: don't need this :)

src/Maker/MakeFixture.php
```diff
@@ -0,0 +1,76 @@
...
+            'Next: Open your new fixtures class and start customizing it.',
+            'See <fg=yellow>https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html</>',
+        ]);
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies)
+    {
+        $dependencies->addClassDependency(
+            Column::class,
+            'orm-pack'
```
weaverryan: usually we use the `doctrine` alias for this. Should we try to centralize on using that?

weaverryan:
Oh, one more thing: I think it *should* have a `class-name` argument. Let's default it to `AppFixture` (actually default it - in most cases, we don't give the user a default value, just a "suggestion" - we should actually default here). I think it makes the command make more sense: I am asked a class name (but given an easy suggestion) and then that class was created.

weaverryan:
Thanks Javier!

PR #107, opened 2018-01-14
Remove check if symfony/flex is installed

rosier:
This bundle doesn't use flex or require flex to work. It just makes assumptions about the directory structure based on the latest recommendations for Symfony.

This bundle can also be useful for projects that don't use flex yet ;-)


fabpot:
👍

weaverryan:
Hmm, ok then :). We may later need to add warnings in other places about the directory structure specifically (to help people realize if they don't have the correct structure for this bundle. Thanks!

PR #108, opened 2018-01-16
Add description and homepage to composer.json

rosier:
According to `composer validate` the property description is required.

I did also add the property homepage in an separate commit. So you can cherry pick if you want.


weaverryan:
Good idea :). Thanks @rosier!

PR #110, opened 2018-01-17
AbstractMaker link

JuanLuisGarciaBorrego:
The link AbstractMaker goes to the current page and is not going to AbstractMaker Class on Github.
I think it should be like that.


weaverryan:
Good catch! Thanks Juan!

PR #112, opened 2018-01-24
Use the core Route annotation in controllers

### LINKED ISSUE #88: Use AbstractController as the base controller class
This has been something we've debated. I just had a long conversation with @nicolas-grekas about it, who feels pretty strongly about pushing this best practice now. I don't feel too strongly, and certainly we need to do this at some point. Since MakerBundle is for Symfony 4 (so new projects, or people who have upgraded and are fully using the new features), this is a great place to push the best practice.

javiereguiluz:


weaverryan:
I think I’d like to delay this until 4.1, where getParameter() will work correctly. In other words, wait until 4.1 is released before merging this.

stof:
The Route annotation change could be done separately though

javiereguiluz:
I've reverted the abstract controller changes and left only the core Route annotation change. Thanks!

Issue #3, opened 2017-11-10
CRUD generator?


fabpot:
This bundle actually does not replace the current generator bundle as it was more focused on a CRUD generator. Is it something we want to do here as well?

mneuhaus:
yes please :) 😃

MarioBlazek:
It would be a nice feature to have +1

ad3n:
Yes we need but with more clearly... Sensio Generator is ugly, redundant and repeated code anywhere

ogizanagi:
How is the old CRUD generator used nowadays? I mean, isn't an admin CRUD generator like EasyAdmin more efficient & simpler? Or is it really used to bootstrap code in an application?
For my part, I never used the old CRUD generator.

dsentker:
The crud Generator is used often in my projects. I would really love a modern crud Generator here. 
Btw i do not think that the sensio generator bundle creates 'ugly' Code.
+1 for a modern crud Generator 👌

Leuloch:
I'm about to start a new project with symfony 4, and waiting for the CRUD generator feature! I'm a freelancer and this would help me a lot to bootstrap apps

codedmonkey:
I like the idea of a CRUD generator, the old one provides enough basic utility for a functional starting point. When looking at the generated code, the PHP-code is just a basic Symfony controller, which seems fine. The templates on the other side, are something to be more opinionated about. Obviously it would be unwise to include CSS with the generated code and it would increase complexity for the user.

One solution to this could be a mechanic where third-party bundles are allowed to override the skeleton files (https://github.com/symfony/maker-bundle/issues/33). A dedicated bundle (like a Bootstrap bundle) would be able to provide more modern templates. The Bootstrap example is just one of the use cases, an official bundle would be unlikely to make everybody happy on this.

lemorragia:
yesss!!

inmarelibero:
I consider this an extremely important feature to have; "old" generator bundle does not provide ugly code, but it has come issue I personally handle very time I use it:
* can't choose in what folder generate the controller, form and templates
* different `new.html.twig` and `edit.html.twig` templates: first things I do is delete one of them and use the other for both `newAction()` and `editAction`
* rename variables in generated controller: `'edit_form'` => `form`, `$editForm` => `$form`
* remove delete form from edit template
* use uuid instead of id

so +1 for a new generator, hopefully extendable giving the possibility to the user to choose options above

ksn135:
[...]

PR #113, opened 2018-01-25
Add make:crud & improve make:form commands

sadikoff:
Hi all, 

## Improved make:form command
Main improvement is that command works in two ways. First way simple, it just generates skeleton form class and user can do with it whatever he want. Second way is generation from entity class. Generator checks if entity exists and ask user if he wants to generate form from entity or user can use `--entity` flag if he knows that entity exists.
### Generation features:
- Simple skeleton form class
- Doctrine entity form class
### Examples:
**Simple class generation**
```bash
root@striX: ~/project_root# bin/console make:form

 The name of the form class (e.g. GentleElephantType):
 > SimpleForm

Enter the class or entity name that the new form will be bound to (empty for none):
 >

 created: src/Form/SimpleFormType.php


  Success!
```
**Entity form generation**
```bash
root@striX: ~/project_root# bin/console make:form 

 The name of the form class (e.g. GentleElephantType):
 > EntityFormType

Enter the class or entity name that the new form will be bound to (empty for none):
 > Entity

 created: src/Form/EntityFormType.php


  Success!
```
Closes: #114

## New make:crud command
Want to introduce `make:crud` command it is similar to doctrine:generate:crud from SensioGeneratorBundle. 

All you need to generate cool crud is Entity :)
### Features
- complete crud generation
### Example:
```bash
root@striX: ~/project_root# bin/console make:crud SweetFood

 created: src/Controller/SweetFoodController.php
 created: src/Form/SweetFoodType.php
 created: templates/sweet_food/_delete_form.html.twig
 created: templates/sweet_food/_form.html.twig
 created: templates/sweet_food/index.html.twig
 created: templates/sweet_food/show.html.twig
 created: templates/sweet_food/new.html.twig
 created: templates/sweet_food/edit.html.twig


  Success!

```
Closes: #3


komujohn:
good work @sadikoff  just what I needed

sblondeau:
Thanks @sadikoff for this great work. However it seems to be exactly the same crud as in sf3, with the same drawbacks, like the ugly html code with ul/li, the unuseful "action" col in the index table (a link on the id should be much more convenient ...).
I know it is impossible to satisfy everybody with this kind of generator but I think that the sf4 crud should be an enhancement of sf3 version, not just  the same. ;-)

YannickFricke:
@sblondeau But what about custom designs?
I would not like to have prewritten CSS for that.
Maybe there should be an option for the command (but please set the standard to false).

sadikoff:
@sblondeau I agree that we need some enhancements. May be we can discuss it?

sadikoff:
I have a question. Is it right to inject EntityManager to maker? or may be there is another way to get Entity metadata?

src/Resources/skeleton/crud/templates/index.tpl.php
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.html.twig' %}
+
+{% block body %}
+    <h1><?= $entity_class_name; ?> index</h1>
+    <table>
```
codedmonkey: Maybe add `thead` and `tbody` tags? They seem to be pretty popular these days.

src/Resources/skeleton/crud/templates/new.tpl.php
```diff
@@ -0,0 +1,9 @@
+{% extends 'base.html.twig' %}
+
+{% block body %}
+    <h1>Create new <?= $entity_class_name; ?></h1>
+    {{ form_start(form) }}
+    {{ form_widget(form) }}
```
codedmonkey: <?php foreach ($entity_fields as $field): ?>{{ form_row(form.<?= ucfirst($field['fieldName']); ?>) }}
    <?php endforeach; ?>

would be my preference here, and I would use a seperate `_form.html.twig` to generate the same form for the two occurrences. https://github.com/symfony/demo/blob/master/templates/admin/blog/_form.html.twig

src/Resources/skeleton/crud/controller/ControllerWithTwig.tpl.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Creates a form to delete a <?= $entity_class_name; ?> entity.
+     *
+     * @param <?= $entity_class_name; ?> $<?= $entity_var_singular; ?> The <?= $entity_class_name; ?> entity
+     *
+     * @return \Symfony\Component\Form\FormInterface The form
+     */
+    private function createDeleteForm(<?= $entity_class_name; ?> $<?= $entity_var_singular; ?>)
```
codedmonkey: Maybe follow the Symfony demo in this case by using a seperate template to generate a delete form? https://github.com/symfony/demo/blob/master/templates/admin/blog/_delete_form.html.twig

src/Resources/skeleton/crud/templates/show.tpl.php
```diff
@@ -0,0 +1,28 @@
...
+    <table>
+    <?php foreach ($entity_fields as $field): ?>
+        <tr>
+            <th><?= ucfirst($field['fieldName']); ?></th>
+            <td>{{ <?= $entity_var_singular; ?>.<?= $field['fieldName']; ?> }}</td>
+        </tr>
+    <?php endforeach; ?>
+    </table>
+
+    <ul>
```
codedmonkey: Just remove the `ul` and `li` tags. They are unnecessary and if there is debate about the styling, just keep it as minimal as possible (a `<br>` could be enough).

src/Resources/skeleton/crud/controller/ControllerWithTwig.tpl.php
sadikoff: There is a big question with delete form for me. May be there is a way to create a delete link in order to delete entities from list and more customization in other templates? Delete form is a good and secure way, but we can include it only in edit or show action.

src/Resources/skeleton/crud/templates/show.tpl.php
sadikoff: agreed. They are really unnecessary.

src/Resources/skeleton/crud/templates/new.tpl.php
sadikoff: do we really need to generate whole form with all fields separated?

src/Resources/skeleton/crud/templates/show.tpl.php
sblondeau: indeed, ul and li should be remove. However, br is not a good id (should not be use outside of `<p>`) I think a simple div would be enough
I also agree about a separate file for delete form. 
And about CSS, it should not be include (or with an option for included bootstrap classes maybe? not by default)

src/Resources/skeleton/crud/controller/ControllerWithTwig.tpl.php
codedmonkey: With a seperate template this is easily achievable with

    {% include '.../_delete_form.html.twig with {'post': post.id} %}  

but I don't know whether there are security concerns when using multiple forms with the same CSRF token.

codedmonkey:
@sblondeau I immediately think of Bootstrap templates as well, but that's something that should be outside this bundle as these differences in styling are trivial at best when it's so easy to create your own maker. In fact, I'm planning on doing that exact thing if @sadikoff doesn't mind if I run off with some of his code.

featuriz:
Thanks for making CRUD friends. I'm using symfony4.0.4 . I created maker extension as https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html#creating-your-own-makers . It works great. The problems i faced are 

1. EntityManagerInterface::class, 'orm-pack' ;; @ MakeCrud Controller -> configureDependencies
**Error:** 

>           [ERROR] Missing package: to use the make:crud command, run:                    
>                                                                                 
>          composer require orm-pack

Solution: I comment this check.

2. ControllerWithTwig.tpl.php after generate, the controller throws error for new().
Solution: I changed new() to newAction(). I added Action at the end of all the public (crud) methods.

3. Template index.twig.html ;; Error for datetime entity.
Solution: I manually added |date('m d y');

How to add date  at index.tpl.php ;;

sadikoff:
@featuriz 
1. was fixed in last code revision
2. which php version do you use? from php 7.0 we can use reserved words as function names
3. will think what I can do with this, I think the solution is to check fields and generate specific code for some types

weaverryan:
Thanks for getting this rolling! It’s on my radar - I’ll review this after I finish the make:entity changes :)

featuriz:
Good Every thing works.
My suggestion is that. 
1. We are just creating the frame. So at create let it be " |date() ", timezone is not required, let them add it. Any how we have to update template as our taste.
2. return must need " 'delete_form' => $deleteForm->createView(), " this, only then the generated twig will be complete. I think this is must for a good template.
3. At show() route add " , requirements={"id"="\d+"} ". This will prevent parameter exception. i.e. /{id} and /new are same. Conflict!.
4. Need confirmation at delete. " <input type="submit" value="Delete" onclick="return confirm('Are you sure you want to delete this item?');">  "

These are the things I updated in my work.

sadikoff:
@featuriz 
1, 6 fixed.

> 3 ... requirements={"id"="\d+"} ...

it's not very good idea, in my project, for example, primary key is uuid and this requirement won't work. I fixed this problem with simple moving new action before show.

> 5 ... Need confirmation at delete

agreed but not on click, I think better will be to place it in onSubmit event

Need more discussion about 2 and 4. May be there are more ideas?

chapterjason:
Wouldn't it usefull to prefix the route name? It would prevent duplicate names if you generate more than one entitiy.

src/Maker/MakeCrud.php
```diff
@@ -0,0 +1,146 @@
...
+
+        $metadata = $this->getEntityMetadata($entityClassName);
+
+        return array(
+            'controller_class_name' => $controllerClassName,
+            'entity_var_plural' => lcfirst(Inflector::pluralize($entityClassName)),
+            'entity_var_singular' => lcfirst(Inflector::singularize($entityClassName)),
+            'entity_class_name' => $entityClassName,
+            'entity_identifier' => $metadata->identifier[0],
+            'entity_fields' => $metadata->fieldMappings,
```
weaverryan: So this is a great start. And actually, we need to improve `make:form` to be smarter and do stuff like this - see #114.

But, we need to be a bit smarter, and fortunately, this work was done already in SensioGeneratorBundle :). See: https://github.com/sensiolabs/SensioGeneratorBundle/blob/master/Generator/DoctrineFormGenerator.php#L99

The eliminate the identifier fields more cleanly and avoid adding OneToMany relationships - which will definitely not work :). I think we could get even crazier, but let's just match what was done before.

And, we could (in this same PR for convenience) update `make:form` to use this new awesomeness as well :). But, we can do it later if you'd like.

src/Resources/skeleton/crud/templates/index.tpl.php
```diff
@@ -0,0 +1,31 @@
...
+            <tr>
+                <?php foreach ($entity_fields as $field): ?><th><?= ucfirst($field['fieldName']); ?></th>
+                <?php endforeach; ?>
+                <th>actions</th>
+            </tr>
+        </thead>
+        <tbody>
+        {% for <?= $entity_var_singular; ?> in <?= $entity_var_plural; ?> %}
+            <tr>
+                <?php foreach ($entity_fields as $field): ?><td>{{ <?= $entity_var_singular; ?>.<?= $field['fieldName'].(false === strpos($field['type'], 'date') ? '' : '|date()'); ?> }}</td>
```
weaverryan: For printing out the fields, we should re-use the logic from here: https://github.com/sensiolabs/SensioGeneratorBundle/blob/master/Resources/skeleton/crud/views/index.html.twig.twig#L25

Honestly, that code is *ugly*! It would be ok to create a new PHP class to do a lot of this work, then just use that class here - e.g. 

```php
{{ <?= $helper->getPrintCode($entity_var_singular, $field['fieldName'], $field['type']); ?> }}
```

... which would print:

```twig
{{ product.createdAt ? product.createdAt|date('Y-m-d') : '' }}
```

src/Resources/skeleton/crud/templates/index.tpl.php
weaverryan: Haha, I agree - they're super hipster :)

sadikoff:
@weaverryan thx for your review, I'll try to fix everything tomorrow.

weaverryan:
That's great @sadikoff! I'll review again after that - I haven't looked too closely yet at the actual generated controller or template code :)

src/Maker/MakeCrud.php
sadikoff: I can improve make:form in this PR, but I need an advise. How smarter this command must be?
For example:
We create 2 ways for command, 
[1] there is Entity with enetered name
[2] there is **NO** Entity

If we have entity, than we generate complete form with all fields, otherwise we generate only stub class.

Is this will be enough? or we need something more?

sadikoff:
@weaverryan could you please check GeneratorHelper class? Is it good or may be we need something else?

src/Maker/MakeCrud.php
weaverryan: That’s it :).

Except, I also think we need to change the initial question:

> Do you want to generate a form for an entity?

If yes, ask them for the entity class (and build the form class name from that). If no, ask them the form class.

sadikoff: Ok, that's clear.

And one question about using this command from crud generation. I think we need at first generate form, than crud. That's why I think we need to add something like beforeGenerate method. What do you think about it?

weaverryan: Actually, in one of the PR’s, we completely simplify into just a generate() method :). But I get your point: we’re generating a form, then a crud. But this should be invisible to the user: they’re just thinking: in generating a crud based on an entity. So we shouldn’t try to call the make:form command from make:crud. But we should isolate any generation logic that needs to be reused into a service and then call it from both commands.

sadikoff: If you don't mind I took DoctrineMetadataFactory from your PR =) and pushed improved make:form command

sadikoff:
Hi all, need review for improved make:form command and GeneratorHelper.

my next step is to write tests for crud generation to make sure everything works as expected.

src/Maker/MakeCrud.php
weaverryan: It's no problem - good idea if you needed it :)

src/Doctrine/DoctrineEntityHelper.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Symfony\Bundle\MakerBundle\Doctrine;
+
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
+
+/**
+ * @author Sadicov Vladimir <sadikoff@gmail.com>
+ */
+final class DoctrineEntityHelper
```
weaverryan: Mark it with `@internal` also

src/Doctrine/DoctrineEntityHelper.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * @param $entityClassName
+     *
+     * @return \Doctrine\ORM\Mapping\ClassMetadata|null
+     */
+    public function getEntityMetadata($entityClassName)
+    {
+        return $this->metadataFactory->getMetadataForClass('App\\Entity\\'.$entityClassName);
```
weaverryan: Throw a clear exception at the start of this method if `metadataFactory` is null

src/Doctrine/DoctrineEntityHelper.php
```diff
@@ -0,0 +1,63 @@
...
+
+        return $fields;
+    }
+
+    /**
+     * @param $entityClassName
+     *
+     * @return \Doctrine\ORM\Mapping\ClassMetadata|null
+     */
+    public function getEntityMetadata($entityClassName)
```
weaverryan: Make this be the *full* class name. I mean, don't add `App\\Entity\\` to it below. If that needs to be done, it should be done in the command itself. And actually, now that #120 is merged, there is only 1 spot in the entire code base that mentions the `App` namespace.

src/Maker/MakeForm.php
```diff
@@ -38,19 +54,62 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
             ->setDescription('Creates a new form class')
             ->addArgument('name', InputArgument::OPTIONAL, sprintf('The name of the form class (e.g. <fg=yellow>%sType</>)', Str::asClassName(Str::getRandomTerm())))
+            ->addOption('entity', null, InputOption::VALUE_NONE, 'Generate form class from existing entity')
             ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeForm.txt'))
         ;
     }
 
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
+    {
+        $entityFormGeneration = $input->getOption('entity');
```
weaverryan: Hmm. This isn't quite what I had in mind, but we can talk about it :). What about this?

> Enter the class or entity name that the new form will be bound to (empty for none):

... and maybe that's it :). If they do not enter anything, then we can ask them for a form class name. Otherwise, we will select a name for them. We can auto-complete this field with entity names, but they should be allowed to enter *any* class name.

Also, I don't think we need the `entity` option. Let's *assume* that the `name` argument (if they pass it in the command) is the "class name" that they want the form bound to. In other words, we'll change the way this command words. It's not really a very useful command *anyways* if you're not binding it to a class.

src/GeneratorHelper.php
```diff
@@ -0,0 +1,104 @@
...
+{
+    public function getFormFieldsPrintCode(array $formFields): string
+    {
+        $printCode = '';
+
+        foreach ($formFields as $fieldName) {
+            $printCode .= <<<PHP
+            ->add('$fieldName')
+PHP;
+        }
```
weaverryan: I'd actually prefer for this to live inside the template: it's not very complex - just a loop.

src/GeneratorHelper.php
```diff
@@ -0,0 +1,104 @@
...
+            $printCode = $printCode.' ? \'Yes\' : \'No\'';
+        }
+
+        return $printCode;
+    }
+
+    public function getHead($baseLayoutExists, $title): string
+    {
+        if ($baseLayoutExists) {
+            $code = <<<TWIG
```
weaverryan: If you just `return <<<TWIG`, then the second part doesn't need to live in an `else` :)

src/GeneratorHelper.php
```diff
@@ -0,0 +1,104 @@
...
+    <head>
+        <title>$title</title>
+    </head>
+HTML;
+        }
+
+        return $code;
+    }
+
+    public function getBodyStart($baseLayoutExists): string
```
weaverryan: Actually, I think we can simplify: just remove `getBodyStart` and `getBodyEnd` and *always* put `{% block body %}` and `{% endblock %}`. We already do this in `make:controller`.

The reason is that it's technically valid HTML (at least, I believe, or valid "enough") to not have a `<body>` tag. And it's also valid to have a `{% block body %}` when you are not extending anything. So let's simplify.

src/GeneratorHelper.php
sadikoff: agreed, don't know why I added this loop in helper :)

src/Maker/MakeForm.php
sadikoff: I have just one question is this interaction with user useful? Lets do simple generation 
```
if (class_exists($entity_class_name)): 
  goto: generate_entity_form; 
else:
  goto: generate_simple_form; 
endif;
```
As I think interaction with user may allow us only generate several forms for one entity. If we need this, lets do 2 questions
First:
> The name of the form class (e.g. GentleElephantType):

and Second:
> Enter the class or entity name that the new form will be bound to (empty for none):

And if user passes name as argument than will assume that the inputs are equal.

weaverryan: Hmm, ok: let’s do your second option: ask the form class and then ask for the entity class. It’s simple and makes sense

src/GeneratorHelper.php
sadikoff: @weaverryan how do you think, may be it will be better to inject this helper in generator with `fileManager` and hide base_layout detection?

src/Maker/MakeForm.php
sadikoff: @weaverryan how do you think what way is better
- Question with autocomplete
```bash
 Enter the class or entity name that the new form will be bound to (empty for none):
 > SourFood
```
or
- Choice question
```bash
 Enter the class or entity name that the new form will be bound to [none]:
  [0] none
  [1] SourFood
 >
```

sadikoff:
`make:crud` is ready for next review

need some more time for `make:form`

sadikoff:
Lets discuss CRUD controller
is this good? or we need to change something?
```
 ------------------ ----------- ----------------- ------------------------
  Name               Method      Path              Controller
 ------------------ ----------- ----------------- ------------------------
  post_index         GET         /post/            PostController:index
  post_new           GET         /post/new         PostController:new
  post_create        POST        /post/            PostController:new
  post_show          GET         /post/{id}        PostController:show
  post_edit          GET         /post/{id}/edit   PostController:edit
  post_update        PATCH|PUT   /post/{id}        PostController:edit
  post_delete        DELETE      /post/{id}        PostController:delete
 ------------------ ----------- ----------------- ------------------------
```

weaverryan:
Hey @sadikoff!

I would say that we do not need separate routes for `post_new` and `post_create` and also for update. Let's use `post_new` and `post_edit` :)

sadikoff:
@weaverryan Agreed, just tried to implement crud as was mentioned in #3 and check if we need it.

I think that old implementation is good, and finally it will be something like: 
```
 ------------------ ----------- ----------------- ------------------------
  Name               Method      Path              Controller
 ------------------ ----------- ----------------- ------------------------
  post_index         GET         /post/            PostController:index
  post_new           GET|POST    /post/new         PostController:new
  post_show          GET         /post/{id}        PostController:show
  post_edit          GET|POST    /post/{id}/edit   PostController:edit
  post_delete        DELETE      /post/{id}        PostController:delete
 ------------------ ----------- ----------------- ------------------------
```

weaverryan:
I like this route setup indeed :).

So what are the next steps from here? Keep me posted so I can continue helping, reviewing, etc

sadikoff:
Today I'll push an update for make:form command, and than everything will be ready for complete review.

src/Maker/MakeCrud.php
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureCommand(Command $command, InputConfiguration $inputConfig)
+    {
+        $command
+            ->setDescription('Creates crud for Doctrine entity class')
+            ->addArgument('entity-class', InputArgument::OPTIONAL, sprintf('The class name of the entity to create crud (e.g. <fg=yellow>%s</>)', Str::asClassName(Str::getRandomTerm())))
```
weaverryan: We'll need to add the auto-complete for this as well, just like you did for `make:form`.

src/Maker/MakeCrud.php
```diff
@@ -0,0 +1,214 @@
...
+            $entityClassNameDetails->getRelativeNameWithoutSuffix(),
+            'Controller\\',
+            'Controller'
+        );
+
+        $formClassNameDetails = $generator->createClassNameDetails(
+            $entityClassNameDetails->getRelativeNameWithoutSuffix(),
+            'Form\\',
+            'Type'
+        );
```
weaverryan: One tricky thing is if you are generating a crud for `SourFood`, but you already have a `SourFoodType` class. It just fails with:

> The file "src/Form/SourFoodType.php" can't be generated because it already exists

As a user, that feels weird because I never asked for this class to be generated directly. I wonder if we need some logic to increment numbers - e.g. `SourFood2Type` - until we find one available. It's likely useful in other commands as well.

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -0,0 +1,104 @@
...
+use <?= $entity_full_class_name; ?>;
+use <?= $form_full_class_name; ?>;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @Route("<?= $route_path; ?>", name="<?= $route_name; ?>_")
```
weaverryan: Hmm, putting the `name=` here is pretty clever... but I personally would rather put the *full* route names below above each method. So, I like the path *prefix* up here, but not the name prefix.

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -0,0 +1,104 @@
+<?= "<?php\n"; ?>
+
+namespace <?= $namespace; ?>;
+
+use <?= $entity_full_class_name; ?>;
+use <?= $form_full_class_name; ?>;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
```
weaverryan: This stuff was actually JUST deprecated from that bundle. Instead, use the `Route` class from the core of Symfony : https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Routing/Annotation/Route.php

There is no `@Method` in the core - instead you should just use the `methods="` option.

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -0,0 +1,104 @@
...
+ * @Route("<?= $route_path; ?>", name="<?= $route_name; ?>_")
+ */
+class <?= $class_name ?> extends Controller
+{
+    /**
+     * @Route("/", name="index")
+     *
+     * @return Response
+     */
+    public function index()
```
weaverryan: I have another trick one for you 😇. IF there is a custom repository class, I would rather type-hint it as an argument to the method:

```php
public function index(SourFoodRepository $repository)
{

}
```

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -0,0 +1,104 @@
...
+        $<?= $entity_var_singular; ?> = new <?= $entity_class_name; ?>();
+        $form = $this->createForm(<?= $form_class_name; ?>::class, $<?= $entity_var_singular; ?>);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $em = $this->getDoctrine()->getManager();
+            $em->persist($<?= $entity_var_singular; ?>);
+            $em->flush();
+
+            return $this->redirectToRoute('<?= $route_name; ?>_edit', ['<?= $entity_identifier; ?>' => $<?= $entity_var_singular; ?>->get<?= ucfirst($entity_identifier); ?>()]);
```
weaverryan: What do you think about flash messages? It would be great to add them... but we don't actually know whether or not the user is printing flash messages in `base.html.twig`. So, it's tricky... I'm not sure personally...

src/Resources/skeleton/crud/templates/edit.tpl.php
```diff
@@ -0,0 +1,13 @@
+<?= $helper->getHead($base_layout_exists, 'Edit '.$entity_class_name); ?>
+
+{% block body %}
+
+    <h1>Edit <?= $entity_class_name; ?></h1>
+
+    {% include '<?= $route_name; ?>/_form.html.twig' with {'form': form, 'button_label': 'Edit'} only %}
```
weaverryan: use `{{ include() }}` instead - considered a better practice :)

src/Resources/skeleton/crud/templates/show.tpl.php
```diff
@@ -0,0 +1,23 @@
+<?= $helper->getHead($base_layout_exists, $entity_class_name) ?>
+
+{% block body %}
+
+    <h1><?= $entity_class_name ?></h1>
+
+    <table>
```
weaverryan: Missing the `<tbody>`

src/Resources/skeleton/crud/templates/show.tpl.php
```diff
@@ -0,0 +1,23 @@
...
+<?php foreach ($entity_fields as $field): ?>
+        <tr>
+            <th><?= ucfirst($field['fieldName']) ?></th>
+            <td>{{ <?= $helper->getEntityFieldPrintCode($entity_var_singular, $field) ?> }}</td>
+        </tr>
+<?php endforeach; ?>
+
+    </table>
```
weaverryan: Extra line break here

src/Resources/skeleton/crud/templates/show.tpl.php
```diff
@@ -0,0 +1,23 @@
...
+
+    <a href="{{ path('<?= $route_name ?>_index') }}">back to list</a>
+
+    <a href="{{ path('<?= $route_name ?>_edit', {'<?= $entity_identifier ?>': <?= $entity_var_singular ?>.<?= $entity_identifier ?>}) }}">edit</a>
+
+    {% include '<?= $route_name ?>/_delete_form.html.twig' with {'identifier': <?= $entity_var_singular ?>.<?= $entity_identifier ?>} only %}
```
weaverryan: Any special reason to take the effort to specifically pass in the `identifier` variable AND use the `only` part? I mean it's more explicit, but looks more complex. But I thought you might also have a specific reason.

chapterjason:
What about the possibility that `/new` getting matched instead of `/{Id}` the requirement to be an number could be possible, but what if the user changes the id to anything like a slug/name?

sadikoff:
@chapterjason lets think that `/new` is reserved route. I don't think that user will have some `slug` with `new` value.

src/Doctrine/DoctrineEntityHelper.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        return null !== $this->metadataFactory;
+    }
+
+    /**
+     * @return array
+     */
+    public function getEntitiesForAutocomplete(): array
+    {
+        $entities = ['none'];
```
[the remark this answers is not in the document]
sadikoff: sorry "none" value is a junk code, from one of my experiments.

src/Maker/MakeForm.php
```diff
@@ -41,6 +51,25 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
             ->addArgument('name', InputArgument::OPTIONAL, sprintf('The name of the form class (e.g. <fg=yellow>%sType</>)', Str::asClassName(Str::getRandomTerm())))
             ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeForm.txt'))
         ;
+
+        $inputConf->setArgumentAsNonInteractive('name');
```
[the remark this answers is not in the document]
sadikoff: yes it can be interactive and it was interactive. There were some problems with detecting if user entered `name` as argument. I need it to fast form generation with only one argument. 

I think if we add a second argument it will be more simple.

src/Resources/skeleton/crud/controller/Controller.tpl.php
sadikoff: It was already  done :) I had to test something before push it

src/Resources/skeleton/crud/controller/Controller.tpl.php
sadikoff: agreed 👍

src/Resources/skeleton/crud/controller/Controller.tpl.php
sadikoff: I have an idea about it. I'll push it soon

src/Resources/skeleton/crud/templates/show.tpl.php
sadikoff: Just thought about some template optimization. I don't think that we must pass whole context to template which doesn't use it at all

src/Maker/MakeForm.php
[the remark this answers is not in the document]
weaverryan: I think my preference would be: 2 arguments (`name` and `bound-class`). Then I think we can let the base maker command do all the normal interactive stuff for us. Actually, we could make `bound-class` an option - e.g .`--bound-class`, as it might make passing "nothing" easier if you don't want one: e.g. `--bound-class=`.

Btw, in general, I'm not very concerned with making the commands work in a *non* interactive state. It causes too much complexity.

src/Maker/MakeForm.php
```diff
@@ -41,6 +51,25 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
+    {
+        if (null === $input->getArgument('name')) {
+            $argument = $command->getDefinition()->getArgument('name');
+            $question = new Question($argument->getDescription());
+            $value = $io->askQuestion($question);
+
+            $input->setArgument('name', $value);
+
+            if ($this->entityHelper->isDoctrineInstalled()) {
```
weaverryan: We need to ask the below question in ALL cases, because the user may not use Doctrine, but may want to bind their form to a model class. In that case, of course, the `$this->entityHelper->getEntitiesForAutocomplete()` should return an empty array.

src/Resources/skeleton/crud/templates/edit.tpl.php
```diff
@@ -0,0 +1,24 @@
...
+<style>
+    .example-wrapper { margin: 1em auto; max-width: 1018px; width: 95%; }
+    .example-wrapper form div { margin: 1em 0; }
+    .example-wrapper form label { display: block; }
+    .example-actions form, .example-actions a { display: inline-block; margin: 0 2px; }
+</style>
+{% endblock %}
+
+{% block body %}
+    <div class="example-wrapper">
```
weaverryan: I don't think we should add this class, or the CSS above it. I *do* like the idea of the CRUD not looking *horrible* initially... but the user WILL need to customize this and make it their own. So, we should not add any extra stuff that they will need to remove (or, we should be VERY minimalistic).

Actually, I talked to Javier, and we both agree that we should add a few *basic* Bootstrap classes  e.g .`<table class="table"` and `<button class="btn">`. But, I think that's all we should include - we shouldn't package our own CSS.

src/Resources/skeleton/crud/templates/edit.tpl.php
```diff
@@ -0,0 +1,24 @@
...
+        <h1>Edit <?= $entity_class_name ?></h1>
+
+        {{ include('<?= $route_name ?>/_form.html.twig', {'form': form, 'button_label': 'Edit'}) }}
+
+        <div class="example-actions">
```
weaverryan: And same here - you can remove this

src/Resources/skeleton/crud/templates/edit.tpl.php
```diff
@@ -0,0 +1,24 @@
...
+            <a href="{{ path('<?= $route_name ?>_index') }}">back to list</a>
+
+            {{ include('<?= $route_name ?>/_delete_form.html.twig', {'<?= $entity_var_singular ?>': <?= $entity_var_singular ?>}) }}
```
weaverryan: No need to pass the variable in explicitly - just `{{ include(....) }}` and that's it - no second argument :)

src/Resources/skeleton/crud/templates/_form.tpl.php
```diff
@@ -0,0 +1,6 @@
+{{ form_start(form) }}
+    {{ form_widget(form) }}
+    <div>
```
weaverryan: Remove the div?

src/Resources/skeleton/crud/templates/_delete_form.tpl.php
```diff
@@ -0,0 +1,5 @@
+<form method="post" action="{{ path('<?= $route_name ?>_delete', {'<?= $entity_identifier ?>': <?= $entity_var_singular ?>.<?= $entity_identifier ?>}) }}" onsubmit="return confirm('Are you sure you want to delete this item?');">
+    <input type="hidden" name="_method" value="DELETE">
+    <input type="hidden" name="_token" value="{{ csrf_token('delete' ~ <?= $entity_var_singular ?>.<?= $entity_identifier ?>) }}">
+    <input type="submit" value="Delete">
```
weaverryan: Could we use a `<button` here?

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -0,0 +1,105 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @Route("<?= $route_path ?>")
+ */
+class <?= $class_name ?> extends Controller
+{
+    /**
+     * @Route("/", name="<?= $route_name ?>_index", methods="GET"): Response
```
weaverryan: The `: Response` shouldn't be here

src/GeneratorTwigHelper.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @internal
+ */
+class GeneratorTwigHelper
+{
+    private $baseLayoutExists;
+
+    public function __construct(FileManager $fileManager)
+    {
+        $this->baseLayoutExists = $fileManager->fileExists('templates/base.html.twig');
```
weaverryan: This is something you should calculate not in `__construct()` but only when you need it (so, down in `getHeadPrintCode()`). It doesn't really make a difference, but it's a better practice.

src/GeneratorTwigHelper.php
```diff
@@ -0,0 +1,65 @@
...
+            $printCode = $printCode.' ? '.$printCode.'|date(\'Y-m-d H:i:s\') : \'\'';
+        } elseif (in_array($field['type'], ['date'])) {
+            $printCode = $printCode.' ? '.$printCode.'|date(\'Y-m-d\') : \'\'';
+        } elseif (in_array($field['type'], ['time'])) {
+            $printCode = $printCode.' ? '.$printCode.'|date(\'H:i:s\') : \'\'';
+        } elseif (in_array($field['type'], ['array'])) {
+            $printCode = $printCode.' ? '.$printCode.'|join(\', \') : \'\'';
+        } elseif (in_array($field['type'], ['boolean'])) {
+            $printCode = $printCode.' ? \'Yes\' : \'No\'';
+        }
```
weaverryan: We could clean this up by changing this to a `switch ($field['type'])`

src/GeneratorTwigHelper.php
```diff
@@ -0,0 +1,65 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle;
+
+/**
+ * @author Sadicov Vladimir <sadikoff@gmail.com>
+ *
+ * @internal
```
weaverryan: This one is actually *not* internal, as we're exposing it inside of the templates as a variable

src/GeneratorTwigHelper.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+class GeneratorTwigHelper
```
weaverryan: Add `final class...`

src/Maker/MakeForm.php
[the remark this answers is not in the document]
weaverryan: Well actually, you might still need to ask the `bound-class` value manually in `interact` so we can do some auto-completion :)

src/Maker/MakeCrud.php
```diff
@@ -0,0 +1,226 @@
...
+        $entityClassDetails = $generator->createClassNameDetails(
+            $input->getArgument('entity-class'),
+            'Entity\\'
+        );
+
+        $repositoryClassDetails = $generator->createClassNameDetails(
+            $entityClassDetails->getRelativeName(),
+            'Repository\\',
+            'Repository'
+        );
```
weaverryan: I think this is not quite right. The user could have setup their repository class to be *any* class - it may not follow the pattern we're expecting (and that's ok!). We need to read the actual customRepositoryClass from the entity metadata and use that.

src/Resources/skeleton/crud/templates/new.tpl.php
```diff
@@ -0,0 +1,22 @@
+<?= $helper->getHeadPrintCode('New '.$entity_class_name) ?>
+
+{% block stylesheets %}
+<style>
```
sblondeau: style should not be here
I think it would be better to remove it and classes example-*

fictiont:
@sadikoff how could i use maker:crud now in symfony4?

src/Resources/skeleton/crud/templates/edit.tpl.php
sadikoff: ok good. It was just my experiment :) will add simple bootstrap classes

src/Resources/skeleton/crud/templates/new.tpl.php
sadikoff: Yes =) just experiment I'll remove this stuff completely :)

sadikoff:
@fictiont while this PR is not ready you can try to use my [koff/crud-maker-bundle](https://packagist.org/packages/koff/crud-maker-bundle)

src/Maker/MakeCrud.php
sadikoff: really I didn't think about it. Will improve it. And I have one little idea about `Form`.

sadikoff:
@javiereguiluz thanks for your review, it's really important for me! 
I would like to discuss your last point about minimal CSS. Is it will be enough to style links as buttons with `class="bth"` and may be add `class="form-inline"` to delete form?

javiereguiluz:
Not sure what to do, honestly. I'd say adding some minimal CSS classes (compatible with Bootstrap) would be nice for most people ... but maybe my guess is wrong and we make most people angry. @weaverryan what do you think about this? Thanks!

jlcd:
Hey there @sadikoff , first of all thanks for the great work you've put in.

I've just tried to make a crud here and noticed that for my `Location` Entity it didn't quite work as expected.
The Entity in question has only one field, `location`, being also the Entity's ID.

When I ran `make:crud Location`, the controller and forms were created, but the new and edit forms are broken as they don't have the only table's field. Maybe it has something to do with the field being the ID field, not sure though.

Here's the Entity:

```
<?php

namespace App\Entity;

/**
 * @ORM\Entity(repositoryClass="App\Repository\LocationRepository")
 */
class Location
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="NONE")
     * @ORM\Column(type="string", length=60)
     */
    private $location;

    /**
     * @return string
     */
    public function getLocation()
    {
        return $this->location;
    }

    /**
     * @param string $location
     *
     * @return self
     */
    public function setLocation($location)
    {
        $this->location = $location;

        return $this;
    }
}
```

sadikoff:
Hello @jlcd I've just tested your entity with last commited version here and everything works perfect. May be you are using my `koff\crud-maker-bundle`? Possibly there is a bug with some type of entities.

src/Maker/MakeCrud.php
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $entityClassDetails = $generator->createClassNameDetails(
+            $input->getArgument('entity-class'),
+            'Entity\\'
+        );
+
+        $entityDoctrineDetails = $this->entityHelper->createDoctrineDetails($entityClassDetails->getFullName());
```
weaverryan: I think we're missing some validation here to make sure that the class the user passed IS indeed a valid entity.

src/Doctrine/DoctrineEntityDetails.php
```diff
@@ -0,0 +1,52 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Doctrine;
+
+/**
+ * @author Sadicov Vladimir <sadikoff@gmail.com>
+ */
+final class DoctrineEntityDetails
```
weaverryan: I think this class can also be internal - i don't think you're passing it directly into the templates (which is good)

sadikoff:
Everything ready!

Thanks to all for reviews and discussion!

weaverryan:
Merged! And already tagged: https://github.com/symfony/maker-bundle/releases/tag/v1.2.0

Thanks you VERY much @sadikoff for very hard work, handling a lot of feedback and doing it quickly! I hope we can keep improving MakerBundle together :).

PR #116, opened 2018-02-07
Type hints for repositories

enleur:
It helps IDE to get correct type if you do calls like this

```php
$user = $this->userRepository->find(1);
$user->getId(); //IDE knows that this is a User class and gives correct autocomplete
```


enleur:
Should I fix code style?

weaverryan:
Yes please :). Unless it’s recommending to use array() instead of [] - that is currently a false recommendation (we use the short syntax)

enleur:
Done

weaverryan:
Thank you Sergey!

PR #119, opened 2018-02-08
Running phpcs on Travis

weaverryan:
Usually this is done by fabbot, but it mis-reports all of our `[]` as invalid. I'd rather not use fabbot (it makes all of our PR's look like build failures) and just validate the CS on the tests.


weaverryan:
@fabpot could you disable fabbot for this repository?

PR #127, opened 2018-02-27
Fixed CS of generated entity

lyrixx:


javiereguiluz:
For what it's worth, in the Symfony Docs we don't add parenthesis when there's no arguments. Example: https://symfony.com/doc/current/doctrine.html#creating-an-entity-class

Maybe we must change Symfony Docs. I'm not sure.

lyrixx:
> Maybe we must change Symfony Docs. I'm not sure.

I think. We never use `$date = new DateTime`. It's the same for annotation ;)

weaverryan:
This is cool with me - I slightly prefer with `()`

weaverryan:
Thanks Grégoire!

PR #129, opened 2018-03-01
Update index.rst

tavy315:
$ composer require maker --dev

  [InvalidArgumentException]
  Could not find package maker.

  Did you mean one of these?
      yab/crudmaker
      yab/formmaker
      greabock/maker
      osflab/doc-maker
      nickwest/form-maker

### DIFF
--- a/src/Resources/doc/index.rst
+++ b/src/Resources/doc/index.rst
@@ -13,7 +13,7 @@ Run this command to install and enable this bundle in your application:
 
 .. code-block:: terminal
 
-    $ composer require maker --dev
+    $ composer require symfony/maker-bundle --dev
 
 Usage
 -----


stof:
This example in the readme assumes that this is used in a Flex project (flex will make `maker` work)

weaverryan:
Hmm. Stof is correct. The question is, should we use the longer, but more universal package name?

tavy315:
Documentation for all the rest of the bundles use the longer package name (ex: https://symfony.com/doc/current/bundles/DoctrineBundle/installation.html, https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/index.html, https://symfony.com/doc/current/bundles/SensioGeneratorBundle/index.html etc).

weaverryan:
Yea, let's use the version that works for everyone :). Thanks @tavy315!

PR #132, opened 2018-03-15
make:crud skeleton optimization

sadikoff:
Lets start from `Controller::delete` action. I don't think we need 2 return statements here.

Maybe it will be better to add some flash messages?

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -23,7 +23,7 @@ class <?= $class_name ?> extends Controller
 <?php if (isset($repository_full_class_name)): ?>
     public function index(<?= $repository_class_name ?> $<?= $repository_var ?>): Response
     {
-        return $this->render('<?= $route_name ?>/index.html.twig', ['<?= $entity_var_plural ?>' => $<?= $repository_var ?>->findAll()]);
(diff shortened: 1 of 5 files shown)


sadikoff:
I think we don't need to know if the user is printing them or not, We just add them in `index` and `edit` templates and each user will decide to use it or not.

weaverryan:
@sadikoff I'm not sure about printing the flash messages in `index.html.twig` and `edit.html.twig`... because the user will *always* need to remove those... as flash messages should *really* live in the base layout.... That's what complicates things for me.

sadikoff:
@weaverryan maybe than we can ask user to use flash or not, and inject them in `base` layout including something like `_flash.html.twig`?

weaverryan:
@sadikoff We don't have any ability to modify Twig files right now. Also, I want to avoid asking unnecessary questions.... but *maybe* we should ask them if they want flash messages or not.

Currently, I'm leaning towards this: we always add flash messages in the controller. Then, we can do a quick check in their `base.html.twig`: if we do not see the flash messages being printed there (we could do some simple string check, not perfect, but good enough), we could print out a warning that tells them they should add this to their base.html.twig.

sadikoff:
@weaverryan If you don't mind lets merge this PR. It will fix some bugs. After I will prepare PR for flash messages and we will discuss it there.

weaverryan:
Thank you Vladimir!

PR #135, opened 2018-03-17
Add a trailing slash to $route_path to avoid "No route found" error

Dotcra:
Request to a controller which generated by "make:controller" with a trailing slash cause "No route found" error.
Add a trailing slash to $route_path in src/Resources/skeleton/controller/Controller.tpl.php since ["adding a trailing slash in the route path is the best way to ensure that both URLs work."](http://symfony.com/doc/current/routing.html#redirecting-urls-with-trailing-slashes)


Dotcra:
sorry about missing the quotes, close this one and create another pl

PR #136, opened 2018-03-17
Add a trailing slash to $route_path to avoid "No route found" error

Dotcra:
Request with a trailing slash to a controller generated by "make:controller" cause "No route found" error.

Add a trailing slash to $route_path in 'src/Resources/skeleton/controller/Controller.tpl.php' since ["adding a trailing slash in the route path is the best way to ensure that both URLs work."](http://symfony.com/doc/current/routing.html#redirecting-urls-with-trailing-slashes)

### DIFF
--- a/src/Resources/skeleton/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/controller/Controller.tpl.php
@@ -8,7 +8,7 @@ use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 class <?= $class_name; ?> extends Controller
 {
     /**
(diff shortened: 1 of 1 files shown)


weaverryan:
Hmm, this is true: creating a route like `/foo/` means that going to `/foo` or `/foo/` will work. But the opposite situation does not work.

However, this was fixed in Symfony 4.1: https://symfony.com/blog/new-in-symfony-4-1-smarter-url-redirections. I think we should keep it as is: the user is free to add the trailing slash if they want.

Dotcra:
It's great.
Okay, so I guess I can close this pr.

PR #139, opened 2018-03-20
Fix expected variable type

IndraGunawan:
error when run `doctrine:schema:update --dump-sql` after creating entity using maker-bundle because of the attribute type error.

for example when generate field with length 100 the annotation will be generate like this
`@ORM\Column(type="string", length="100", nullable=true)` and when run the schema update command will throw exception 
```
In AnnotationException.php line 83:

  [Type Error] Attribute "length" of @ORM\Column declared on property App\Entity\Entity::$field expects a(n) integer, but got string.
```

and it is also affect for `precision` and `scale`

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -19,4 +19,10 @@
             <exclude>tests/tmp</exclude>
         </testsuite>
     </testsuites>
+
+    <filter>
+        <whitelist>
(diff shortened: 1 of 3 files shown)


weaverryan:
Wow! This is an absolutely *wonderful* PR! Fixed the bug & a great test. Thank you @IndraGunawan!

PR #140, opened 2018-03-20
[WIP - Experimental] Superfast tests

sadikoff:
Test PR to see if tests work as expected

### DIFF
--- a/src/Test/MakerTestDetails.php
+++ b/src/Test/MakerTestDetails.php
@@ -22,6 +22,8 @@ final class MakerTestDetails
 
     private $fixtureFilesPath;
 
+    private $ignoredFiles = [];
+
     private $replacements = [];
 
     private $preMakeCommands = [];
@@ -72,6 +74,18 @@ final class MakerTestDetails
         $this->postMakeCommands[] = $postMakeCommand;
 
         return $this;
+    }
+
+    public function ignoreFile(string $filename): self
+    {
+        $this->ignoredFiles[] = $filename;
+
+        return $this;
+    }
+
+    public function getIgnoredFiles(): array
+    {
+        return $this->ignoredFiles;
     }
 
     public function addReplacement(string $filename, string $find, string $replace): self
--- a/src/Test/MakerTestDetails.php
+++ b/src/Test/MakerTestDetails.php
@@ -22,7 +22,7 @@ final class MakerTestDetails
 
     private $fixtureFilesPath;
 
-    private $ignoredFiles = [];
+    private $deletedFiles = [];
 
     private $replacements = [];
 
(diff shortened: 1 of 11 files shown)


src/Test/MakerTestDetails.php
```diff
@@ -74,6 +76,18 @@ public function addPostMakeCommand(string $postMakeCommand): self
...
+    {
+        $this->deletedFiles[] = $filename;
+
+        return $this;
+    }
+
+    public function getFilesToDelete(): array
+    {
+        return $this->deletedFiles;
+    }
```
weaverryan: Nice addition

src/Test/MakerTestEnvironment.php
```diff
@@ -0,0 +1,345 @@
...
+        if (!$this->fs->exists($cachePath)) {
+            $this->fs->mkdir($cachePath);
+        }
+
+        $this->cachePath = realpath($cachePath);
+        $this->flexPath = $this->cachePath.DIRECTORY_SEPARATOR.'flex_project';
+
+        $this->path = $this->cachePath.DIRECTORY_SEPARATOR.$testDetails->getUniqueCacheDirectoryName();
+
+        $this->snapshotFile = $this->path.DIRECTORY_SEPARATOR.basename($this->path).'.json';
```
weaverryan: I'm 99% sure there should be no problem with using `/` instead of `DIRECTORY_SEPARATOR`. In fact, I think using `/` is a Symfony standard.

src/Test/MakerTestEnvironment.php
```diff
@@ -0,0 +1,345 @@
...
+namespace Symfony\Bundle\MakerBundle\Test;
+
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\InputStream;
+
+/**
+ * @author Sadicov Vladimir <sadikoff@gmail.com>
+ */
```
weaverryan: `@internal`

Nice idea to extract this to a class, btw.

src/Test/MakerTestProcess.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+namespace Symfony\Bundle\MakerBundle\Test;
+
+use Symfony\Component\Process\Process;
+
+/**
+ * @author Sadicov Vladimir <sadikoff@gmail.com>
+ */
+final class MakerTestProcess
```
weaverryan: `@internal`

src/Test/MakerTestEnvironment.php
sadikoff: Agreed! Just some perfectionism :)

sadikoff:
Using `/` instead of `DIRECTORY_SEPARATOR` breaks AppVeyor tests :( I think we got 1% problem with paths on Windows :)

PS: I knew that it would be so, but I did not know where exactly it would break :)

sadikoff:
That was my fault. I use snapshots to avoid copy operations. There are full filenames and that's why I need paths with the correct slashes.

weaverryan:
Thank you @sadikoff! You've made the tests faster than I thought possible :)

PR #147, opened 2018-03-31
Ignorning tests so it is not downloaded via Composer

### LINKED ISSUE #144: IDE inspection multiple definitions of tests namespace
Since the latest updates I've got some issues with the namespace in my IDE (PHPStorm).

I always get the `multiple definitions found for class User` warning, cause I have my own `App\Entity\User` class and the tests also have this namespace. It would be really useful to change the namespaces to something like `Symfony\Bundle\MakerBundle\Tests\Fixtures\Entity\User` in the tests.

weaverryan:
This is different than the policy in symfony/symfony. But, this library is a bit different:  the tests do *not* contain example usage code, as most of the classes are internal. By not packing the tests/ directory, we avoid polluting the user's project with extra User classes (etc) that confuse IDE's.

Fixes #144

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/tests export-ignore


javiereguiluz:
Thanks Ryan! I agree that it makes sense to make an exception about Symfony's policy in this repo.

PR #149, opened 2018-04-11
Improve Doctrine integration Fix #146, #152

### LINKED ISSUE #146: make:* commands fail on entities using Single Table Inheritance
I have the following two entities in a Symfony 4.0.6 application (renamed to match the docs for the sake of this issue) using Doctrine's [Single Table Inheritance](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html#single-table-inheritance) feature;

```
App\Entity\Person:
  type: entity
  table: person
  fields:
    id:
      type: integer
      id: true
      generator:
        strategy: AUTO
    name:
      type: string
      length: 250
      nullable: true
  inheritanceType: SINGLE_TABLE
  discriminatorColumn:
    name: person_type
    type: string
    length: 250
  discriminatorMap:
    thread: App\Entity\Employee
    reply: App\Entity\Manager
```

```
App\Entity\Employee:
  type: entity
  table: ~
  fields:
    ...
```

Running, for example `bin/console make:form EmployeeType Employee` presents the following error;

>In DoctrineEntityHelper.php line 67:
  Notice: Undefined offset: 0

The relevant code;

```
$metadata = $this->getEntityMetadata($entityClassName);

if (null !== $metadata) {
    return new DoctrineEntityDetails(
        $metadata->customRepositoryClassName,
        $metadata->identifier[0], // Error line
        $metadata->fieldMappings,
        $this->getFormFieldsFromEntity($metadata)
    );
}
```

Basically, because the entity metadata does not have an ID (as it exists in the parent) it errors. It doesn't look as though the `getIdentifier` method of `DoctrineEntityDetails` is in use anywhere anyway so I'm not sure what the identifier is required for. Temporarily replacing the line causing error results in the form being successfully created as expected. 

If the `make` features can't work without entities having an identifier, it may be worth adding an error to describe that. Thanks!

sadikoff:
This will close #146 

in progress...


src/Doctrine/EntityRegenerator.php
```diff
@@ -24,39 +23,31 @@
...
-    public function __construct(ManagerRegistry $doctrineRegistry, FileManager $fileManager, Generator $generator, string $projectDirectory, bool $overwrite)
+    public function __construct(DoctrineHelper $doctrineHelper, FileManager $fileManager, Generator $generator, string $projectDirectory, bool $overwrite)
     {
-        $this->doctrineRegistry = $doctrineRegistry;
+        $this->doctrineHelper = $doctrineHelper;
         $this->fileManager = $fileManager;
         $this->generator = $generator;
         $this->projectDirectory = $projectDirectory;
         $this->overwrite = $overwrite;
-        $this->metadataFactory = new DoctrineMetadataFactory($this->doctrineRegistry);
```
weaverryan: The `DoctrineMetadataFactory` calls should be deleted entirely now, correct?

src/Doctrine/DoctrineHelper.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * @param string $entityClassName
+     *
+     * @return null|EntityDetails
+     *
+     * @throws \Exception
+     */
+    public function createDoctrineDetails(string $entityClassName)
```
weaverryan: you can use a return type here instead of `@returns`

src/Doctrine/DoctrineHelper.php
```diff
@@ -0,0 +1,172 @@
...
+            foreach (array_keys($allMetadata) as $classname) {
+                $entityClassDetails = new ClassNameDetails($classname, 'App\\Entity');
+                $entities[] = $entityClassDetails->getRelativeName();
+            }
+        }
+
+        return $entities;
+    }
+
+    public function getMetadata(string $classOrNamespace = null, bool $disconnected = false)
```
weaverryan: Add a return type here too

tests/fixtures/MakeFormEmbedableEntity/src/Entity/Recept.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * @ORM\Embeddable()
+ */
+class Recept
+{
+    /**
+     * @ORM\Column(type="string", length=255)
+     */
+    private $ingridients;
```
weaverryan: Minor: `ingredients` :)

src/Doctrine/EntityRegenerator.php
sadikoff: yes, and DoctrineMetadataFactory class too I think

tests/fixtures/MakeFormEmbedableEntity/src/Entity/Recept.php
sadikoff: oops :)

weaverryan:
Very nice, very careful work. Thank you @sadikoff!

MrNicodemuz:
@weaverryan I think this introduced a bug:

Undefined class constant 'ONE_TO_MANY'

Issue #148, opened 2018-04-08
make:entity class name got trimmed by D


chapterjason:
While I worked on a project I found the following. The class name got trimmed by the `D`.

```
 Add another property? Enter the property name (or press <return> to stop adding fields):
 > directory

 Field type (enter ? to see all types) [string]:
 > relation

 What class should this entity be related to?:
 > Directory

What type of relationship is this?
 ------------ --------------------------------------------------------------- 
  Type         Description                                                    
 ------------ --------------------------------------------------------------- 
  ManyToOne    Each File relates to (has) one irectory.                       
               Each irectory can relate/has to (have) many File objects       
                                                                              
  OneToMany    Each File relates can relate to (have) many irectory objects.  
               Each irectory relates to (has) one File                        
                                                                              
  ManyToMany   Each File relates can relate to (have) many irectory objects.  
               Each irectory can also relate to (have) many File objects      
                                                                              
  OneToOne     Each File relates to (has) exactly one irectory.               
               Each irectory also relates to (has) exactly one File.          
 ------------ --------------------------------------------------------------- 

 Relation type? [ManyToOne, OneToMany, ManyToMany, OneToOne]:
 > ManyToOne

 Is the File.directory property allowed to be null (nullable)? (yes/no) [yes]:
 > no

 Do you want to add a new property to irectory so that you can access/update File objects from it - e.g. $irectory->getFiles()? (yes/no) [yes]:
 > 

 A new property will also be added to the irectory class so that you can access the related File objects from it.

 New field name inside irectory [files]:
 > 

 Do you want to activate orphanRemoval on your relationship?
 A File is "orphaned" when it is removed from its related irectory.
 e.g. $irectory->removeFile($file)
 
 NOTE: If a File may *change* from one irectory to another, answer "no".

 Do you want to automatically delete orphaned App\Entity\File objects (orphanRemoval)? (yes/no) [no]:
 > yes


In ClassSourceManipulator.php line 662:
                                 
  Could not find namespace node  
                                 

make:entity [--regenerate] [--overwrite] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<name>]


```

weaverryan:
Wow! This is totally weird! Not only does the “d” become missing suddenly in the messages, but it erroree at the end! What entity were you generating this into? Can you show the source?

[...]

PR #150, opened 2018-04-11
make entity, related class without namespace

LeJeanbono:
For #148


weaverryan:
This is great! Definitely a bug fix, and with a test. Awesome and thank you @LeJeanbono!

LeJeanbono:
Your welcome !
Nice bundle

PR #155, opened 2018-04-19
Re-adding PHP 7.0 support, which works for most commands

weaverryan:
PHP 7.0 support was removed when I updated the make:entity command. This is because the make:entity command generates PHP 7.1+ compatible code only (it uses nullable ?).

But, having 7.0 support is easy... because 7.1 is only needed for this one command.


src/Maker/MakeEntity.php
```diff
@@ -110,6 +110,10 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
 
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
     {
+        if (version_compare(PHP_VERSION, '7.1.0') < 0) {
```
stof: use `\PHP_VERSION_ID` instead, as done in Symfony.

src/Test/MakerTestCase.php
```diff
@@ -12,11 +12,16 @@
 namespace Symfony\Bundle\MakerBundle\Test;
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\MakerBundle\Maker\MakeEntity;
 
 class MakerTestCase extends TestCase
 {
     protected function executeMakerCommand(MakerTestDetails $testDetails)
     {
+        if ($testDetails->getMaker() instanceof MakeEntity && version_compare(PHP_VERSION, '7.1.0') < 0) {
```
stof: This MakerTestCase is a reusable one for any maker (including third-party ones). Hardcoding a special case for MakeEntity looks weird to me. I suggest adding a way to mark the test as skipped in MakerTestDetails instead.

src/Util/ClassSourceManipulator.php
```diff
@@ -504,7 +504,11 @@ private function addStatementToConstructor(Node\Stmt $stmt)
         $this->updateSourceCodeFromNewStmts();
     }
 
-    private function getConstructorNode(): ?Node\Stmt\ClassMethod
+    /**
+     * @return null|Node\Stmt\ClassMethod
```
stof: we generally put `|null` at the end in Symfony.

src/Test/MakerTestCase.php
```diff
@@ -12,14 +12,13 @@
...
-use Symfony\Bundle\MakerBundle\Maker\MakeEntity;
 
 class MakerTestCase extends TestCase
 {
     protected function executeMakerCommand(MakerTestDetails $testDetails)
     {
-        if ($testDetails->getMaker() instanceof MakeEntity && version_compare(PHP_VERSION, '7.1.0') < 0) {
-            $this->markTestSkipped('MakeEntity tests require 7.1 or higher');
+        if (!$testDetails->isSupportedByCurrentPhpVersion()) {
+            $this->markTestSkipped();
```
stof: skipping should have a message.

src/Test/MakerTestDetails.php
```diff
@@ -245,4 +254,9 @@ public function isCommandAllowedToFail(): bool
     {
         return $this->commandAllowedToFail;
     }
+
+    public function isSupportedByCurrentPhpVersion()
```
stof: `: bool`

Issue #156, opened 2018-04-19
Undefined class constant 'ONE_TO_MANY'


MrNicodemuz:
I'm trying to run the following command:

    php bin/console make:crud

But end up getting the following error:

    In EntityDetails.php line 66:

    [Symfony\Component\Debug\Exception\FatalThrowableError]
    Undefined class constant 'ONE_TO_MANY'

    Exception trace:
    Symfony\Bundle\MakerBundle\Doctrine\EntityDetails->getFormFields() at /symfony/vendor/symfony/maker-bundle/src/Maker/MakeCrud.php:155
    Symfony\Bundle\MakerBundle\Maker\MakeCrud->generate() at /symfony/vendor/symfony/maker-bundle/src/Command/MakerCommand.php:93
    Symfony\Bundle\MakerBundle\Command\MakerCommand->execute() at /symfony/vendor/symfony/console/Command/Command.php:252
    Symfony\Component\Console\Command\Command->run() at /symfony/vendor/symfony/console/Application.php:883
    Symfony\Component\Console\Application->doRunCommand() at /symfony/vendor/symfony/framework-bundle/Console/Application.php:86
    Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /symfony/vendor/symfony/console/Application.php:241
    Symfony\Component\Console\Application->doRun() at /symfony/vendor/symfony/framework-bundle/Console/Application.php:74
    Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /symfony/vendor/symfony/console/Application.php:143
    Symfony\Component\Console\Application->run() at /symfony/bin/console:39`

sadikoff:
Sorry it's my fault :( already fixed, but we need to wait @weaverryan

PR #157, opened 2018-04-19
fix ClassMetadata namespace

sadikoff:
fix: #156


weaverryan:
Thank you @sadikoff! I guess our tests missed this? Could we add some coverage in another PR to avoid issues in the future?

sadikoff:
I'll do it tomorrow!

PR #161, opened 2018-04-21
Decimal Type: Fix small typo

mercuryseries:
Scale value currently overwrites the previously entered precision value.


weaverryan:
Oh no! Great catch, and thanks for the PR!

PR #163, opened 2018-04-22
Get the autoloader class from current registered autoload functions

silvester:
If the default folder for vendor is changed the AutloaderUtil class is no more able to find the ClassLoader. Testing the failure can be easily done by changing the **vendor-dir** in composer.json. After adjusting the autoload require in console file, all makers that use AutoloadUtil will fail.

Motivation behind this is development with docker. The application is a lot faster on windows and mac if the vendor folder is not a shared folder.

### DIFF
--- a/src/Util/AutoloaderUtil.php
+++ b/src/Util/AutoloaderUtil.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Util;
 
 use Composer\Autoload\ClassLoader;
+use Symfony\Component\Debug\DebugClassLoader;
 
 /**
  * @author Ryan Weaver <weaverryan@gmail.com>
@@ -20,13 +21,10 @@ use Composer\Autoload\ClassLoader;
  */
 class AutoloaderUtil
 {
-    private static $classLoader;
-    private $rootDir;
-
-    public function __construct(string $rootDir)
-    {
-        $this->rootDir = $rootDir;
-    }
+    /**
+     * @var ClassLoader
+     */
+    private $classLoader;
 
     /**
      * Returns the relative path to where a new class should live.
@@ -76,16 +74,27 @@ class AutoloaderUtil
 
     private function getClassLoader(): ClassLoader
     {
-        if (null === self::$classLoader) {
-            $autoloadPath = $this->rootDir.'/vendor/autoload.php';
-
-            if (!file_exists($autoloadPath)) {
-                throw new \Exception(sprintf('Could not find the autoload file: "%s"', $autoloadPath));
+        if (null === $this->classLoader) {
+            $autoloadFunctions = spl_autoload_functions();
+            foreach ($autoloadFunctions as $autoloader) {
+                if (is_array($autoloader) && isset($autoloader[0]) && is_object($autoloader[0])) {
+                    if ($autoloader[0] instanceof ClassLoader) {
+                        $this->classLoader = $autoloader[0];
+                        break;
+                    }
+                    if ($autoloader[0] instanceof DebugClassLoader
+                        && is_array($autoloader[0]->getClassLoader())
+                        && $autoloader[0]->getClassLoader()[0] instanceof ClassLoader) {
+                        $this->classLoader = $autoloader[0]->getClassLoader()[0];
+                        break;
+                    }
+                }
             }
-
-            self::$classLoader = require $autoloadPath;
+            if (null === $this->classLoader) {
+                throw new \Exception('Composer ClassLoader not found!');
+            }
         }
 
-        return self::$classLoader;
+        return $this->classLoader;
     }
 }
--- a/tests/Util/AutoloaderUtilTest.php
+++ b/tests/Util/AutoloaderUtilTest.php
@@ -2,10 +2,10 @@
 
 namespace Symfony\Bundle\MakerBundle\Tests\Util;
 
+use Composer\Autoload\ClassLoader;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bundle\MakerBundle\Util\AutoloaderUtil;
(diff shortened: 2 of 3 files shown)


weaverryan:
Tests are failing. But, yea, I think this approach makes even more sense!

silvester:
Just fixed the closely related test. I will look into others.

src/Util/AutoloaderUtil.php
```diff
@@ -77,13 +72,23 @@ public function getNamespacePrefixForClass(string $className): string
...
-
-            if (!file_exists($autoloadPath)) {
-                throw new \Exception(sprintf('Could not find the autoload file: "%s"', $autoloadPath));
+            $autoloadFunctions = spl_autoload_functions();
+            foreach ($autoloadFunctions as $autoloader) {
+                if (is_array($autoloader) && isset($autoloader[0]) && is_object($autoloader[0])) {
+                    if ($autoloader[0] instanceof ClassLoader) {
+                        self::$classLoader = $autoloader[0];
+                        break;
+                    } elseif (class_exists('Symfony\Component\Debug\DebugClassLoader')
```
stof: `class_exists` is useless. `$a instanceof NonExistentClass` is perfectly valid (it won't ever return `true` for non-existent classes) and this will avoid triggering the autoloader when it does not exist.

tests/Util/AutoloaderUtilTest.php
```diff
@@ -31,39 +30,39 @@ public function testGetPathForFutureClass()
...
+            }
+        }
+        $autoloader->setUseIncludePath(true);
 
-        $autoloaderUtil = new AutoloaderUtil(self::$currentRootDir);
         foreach ($this->getPathForFutureClassTests() as $className => $expectedPath) {
             $this->assertSame(
-                // the paths will start in vendor/composer and be relative
-                str_replace('\\', '/', self::$currentRootDir.'/vendor/composer/../../'.$expectedPath),
+            // the paths will start in vendor/composer and be relative
```
stof: you should revert this indentation change

src/Util/AutoloaderUtil.php
silvester: Good to know. Changed it.

src/Util/AutoloaderUtil.php
```diff
@@ -76,16 +74,25 @@ public function getNamespacePrefixForClass(string $className): string
...
-                throw new \Exception(sprintf('Could not find the autoload file: "%s"', $autoloadPath));
+        if (null === $this->classLoader) {
+            $autoloadFunctions = spl_autoload_functions();
+            foreach ($autoloadFunctions as $autoloader) {
+                if (is_array($autoloader) && isset($autoloader[0]) && is_object($autoloader[0])) {
+                    if ($autoloader[0] instanceof ClassLoader) {
+                        $this->classLoader = $autoloader[0];
+                        break;
+                    } elseif ($autoloader[0] instanceof DebugClassLoader
+                        && $autoloader[0]->getClassLoader()[0] instanceof ClassLoader) {
```
stof: what if the wrapped loader is not an array callable ?

src/Util/AutoloaderUtil.php
```diff
@@ -76,16 +74,25 @@ public function getNamespacePrefixForClass(string $className): string
...
-            if (!file_exists($autoloadPath)) {
-                throw new \Exception(sprintf('Could not find the autoload file: "%s"', $autoloadPath));
+        if (null === $this->classLoader) {
+            $autoloadFunctions = spl_autoload_functions();
+            foreach ($autoloadFunctions as $autoloader) {
+                if (is_array($autoloader) && isset($autoloader[0]) && is_object($autoloader[0])) {
+                    if ($autoloader[0] instanceof ClassLoader) {
+                        $this->classLoader = $autoloader[0];
+                        break;
+                    } elseif ($autoloader[0] instanceof DebugClassLoader
```
stof: this can be `if` rather than `elseif` as there is a `break` above

tests/Util/AutoloaderUtilTest.php
```diff
@@ -31,39 +30,39 @@ public function testGetPathForFutureClass()
...
+                    $autoloader->add($prefix, self::$currentRootDir.$path);
+                }
+            }
+        }
+        $autoloader->setUseIncludePath(true);
 
-        $autoloaderUtil = new AutoloaderUtil(self::$currentRootDir);
         foreach ($this->getPathForFutureClassTests() as $className => $expectedPath) {
             $this->assertSame(
                 // the paths will start in vendor/composer and be relative
```
stof: this comment is now wrong

tests/Util/AutoloaderUtilTest.php
```diff
@@ -31,39 +30,39 @@ public function testGetPathForFutureClass()
...
+        foreach ($composerJson['autoload'] as $psr => $dirs) {
+            foreach ($dirs as $prefix => $path) {
+                if ($psr == 'psr-4') {
+                    $autoloader->addPsr4($prefix, self::$currentRootDir.$path);
+                } else {
+                    $autoloader->add($prefix, self::$currentRootDir.$path);
+                }
+            }
+        }
+        $autoloader->setUseIncludePath(true);
```
stof: mutating the actual autoloader of PHP is a bad idea, as it will impact the test process (until the end of the testsuite).

a better idea would be to create a `Composer\ClassLoader` instance with your fake configuration, and put it in the `$autoloaderUtil` private property (as if the class loader had already been located previously).
And then, you could have a separate test triggering the actual class loader retrieval code, but without altering the instance itself.

silvester:
Any other changes needed?

weaverryan:
No other changes needed - this is great work! Thank you Silvester!

Issue #133, opened 2018-03-16
'Make:entity' failed when don't use the default namespace 'App'.


yuhuaian:
src/Doctrine/DoctrineEntityHelper.php
line 45
`$entityClassDetails = new ClassNameDetails($metadata->name, 'App\\Entity');`

src/Maker/MakeCommand.php
line 91
`$generator = new Generator($this->fileManager, 'App\\');`

weaverryan:
Hmm, the command should work if you pass a fully-qualified class name to the first step - e.g. `Acme\Foo\Bar`, but it looks like we missed at least one thing:

Here's the error I'm currently getting:

```
./bin/console make:entity

 Class name of the entity to create or update (e.g. GentleGnome):
 > \App\Foo\Bar\Entity\Baz

 created: src/Foo/Bar/Entity/Baz.php
 created: src/Repository/Foo/Bar/Entity/BazRepository.php

 [ERROR] Only annotation mapping is supported by make:entity, but the                            
         <info>App\Foo\Bar\Entity\Baz</info> class uses a different format. If you would like    
         this command to generate the properties & getter/setter methods, add your mapping       
         configuration, and then re-run this command with the <info>--regenerate</info> flag.    
```

kilianklein:
Hi guys, first of all thanks for this amazing and promising bundle!

I am testing the bundle with symfony3.4 (as the documentation mentioned it should work with it) and indeed I had to go mess directly with the 2 lignes @yuhuaian mentioned to change "App" to "AppBundle" (unless I am wrong this is the convention for the bundle name for 3.4?). Otherwise, I would not get prompted for the entity's fully qualified name and would get this error instead:

> Could not determine where to locate the new class "App\Entity\Test".

Hope this helps.

weaverryan:
Hmm, this does help! What you’re describing is the expected behavior: we assume App unless you pass a full class name (including the beginning slash iirc). But, I think if we guess App, but we can’t determine where to generate the file (which means that you have no App namespace in your autoloading), we could THEN ask you to clarify by entering the full class name. Then you could easily know how to use the command in these cases :).

kilianklein:
Thanks  for your answer, the solution sounds good to me! :)

helariL:
Wouldn't it be possible to reference namespace directly from composer.json ? If we anyway have to define it for autoload.

upyx:
Hello!

[...]

PR #171, opened 2018-05-03
Better message with full namespace

LeJeanbono:
#133


weaverryan:
Thank you @LeJeanbono!

PR #173, opened 2018-05-04
Add configuration of root namespace

upyx:
This PR adds configuration for namespaces used in `make:*` commands that was discussed in #133.
I've used [this doc](https://symfony.com/doc/current/bundles/configuration.html) in work and examples from `FrameworkBundle`.

New parameter:
`maker.root_namespace` - application root namespace, default: `App`

To change it, you should create configuration file in `config/packages/dev` with name `maker.yaml` or `maker.xml` or `maker.php` with root node `maker`. For example YAML file:
```yaml
# config/packages/dev/maker.yaml
maker:
    root_namespace: 'App'
```

I've added some tests. I think we don't need Flex recipe but just document new feature.
I need help for write documentation because my English.

TODO:
- [x] Tests
- [x] Documentations
- [ ] Squash commits

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -92,7 +92,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
                 'This command will generate any missing methods (e.g. getters & setters) for a class or all classes in a namespace.',
                 'To overwrite any existing methods, re-run this command with the --overwrite flag',
             ], null, 'fg=yellow');
-            $classOrNamespace = $io->ask('Enter a class or namespace to regenerate', $this->entityNamespace, [Validator::class, 'notBlank']);
+            $classOrNamespace = $io->ask('Enter a class or namespace to regenerate', $this->getEntityNamespace(), [Validator::class, 'notBlank']);
 
             $input->setArgument('name', $classOrNamespace);
 
@@ -518,14 +518,14 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
             $targetEntityClass = $io->askQuestion($question);
 
             if (!class_exists($targetEntityClass)) {
-                if (!class_exists($this->entityNamespace.'\\'.$targetEntityClass)) {
+                if (!class_exists($this->getEntityNamespace().'\\'.$targetEntityClass)) {
                     $io->error(sprintf('Unknown class "%s"', $targetEntityClass));
                     $targetEntityClass = null;
 
                     continue;
                 }
 
-                $targetEntityClass = $this->entityNamespace.'\\'.$targetEntityClass;
+                $targetEntityClass = $this->getEntityNamespace().'\\'.$targetEntityClass;
             }
         }
 
@@ -840,5 +840,10 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         $driver = $this->doctrineHelper->getMappingDriverForClass($className);
 
         return $driver instanceof AnnotationDriver;
+    }
+
+    private function getEntityNamespace(): string
+    {
+        return $this->doctrineHelper->getEntityNamespace();
     }
 }


LeJeanbono:
Doc and Recipe sound good !
Nice job ;)

LeJeanbono:
You did a conf for entity namespace, should we have for others classes (Form, Repo, Twig, etc)?

upyx:
@LeJeanbono Yes, I think we should. This PR is WIP.

weaverryan:
Hey @upyx!

Woo! Thanks for starting this! I have some thoughts:

1) I think we should only make the `root_namespace` configurable, at least for now. This bundle is highly opinionated on purpose, and isn't intended to work for all edge cases. Besides, if you really do want to, for example, generate a form in a different directory, you can do that by passing the `\Full\Namespace`. So, only include `root_namespace`. It will be much faster to finish  & merge this PR anyways.

2) I don't think we should update the recipe to create the config file, because 99% of people won't need it. Let's just document it. Actually, (maybe in a future PR), it would be cool to detect when the user is having problems using the `App\` namespace (e.g. they try to generate some entity called `Foo`, and we determine that there is no `App\` in their autoloader), ask them if they want to define a different root namespace, then create the config file for them. After all, MakerBundle is all about generating code :)

Cheers!

upyx:
Hi @weaverryan !
At the first look I thought to change only root namespace. But in our project we have many legacy code and have non standard directories. So configuration of many namespaces will be very helpful for our project and may be other. But I understand that our project only one of all. If you are absolutely sure this feature isn't practical then I remove configuration of entity namespace. But think again. We can do bundle more configurable without any disadvantages. Why not do it?

src/DependencyInjection/MakerExtension.php
```diff
@@ -34,7 +34,29 @@ public function load(array $configs, ContainerBuilder $container)
...
             ->addTag(MakeCommandRegistrationPass::MAKER_TAG);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAlias()
+    {
+        return 'maker';
+    }
```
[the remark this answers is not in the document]
upyx: Oh, I'm not sure... I will be grateful if you explain it. I see doctrine and many others do the same. As I understand, result of this method won't cached. So avoiding it (as you say earlier) has minor negative performance impact.

src/DependencyInjection/MakerExtension.php
```diff
@@ -34,7 +34,29 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('services.xml');
         $loader->load('makers.xml');
 
+        $configuration = $this->getConfiguration($configs, $container);
+        $config = $this->processConfiguration($configuration, $configs);
+
+        $container->setParameter('maker.root_namespace', $config['root_namespace']);
```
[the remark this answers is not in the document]
upyx: Use abstract service is great idea! I'll try it.

However where I can get `root_namespace` in DI context (compiler pass or `service.xml`) if it is not stored as "parameter"?

The second, if we are deciding to configure root namespace only, it can be more practical - to remove config and use just a parameter. What do think?

P.S.
My English leaves much to be desired. Please, fill free to correct me and ask to explain.

JSkalskiSBG:
Other ways to fix it for 'advanced users with other namespace then App' would be:
1. check namespace of Kernel.php
2. handle namespace parameter:
```
bin/console maker:command my-command --namespace Custom\\Namespace\\Project
```
3. handle namespace given in name:
```
bin/console maker:command \\Custom\\Namespace\\Project\\my-command
-- OR:
bin/console maker:command \\Custom\\Namespace\\Project\\Command\\my-command
```
Right now 3rd option gives error:

> [ERROR] The "\Custom\Namespace\Project\Command\my-command" command name is not valid because it would be implemented by
>          "\Custom\Namespace\Project\Command\myCommand" class, which is not valid as a PHP class name (it must start with
>          a letter or underscore, followed by any number of letters, numbers, or underscores).

weaverryan:
About that error - that looks like a bug to me - we may have too much validation on make:command. We need to fix that :)

upyx:
Guys! Let's finish this work.

Could you help me with these questions:
1. Does implementation of `getAlias()` and `getConfiguration()` methods interfere to performance? Should I really delete these method?
2. What do we configure the root namespace only or namespaces for all commands?
3. If we configure root namespace only can I remove configuration file and use `parameters`?
4. If we continue to use the configuration file how parameters can be removed?
5. Will we use a recipe?
6. What tests should I add?

@weaverryan what do you think about it?

JSkalskiSBG:
**About this solution:**
It looks bad with current config file:
```
root_namespace: 'App'
entity_namespace: '%maker.root_namespace%\Entity'
```
If we got special config for 'entity', why not special config for 'command'?
Someone may want to keep it in **MySpace\Console**, so adding 'Command\' to namespace is bad idea.
Someone may also use other directory than 'src/Command'.

I think that options '--namespace My\\Name\\Space\\CustomDir' and '--dirpath src/CustomDir' would be much better (directory path relational to folder with composer.json).


Issue should be renamed to 'Add configurable namespace and file path'.
Code that fix this issue should fix it in all 'make' commands, not only 'make:entity' and 'make:command'.


**About performance:**
maker-bundle should be installed only in 'dev' environment.
Does anyone know, if loading config will affect perfomance after installation on production without 'require-dev' from composer?

upyx:
Hello @LeJeanbono! I've already removed `entity_namespace` from configuration as @weaverryan advised. But I prefer to add more configurable options. However we still able to add it later. You are right about `dev` environment so we shouldn't care about <1ms performance impact. I don't like idea about options because it requires to type more letters every time. But really benefit of this bundle it is creation files quick and simple.

Symfony uses aggressive caching of configuration so performance very weak depends on configuration size. However default `App\Kernel` class has that string:
```php
$loader->load($confDir . '/{packages}/*' . self::CONFIG_EXTS, 'glob');
$loader->load($confDir . '/{packages}/' . $this->environment . '/**/*' . self::CONFIG_EXTS, 'glob');
```
So if you will place `maker.yaml` to `config/packages` directory it would be affected on production environment. But you can place one in `config/packages/dev` to fix it.

upyx:
I've added tests for changed namespace cases. @weaverryan could you check it, please?

LeJeanbono:
Done for dev env @upyx

src/Maker/MakeEntity.php
```diff
@@ -43,12 +43,14 @@
     private $fileManager;
     private $doctrineHelper;
     private $projectDirectory;
+    private $entityNamespace;
 
-    public function __construct(FileManager $fileManager, string $projectDirectory, DoctrineHelper $doctrineHelper)
+    public function __construct(FileManager $fileManager, DoctrineHelper $doctrineHelper, string $projectDirectory, string $entityNamespace)
```
weaverryan: To avoid needing to inject the `$entityNamespace` ALSO here, let's add a `getEntityNamespace()` method to `DoctrineHelper`.

src/Test/MakerTestDetails.php
```diff
@@ -64,6 +68,47 @@ public function setFixtureFilesPath(string $fixtureFilesPath): self
...
+            ->addReplacement(
+                'config/services.yaml',
+                'App\\',
+                $rootNamespace.'\\'
+            )
+            ->addReplacement(
+                'phpunit.xml.dist',
+                '<env name="KERNEL_CLASS" value="App\\Kernel" />',
+                '<env name="KERNEL_CLASS" value="'.$rootNamespace.'\\Kernel" />'
+            );
```
weaverryan: I don't know if we need to do all of this. I should be enough to create the `maker.yaml` file with the config AND add one new PSR4 namespace autoload to `composer.json`. I don't think we need to completely eliminate and convert the rest of the app from the `App\` namespace.

tests/Maker/FunctionalTest.php
```diff
@@ -208,6 +258,17 @@ public function getCommandTests()
...
+            $this->getMakerInstance(MakeFunctionalTest::class),
+            [
+                // functional test class name
+                'FooBar',
+            ])
+            ->setFixtureFilesPath(__DIR__.'/../fixtures/MakeFunctionalInCustomRootNamespace')
+            ->changeRootNamespace('Custom')
+            ->addPreMakeCommand('composer dump-autoload')
+        ];
+
```
weaverryan: I think we only really need to test one case, because each maker is using the exact same code to place files. Well, maybe 2 cases: `MakeCommand` (as a general case) and `MakeEntity`, because it *is* indeed a bit more complex.

src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
```diff
@@ -32,13 +33,12 @@ public function process(ContainerBuilder $container)
...
 
-            $container->register(
-                sprintf('maker.auto_command.%s', Str::asTwigVariable($class::getCommandName())),
-                MakerCommand::class
-            )->setArguments([
-                new Reference($id),
-                new Reference('maker.file_manager'),
-            ])->addTag('console.command', ['command' => $class::getCommandName()]);
+            $commandDefinition = new ChildDefinition('maker.auto_command.abstract');
+            $commandDefinition->setClass(MakerCommand::class);
```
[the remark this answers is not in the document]
upyx: It is needed - actually `class` not inherits to child services. However it may be bug. [Documentation](https://symfony.com/doc/current/service_container/parent_services.html) says:
> All attributes on the parent service are shared with the child except for `shared`, `abstract` and `tags`. These are not inherited from the parent.

src/Test/MakerTestDetails.php
upyx: At first I did so. But such solution has issue: tests pass when it mustn't because autoloader. After debugging tests we can remove it. But I think complete converting namespace is better because it is case for we are doing all of this :smile:

tests/Maker/FunctionalTest.php
upyx: It is. I will remove useless tests

upyx:
Thank you for review @weaverryan! I've done changes by your requests. Could you review again, please?
And can I ask you to do PR in documentation?

src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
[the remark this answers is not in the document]
upyx: I debugged it. Really `class` inherits. It is resolved in `ResolveChildDefinitionsPass` but `AddConsoleCommandPass` calls earlier (when `class` is set to `null`).

src/Command/MakerCommand.php
```diff
@@ -88,7 +90,17 @@ protected function interact(InputInterface $input, OutputInterface $output)
...
-        $generator = new Generator($this->fileManager, 'App\\');
+        if (!$this->fileManager->isNamespaceConfiguredToAutoload($this->rootNamespace)) {
+            throw new \LogicException(
+                sprintf(
+                    'It looks like your app may be using a namespace other than "%s".'.
+                    ' See https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html for how to configure this.',
+                    $this->rootNamespace
+                )
+            );
+        }
```
weaverryan: Instead of an exception, let's convert this to a warning. Something like:

```php
$this->io->note([
    sprintf('It looks like your app may be using a namespace other than "%s".', $this->rootNamespace),
    'See https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html#configuration to configure this.
]);
```

The `#configuration` part of the link doesn't exist yet, but we'll add it.

src/FileManager.php
```diff
@@ -163,6 +163,11 @@ public function getNamespacePrefixForClass(string $className): string
         return $this->autoloaderUtil->getNamespacePrefixForClass($className);
     }
 
+    public function isNamespaceConfiguredToAutoload(string $namespace): string
```
weaverryan: `: bool` return type

src/Util/AutoloaderUtil.php
```diff
@@ -72,29 +78,35 @@ public function getNamespacePrefixForClass(string $className): string
...
 
-    private function getClassLoader(): ClassLoader
+    /**
+     * Returns if the namespace is configured by composer autoloader.
+     *
+     * @param string $namespace
+     *
+     * @return bool
+     */
+    public function isNamespaceConfiguredToAutoload(string $namespace)
```
weaverryan: Add `: bool` return type, and remove the `@param` and `@return` - they are duplicated by the code itself.

tests/Command/MakerCommandTest.php
```diff
@@ -26,10 +26,27 @@ public function testExceptionOnMissingDependencies()
...
+    {
+        $maker = $this->createMock(MakerInterface::class);
+
+        $fileManager = $this->createMock(FileManager::class);
+
+        $command = new MakerCommand($maker, 'Unknown', $fileManager);
+        // needed because it's normally set by the Application
+        $command->setName('make:foo');
+        $tester = new CommandTester($command);
+        $tester->execute(array());
```
weaverryan: I like this test. We'll just need to instead look at the output, instead of the exception when we convert it to just a message.

weaverryan:
Actually, I just made all the changes - including the docs - because they were so minor. Let's see if the tests still pass!

weaverryan:
Also updated after a nasty merge conflict with a related PR. Let's see if the tests pass. @upyx can you check out the latest code to make sure the conflict didn't mess anything up? A lot of the same lines of code were changed.

upyx:
Oh! Great! I see you strongly determined to comlete this :smiley:

weaverryan:
Ha - I am! It’s a missing feature. I knew that :). And I’ll make a release as soon as it’s done.

upyx:
I am glad to work with you :)

PR #174, opened 2018-05-05
[WIP] Improve make:functional-test by using routes and security configuration

tgalopin:
I would like to propose with this PR an improvement over the generation of functional tests. Instead of creating a simple class, this maker goes further and read your application routes and security configuration to create more things for you.

Basically, it reads all your routes and creates a test for each action in your application:

```
tests/Controller
├── AccountController
│   ├── InformationsTest.php
│   ├── NotificationsTest.php
│   ├── PictureClearTest.php
│   ├── PictureCropTest.php
│   ├── PictureDeleteTest.php
│   └── PictureTest.php
├── Api
│   ├── ProjectApiController
│   │   └── SearchTest.php
├── AssetController
│   └── ImageTest.php
├── CompanyController
│   └── CreateTest.php
├── DashboardController
│   └── IndexTest.php
├── InviteController
│   └── InviteTest.php
├── MarketingController
│   ├── AboutTeamTest.php
│   ├── AboutWhyTest.php
│   ├── AnalyzeTest.php
│   ├── BorrowTest.php
│   └── LendTest.php
├── ProfileController
│   ├── CompanyTest.php
│   ├── UserActivityTest.php
│   ├── UserBadgesTest.php
│   └── UserProjectsTest.php
├── ProjectController
│   ├── CreateTest.php
│   ├── EditInformationsTest.php
│   ├── EditLoanTest.php
│   ├── EditPictureTest.php
│   ├── EditSocialsTest.php
│   └── ViewTest.php
└── WebTestCase.php
```

The names of the test classes are based on the controllers and actions names:

```
App\Controller\User\AccountController::informations
=> Tests\Controller\User\AccountController\InformationsTest
```

Each of these files looks like this:

``` php
<?php

namespace App\Tests\Controller\DashboardController;

use App\Tests\Controller\WebTestCase;

/**
 * @covers \App\Controller\DashboardController::index
 */
class IndexTest extends WebTestCase
{
    public function testGet()
    {
        $client = static::createAnonymousClient();
        $crawler = $client->request('GET', '/');

        $this->assertSame(200, $client->getResponse()->getStatusCode());
    }
}
```

The maker reads your route methods configuration: if there is one, it creates appropriate test methods, otherwise it creates test a method only for GET. The maker also try to access GET routes during the build, and if a GET route returns 200 it enables an assertion to check it keeps working in the future.

Additionally, this prototype also creates a base WebTestCase class. This class gives a few helpers to easily authenticate users in a functional test, by reading security configuration:

``` php
<?php

namespace App\Tests\Controller;

use Symfony\Bundle\FrameworkBundle\Test\WebTestCase as BaseWebTestCase;
use Symfony\Bundle\FrameworkBundle\Client;
use Symfony\Component\BrowserKit\Cookie;
use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
use Symfony\Component\Security\Core\User\UserInterface;

abstract class WebTestCase extends BaseWebTestCase
{
    protected static function createAnonymousClient(array $options = [], array $server = [])
    {
        return static::createClient($options, $server);
    }

    protected static function createAnalystClient(array $options = [], array $server = [])
    {
        $client = static::createClient($options, $server);
        $container = $client->getContainer();

        // Uncomment the following lines and configure how to create a User with role ROLE_ANALYST

        // $user = $container->get(UserRepository::class)->findOneBy(['username' => 'analyst@example.com']);
        // self::logInUser($client, 'main', $user);

        throw new \LogicException(
            __CLASS__."::createAnalystClient() is not implemented\n".
            'Open the WebTestCase class and customize it!'
        );

        return $client;
    }

    protected static function createUserClient(array $options = [], array $server = [])
    {
        $client = static::createClient($options, $server);
        $container = $client->getContainer();

        // Uncomment the following lines and configure how to create a User with role ROLE_USER

        // $user = $container->get(UserRepository::class)->findOneBy(['username' => 'user@example.com']);
        // self::logInUser($client, 'main', $user);

        throw new \LogicException(
            __CLASS__."::createUserClient() is not implemented\n".
            'Open the WebTestCase class and customize it!'
        );

        return $client;
    }

    protected static function createAdminClient(array $options = [], array $server = [])
    {
        $client = static::createClient($options, $server);
        $container = $client->getContainer();

        // Uncomment the following lines and configure how to create a User with role ROLE_ADMIN

        // $user = $container->get(UserRepository::class)->findOneBy(['username' => 'admin@example.com']);
        // self::logInUser($client, 'main', $user);

        throw new \LogicException(
            __CLASS__."::createAdminClient() is not implemented\n".
            'Open the WebTestCase class and customize it!'
        );

        return $client;
    }

    protected static function logInUser(Client $client, $firewallName, UserInterface $user)
    {
        $container = $client->getContainer();
        $container->get('security.user_checker')->checkPreAuth($user);

        $token = new UsernamePasswordToken($user, null, $firewallName, $user->getRoles());

        $request = $container->get('request_stack')->getCurrentRequest();
        if (null !== $request) {
            $container->get('security.authentication.session_strategy')->onAuthentication($request, $token);
        }

        $container->get('security.token_storage')->setToken($token);

        $session = $container->get('session');
        $session->set('_security_'.$firewallName, serialize($token));
        $session->save();

        $client->getCookieJar()->set(new Cookie($session->getName(), $session->getId()));
    }
}
```

These features are not entirely functional right after the generation because the rely on users and the maker can't know how to create users. However, in the future, this may be the case when we will have the `make:user` command.


javiereguiluz:
Thanks for this contribution. I've reviewed it a bit and there are certain things that look strange to me:

* Why use `App\Tests\Controller\Http\` instead of `App\Tests\Controller\`. The `Http\` thing is not what we show, use and recommend in official Symfony docs.
* Creating a `GET` and `POST` test when actions don't define their methods may not be the best idea. 99% of the apps I've seen don't set the `@Method` annotation when the controllers are simple GET requests. For example, the `homepage` route  never defines that, so we'd end up with a `GET` and `POST` tests for the homepage.
* Generating the test based on the route name instead of the controller name may not be the best choice either. It's common to have beautiful method names `showUser()` and weird route names `app_public_user_actions_show` ... so the test method would be `class AppPublicUserActionsShowTest` instead of `class UserTest`

tgalopin:
I updated the PR from feedbacks:

- The test classes are now names using controllers and actions names, which is especially nice as it matches the app Controller namespace
- The generation is now based on actions in the app directory: routes from outside of it are not generated, to avoid generating EasyAdmin of FOSUserBundle tests
- Instead of relying on a specific structure to detect already present function tests, the matcher adds a `@cover` annotation on tests and check it when it needs to know if a route is already tested. This technique let developers organize their tests as they wish and let them ignore some actions if they want to.
- By default, only tests for GET are created if not Methods configuration is present on the route
- Everything is directly under Controller now

PR #178, opened 2018-05-10
API Platform support: add an --api-resource flag for the make:entity command

### LINKED ISSUE #175: Add a flag to the entity generator to mark it as an API Platform resource
As an alternative to #158, what do you think about adding a new `--api-platform-resource` option to the entity generator that will automatically:

* Add the `@ApiResource` annotation to the generated entity
* Tell the user to run `composer req api` if API Platform isn't already installed

dunglas:
This PR adds a new `--api-resource` flag to the `make:entity` command. It allows to automatically a REST endpoint with CRUD operations for the given resource class using API Platform:

screencast

I need some guidance about how to add test for this (actually, a few tests are red locally on my Mac even on master).

Fixes #175.


weaverryan:
Ignoring your question about testing for a moment, I don't like the `--api-resource` flag from a user's perspective. I think it only makes sense to add a new interactive question to `make:entity` automatically IF ApiPlatform is installed.

weaverryan:
Or, the option AND automatically asking if the flag isn’t passed and API Platform is installed

dunglas:
@weaverryan done

capture d ecran 2018-05-11 a 08 00 55

src/Maker/MakeEntity.php
```diff
@@ -106,6 +111,14 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
 
         $input->setArgument('name', $value);
+
+        if (class_exists(ApiResource::class) && !$input->getOption('api-resource')) {
+            $description = $command->getDefinition()->getOption('api-resource')->getDescription();
+            $question = new ConfirmationQuestion($description, false);
+            $value = $io->askQuestion($question);
+
+            $input->setOption('api-resource', $value);
+        }
```
weaverryan: We should only ask this question when the entity does NOT exist yet. Unfortunately, that means we'll need to duplicate a bit of logic from `generate()` up here:

```php
$entityClassDetails = $generator->createClassNameDetails(
    $input->getArgument('name'),
    'Entity\\'
);
$classExists = class_exists($entityClassDetails->getFullName());
```

dunglas: Fixed in d003a487f5f47f99b3276ed45c21f3958d27dffa (with some more Voodoo...)

dunglas:
Tests added.

I also transformed `Generator` in a service: it's better from a design POV (this class should be a service anyway), and it prevents to add a new magic interface to be able to retrieve the `Generator` instance in the `interact` method.

IMO we should also remove the `Generator` argument from the `MakerInterface::generate()` method and recommend using the service instead (it's allowed by our BC policy, and won't break existing external implementation as long a we still pass this argument even if not defined in the interface). But I'll handle that in another PR, because all other makers need to be updated, and it's not related at all with the purpose of this PR.

Issue #167, opened 2018-05-02
make:entity --regenerate and classes using traits


nandorstanko:
I'm using traits in my entities like blameable, timestampable etc. I have getters/setters in those traits, so I don't have to define them in every entity. The command adds the getters/setters to my entities, although I have them already in the traits, and even adds the private property definitions again... I have tried to rewrite everything to exactly match the generated code, but the generator still inserts the functions again, so my guess is that it ignores the trait.

Maybe there is some other solution, and using traits is not the way to move forward, or this is just a bug. I need some advice or confirmation in both cases :) Thanks!

weaverryan:
Hey @nandorstanko!

Looks like a bug to me - using a trait should totally be allowed, and the system should notice the property & getter/setter, and not regenerate it. We'll need to fix this :).

Cheers!

andrewtch:
Yes, this is confirmed, and affects every entity that has, for example Gedmo's TimestampableTrait.

andrewtch:
Also, the same thing happens then using MappedSuperclass - the fields are copied from parent class.

andrewtch:
@weaverryan , can this be fixed by filtering out inherited fields here - https://github.com/symfony/maker-bundle/blob/master/src/Maker/MakeEntity.php#L781 or should it better be done in https://github.com/symfony/maker-bundle/blob/master/src/Doctrine/EntityRegenerator.php ? I suppose the first options is more bulletproof.

stof:
filtering out inherited fields would solve the case of MappedSuperclass, but it won't solve the case of traits. Such fields are *not* inherited.

andrewtch:
@stof indeed. I'll try to find some time to dig into tests of maker and probably will make a PR for that.

andrewtch:
@weaverryan , I am having some struggle with running test for this bundle. Specifically, running just `EntityRegeneratorTest` leads to class resolution error:

```
vendor/bin/simple-phpunit tests/Doctrine/EntityRegeneratorTest.php 
PHPUnit 6.5.8 by Sebastian Bergmann and contributors.

Testing Symfony\Bundle\MakerBundle\Tests\Doctrine\EntityRegeneratorTest
..E                                                                 3 / 3 (100%)

Time: 6.13 seconds, Memory: 26.00MB

There was 1 error:

1) Symfony\Bundle\MakerBundle\Tests\Doctrine\EntityRegeneratorTest::testXmlRegeneration
[...]

PR #179, opened 2018-05-11
Fix the EntityRegenerator on Mac

dunglas:
On Mac, an instance of `Doctrine\Common\Persistence\Mapping\MappingException` is thrown by Doctrine when a mapping file references a non-existing class. Currently, the code only catches `Doctrine\ORM\Mapping\MappingException` instances.

This PR makes the tests green on Mac (https://github.com/symfony/maker-bundle/issues/167#issuecomment-387976395). This also probably fixes a real bug.

I've no idea why Doctrine throws a different exception on Mac than on Linux.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -3,6 +3,18 @@
 if (!file_exists(__DIR__.'/src')) {
     exit(0);
 }
+
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__.'/src')
+    // the PHP template files are a bit special
+    ->notName('*.tpl.php')
+;
+
+if (\PHP_VERSION_ID < 70100) {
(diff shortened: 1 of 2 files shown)


dunglas:
@andrewtch see https://github.com/symfony/maker-bundle/pull/179#discussion_r187522288

`make:entity` already requires PHP 7.1: https://github.com/symfony/maker-bundle/blob/master/src/Maker/MakeEntity.php#L113

andrewtch:
@dunglas , ah, indeed. The current fix seems to work for me just fine.

weaverryan:
Fantastic! Super weird, but the code/fix gives us code that still makes sense: if there’s a mapping exception in general, fallback to the disconnected metadata lookup.

Hopefully this will unblock some things!

PR #180, opened 2018-05-11
Remove some unused props

dunglas:


weaverryan:
Good catch! But, let's remove the argument entirely from the constructor. Then, there are 2 other references to `new EntityRegenerator()` that need to be updated: in `MakeEntity` and `EntityRegeneratorTest`.

dunglas:
Removing the argument is a BC break. Is it allowed?

weaverryan:
Ah, yes. The class is marked as `@internal`. Almost the entirely library is marked as `@internal`. We purposely don't support using the classes of the library to build on top of.

src/Maker/MakeEntity.php
```diff
@@ -42,12 +42,10 @@
 {
     private $fileManager;
     private $doctrineHelper;
-    private $projectDirectory;
 
     public function __construct(FileManager $fileManager, string $projectDirectory, DoctrineHelper $doctrineHelper)
```
dunglas: This one cannot be removed because this class isn't `@internal` and  `$projectDirectory` isn't the last argument.

weaverryan:
Thanks @dunglas!

PR #185, opened 2018-05-14
fix: add src option for not standard paths

rampmaster:
I add custom option `src` for entity generator


weaverryan:
Hi there!

I agree that we can make this better :). But, I think there is a different solution that is better for DX. Instead of using Finder to look in the src/ directory, we should read the Doctrine metadata to load a real list of all of the entities. This should not be too difficult by using the DoctrineHelper class. Then, you will get the nice functionality without needing to know about or pass an option.

Do you want to update your PR to try this?

Cheers!

rampmaster:
Yes, not problem @weaverryan and thanks for your idea :-)

mssimi:
@rampmaster Hey, any chance updating this PR? I need this feature.

rampmaster:
Yes, working.

bastos71:
Hi @rampmaster ! I was about to create the same fix ^^
When this PR would be finish and accepted ?

weaverryan:
Closing in favor of #310 or #311

Issue #176, opened 2018-05-07
Wrong generation during make:crud when entity field name has an underscore in it


MrNicodemuz:
Template is wrongly generated when one of the field names has an underscore. The following error is raised:

    Neither the property "field_1" nor one of the methods "field_1()", "getfield_1()"/"isfield_1()"/"hasfield_1()" or "__call()" exist and have public access in class "App\Entity\Test".

LeJeanbono:
If your database field is `field_1`, you should have a Doctrine mapping like this : 
` @ORM\Column(name="field_1")
  private $field1;
`
And a getter : 
`public function getField1()`
Maybe the command should detect `_` and use this template

weaverryan:
The user’s property name is probably field_1, which is weird, but legal. There is probably a bug here if your properties have an underscore. Can someone check into it?

LeJeanbono:
I confirm, the entity (from make:entity) got : 
`private $field_1;`
`public function getField1()`
and in _index_ and _show_ template : 
`{{ entity.field_1 }}`

So, what should we want ? 
`public function getField_1()`
or
`{{ entity.field1 }}`

weaverryan:
Excellent! I think the getter is ok, the template code should be fixed. Since you are already on top of the issue, could you open a PR?

PR #186, opened 2018-05-15
fix crud template - entity field with underscore

LeJeanbono:
#176

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -176,7 +176,7 @@ final class MakeCrud extends AbstractMaker
                 'entity_twig_var_plural' => $entityTwigVarPlural,
                 'entity_twig_var_singular' => $entityTwigVarSingular,
                 'entity_identifier' => $entityDoctrineDetails->getIdentifier(),
-                'entity_fields' => $entityDoctrineDetails->getDisplayFields(),
+                'entity_fields' => str_replace('_', '', $entityDoctrineDetails->getDisplayFields()),
                 'route_name' => $routeName,
             ],
             'new' => [
@@ -187,7 +187,7 @@ final class MakeCrud extends AbstractMaker
                 'entity_class_name' => $entityClassDetails->getShortName(),
                 'entity_twig_var_singular' => $entityTwigVarSingular,
                 'entity_identifier' => $entityDoctrineDetails->getIdentifier(),
-                'entity_fields' => $entityDoctrineDetails->getDisplayFields(),
+                'entity_fields' => str_replace('_', '', $entityDoctrineDetails->getDisplayFields()),
                 'route_name' => $routeName,
             ],
         ];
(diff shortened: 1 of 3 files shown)


weaverryan:
Awesome! And even with a great test. There are some failures. When those are green, we are good :)

src/Maker/MakeCrud.php
```diff
@@ -187,7 +187,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
                 'entity_class_name' => $entityClassDetails->getShortName(),
                 'entity_twig_var_singular' => $entityTwigVarSingular,
                 'entity_identifier' => $entityDoctrineDetails->getIdentifier(),
-                'entity_fields' => $entityDoctrineDetails->getDisplayFields(),
+                'entity_fields' => str_replace('_', '', $entityDoctrineDetails->getDisplayFields()),
                 'route_name' => $routeName,
```
sadikoff: I think this is not best solution. We need to fix it in `Doctrine\EntityDetails`

LeJeanbono: Just for twig template, we don't want here to have _ (to match the getter).
@sadikoff If 'getDisplayFields()' is only for templating, I'll put replace in it !
Can you confirm ?

LeJeanbono:
@sadikoff What do you think I miss for the failling test ? I dont' get it, please.

src/Maker/MakeCrud.php
sadikoff: Yes it's for templating. But you should change it carefully. `getDisplayFields()` uses `metadata->fieldMappings` it is not simple array we will need part of this array to detect correct field types

src/GeneratorTwigHelper.php
```diff
@@ -25,20 +25,23 @@ public function __construct(FileManager $fileManager)
 
     public function getEntityFieldPrintCode($entity, $field): string
     {
-        $printCode = $entity.'.'.$field['fieldName'];
+        $twigField = preg_replace_callback('/(?!^)_([a-z0-9])/', function ($s) {
+            return strtoupper($s[1]);
+        }, $field['fieldName']);
+        $printCode = $entity . '.' . str_replace('_', '', $twigField);
```
sadikoff: May be we can use `Str::asLowerCamelCase` here?

src/GeneratorTwigHelper.php
```diff
@@ -28,20 +28,20 @@ public function getEntityFieldPrintCode($entity, $field): string
         $twigField = preg_replace_callback('/(?!^)_([a-z0-9])/', function ($s) {
             return strtoupper($s[1]);
         }, $field['fieldName']);
-        $printCode = $entity . '.' . str_replace('_', '', $twigField);
+        $printCode = $entity.'.'.str_replace('_', '', $twigField);
```
sadikoff: May be we can use Str::asLowerCamelCase here?

sadikoff: May be we can use Str::asLowerCamelCase here?

weaverryan:
Really great PR! Thank you @LeJeanbono!

PR #190, opened 2018-05-19
Fix: Case mismatch

localheinz:
This PR

* [x] fixes a case mismatch when referencing a class


weaverryan:
Thank you Andreas!

localheinz:
Thank you, @weaverryan!

PR #197, opened 2018-06-03
Preserve existing class metadata

ro0NL:
Hi,

Im facing an issue related to `make:entity`. I noticed it uses `DisconnectedClassMetadataFactory` which at this point basically forgets about existing metadata (metadata that on itself is loaded during the `loadClassMetadata` event).

Real life example is e.g. https://github.com/msgphp/symfony-demo-app/blob/b9536243f4a0803b36acfef839ac0c3559a91bd6/src/Entity/User/UserRole.php#L11 where the `roles` field is mapped during `loadClassMetadata`.

Currently its mapping is not loaded, causing an invalid field override here.

This solves it.


weaverryan:
Hey @ro0NL!

We load the disconnected metadata factory when/because we can’t use the normal metadata factory, usually (entirely?) when we’re readijg metadata for an entity whose class doesn’t yet exist (so the normal metadata factory throws an error). Won’t using the metadata factory here trigger that error? It looks like the tests may be failing due to this.

ro0NL:
after submitting the PR i realized a call to `$em->getMetadataFactory()->getAllMetadata()` is probably unintended in disconnected state. Now reverted as such to see if test pass

The current approach (try/catching) works for me.

PR #202, opened 2018-06-13
Do not remove escaping on TwigFilter by default

### LINKED ISSUE #200: Why twig extension are "HTML safe" by default ?
I created a new twig extension with the maker bundle, and I saw that

```php
    public function getFilters(): array
    {
        return [
            new TwigFilter('my_filter', [$this, 'doSomething'], ['is_safe' => ['html']]),
        ];
    }
```

IMHO, this should not be added because it's a major security issue.
If you really want to make things easy for new comer, it should be commented by default.

lyrixx:
fixes #200


src/Resources/skeleton/twig/Extension.tpl.php
```diff
@@ -11,7 +11,10 @@ class <?= $class_name ?> extends AbstractExtension
     public function getFilters(): array
     {
         return [
-            new TwigFilter('filter_name', [$this, 'doSomething'], ['is_safe' => ['html']]),
+            // If your filter generates SAFE HTML, you should add a third
+            // parameter: ['is_safe' => ['html']]
+            // Reference: https://twig.symfony.com/doc/2.x/advanced.html#automatic-escaping
```
nicolas-grekas: Or add an --safe-html option on the command? Such comments will be committed usually, thus be noise to many lazy fools (all of us ;) )

lyrixx: Almost all code generated by this bundle should be updated by the end user. So it prefer to not add a new option
For example the callback on this next **next** line should be updated.

PR #203, opened 2018-06-13
Better CS in make:crud command

lyrixx:
* Make smaller code line

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -23,7 +23,9 @@ class <?= $class_name ?> extends <?= $parent_class_name; ?><?= "\n" ?>
 <?php if (isset($repository_full_class_name)): ?>
     public function index(<?= $repository_class_name ?> $<?= $repository_var ?>): Response
     {
-        return $this->render('<?= $templates_path ?>/index.html.twig', ['<?= $entity_twig_var_plural ?>' => $<?= $repository_var ?>->findAll()]);
+        return $this->render('<?= $templates_path ?>/index.html.twig', [
+            '<?= $entity_twig_var_plural ?>' => $<?= $repository_var ?>->findAll(),
+        ]);
     }
 <?php else: ?>
     public function index(): Response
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -43,10 +47,9 @@ public function new(Request $request): Response
     {
         $<?= $entity_var_singular ?> = new <?= $entity_class_name ?>();
         $form = $this->createForm(<?= $form_class_name ?>::class, $<?= $entity_var_singular ?>);
-        $form->handleRequest($request);
 
-        if ($form->isSubmitted() && $form->isValid()) {
-            $em = $this->getDoctrine()->getManager();
+        if ($form->handleRequest($request)->isSubmitted() && $form->isValid()) {
+            $em = $this->get('doctrine.orm.entity_manager');
```
nicolas-grekas: This works only because the service is public? If yes, then I think that's not a good idea, we should encourage ppl to not rely on any public services, so that they learn it's better to declare your dependencies.

lyrixx: It relies one the same behaviour as the previous code. It just saves some calls.
I might agree with you anyway . Could you open an issue about it ?
Btw it's the same for the form factory

weaverryan: In the docs, we've so far decided to never use `$this->get()` anymore, and instead to:

1) Use a shortcut method if it's available (e.g. `$this->getDoctrine()`)
2) if there is no shortcut, use action DI

And I think that is the right thing to do. So I think we should not make this change. And if we *do* want to make this change, let's discuss on symfony-docs - so we can keep the generated code consistent with the docs :)

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -43,10 +47,9 @@ public function new(Request $request): Response
     {
         $<?= $entity_var_singular ?> = new <?= $entity_class_name ?>();
         $form = $this->createForm(<?= $form_class_name ?>::class, $<?= $entity_var_singular ?>);
-        $form->handleRequest($request);
 
-        if ($form->isSubmitted() && $form->isValid()) {
-            $em = $this->getDoctrine()->getManager();
+        if ($form->handleRequest($request)->isSubmitted() && $form->isValid()) {
```
weaverryan: We should revert this too. This change should be made to symfony-docs first, if we want it.

lyrixx:
I updated the PR to address your comments

lyrixx:
I know I have to rebase, But What do you think about this PR ?

weaverryan:
Thanks Grégoire!

PR #205, opened 2018-06-13
Do not prefix Command Class Name by 'App' if the prefix is 'app:'

lyrixx:


lyrixx:
Thanks @javiereguiluz

javiereguiluz:
Thanks Grégoire.

PR #206, opened 2018-06-13
Put the Form Type in a Type folder

lyrixx:
It makes app scales with ease:

* It's really common to have some extension / transformer / listener:
  It's a mess if you mix extensions / transformers / litener in the same
  folrders. It's better to organize class (/file) responability in
  specific NS / folder
* It's not harmful since people are using (should ?) autocomplete

You may argue it's overkill. But really it's really common to extends
forms capability. So IMHO it's better to put type in a separate folders.
Thus, the only diff in userland is an extra NS level... Not a big deal!


weaverryan:
Hey @lyrixx!

I personally don't really want this change, but it is definitely subjective. However, we document it on symfony-docs just in the `Form` directory. So, we should create an issue there first, then update the maker to reflect it.

Cheers!

PR #210, opened 2018-06-17
Request: Added appending method to Generator

Skandalik:


src/Generator.php
```diff
@@ -120,13 +136,15 @@ public function createClassNameDetails(string $name, string $namespacePrefix, st
...
 
-    private function addOperation(string $targetPath, string $templateName, array $variables)
+    private function addOperation(string $targetPath, string $templateName, array $variables, bool $append = false)
     {
         if ($this->fileManager->fileExists($targetPath)) {
-            throw new RuntimeCommandException(sprintf(
-                'The file "%s" can\'t be generated because it already exists.',
-                $this->fileManager->relativizePath($targetPath)
-            ));
+            if (!$append) {
```
weaverryan: We can put this into just one if statement:

```php
if ($this->fileManager->fileExists($targetPath) && !$append) {
```

src/Generator.php
```diff
@@ -120,13 +136,15 @@ public function createClassNameDetails(string $name, string $namespacePrefix, st
...
+                throw new RuntimeCommandException(sprintf(
+                    'The file "%s" can\'t be generated because it already exists.',
+                    $this->fileManager->relativizePath($targetPath)
+                ));
+            }
         }
 
         $variables['relative_path'] = $this->fileManager->relativizePath($targetPath);
```
weaverryan: I think, below, when we add this to the `pendingOperations` array, we should add a new `append` key set to `true` or `false` so that we know if this is an append operation. That would make the code in `FileManager::dumpFile()` safer, because we could give *that* method a new, 3rd argument: `$append = false` (and we would use that new `append` key on the array to set this argument).

weaverryan:
Closing this for now. It may still be a sound idea - but maybe not, now that YamlSourceManipulator exists, which helps (at least) for updating Yaml files.

Cheers!

PR #215, opened 2018-06-21
make:entity: Add return type to getId()

gharlan:


lyrixx:
What if id are uuid ?

javiereguiluz:
Thank you Gregor.

PR #221, opened 2018-07-04
Extend from AbstractController when using Symfony 4.1 or higher

javiereguiluz:
Related to https://github.com/symfony/symfony-docs/pull/9991 ... let's extend from `AbstractController` when possible.

### DIFF
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -21,6 +21,7 @@ use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\HttpKernel\Kernel;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
@@ -55,6 +56,7 @@ final class MakeController extends AbstractMaker
             $controllerClassNameDetails->getFullName(),
             'controller/Controller.tpl.php',
             [
+                'parent_class_name' => (Kernel::VERSION_ID) >= 40100 ? 'AbstractController' : 'Controller',
                 'route_path' => Str::asRoutePath($controllerClassNameDetails->getRelativeNameWithoutSuffix()),
                 'route_name' => Str::asRouteName($controllerClassNameDetails->getRelativeNameWithoutSuffix()),
                 'twig_installed' => $this->isTwigInstalled(),
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Maker;
 
 use Doctrine\Common\Annotations\Annotation;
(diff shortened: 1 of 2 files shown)


src/Maker/MakeController.php
```diff
@@ -55,6 +56,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             $controllerClassNameDetails->getFullName(),
             'controller/Controller.tpl.php',
             [
+                'parent_class_name' => (Kernel::VERSION_ID) >= 40100 ? 'AbstractController' : 'Controller',
```
stof: you should detect the FrameworkBundle version actually, not the HttpKernel one

javiereguiluz: I don't know how to do that. Which is the constant in `FrameworkBundle` that stores its version? Thanks!

chalasr: `class_exists(AbstractController::class)` should be good

src/Maker/MakeController.php
```diff
@@ -55,6 +56,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             $controllerClassNameDetails->getFullName(),
             'controller/Controller.tpl.php',
             [
+                'parent_class_name' => class_exists(AbstractController::class) ? 'AbstractController' : 'Controller',
```
nicolas-grekas: AbstractController exists also in 3.4, so the check doesn't achieve what the title of this PR tells

javiereguiluz: You are right 😭  Should we revert to the original code then?

chalasr: What's wrong with using `AbstractController` on 3.4 code?

javiereguiluz: That it doesn't have the `->getParameter()` shortcut ... so it's not really equivalent to `Controller`. In 4.1 it is.

chalasr: Ok, change the check to `\method_exists(AbstractController::class, 'getParameter`) then?

src/Resources/skeleton/controller/Controller.tpl.php
```diff
@@ -2,10 +2,10 @@
 
 namespace <?= $namespace; ?>;
 
+use Symfony\Bundle\FrameworkBundle\Controller\<?= $parent_class_name; ?>;
 use Symfony\Component\Routing\Annotation\Route;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 
-class <?= $class_name; ?> extends Controller
+class <?= $class_name; ?> extends <?= $parent_class_name; ?><?= "\n" ?>
```
chalasr: just wondering: is the `"\n"` necessary?

javiereguiluz: In theory it's not and I didn't add it at first ... but the generated controller class was like this:

```php
class CoolController extends AbstractController {
  // ...
}
```

instead of this:

```php
class CoolController extends AbstractController
{
  // ...
}
```

Can any of you reproduce my wrong behavior? If yes, can you explain to me why, because I don't understand this 😅

PR #226, opened 2018-07-09
Don't use :contains in the functional test tpl

dunglas:
`:contains` is a proprietary jQuery selector, and has been explicitly removed from the CSS 3 spec: https://stackoverflow.com/questions/4781141/why-doesnt-the-selector-h3nth-child1containsa-work/4781167#4781167

Then, it doesn't work with Panthère (it's actually supported by Selenium, but not by WebDriver).

I propose to use Xpath instead.

### DIFF
--- a/src/Resources/skeleton/test/Functional.tpl.php
+++ b/src/Resources/skeleton/test/Functional.tpl.php
@@ -12,6 +12,6 @@ class <?= $class_name ?> extends WebTestCase
         $crawler = $client->request('GET', '/');
 
         $this->assertSame(200, $client->getResponse()->getStatusCode());
-        $this->assertCount(1, $crawler->filterXPath('//h1:[contains(text(), "Hello World")]'));
+        $this->assertContains('Hello World', $crawler->filter('h1')->text());
     }
 }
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/test/Functional.tpl.php
```diff
@@ -12,6 +12,6 @@ public function testSomething()
         $crawler = $client->request('GET', '/');
 
         $this->assertSame(200, $client->getResponse()->getStatusCode());
-        $this->assertSame(1, $crawler->filter('h1:contains("Hello World")')->count());
+        $this->assertCount(1, $crawler->filterXPath('//h1:[contains(text(), "Hello World")]'));
```
lyrixx: What about
```php
$this->assertContains('Hello World', $crawler->filter('h1'));
```

stof: should be this though:

```php
$this->assertContains('Hello World', $crawler->filter('h1')->text());
```

And this assumes that it is the only h1 on the page btw (or at least the first one), otherwise the test would fail

lyrixx: You are right ;)

But anyway, this is boiler plate code, and it should be updated...
So I'm totally :+1: with stof's proposal

weaverryan:
I updated the selector to the simpler version, recommended by Stof, but still without the :contains. Thanks @dunglas!

PR #227, opened 2018-07-09
Add an E2E test maker

dunglas:
Add a new `make:e2e-test` command that uses Panthère. Basically the same than the "functional" maker.

### DIFF
--- a/src/Maker/MakeE2eTest.php
+++ b/src/Maker/MakeE2eTest.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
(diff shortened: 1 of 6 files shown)


weaverryan:
Hey @dunglas!

Sorry for the delay :). I really want to add this. But, could we make it even *more* integrated? We already have a `make:functional-test` maker. Is an E2E test really any different than a functional test that uses Panthere? What if we instead did *not* make a new command, but just enhanced the existing `make:functional-test` command to:

A) If `symfony/panthere` is installed, automatically use it
B) If it is not installed, ask if the user wants to test JavaScript with panthere. If yes, make sure they have the dependency and generate a test using Panthere

We could also still create a `make:e2e-test` command, but it would just be a wrapper around `make:functional-test` that basically answers "yes" automatically to whether or not the user wants to use panthere.

WDYT?

patrickvale:
This appears to be mostly resolved by https://github.com/symfony/maker-bundle/commit/3333d6ed324ce58104b8bf54a6d6a31cbc4c6a22#diff-705024daa6707efe0a1e3694f923414f

But that commit doesn't include a `make:e2e-test` wrapper command - is that something that is still wanted?

dunglas:
Indeed, #417 is better! Closing.

PR #237, opened 2018-08-07
[WIP] Fix failing tests

sadikoff:
Suppress deprecated doctrine class error

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,6 +10,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="TEST_DATABASE_DSN" value="mysql://root:@127.0.0.1:3306/test_maker" />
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Which failing test? Which deprecation?
I'm usually :-1: as this removes an opportunity to fix symfony/symfony when it has issues.

nicolas-grekas:
The root issue is fixed in https://github.com/doctrine/doctrine2/pull/7307, which will be part of doctrine/orm v2.6.3
I'm fine with this change as long as the weak_vendor mode is removed once v2.6.3 is out personnaly.

weaverryan:
Ah, I thought we needed to wait until Symfony 4.2 for the deprecation to be removed. Let's close and ignore the deprecation until 2.6.3 is out.

Thanks :)

PR #241, opened 2018-08-10
make:crud | Improvements

### LINKED ISSUE #234: make:crud does not handle DateInterval objects
The generated index and show pages of an entity with a `DateInterval` will crash, due to `DateInterval` not being able to be cast to string.

DateTimeInterfaces generate something like this:
`{{ obj.endDate ? obj.endDate|date('Y-m-d') : '' }}`

Intervals generate something like this:
`<td>{{ obj.period }}</td>`

The following should do the trick, at least as a default.
`<td>{{ obj.period.format('%Y-%m-%d') }}</td>`

### LINKED ISSUE #270: make:crud XType (or XController)
I tried to execute this command except that it did not work as expected ...
make:crud XType
When command make:crud ends with "Type", the form isn't created correctly. (XType.php instead of XTypeType.php).
Bug may be reproduced with "Controller" instead of "Type".

sadikoff:
Added datetimetz, *_immutable, dateinterval and json fileds processing.

closes #234
closes #270

### DIFF
--- a/src/GeneratorTwigHelper.php
+++ b/src/GeneratorTwigHelper.php
@@ -31,14 +31,28 @@ final class GeneratorTwigHelper
         $printCode = $entity.'.'.str_replace('_', '', $twigField);
 
         switch ($field['type']) {
+            case 'datetimetz_immutable':
+            case 'datetimetz':
+                $printCode .= ' ? '.$printCode.'|date(\'Y-m-d H:i:s T\') : \'\'';
+                break;
+            case 'datetime_immutable':
             case 'datetime':
                 $printCode .= ' ? '.$printCode.'|date(\'Y-m-d H:i:s\') : \'\'';
                 break;
+            case 'dateinterval':
+                $printCode .= ' ? '.$printCode.'.format(\'%y year(s), %m month(s), %d day(s)\') : \'\'';
+                break;
+            case 'date_immutable':
             case 'date':
                 $printCode .= ' ? '.$printCode.'|date(\'Y-m-d\') : \'\'';
                 break;
+            case 'time_immutable':
             case 'time':
                 $printCode .= ' ? '.$printCode.'|date(\'H:i:s\') : \'\'';
+                break;
+            case 'json':
+            case 'json_array':
+                $printCode .= ' ? '.$printCode.'|json_encode : \'\'';
                 break;
             case 'array':
                 $printCode .= ' ? '.$printCode.'|join(\', \') : \'\'';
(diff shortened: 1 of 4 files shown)


src/GeneratorTwigHelper.php
```diff
@@ -31,15 +31,29 @@ public function getEntityFieldPrintCode($entity, $field): string
...
                 $printCode .= ' ? '.$printCode.'|date(\'Y-m-d\') : \'\'';
                 break;
+            case 'time_immutable':
             case 'time':
                 $printCode .= ' ? '.$printCode.'|date(\'H:i:s\') : \'\'';
                 break;
+            case 'json':
+            case 'json_array':
+                $printCode .= ' is iterable ? '.$printCode.'|join(\', \') : '.$printCode;
+                break;
```
weaverryan: Interesting! From looking at the docs, it seems like the value in this field will always be an `array` or `null`. Which case were you covering with the `is iterable`?

And, because the array can be an associative array, the `|join` won't look quite right in those cases. We could use `dump()` WDYT?

weaverryan:
> How about adding new test cases to make sure these behave as intended?
It should be easy enough to add them right here:

I agree! And actually, this test could be even a little bit cooler if we actually "rendered" the result through Twig (passing in some real `entity` object) and asserted the final *output* (instead of or in addition to asserting the Twig code that should be printed). But... that's not a requirement :).

src/GeneratorTwigHelper.php
sadikoff: I'll check it again. I thought there could be string value.

`dump()` is a good idea, but will it work without debug package?

weaverryan: It believe that it DOES work without `debug` - it reverts to `var_dump()`. We can add a `json` field to our entity in the functional test to be sure.

weaverryan: Oh, you know what? Let's use `|json_encode`. That will print nicely, and make it very obvious that this is a JSON field. And, it will definitely work :).

@sadikoff if you agree, can you make that change? Also, if you rebase from the latest master, the tests should be closer to passing.

sadikoff:
[c769605](https://github.com/symfony/maker-bundle/pull/241/commits/c769605d55d7dafb1b174e828da16848fd357069) will close #277

weaverryan:
@sadikoff Can you rebase this? I want to make sure the tests pass with the latest changes on master. Also, there are 2 changes in this PR to `.travis.yml` - do we want those?

sadikoff:
@weaverryan ok, I'll rebase it

P.S. I think travis will be more stable with this changes

weaverryan:
Thanks Vladimir!

PR #243, opened 2018-08-10
Fixes conflict between namespace psr-4 prefix and target directory structure

### LINKED ISSUE #242: Conflict between namespace psr-4 prefix and target directory structure
Steps to reproduce:

- Create a symfony project and leave App\\ as a psr-4 prefix
- Create a controller with make:controller App\\MyController

Expected behaviour:

- Generated class named MyController
- Generated namespace App\Controller\App\MyController
- Generated file src/Controller/App/MyController.php

Actual result:

- Generated class named MyController
- Generated namespace App\Controller\App\MyController
- Generated file src/Controller/MyController.php

This is because when generating the file path, the psr-4 prefix is replaced in all of the path string like so (Symfony\Bundle\MakerBundle\Util\AutoloaderUtil::getPathForFutureClass()):
str_replace($prefix, '', $className)

If at any point in your directory structure, a directory is named exactly as your psr-4 prefix, it will be removed from the file path.

A simple fix would be to only remove the prefix at the begining of the string. Suggested implementation would be:
substr($className, strlen($prefix))

gmanen:
Fixes #242

### DIFF
--- a/src/Util/AutoloaderUtil.php
+++ b/src/Util/AutoloaderUtil.php
@@ -46,7 +46,7 @@ class AutoloaderUtil
         // lookup is obviously modeled off of Composer's autoload logic
         foreach ($classLoader->getPrefixesPsr4() as $prefix => $paths) {
             if (0 === strpos($className, $prefix)) {
-                return $paths[0].'/'.str_replace('\\', '/', str_replace($prefix, '', $className)).'.php';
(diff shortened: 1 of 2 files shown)


weaverryan:
Hey Guillaume! Wow, great job! Awesome issue - very clear to follow. I was able to easily use it to add a test case for the bug, which WAS indeed failing :). Thank you!

PR #245, opened 2018-08-10
fix MakerEntity when generating entity inside folder

### LINKED ISSUE #192: Generating entity inside folder eg: Attr\Values, and entity with name Attr already exists
I'm trying to generate an entity, using: 

`php bin/console make:entity`

When in the Class name of the entity I put the following value: Attr\Values, and the Attr class already exists, so:
\App\Entity\Attr.php
\App\Entity\Attr\Values.php ( the new one I'm creating ), the following happens:

 

> created: src/Entity/Attr/Values.php
>  created: src/Repository/Attr/ValuesRepository.php
> 2018-05-29T15:47:54+00:00 [error] Error thrown while running command "make:entity". Message: "Call to a member function getName() on string"
> 2018-05-29T15:47:54+00:00 [critical] Call to a member function getName() on string
> 
> In MakeEntity.php line 837:
> 
>   Call to a member function getName() on string

The curious thing, is that the entity is created, nothing is changed on the Attr entity, and if run the command again to edit the entity Attr\Values, everything is fine, and everything works perfectly.

I'm on Windows 10, and running v1.5.0 of maker-bundle.

nikophil:
fixes #192 

When we're trying to generate an entity inside a folder, and when the namespace of the newly created correspond to an entity that already exists, an error occurs : 

> In MakeEntity.php line 840:
> Call to a member function getName() on string

Ex :
An entity called App\Entity\BlogPost exists, and we try to create an entity named "BlogPost\Comment" 

Actually `Str::getNamespace($className)` refers to an existing class and not to a namespace, then $doctrineHelper->getMetadata() won't return an array, but a single metadata, and this leads to an error.

To prevent this behavior, the easier way is to put a trailing backslash at the end of the namespace name, so we know it won't refer to a classname.

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -827,7 +827,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
     private function doesEntityUseAnnotationMapping(string $className): bool
     {
         if (!class_exists($className)) {
-            $otherClassMetadatas = $this->doctrineHelper->getMetadata(Str::getNamespace($className), true);
+            $otherClassMetadatas = $this->doctrineHelper->getMetadata(Str::getNamespace($className) . '\\', true);
 
             // if we have no metadata, we should assume this is the first class being mapped
             if (empty($otherClassMetadatas)) {


weaverryan:
Wow, awesome diagnosis and fix! This rocks! I added a failing test case which confirmed the bug (and your patch DOES fix it of course). Great PR - thank you Nicolas! I hope to see you back for some more PR's ;)

nikophil:
haha thanks :+1:

PR #246, opened 2018-08-11
Extend from AbstractController and use Symfony's @Route

### LINKED ISSUE #240: make:crud not using AbstractController
Seeing the recent change, I believe this one should also be updated to use the `AbstractController`

https://github.com/symfony/maker-bundle/blob/a1b0d9ff35219ec36c44fcc6213056cc339a1cf8/src/Resources/skeleton/crud/controller/Controller.tpl.php#L18

royklutman:
This PR would close #240 and update the `make:crud` command to correctly use Symfony's Routing component instead of the deprecated `FrameworkExtraBundle`'s. Would appreciate a thorough review as all I did was basically look at @javiereguiluz's PR #221 for the AbstractController fix.

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -13,7 +13,7 @@ namespace Symfony\Bundle\MakerBundle\Maker;
 
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use Doctrine\Common\Inflector\Inflector;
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks Roy! It looks like you did a very good job! We have some great tests to make sure the functionality of the generated code actually works - so we don't need to worry. Thanks for noticing and fixing this!

PR #254, opened 2018-09-02
Generate forms in Form\Type namespace

chindit:
According to rule 37-045 of Insights, Forms should by in `Form\Type` namespace
[See Insights rule](https://insight.sensiolabs.com/what-we-analyse/symfony.form.form_type_not_in_type_form_folder)

### DIFF
--- a/src/Maker/MakeForm.php
+++ b/src/Maker/MakeForm.php
@@ -76,7 +76,7 @@ final class MakeForm extends AbstractMaker
     {
         $formClassNameDetails = $generator->createClassNameDetails(
             $input->getArgument('name'),
-            'Form\\',
+            'Form\\Type\\',
             'Type'
         );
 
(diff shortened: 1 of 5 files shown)


weaverryan:
Hmm. I disagree with this actually. Not strongly, but I still disagree :). @tgalopin do you know who I might chat to about this rule?

chindit:
It's just a matter of coherency.
If a SF tool generates a code which is not good for SF inspection... Well, that's bad.

weaverryan:
I totally agree :). Thanks for reporting it - we should figure out which way to go.

tgalopin:
According to https://symfony.com/doc/current/best_practices/forms.html, the Maker is indeed right. The best practice changed, I will update Insight :) . Thanks for the feedback @chindit !

weaverryan:
Cheers people!

truckee:
I'm stumped. The [reference ](https://symfony.com/doc/current/best_practices/forms.html) above says nothing about a Form\Type directory (at least as of this writing). Also, a [discussion ](https://github.com/symfony/demo/issues/6) a while back suggests that Form\Type is preferred. I've found that `make:form` fails if any form type exists in App\Form\Type. So I leave that folder empty until I've made my forms. Should this be reopened?

PR #256, opened 2018-09-03
Add framework-extra-bundle dependency

### LINKED ISSUE #255: make:crud requires framework-extra-bundle
make:crud generates a controller which uses ParamConverter to load entities.

Steps to reproduce:
```bash
composer create-project symfony/skeleton
composer req twig orm maker form validator server symfony/security-csrf
```

Update .env to have some db connectivity:
`DATABASE_URL="sqlite:///%kernel.project_dir%/var/app.db"`

```bash
php bin/console make:entity Foobar // add a name field or something
php bin/console doctrine:schema:create
php bin/console make:crud Foobar
php bin/console server:start
```
Now open the webserver and browse to the route, probably http://127.0.0.1:8000/foobar/new
Add an entity
Click on the entity show or edit link, probably http://127.0.0.1:8000/foobar/1

> Cannot autowire argument $foobar of "App\Controller\FoobarController::show()": it references class "App\Entity\Foobar" but no such service exists.

Entities are not supposed to be autowired, in this case they should be loaded by ParamConverter.

Fix this by installing the bundle containing the paramconverter:
`composer require sensio/framework-extra-bundle`

Note that I started with `symfony/skeleton`instead of `symfony/website-skeleton`, which includes the framework-extra-bundle. However the dependency should be included in `MakeCrud::configureDependencies`.

ckrack:
The make:crud controllers rely on ParamConverters,
which are provided by framework-extra-bundle.
Fixes #255

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -216,7 +216,7 @@ final class MakeCrud extends AbstractMaker
     {
         $dependencies->addClassDependency(
             Route::class,
-            'annotations'
+            'router'
         );
 
         $dependencies->addClassDependency(
@@ -246,7 +246,7 @@ final class MakeCrud extends AbstractMaker
 
         $dependencies->addClassDependency(
             ParamConverter::class,
-            'sensio/framework-extra-bundle'
+            'annotations'
         );
     }
 }
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -243,5 +243,10 @@ final class MakeCrud extends AbstractMaker
             CsrfTokenManager::class,
             'security-csrf'
         );
+
+        $dependencies->addClassDependency(
+            ParamConverter::class,
+            'sensio/framework-extra-bundle'
+        );
     }
 }


weaverryan:
Hey @ckrack!

Hmm, this makes sense. But, we have a test to catch this.... @sadikoff what do you think? It looks like our CRUD test definitely checks controllers that use this - https://github.com/symfony/maker-bundle/blob/master/tests/fixtures/MakeCrud/tests/GeneratedCrudControllerTest.php - so if we're missing a dependency, shouldn't that test fail?

ckrack:
Hi @weaverryan 
I haven't found out all details about the tests, but:
There is an exclude for `tests/fixtures` in `phpunit.xml.dist`

I found one other test here, but it's only making a GET to the list method. https://github.com/symfony/maker-bundle/blob/master/tests/fixtures/MakeCrudRepository/tests/GeneratedCrudControllerTest.php

In https://github.com/symfony/maker-bundle/blob/master/tests/Maker/FunctionalTest.php, there are a couple of comments `// need for crud web tests`, maybe that's an indication of missing tests?

`MakeCrud::configureDependencies` only checks for the existance of `Symfony\Component\Routing\Annotation\Route`, which is part of the routing component.
If this fails, it suggests installing "annotations". I'd change this to suggest the router in this place and add the block in this pull request for the annotations.

By the way: I can't run the tests properly.
When i leave the example db DSN in phpunit.xml.dist, it complains about Access denied. When i replace it with sqlite, it complains about: 

> 1) Symfony\Bundle\MakerBundle\Tests\Maker\FunctionalTest::testCommands with data set "crud_with_no_base" (Symfony\Bundle\MakerBundle\Test\MakerTestDetails Object (...))
Exception: Could not find "mysql://db_user:db_password@127.0.0.1:3306/db_name" inside "phpunit.xml.dist"

weaverryan:
Oh, one more thing:

> By the way: I can't run the tests properly.
When i leave the example db DSN in phpunit.xml.dist, it complains about Access denied. When i replace it with sqlite, it complains about:
> Symfony\Bundle\MakerBundle\Tests\Maker\FunctionalTest::testCommands with data set "crud_with_no_base" (Symfony\Bundle\MakerBundle\Test\MakerTestDetails Object (...))
Exception: Could not find "mysql://db_user:db_password@127.0.0.1:3306/db_name" inside "phpunit.xml.dist"

I need your help with this :). You are the 2nd person today who has reported this. So, I think there is a legitimate issue. However, I can't repeat it locally or on TravisCI. What does your `tests/tmp/cache/flex_project/phpunit.xml.dist` file look like? Also, what does this same `phpunit.xml.dist` file look like inside one of the `maker_*` directories inside that same `tests/tmp/cache` folder (e.g. what does `tests/tmp/cache/maker_98f004a33a/phpunit.xml.dist`)?

We create a mini Flex app during the tests. We run `composer require phpunit`, and that recipe should give us a `phpunit.xml.dist` file that contains the string in the error that (for some reason) cannot be found.

If you can help a bit debugging this, it would be awesome. The test suite is super powerful... but some people are having trouble running it, which is no good :).

Cheers!

PR #261, opened 2018-09-06
MakeAuthenticator : update security.yaml

nikophil:
This PR adds the newly created authenticator to `security.yaml`

The process is the following :
- if no firewall exist, a "main" firewall is automatically created (this should never happen)
- if there is only one "real" firewall (i.e. : a firewall withou security: false"), the authenticator is added to this firewall
- if there are more than one firewall, the user is asked to which firewall the authenticator should be added
- if the target firewall already has an authenticator, the user is asked which authenticator should be the entry point.

### DIFF
--- a/src/Security/InteractiveSecurityHelper.php
+++ b/src/Security/InteractiveSecurityHelper.php
@@ -19,7 +19,7 @@ use Symfony\Component\Console\Style\SymfonyStyle;
  */
 final class InteractiveSecurityHelper
 {
-    public function guessFirewallName(SymfonyStyle $io, array $securityData)
+    public function guessFirewallName(SymfonyStyle $io, array $securityData): string
     {
         $realFirewalls = array_filter(
             $securityData['security']['firewalls'] ?? [],
--- a/src/Security/InteractiveSecurityHelper.php
+++ b/src/Security/InteractiveSecurityHelper.php
@@ -11,18 +11,25 @@
 
 namespace Symfony\Bundle\MakerBundle\Security;
 
-use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
  * @internal
  */
 final class InteractiveSecurityHelper
 {
-    public function guessFirewallName(InputInterface $input, ConsoleStyle $io, array $securityData)
+    /**
+     * @param InputInterface $input
+     * @param SymfonyStyle   $io
+     * @param array          $securityData
+     */
+    public function guessFirewallName(InputInterface $input, SymfonyStyle $io, array $securityData)
     {
         $firewalls = array_filter(
-            $securityData['security']['firewalls'],
+            $securityData['security']['firewalls'] ?? [],
             function ($item) {
                 return !isset($item['security']) || true === $item['security'];
             }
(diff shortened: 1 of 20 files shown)


tests/Maker/FunctionalTest.php
```diff
@@ -282,12 +283,124 @@ public function getCommandTests()
...
+            MakerTestDetails::createTest(
+                $this->getMakerInstance(MakeAuthenticator::class),
+                [
+                    // class name
+                    'AppCustomAuthenticator',
+                ]
+            )
+                ->addExtraDependencies('security')
+                ->setFixtureFilesPath(__DIR__.'/../fixtures/MakeAuthenticator')
+                ->setRequiredPhpVersion(70100)
```
weaverryan: I don't think we need this. I added it for `MakeUser`, but that was because we were using MakeEntity in the background, which generates PHP 7.1 code (nullable types). But, in this case, the authenticator doesn't have nullable types, and updating the YAML obviously is ok.

tests/Maker/FunctionalTest.php
```diff
@@ -282,12 +283,124 @@ public function getCommandTests()
...
+                ->assert(
+                    function (string $output, string $directory) {
+                        $this->assertContains('Success', $output);
+
+                        $finder = new Finder();
+                        $finder->in($directory.'/src/Security')
+                            ->name('AppCustomAuthenticator.php');
+                        $this->assertCount(1, $finder);
```
weaverryan: We can simplify to just check `$this->assertFileExists()` and check for the `$directory.'/src/Security/AppCustomAuthenticator.php'.

tests/Maker/FunctionalTest.php
```diff
@@ -282,12 +283,124 @@ public function getCommandTests()
...
+
+                        $securityConfig = Yaml::parse(file_get_contents("$directory/config/packages/security.yaml"));
```
weaverryan: sprintf with single quotes

tests/Maker/FunctionalTest.php
```diff
@@ -282,12 +283,124 @@ public function getCommandTests()
...
+                ),
+        ];
+
+        yield 'auth_empty_multiple_firewalls' => [
+            MakerTestDetails::createTest(
+                $this->getMakerInstance(MakeAuthenticator::class),
+                [
+                    // class name
+                    'AppCustomAuthenticator',
+                    1
```
weaverryan: Put a comment about what question we're answering

tests/Maker/FunctionalTest.php
```diff
@@ -282,12 +283,124 @@ public function getCommandTests()
...
+                ->setFixtureFilesPath(__DIR__.'/../fixtures/MakeAuthenticatorMultipleFirewalls')
+                ->setRequiredPhpVersion(70100)
+                ->assert(
+                    function (string $output, string $directory) {
+                        $this->assertContains('Success', $output);
+
+                        $finder = new Finder();
+                        $finder->in($directory.'/src/Security')
+                            ->name('AppCustomAuthenticator.php');
+                        $this->assertCount(1, $finder);
```
weaverryan: I think we can avoid this check on all the tests, except for the frist.

tests/Maker/FunctionalTest.php
```diff
@@ -282,12 +283,124 @@ public function getCommandTests()
...
+                ),
+        ];
+
+        yield 'auth_empty_existing_authenticator' => [
+            MakerTestDetails::createTest(
+                $this->getMakerInstance(MakeAuthenticator::class),
+                [
+                    // class name
+                    'AppCustomAuthenticator',
+                    1
```
weaverryan: same here - comment for what the option is

tests/Maker/FunctionalTest.php
```diff
@@ -282,12 +283,124 @@ public function getCommandTests()
...
+        ];
+
+        yield 'auth_empty_multiple_firewalls_existing_authenticator' => [
+            MakerTestDetails::createTest(
+                $this->getMakerInstance(MakeAuthenticator::class),
+                [
+                    // class name
+                    'AppCustomAuthenticator',
+                    1,
+                    1
```
weaverryan: comments here

tests/Security/InteractiveSecurityHelperTest.php
```diff
@@ -0,0 +1,175 @@
...
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class InteractiveSecurityHelperTest extends TestCase
+{
+    /**
+     * @dataProvider getFirewallNameTests
+     *
+     * @param array  $securityData
+     * @param string $expectedFirewallName
+     * @param bool   $multipleValues
```
weaverryan: Remove these `@param`

tests/Security/InteractiveSecurityHelperTest.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public function testGuessFirewallName(array $securityData, string $expectedFirewallName, $multipleValues = false)
+    {
+        /** @var InputInterface|\PHPUnit_Framework_MockObject_MockObject $input */
+        $input = $this->createMock(InputInterface::class);
+        $input->expects($this->once())
+            ->method('setOption')
+            ->with('firewall-name', $expectedFirewallName);
```
weaverryan: This will simplify a but, when we start passing in `$firewallName` directly :)

src/Security/InteractiveSecurityHelper.php
```diff
@@ -0,0 +1,94 @@
...
+        $firewalls = $securityData['security']['firewalls'];
+        if (!isset($firewalls[$firewallName])) {
+            throw new RuntimeCommandException("Firewall \"$firewallName\" does not exist");
+        }
+
+        if (!isset($firewalls[$firewallName]['guard'])
+            || !isset($firewalls[$firewallName]['guard']['authenticators'])
+            || !$firewalls[$firewallName]['guard']['authenticators']) {
+            return;
+        }
```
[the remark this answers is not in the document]
nikophil: isn't it better to let the user chose ?

src/Security/InteractiveSecurityHelper.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @internal
+ */
+final class InteractiveSecurityHelper
+{
+    public function guessFirewallName(SymfonyStyle $io, array $securityData)
```
weaverryan: Return type: `: string`

src/Security/InteractiveSecurityHelper.php
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        if (1 === \count($realFirewalls)) {
+            return key($realFirewalls);
+        }
+
+        return $io->choice('Which firewall do you want to update ?', array_keys($realFirewalls), key($realFirewalls));
+    }
+
+    public function guessEntryPoint(SymfonyStyle $io, array $securityData, string $authenticatorClass, string $firewallName)
```
weaverryan: Good job with no return type here... because we can't use the `: ?string` yet.

weaverryan:
Test failures are unrelated - working on them in a different PR :)

weaverryan:
This is awesome! Thank you Nicolas!

PR #263, opened 2018-09-07
[Controller] Adding the --no-annotations option to make:controller

### LINKED ISSUE #262: [Controller] Add --no-annotations option
I think it might be a nice addition to add the option `--no-annotations` to the `make:controller` command so the following lines are not included if you want to use `routes.yaml` instead of annotations for routes:

https://github.com/symfony/maker-bundle/blob/22507228ecc97f7c843915da99d0a8a517185130/src/Resources/skeleton/controller/Controller.tpl.php#L6

https://github.com/symfony/maker-bundle/blob/22507228ecc97f7c843915da99d0a8a517185130/src/Resources/skeleton/controller/Controller.tpl.php#L10-L12

I'd like to hear what you guys think about that.

royklutman:
This closes #262.

### DIFF
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -22,6 +22,7 @@ use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
@@ -39,6 +40,7 @@ final class MakeController extends AbstractMaker
         $command
             ->setDescription('Creates a new controller class')
             ->addArgument('controller-class', InputArgument::OPTIONAL, sprintf('Choose a name for your controller class (e.g. <fg=yellow>%sController</>)', Str::asClassName(Str::getRandomTerm())))
+            ->addOption('no-annotations', null, InputOption::VALUE_NONE, 'Do not use annotations for declaring routes')
             ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeController.txt'))
         ;
     }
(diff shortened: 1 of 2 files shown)


src/Maker/MakeController.php
```diff
@@ -80,14 +83,16 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
-    public function configureDependencies(DependencyBuilder $dependencies)
+    public function configureDependencies(DependencyBuilder $dependencies, InputInterface $input = null)
     {
-        $dependencies->addClassDependency(
-            // we only need doctrine/annotations, which contains
-            // the recipe that loads annotation routes
-            Annotation::class,
-            'annotations'
-        );
+        if (null === $input) {
```
royklutman: I am not completely sure about this myself. If there is a better way to handle this, please feel free to edit or comment.

weaverryan: Don't we need to check for the option? `if (null !== $input && !$input->getOption('no-annotations'))` - then the `Annotation` dependency is optional.

src/Maker/MakeController.php
```diff
@@ -60,6 +62,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
                 'route_path' => Str::asRoutePath($controllerClassNameDetails->getRelativeNameWithoutSuffix()),
                 'route_name' => Str::asRouteName($controllerClassNameDetails->getRelativeNameWithoutSuffix()),
                 'twig_installed' => $this->isTwigInstalled(),
+                'use_annotations' => !$input->getOption('no-annotations'),
                 'template_name' => $templateName,
             ]
```
weaverryan: If we're going to do this, then you should also add the route to `routes.yaml`. There is a `YamlSourceManipulator` capable of updating YAML files without losing comments, formatting, etc.

weaverryan:
On second thought, I think we can't do this. It's a slippery slope. We purposely not supporting *too* many options in this bundle, because that's what its predecessor did (SensioGeneratorBundle) and it suffered from being bloated, and it was difficult to add features. If we add this to `make:controller`, then we need it also in `make:crud`. AND, in the new "login form" part of `make:auth` - #266. You can see the problem - it's not just one small place where we need this. 

For that reason, I'm 👎 . We have to keep a focus on the "main" workflow that people use.

Thanks!

PR #265, opened 2018-09-10
"make:entity" root directory location improvment

bastos71:
As a developer using Symfony 3.x, I got issues using `make:entity`

I made this pull request to allow users to override the hardcoded `src/Entity/` as the root directory for entities.
Now there are two ways to override it : 

* define the new available configuration in your `config.yml`
```yaml
maker:
    root_namespace: Acme\AwesomeBundle
    root_entity_directory: src/Acme/AwesomeBundle/Entity/
```
* use the new option of the `make:entity` command : 
```shell
php bin/console make:entity --rootdir=src/Acme/AwesomeBundle/Entity/
```

Anyway the file will be created to the correct location (corresponding to the class namespace), that was already the case before my pull request.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -27,6 +27,7 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->children()
                 ->scalarNode('root_namespace')->defaultValue('App')->end()
+                ->scalarNode('root_entity_directory')->defaultValue('src/Entity/')->end()
             ->end()
         ;
 
--- a/src/DependencyInjection/MakerExtension.php
+++ b/src/DependencyInjection/MakerExtension.php
@@ -37,6 +37,11 @@ class MakerExtension extends Extension
         $configuration = $this->getConfiguration($configs, $container);
         $config = $this->processConfiguration($configuration, $configs);
 
+        $rootEntityDirectory = trim($config['root_entity_directory']);
+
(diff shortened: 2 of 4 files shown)


bastos71:
The error returned by the AppVeyor build seems not to be dued to my development : 

_Error: "PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\ClassNotFoundException: Attempted to load class "Kernel" from namespace "Custom"_

weaverryan:
Hey @bastos71!

Ah, I think I understand the issue: it's just that the auto-complete questions don't fill from the correct entity directory - that part IS indeed hardcoded. But, I think there's a better solution - and I think you already saw the PR I'm talking about :) https://github.com/symfony/maker-bundle/pull/185#issuecomment-389009674

As I mentioned there, there's no reason to force the user to think/configure an option at all. We can ask Doctrine for the list of mapped entity and create an auto-complete list from that. In fact, I think we may already have the hard part done :) - see: https://github.com/symfony/maker-bundle/blob/24ae473faece63318f5c4b429005c86668c88fc0/src/Doctrine/DoctrineHelper.php#L101-L116

This code was created after MakeEntity I believe, so we may just be able to point to that and fix it. Would you be willing to try it?

bastos71:
Thanks for your reply @weaverryan 
Ok I will try to implement this, parsing the list of mapped entities to get a list of namespaces.

But what if the user tries to create its very First Entity ? That means there would not have any mapped entities, so no namespace to list :/

weaverryan:
Closing in favor of #310 or #311

PR #266, opened 2018-09-11
Make login form

nikophil:
Improvements for the command `make:auth` in order to create a form login.
Now, the `make:auth` command asks if the user wants an empty authenticator or a form login.
If "form login" is chosen, the user  is asked for the security controller name, and for the user class, if needed.
If the controller already exists, the command adds the login / logout methods in the existing controller.


src/Maker/MakeAuthenticator.php
```diff
@@ -55,50 +70,94 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+
+        $command->addOption('entry-point', null, InputOption::VALUE_OPTIONAL);
         $input->setOption(
             'entry-point',
-            $interactiveSecurityHelper->guessEntryPoint($io, $securityData, $authenticatorClassNameDetails->getFullName(), $firewallName)
+            InteractiveSecurityHelper::guessEntryPoint($io, $securityData, $input->getArgument('authenticator-class'), $firewallName)
         );
+
+        if (self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type')) {
+            $command->addArgument('controller-class', InputArgument::OPTIONAL);
```
weaverryan: Optional? I don't think it matters, because we're always passing it. But you use required above - let's be consistent. This looks required :).

src/Maker/MakeAuthenticator.php
```diff
@@ -55,50 +70,94 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+            $input->getArgument('authenticator-class'),
+            self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type') ?
+                ($this->doctrineHelper->isClassAMappedEntity($input->getArgument('user-class')) ? 'authenticator/LoginFormEntityAuthenticator.tpl.php' : 'authenticator/LoginFormNotEntityAuthenticator.tpl.php')
+                : 'authenticator/Empty.tpl.php',
+            $input->hasArgument('user-class') ?
+                [
+                    'user_fully_qualified_class_name' => $input->getArgument('user-class'),
+                    'user_class_name' => substr(strrchr($input->getArgument('user-class'), '\\'), 1),
+                ]
+                : []
```
weaverryan: I found the ternary you were asking about :p. I think an if statement around this entire block is much more clear. Actually, even a switch-case

```php
switch ($input->getArgument('authenticator-type')) {
    case self::AUTH_TYPE_FORM_LOGIN:
        $generator->generateClass(
            // ...
        );
        break;
    case self::self::AUTH_TYPE_EMPTY_AUTHENTICATOR:
        $generator->generateClass(
            // ...
        );
        break;
    default:
        throw new \Exception('Invalid authenticator type');
}
```

src/Maker/MakeAuthenticator.php
```diff
@@ -55,50 +70,94 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
-            'authenticator/Empty.tpl.php',
-            []
+            $input->getArgument('authenticator-class'),
+            self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type') ?
+                ($this->doctrineHelper->isClassAMappedEntity($input->getArgument('user-class')) ? 'authenticator/LoginFormEntityAuthenticator.tpl.php' : 'authenticator/LoginFormNotEntityAuthenticator.tpl.php')
+                : 'authenticator/Empty.tpl.php',
+            $input->hasArgument('user-class') ?
+                [
+                    'user_fully_qualified_class_name' => $input->getArgument('user-class'),
+                    'user_class_name' => substr(strrchr($input->getArgument('user-class'), '\\'), 1),
```
weaverryan: We should use the `ClassNameDetails` for this

src/Maker/MakeAuthenticator.php
```diff
@@ -107,14 +166,62 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
-                    $classNameDetails->getFullName()
+                    $input->getArgument('authenticator-class')
                 );
                 $generator->dumpFile($path, $newYaml);
                 $securityYamlUpdated = true;
             } catch (YamlManipulationFailedException $e) {
             }
         }
 
+        if (self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type')) {
```
weaverryan: Let's put the body of all of this into a private function for readability - e.g. `$this->generateLoginForm(...);`

src/Maker/MakeAuthenticator.php
```diff
@@ -107,14 +166,62 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+                $this->generator->dumpFile($controllerPath, $manipulator->getSourceCode());
+            } else {
+                // otherwise, create security controller
+                $controllerPath = $generator->generateClass(
+                    $controllerClassNameDetails->getFullName(),
+                    'authenticator/SecurityController.tpl.php',
+                    [
+                        'parent_class_name' => \method_exists(AbstractController::class, 'getParameter') ? 'AbstractController' : 'Controller',
+                    ]
+                );
```
weaverryan: We can simplify this a bit. You've already done the hard work in the method above to *modify* an existing controller... no matter what it looks like. So, here, we should just generate an *empty* class - something *almost* as simple as `Class.tpl.php` (but we'd obviously need the `extends` still). Then, we would run the code in the `if` part of this always. So:

```php
if (!class_exists($controllerClassNameDetails->getFullName())) {
    // make a dead-simple controller
}

// now, in all cases, do some ClassSourceManipulator magic!
```

Btw, I'm a bit concerned that we will now have the `parent_class_name` logic in 3 separate places. I'm not sure immediately how to fix that - but it bothers me.

src/Maker/MakeAuthenticator.php
```diff
@@ -107,14 +166,62 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+                    $this->fileManager->getFileContents($controllerPath = $this->fileManager->getRelativePathForFutureClass($controllerClassNameDetails->getFullName())),
+                    true
+                );
+                $securityControllerBuilder = new SecurityControllerBuilder();
+                $securityControllerBuilder->addLoginMethod($manipulator);
+
+                if (method_exists($controllerClassNameDetails->getFullName(), 'logout')) {
+                    throw new RuntimeCommandException(sprintf('Method "logout" already exists on class %s', $controllerClassNameDetails->getFullName()));
+                }
+                $securityControllerBuilder->addLogoutMethod($manipulator);
```
weaverryan: Logout. That's interesting. It's kinda unrelated to this. But... it *is* kinda a nice feature. My instinct is to *not* have it - too unrelated - it's a separate thing.

src/Maker/MakeAuthenticator.php
```diff
@@ -107,14 +166,62 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+                    $controllerClassNameDetails->getFullName(),
+                    'authenticator/SecurityController.tpl.php',
+                    [
+                        'parent_class_name' => \method_exists(AbstractController::class, 'getParameter') ? 'AbstractController' : 'Controller',
+                    ]
+                );
+            }
+
+            // create login form template
+            $templateName = Str::asFilePath($controllerClassNameDetails->getRelativeNameWithoutSuffix()).'/login.html.twig';
```
weaverryan: There's disconnect here I think. This code looks more or less right to me. But I think when the controller is generated, it's *always* `security/login.html.twig`

src/Resources/help/MakeAuth.txt
```diff
@@ -1,6 +1,4 @@
 The <info>%command.name%</info> command generates an empty
 Guard Authenticator class.
 
-<info>php %command.full_name% AppCustomAuthenticator</info>
-
-If the argument is missing, the command will ask for the class name interactively.
+<info>php %command.full_name%</info>
```
weaverryan: We should update this file with some better info now :)

src/Resources/skeleton/authenticator/LoginFormEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,86 @@
...
+    private $router;
+    private $csrfTokenManager;
+    private $passwordEncoder;
+
+    public function __construct(EntityManagerInterface $entityManager, RouterInterface $router, CsrfTokenManagerInterface $csrfTokenManager, UserPasswordEncoderInterface $passwordEncoder)
+    {
+        $this->router = $router;
+        $this->csrfTokenManager = $csrfTokenManager;
+        $this->passwordEncoder = $passwordEncoder;
+        $this->entityManager = $entityManager;
```
weaverryan: Make sure these have the same order as the arguments and as the properties - for sanity :)

src/Resources/skeleton/authenticator/LoginFormEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,86 @@
...
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Component\Security\Guard\Authenticator\AbstractFormLoginAuthenticator;
+use Symfony\Component\Security\Http\Util\TargetPathTrait;
+
+class <?= $class_name; ?> extends AbstractFormLoginAuthenticator
+{
+    use TargetPathTrait;
```
weaverryan: I think we need a line break after this

src/Resources/skeleton/authenticator/LoginFormEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,86 @@
...
+        if (!$this->csrfTokenManager->isTokenValid($token)) {
+            throw new InvalidCsrfTokenException();
+        }
+
+        return $this->entityManager->getRepository(<?= $user_class_name ?>::class)->findOneBy(['email' => $credentials['email']]);
+    }
+
+    public function checkCredentials($credentials, UserInterface $user)
+    {
+        return $this->passwordEncoder->isPasswordValid($user, $credentials['password']);
```
weaverryan: This is tricky! The user may not have a password. Remember, that's one of the questions in `make:user`! It's totally valid, even if the form has a password (I think it's safe to assume that) that they will send the password to some SSO and IT will check it.

Hmm. So, the ONLY way I think we could "guess" this is by looking at the `encoders` config in `security.yaml` to see if the class (or some sub-class or even interface - it's legal to configure the core `Symfony\...\UserInterface` in that file) is configured. If it is, we can probably guess that a password should be checked (and of course, we can add a comment here - they can always modify it). If not, we should just return `true` here (again, with a comment).

WDYT?

src/Resources/skeleton/authenticator/LoginFormEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,86 @@
...
+    public function supports(Request $request)
+    {
+        return 'app_login' === $request->attributes->get('_route')
+            && $request->isMethod('POST');
+    }
+
+    public function getCredentials(Request $request)
+    {
+        $credentials = [
+            'email' => $request->request->get('email'),
```
weaverryan: Hmm. Another tricky spot. What if the user is actually using a `username` property? Dang these little details! In `MakeUser`, this is something we ask: what is your "identifier" property - e.g. `email` or `user`. When this command is ran, I don't think we know/can guess that info. I think we will need to ask them another question. I'm not totally sure how to phrase it:

> When field on your `User` class will people enter when logging in?

I guess if their `User` class has an `email` or `username` property (but not both), it's safe to guess. But, it *does* need to be configurable I think. Which, affects a few other places.

Who thought generators would be so hard? ;)

src/Resources/skeleton/authenticator/LoginFormEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,86 @@
...
+        return $this->passwordEncoder->isPasswordValid($user, $credentials['password']);
+    }
+
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
+    {
+        if ($targetPath = $this->getTargetPath($request->getSession(), $providerKey)) {
+            return new RedirectResponse($targetPath);
+        }
+
+        return new RedirectResponse($this->router->generate('app_homepage'));
```
weaverryan: This route may not exist. Another tricky spot! Hmm. We could throw an exception with a clear "TODO" error here. I mean we COULD do some magic and look at the routes and find one that looks like it's for the homepage. But, that seems a bit nuts.

src/Resources/skeleton/authenticator/LoginFormNotEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,82 @@
...
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Component\Security\Guard\Authenticator\AbstractFormLoginAuthenticator;
+use Symfony\Component\Security\Http\Util\TargetPathTrait;
+
+class <?= $class_name; ?> extends AbstractFormLoginAuthenticator
+{
+    use TargetPathTrait;
```
weaverryan: empty line here again

src/Resources/skeleton/authenticator/LoginFormNotEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,82 @@
...
+    private $router;
+    private $csrfTokenManager;
+    private $passwordEncoder;
+
+    public function __construct(RouterInterface $router, CsrfTokenManagerInterface $csrfTokenManager, UserPasswordEncoderInterface $passwordEncoder)
+    {
+        $this->router = $router;
+        $this->csrfTokenManager = $csrfTokenManager;
+        $this->passwordEncoder = $passwordEncoder;
```
weaverryan: Same comment for this class - the `passwordEncoder` may not be needed

src/Resources/skeleton/authenticator/LoginFormNotEntityAuthenticator.tpl.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public function getUser($credentials, UserProviderInterface $userProvider)
+    {
+        $token = new CsrfToken('authenticate', $credentials['csrf_token']);
+        if (!$this->csrfTokenManager->isTokenValid($token)) {
+            throw new InvalidCsrfTokenException();
+        }
+
+        return $userProvider->loadUserByUsername($credentials['email']);
```
weaverryan: Hmm, you know how I feel about the user provider ;). Hmm. Ok. Let's do this. But, with a comment above:

> Load / create our user however you need. You can do this by calling the user provider, or with custom logic here.

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,31 @@
+{% extends 'base.html.twig' %}
+
+{% block title %}Login!{% endblock %}
+
+{% block body %}
+<form class="form-signin" method="post">
+    {% if error %}
+    <div class="alert alert-danger">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
```
weaverryan: I like to go in 4 spaces inside the if

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,31 @@
+{% extends 'base.html.twig' %}
+
+{% block title %}Login!{% endblock %}
+
+{% block body %}
+<form class="form-signin" method="post">
```
weaverryan: class can be removed

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,31 @@
...
+    <label for="inputPassword" class="sr-only">Password</label>
+    <input type="password" name="password" id="inputPassword" class="form-control" placeholder="Password" required>
+
+    <input type="hidden" name="_csrf_token"
+           value="{{ csrf_token('authenticate') }}"
+    >
+
+    <div class="checkbox mb-3">
+        <label>
+            <input type="checkbox" name="_remember_me"> Remember me
```
weaverryan: Let's comment this out. And add a comment above it:

> Uncomment this section *and* add a `remember_me` option below your firewall to activate remember me functionality. See https://symfony.com/doc/current/security/remember_me.html

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,31 @@
...
+<form class="form-signin" method="post">
+    {% if error %}
+    <div class="alert alert-danger">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
+    {% endif %}
+
+    <h1 class="h3 mb-3 font-weight-normal">Please sign in</h1>
+    <label for="inputEmail" class="sr-only">Email address</label>
+    <input type="email" value="{{ last_username }}" name="email" id="inputEmail" class="form-control" placeholder="Email address" required autofocus>
+    <label for="inputPassword" class="sr-only">Password</label>
+    <input type="password" name="password" id="inputPassword" class="form-control" placeholder="Password" required>
```
weaverryan: As you noticed, we "lightly" try to add Bootstrap4 generated code. The idea is to help people who rare using that, but not get too annoying. I noticed *some* classes on this for Bootstrap? How does this actually look? Are some Bootstrap classes missing? Or does this more or less have what it needs?

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,31 @@
+{% extends 'base.html.twig' %}
+
+{% block title %}Login!{% endblock %}
```
weaverryan: 2 words I think - I always mess this up (I think "Log In" is the verb form)

src/Security/InteractiveSecurityHelper.php
```diff
@@ -12,14 +12,16 @@
...
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Security\Core\User\UserInterface;
 
 /**
  * @internal
  */
 final class InteractiveSecurityHelper
 {
-    public function guessFirewallName(SymfonyStyle $io, array $securityData): string
+    public static function guessFirewallName(SymfonyStyle $io, array $securityData): string
```
weaverryan: I'd prefer non-statics. Even if it means we're instantiating it... and then just using it.

src/Maker/MakeAuthenticator.php
```diff
@@ -55,50 +70,94 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
-            return;
-        }
+        // authenticator type
+        $authenticatorTypeValues = [
+            'Empty authenticator' => self::AUTH_TYPE_EMPTY_AUTHENTICATOR,
+            'Form login' => self::AUTH_TYPE_FORM_LOGIN,
+        ];
+        $command->addArgument('authenticator-type', InputArgument::REQUIRED);
+        $authenticatorType = $io->choice(
+            'Which authentication type do you want ?',
```
weaverryan: What do you think about:

> What style of authentication do you want?

src/Security/SecurityControllerBuilder.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
```
weaverryan: we don't currently add this in the lib

src/Security/SecurityControllerBuilder.php
```diff
@@ -0,0 +1,68 @@
...
+        'last_username' => $lastUsername,
+        'error' => $error,
+    ]
+);
+CODE
+        );
+        $manipulator->addMethodBuilder($loginMethodBuilder);
+        $manipulator->addUseStatementIfNecessary('Symfony\\Component\\HttpFoundation\\Response');
+        $manipulator->addUseStatementIfNecessary('Symfony\\Component\\Routing\\Annotation\\Route');
+        $manipulator->addUseStatementIfNecessary('Symfony\\Component\\Security\\Http\\Authentication\\AuthenticationUtils');
```
weaverryan: Can we move these `addUseStatementIfNecessary` up a little? It's a nit-pick, but I think it's clearer if they are right near where they are used/needed.

Also, we can use the `::class` syntax with `use` statements on top.

src/Security/InteractiveSecurityHelper.php
```diff
@@ -75,4 +77,22 @@ public function guessEntryPoint(SymfonyStyle $io, array $securityData, string $a
             current($authenticators)
         );
     }
+
+    public static function guessUserClass(SymfonyStyle $io, array $securityData): string
+    {
+        if (1 === \count($securityData['security']['providers']) && isset(current($securityData['security']['providers'])['entity'])) {
+            $entityProvider = current($securityData['security']['providers']);
+            $userClass = $entityProvider['entity']['class'];
```
weaverryan: Just `return $entityProvider['entity']['class'];`. Then the next part doesn't need to be in the `else`

src/Security/InteractiveSecurityHelper.php
```diff
@@ -75,4 +77,22 @@ public function guessEntryPoint(SymfonyStyle $io, array $securityData, string $a
...
+        } else {
+            $userClass = $io->ask(
+                'Enter the User class you want to authenticate (e.g. <fg=yellow>App\\Entity\\User</>)
```
weaverryan: > Enter the User class that you want to authenticate

src/Security/InteractiveSecurityHelper.php
```diff
@@ -75,4 +77,22 @@ public function guessEntryPoint(SymfonyStyle $io, array $securityData, string $a
...
+ (It has to be handled by one of the firewall\'s providers)',
```
weaverryan: Hmm, let's remove this. I know why you have it, but feel like it's just going to confuse people. Probably an edge case that they've successfully created a `UserInterface` User class, but have not setup a provider and need one. And even if they DO make this mistake, it's not really our job to catch here.

src/Security/InteractiveSecurityHelper.php
```diff
@@ -75,4 +77,22 @@ public function guessEntryPoint(SymfonyStyle $io, array $securityData, string $a
...
+                class_exists('App\\Entity\\User') && isset(class_implements('App\\Entity\\User')[UserInterface::class]) ? 'App\\Entity\\User'
+                    : class_exists('App\\Security\\User') && isset(class_implements('App\\Security\\User')[UserInterface::class]) ? 'App\\Security\\User' : null,
```
weaverryan: Oh, that's a nice idea! Let's move this logic into a private function though, and do it without ternaries. It'll make it much easier, because you can return quickly - e.g.

```php
if (class_exists('App\\Entity\\User') && isset(class_implements('App\\Entity\\User')[UserInterface::class])) {
    return 'App\\Entity\\User';
}

// same for other case
```

src/Validator.php
```diff
@@ -184,4 +185,26 @@ public static function entityExists(string $className = null, array $entities =
...
+
+        return $className;
+    }
+
+    public static function classIsUserInterface($userClassName)
+    {
+        self::classExists($userClassName);
+
+        if (!isset(class_implements($userClassName)[UserInterface::class])) {
+            throw new RuntimeCommandException(sprintf('The class "%s" doesn\'t implement "%s"', $userClassName, UserInterface::class));
```
weaverryan: The class "%s" must implement "%s".

src/Validator.php
```diff
@@ -184,4 +185,26 @@ public static function entityExists(string $className = null, array $entities =
 
         return $className;
     }
+
+    public static function classDoesNotExist($className)
```
weaverryan: returns `: string`

src/Validator.php
```diff
@@ -184,4 +185,26 @@ public static function entityExists(string $className = null, array $entities =
...
+        self::notBlank($className);
+
+        if (class_exists($className)) {
+            throw new RuntimeCommandException(sprintf('Class "%s" already exists', $className));
+        }
+
+        return $className;
+    }
+
+    public static function classIsUserInterface($userClassName)
```
weaverryan: returns `: string`

tests/Maker/FunctionalTest.php
```diff
@@ -387,6 +395,94 @@ function (string $output, string $directory) {
...
+                1,
+                // class name
+                'AppCustomAuthenticator',
+                // controller name
+                'SecurityController',
+            ]
+        )
+            ->addExtraDependencies('doctrine')
+            ->addExtraDependencies('security')
+            ->addExtraDependencies('twig')
```
weaverryan: This extra dependency thing should be rare. `doctrine` make sense to me - that's not a requirement of the Maker, but we're testing that case. But, shouldn't security be a dependency? I suppose `twig` is correct to stay here too, because it's only a conditional dependency - i.e. if the choose the "login form" version. Our test system doesn't support those "optional" dependencies. But, we *should* do a manual check in the maker to make sure they have this if the choose the login-form option. I think I do this in MakeUser - I do a manual check for Doctrine, if they choose the entity option.

tests/Maker/FunctionalTest.php
```diff
@@ -387,6 +395,94 @@ function (string $output, string $directory) {
...
+            )
+                ->addExtraDependencies('security')
+                ->addExtraDependencies('twig')
+                ->addExtraDependencies('doctrine/annotations')
+                ->addExtraDependencies('symfony/form')
+                ->setFixtureFilesPath(__DIR__.'/../fixtures/MakeAuthenticatorLoginFormUserNotEntity')
+                ->assert(
+                    function (string $output) {
+                        $this->assertContains('Success', $output);
+                    }
```
weaverryan: I don't think this is necessary. I think we either already check for this always, or, we are at least checking that the process exited successfully, which is basically the same as this.

tests/Maker/FunctionalTest.php
```diff
@@ -387,6 +395,94 @@ function (string $output, string $directory) {
...
+
+                    $fs = new Filesystem();
+                    $this->assertTrue($fs->exists(sprintf('%s/src/Controller/SecurityController.php', $directory)));
+                    $this->assertTrue($fs->exists(sprintf('%s/templates/security/login.html.twig', $directory)));
+                    $this->assertTrue($fs->exists(sprintf('%s/src/Security/AppCustomAuthenticator.php', $directory)));
+                }
+            );
+
+        if (Kernel::VERSION_ID < 40100) {
+            $makerTestAuthenticatorLoginFormUserEntity->addExtraDependencies('symfony/form');
```
weaverryan: I think we could instead add this into the Maker. Then, the user would also see this dependency error. And, I think the tests would also automatically add it, as it WOULD be a dependency when the user is running 4.0.

src/Maker/MakeAuthenticator.php
```diff
@@ -55,50 +70,94 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
 
-        $interactiveSecurityHelper = new InteractiveSecurityHelper();
-
-        $command->addOption('firewall-name', null, InputOption::VALUE_OPTIONAL, '');
-        $input->setOption('firewall-name', $firewallName = $interactiveSecurityHelper->guessFirewallName($io, $securityData));
-
-        $command->addOption('entry-point', null, InputOption::VALUE_OPTIONAL);
+        if (self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type')
+            && !isset($securityData['security']['providers']) || !$securityData['security']['providers']) {
+            throw new RuntimeCommandException('You need to have at least one provider defined in security.yaml');
```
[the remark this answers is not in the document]
nikophil: This is pretty similar than the next comment you've made : once the user has chosen "form login", we already know that we should raise an exception, but, if i put it down before `InteractiveSecurityHelper::guessUserClass()`, we're asking 4 questions before raising the exception... i think this could be annoying for the user.
What's your feeling about that ? - but i'm ok that in the code it seems unrelated to anything... at the beginning this test was right before `guessUserClass()` call...

ok for other comments

src/Maker/MakeAuthenticator.php
```diff
@@ -55,50 +70,94 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+            function ($answer) {
+                Validator::notBlank($answer);
+
+                return Validator::classDoesNotExist(
+                    $this->generator->createClassNameDetails(
+                        $answer,
+                        'Security\\'
+                    )->getFullName()
+                );
+            }
```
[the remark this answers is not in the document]
nikophil: Maybe i could just not use the question's validator ? but we're losing the behavior when the console asks again the answer if the user has made a typo

src/Maker/MakeAuthenticator.php
nikophil: ok, this was quite annoying ;)
because we're asking the user the FQCN of the user class...

src/Maker/MakeAuthenticator.php
nikophil: ok for generating an empty controller and then adding methods, indeed we got less duplicated code :)

about your concern, why don't we create a `Generator::generateController()` ?

src/Resources/skeleton/authenticator/LoginFormNotEntityAuthenticator.tpl.php
nikophil: i knew you would say something about it ;)

src/Resources/skeleton/authenticator/login_form.tpl.php
nikophil: i took the login template file as you've provided ;)
i think some markup is missing to fill bootstrap needs (e.g.: `form-group`), but if we add these, if will be annoying for users that don't use it.
Then whether we remove all boostrap stuff OR we ask the user if he uses bootstrap, and then we provide all bootstrap classes and markup.
But then, maybe we have to handle bootstrap 3 & boostrap 4...

what do you think ?

src/Maker/MakeAuthenticator.php
nikophil: ok that's fair ;)

src/Resources/skeleton/authenticator/LoginFormEntityAuthenticator.tpl.php
nikophil: i think you're  right.
i' ve added two more authenticator templates (i could have done it with ClassSourceManipulator, but it would have been cumbersome...)
I've also provided 2 more functional test cases

src/Resources/skeleton/authenticator/LoginFormEntityAuthenticator.tpl.php
nikophil: ok, that was actually a bit "brain-fuck" ;)
i addition with the previous comment, i had too much authenticator templates, so i've merged all them into one dynamic template.
I'm also automatically guessing username field when the provider is an "entity" type, then we got the field

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,36 @@
...
+<form method="post">
+    {% if error %}
+        <div class="alert alert-danger">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
+    {% endif %}
+
+    <h1 class="h3 mb-3 font-weight-normal">Please sign in</h1>
+    <label for="inputEmail" class="sr-only">Email address</label>
+    <input type="email" value="{{ last_username }}" name="<?= $username_field; ?>" id="inputEmail" class="form-control" placeholder="Email address" required autofocus>
+    <label for="inputPassword" class="sr-only">Password</label>
+    <input type="password" name="password" id="inputPassword" class="form-control" placeholder="Password" required>
```
weaverryan: Ok, if this is the markup I gave you, then it's based off of this - https://getbootstrap.com/docs/4.1/examples/sign-in/ - which actually includes an extra CSS File to get some of the magic. But, I think it still looks fairly close to good, and without the extra markup. So, let's keep it for now.

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,36 @@
...
+        See https://symfony.com/doc/current/security/remember_me.html
+
+        <div class="checkbox mb-3">
+            <label>
+                <input type="checkbox" name="_remember_me"> Remember me
+            </label>
+        </div>
+    #}
+
+    <button class="btn btn-lg btn-primary btn-block" type="submit">
```
weaverryan: Remove btn-block - that only looks right if you have that full login markup/css file I linked to above

composer.json
```diff
@@ -20,8 +20,10 @@
         "symfony/dependency-injection": "^3.4|^4.0",
         "symfony/filesystem": "^3.4|^4.0",
         "symfony/finder": "^3.4|^4.0",
+        "symfony/form": "^3.4|^4.1",
         "symfony/framework-bundle": "^3.4|^4.0",
-        "symfony/http-kernel": "^3.4|^4.0"
+        "symfony/http-kernel": "^3.4|^4.0",
+        "symfony/security": "^3.4|^4.1"
```
weaverryan: `symfony/security` and `symfony/form` shouldn't be dependencies. What did you need them for?

src/Generator.php
```diff
@@ -178,20 +190,27 @@ public function writeChanges()
...
+            $this->fileManager->dumpFile(
+                $targetPath,
+                $this->generateFileContents($targetPath, $templateData)
+            );
         }
 
         $this->pendingOperations = [];
     }
 
+    private function generateFileContents(string $targetPath, array $templateData)
```
weaverryan: `: string`

src/Maker/MakeAuthenticator.php
```diff
@@ -55,64 +70,119 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+            $authenticatorTypeValues[$authenticatorType]
+        );
 
-        $command->addOption('firewall-name', null, InputOption::VALUE_OPTIONAL, '');
-        $input->setOption('firewall-name', $firewallName = $interactiveSecurityHelper->guessFirewallName($io, $securityData));
+        if (self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type')) {
+            $dependencies = new DependencyBuilder();
+            $dependencies->addClassDependency(
+                TwigBundle::class,
+                'tiwg'
```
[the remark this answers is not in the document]
nikophil: oups, sorry for that

src/Maker/MakeAuthenticator.php
```diff
@@ -55,64 +71,128 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
         }
-
         $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
         $securityData = $manipulator->getData();
 
-        $interactiveSecurityHelper = new InteractiveSecurityHelper();
+        // authenticator type
+        $authenticatorTypeValues = [
+            'Empty authenticator' => self::AUTH_TYPE_EMPTY_AUTHENTICATOR,
+            'Form login' => self::AUTH_TYPE_FORM_LOGIN,
```
weaverryan: > Login form authenticator

src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,90 @@
...
+        <?= $user_needs_encoder ? "return \$this->passwordEncoder->isPasswordValid(\$user, \$credentials['password']);\n" : "return true;\n" ?>
+    }
+
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
+    {
+        if ($targetPath = $this->getTargetPath($request->getSession(), $providerKey)) {
+            return new RedirectResponse($targetPath);
+        }
+
+        throw new \Exception('TODO: provide a valid redirection inside '.__FILE__);
```
weaverryan: We could put some example code here in a comment above:

```php
// return new RedirectResponse($this->router->generate('some_route'));
```

src/Generator.php
```diff
@@ -84,6 +85,18 @@ public function dumpFile(string $targetPath, string $contents)
         ];
     }
 
+    public function getFileContents(string $targetPath): string
```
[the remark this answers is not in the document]
nikophil: ype, i don't know why i've done this ;)

src/Maker/MakeAuthenticator.php
```diff
@@ -55,88 +71,275 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
                 null,
-                $classNameDetails->getFullName()
+                $authenticatorClass
             );
-            $text[] = "Your <info>security.yaml</info> could not be updated automatically. You'll need to add the following config manually:\n\n".$yamlExample;
+            $nextTexts[] = '- Your <info>security.yaml</info> could not be updated automatically. You\'ll need to add the following config manually:\n\n'.$yamlExample;
+        }
+
+        if (self::AUTH_TYPE_FORM_LOGIN === $authenticatorType) {
+            $nextTexts[] = sprintf('- You must provide a valid redirection in the method <info>%s::onAuthenticationSuccess()</info>.', $authenticatorClass);
```
weaverryan: > Finish the redirect "TODO" in the <info>%s::onAuthenticationSuccess()</info> method.

src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,94 @@
...
+        throw new \Exception('TODO: check the credentials inside '.__FILE__);\n" ?>
+    }
+
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
+    {
+        if ($targetPath = $this->getTargetPath($request->getSession(), $providerKey)) {
+            return new RedirectResponse($targetPath);
+        }
+
+        // e.g. : return new RedirectResponse($this->router->generate('some_route'));
```
weaverryan:
```php
// For example: return
```

src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,94 @@
...
+        throw new \Exception('TODO: provide a valid redirection inside '.__FILE__);
```
weaverryan: > ... a valid redirect inside ...

src/Maker/MakeAuthenticator.php
```diff
@@ -55,88 +71,275 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+            $nextTexts[] = '- Review & adapt the login template : <info>/templates/security/login.html.twig</info>.';
```
weaverryan: > - Review & adapt the login template: <info>templates/security/login.html.twig</info>.

src/Maker/MakeAuthenticator.php
```diff
@@ -55,88 +71,275 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+
+            if (!$this->doctrineHelper->isClassAMappedEntity($userClass)) {
+                $nextTexts[] = sprintf('- Review <info>%s::getUser()</info>, if it match your needs.', $authenticatorClass);
```
weaverryan: > - Review <info>%s::getUser()</info> to make sure it matches your needs.

src/Maker/MakeAuthenticator.php
weaverryan: And let's move this down below the next two. That way, if there are multiple things you need to do in the authenticator, we should all of them, and THEN the template text

src/Maker/MakeAuthenticator.php
```diff
@@ -55,88 +71,275 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+            if (!$this->userNeedsEncoder($securityData, $userClass)) {
+                $nextTexts[] = sprintf('- Check user\'s password in <info>%s::checkCredentials()</info>.', $authenticatorClass);
+            }
         }
-        $io->text($text);
+
+        return $nextTexts;
+    }
+
+    private function userNeedsEncoder(array $securityData, string $userClass): bool
```
weaverryan: Could we name this `userClassHasEncoder()`? Is that the correct meaning of this?

src/Maker/MakeAuthenticator.php
```diff
@@ -55,88 +71,275 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+        if (self::AUTH_TYPE_FORM_LOGIN === $authenticatorType) {
+            $nextTexts[] = sprintf('- You must provide a valid redirection in the method <info>%s::onAuthenticationSuccess()</info>.', $authenticatorClass);
+            $nextTexts[] = '- Review & adapt the login template : <info>/templates/security/login.html.twig</info>.';
+
+            if (!$this->doctrineHelper->isClassAMappedEntity($userClass)) {
+                $nextTexts[] = sprintf('- Review <info>%s::getUser()</info>, if it match your needs.', $authenticatorClass);
+            }
+
+            if (!$this->userNeedsEncoder($securityData, $userClass)) {
+                $nextTexts[] = sprintf('- Check user\'s password in <info>%s::checkCredentials()</info>.', $authenticatorClass);
```
weaverryan: > - Check the user\'s password in <info>%s::checkCredentials()</info>.

tests/fixtures/MakeAuthenticatorLoginFormCustomUsernameField/src/Security/UserProvider.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Tells Symfony to use this provider for this User class.
+     */
+    public function supportsClass($class)
+    {
+        return User::class === $class;
+    }
+}
```
weaverryan: Nit pick - but could we "slim down" some of these fixture classes, so that they only have what we need?

src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,94 @@
...
+        <?= $user_is_entity ? "return \$this->entityManager->getRepository($user_class_name::class)->findOneBy(['$username_field' => \$credentials['$username_field']]);\n"
+        : "// Load / create our user however you need.
+        // You can do this by calling the user provider, or with custom logic here.
+        return \$userProvider->loadUserByUsername(\$credentials['$username_field']);\n"; ?>
+    }
+
+    public function checkCredentials($credentials, UserInterface $user)
+    {
+        <?= $user_needs_encoder ? "return \$this->passwordEncoder->isPasswordValid(\$user, \$credentials['password']);\n"
+        : "// Check the user’s password or other credentials and return true or false
```
weaverryan: I think the `'` in `user's` is the wrong type of apostrophe?

src/Maker/MakeAuthenticator.php
```diff
@@ -55,88 +71,275 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+            }
         }
-        $io->text($text);
+
+        return $nextTexts;
```
weaverryan: What about one more on the end?

> When you're ready, see your login form by going to `login`.

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -0,0 +1,36 @@
...
+
+{% block body %}
+<form method="post">
+    {% if error %}
+        <div class="alert alert-danger">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
+    {% endif %}
+
+    <h1 class="h3 mb-3 font-weight-normal">Please sign in</h1>
+    <label for="inputEmail" class="sr-only">Email address</label>
+    <input type="email" value="{{ last_username }}" name="<?= $username_field; ?>" id="inputEmail" class="form-control" placeholder="Email address" required autofocus>
```
weaverryan: Ah.... this should only be `type="email"` if their "username" field contains the term `email` I think. I chose `username` for my User class, and so this field should not have been type="email".

Also, though annoying, we should probably make the `id=""` and label's `for=""` a bit dynamic - based on the actual identity field name.

AND... of course - the label's text - `Email address` - should be dynamic - probably we can turn their propery name into title-case words - e.g. `emailAddress` -> `Email Address` or `email` to just `Email`

src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
nikophil: wow you got god eyes :+1:

src/Resources/skeleton/authenticator/login_form.tpl.php
nikophil: ok you're absolutely right

weaverryan:
Wow! Thank you Nicolas! This is really, really cool - and was a lot of work :). As we discussed, there is one new test failing - we'll address it in another PR before we tag.

PR #271, opened 2018-09-12
Tiny fix: Don't end the link with a 'period'.

bobdenotter:
Not having the link end in a period makes it easier to click or copy/paste. This is also consistent with a few other places in the same bundle.


weaverryan:
Good call! Thanks @bobdenotter!

PR #274, opened 2018-09-19
[WIP] Fix windows tests issues and speed

sadikoff:
Trying to fix all tests issues. Still in progress...

PR #275, opened 2018-09-20
Use git to reset the functional test environments

nikophil:
Hi,

With this PR, the first time a test environment is built, a git repository is created and a first commit is done. Then, the next time this environment is used, the repository is reset to the original form.
This fixes some cache problems : the environments didn't get back to their original form for modified files. And this way, we don't have to copy back the fixtures files to run maker commands inside the cache directories for manual testing sake.

Edited by @lyrixx: Better diff with [`?w=1`](https://github.com/symfony/maker-bundle/pull/275/files?w=1)

### DIFF
--- a/src/Test/MakerTestCase.php
+++ b/src/Test/MakerTestCase.php
@@ -26,41 +26,37 @@ class MakerTestCase extends TestCase
         // prepare environment to test
         $testEnv->prepare();
 
-        try {
-            // run tests
-            $makerTestProcess = $testEnv->runMaker();
-            $files = $testEnv->getGeneratedFilesFromOutputText();
+        // run tests
+        $makerTestProcess = $testEnv->runMaker();
+        $files = $testEnv->getGeneratedFilesFromOutputText();
 
-            foreach ($files as $file) {
-                $this->assertTrue($testEnv->fileExists($file));
+        foreach ($files as $file) {
+            $this->assertTrue($testEnv->fileExists($file));
 
-                if ('.php' === substr($file, -4)) {
-                    $csProcess = $testEnv->runPhpCSFixer($file);
+            if ('.php' === substr($file, -4)) {
+                $csProcess = $testEnv->runPhpCSFixer($file);
 
-                    $this->assertTrue($csProcess->isSuccessful(), sprintf('File "%s" has a php-cs problem: %s', $file, $csProcess->getOutput()));
-                }
-
-                if ('.twig' === substr($file, -5)) {
-                    $csProcess = $testEnv->runTwigCSLint($file);
-
-                    $this->assertTrue($csProcess->isSuccessful(), sprintf('File "%s" has a twig-cs problem: %s', $file, $csProcess->getOutput()));
-                }
+                $this->assertTrue($csProcess->isSuccessful(), sprintf('File "%s" has a php-cs problem: %s', $file, $csProcess->getOutput()));
             }
 
-            // run internal tests
-            $internalTestProcess = $testEnv->runInternalTests();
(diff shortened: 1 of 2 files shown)


sadikoff:
Hello,

This is a very interesting idea, how about comparing speed and HDD operations?

nikophil:
Hi @sadikoff 
by comparison to the last passing test on travis, it does not seem to improve speed performance:
https://travis-ci.org/symfony/maker-bundle/builds/428439796?utm_source=github_status&utm_medium=notification

I don't know how to compare HDD operations ?

weaverryan:
Hey @nikophil!

I really like this! @sadikoff did a great job creating our test system so that we could create these "mini apps" without *terrible* performance (my initial version had terrible performance). This `git` strategy has two advantages I like: (A) it simplifies the code and (B) it leaves the directory in a "finished" state after running the test, which is nice because you can move into it and try to debug things on failures.

The tests are failing, however. I'm not sure the issue. The `vendor_target` tests make me wonder if the composer autoloader isn't being re-run... but needs to be (after we make `composer.json` autoload tweaks). About the other tests - not so sure - could even be a Symfony cache issue.

sadikoff:
@nikophil if you don't mind, I'll use your code in my PR for testing and debug purposes

nikophil:
:+1: 
feel free to use it anyway you want

sadikoff:
Ok I finished some tests. and I have some arguments against this implementation.
- We need more disk space ~600mb for `.git` folders
- On windows first run tests are slower 25 minutes full run (8-10 minutes in previous env) (Core i7 7700hq 8gb ram SSD disk)
- Second run is slower too, 8 minutes (5-7 minutes previously)

Maybe there is a way to optimize it, I'll try to push code and test it on travis and appveyor

nikophil:
hi,
i'm pretty sure that if we provide a `.gitignore` file, which exludes the vendors, the disk space will not be a problem anymore
On my side, the first run of `FunctionalTest.php` (this is the only one impacted by this solution) is 8.7 minutes (8.5 when i'm running them from master) (core i7 - 16gb ram - ssd)
Second run is 5.1 minutes (5.0 on master)

nikophil:
i can"t find where you removed the `.gitignore` ?

sadikoff:
@nikophil Check [this](https://github.com/symfony/maker-bundle/pull/274/commits/2951b8c1e2988a8e15de5eaa912747f39d30779e?utf8=%E2%9C%93&diff=unified)

nikophil:
yep, but i don't understand why you need it : if we exclude `vendor` directory from git, then this directory will not change from test to test, and only `composer install` and some `composer require` are needed the first time we build the environment, no ?

sadikoff:
I don't understand it too ) I'm trying to investigate this moment now. I just know that with `.gitignore` we have a lot of failing tests, without everything is good and this is very strange for me :)

nikophil:
isn't it because the `.env` is in `.gitignore` ?

nikophil:
We've forgotten to mention that beside that this PR helps for DX, it would also fix a bug :
Presently, if one command in some functional test modifies a file that is coming from the fixtures files, this one does is not reverted to it's fresh state.
Then, if another test uses the same environment, then i might fail.
For example: https://travis-ci.org/symfony/maker-bundle/jobs/431566618
this `user_security_model_no_password` ran alone is all green. But it shares the it's environment with `auth_empty_multiple_firewalls`, which introduces a modification in `security.yaml`

+1 for making this thing work :)

nikophil:
i'm closing this Pr, as the fix will be integrated with @sadikoff 's PR #274

PR #276, opened 2018-09-21
Add a wizard emoji

tgalopin:
🧙

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -378,7 +378,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
     private function printAvailableTypes(ConsoleStyle $io)
     {
         $allTypes = Type::getTypesMap();
-        
+
         if ('Hyper' === getenv('TERM_PROGRAM')) {
             $wizard = 'wizard 🧙';
         } else {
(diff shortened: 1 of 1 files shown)


lyrixx:
How does it behave on system that does not support it? (we may don't care :) )

javiereguiluz:
Does anyone see a wizard here?

error-1

error-2

tgalopin:
Updated :)

nicolas-grekas:
@javiereguiluz 
image

javiereguiluz:
@tgalopin I'm on a macOS computer and still can't see this emoji:

error-3

javiereguiluz:
@tgalopin yes, I can see all emojis perfectly ... but not the one you added here. I only see an empty square.

ogizanagi:
@javiereguiluz : You might still be on OS X Sierra, but this emoji is part of Unicode 10 for which the Apple Color Emoji font was only updated since High Sierra.

javiereguiluz:
@ogizanagi you are right ... and that was my point: we may be trying to use a too new emoji ... which is not seen by some (lots?) of users. Please, think twice before merging it 🙏  Thanks!

nicolas-grekas:
As time passes, this compat issue will disappear. Given the outcome is a simple square, I'm all for merging this now :)

weaverryan:
We did it! Thanks Tito ;)

Issue #166, opened 2018-04-30
Invalid generation of self-referencing association


codedmonkey:
When trying to generate a self-referencing association on an entity, only the `OneToMany` field gets generated, while I expected both `ManyToOne` and `OneToMany` associations.

Here's my input

```
php bin/console make:entity Test

 created: src/Entity/Test.php
 created: src/Repository/TestRepository.php

 Entity generated! Now let's add some fields!
 You can always add more fields later manually or by re-running this command.

 New property name (press <return> to stop adding fields):
 > hallo

 Field type (enter ? to see all types) [string]:
 >

 Field length [255]:
 >

 Can this field be null in the database (nullable) (yes/no) [no]:
 >

 updated: src/Entity/Test.php

 Add another property? Enter the property name (or press <return> to stop adding fields):
 > headTest

 Field type (enter ? to see all types) [string]:
 > ManyToOne

 What class should this entity be related to?:
 > Test

 Is the Test.headTest property allowed to be null (nullable)? (yes/no) [yes]:
 >

 Do you want to add a new property to Test so that you can access/update Test objects from it - e.g. $test->getTests()? (yes/no) [yes]:
 > yes

 A new property will also be added to the Test class so that you can access the related Test objects from it.

 New field name inside Test [tests]:
 > testings

 updated: src/Entity/Test.php
 updated: src/Entity/Test.php
```

Here's the generated output:

```php
<?php

namespace App\Entity;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity(repositoryClass="App\Repository\TestRepository")
 */
class Test
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue()
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\Column(type="string", length=255)
     */
    private $hallo;

    /**
     * @ORM\OneToMany(targetEntity="App\Entity\Test", mappedBy="headTest")
     */
    private $testings;

    public function __construct()
    {
        $this->testings = new ArrayCollection();
    }

    public function getId()
    {
        return $this->id;
    }

    public function getHallo(): ?string
    {
        return $this->hallo;
    }

    public function setHallo(string $hallo): self
    {
        $this->hallo = $hallo;

        return $this;
    }

    /**
     * @return Collection|Test[]
     */
    public function getTestings(): Collection
    {
[...]

PR #278, opened 2018-09-25
Fix self-referencing entity associations

codedmonkey:
Fixes faulty generation of entities with self-referencing associations https://github.com/symfony/maker-bundle/issues/166. It looks like changes for the owning entity were overwritten by the changes for the inversed entity.

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -201,12 +201,12 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
             } elseif ($newField instanceof EntityRelation) {
                 // both overridden below for OneToMany
                 $newFieldName = $newField->getOwningProperty();
-                if ($newField->getInverseClass() !== $newField->getOwningClass()) {
-                    $otherManipulatorFilename = $this->getPathOfClass($newField->getInverseClass());
-                    $otherManipulator = $this->createClassManipulator($otherManipulatorFilename, $io, $overwrite);
-                } else {
+                if ($newField->isSelfReferencing()) {
                     $otherManipulatorFilename = $entityPath;
                     $otherManipulator = $manipulator;
+                } else {
+                    $otherManipulatorFilename = $this->getPathOfClass($newField->getInverseClass());
+                    $otherManipulator = $this->createClassManipulator($otherManipulatorFilename, $io, $overwrite);
                 }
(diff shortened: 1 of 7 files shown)


src/Maker/MakeEntity.php
```diff
@@ -201,8 +201,13 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             } elseif ($newField instanceof EntityRelation) {
                 // both overridden below for OneToMany
                 $newFieldName = $newField->getOwningProperty();
-                $otherManipulatorFilename = $this->getPathOfClass($newField->getInverseClass());
-                $otherManipulator = $this->createClassManipulator($otherManipulatorFilename, $io, $overwrite);
+                if ($newField->getInverseClass() !== $newField->getOwningClass()) {
```
weaverryan: Could we use some `$newField->isSelfReferencingMethod()` here?

src/Util/ClassSourceManipulator.php
```diff
@@ -447,7 +447,7 @@ private function addSingularRelation(BaseRelation $relation)
 
     private function addCollectionRelation(BaseCollectionRelation $relation)
     {
-        $typeHint = $this->addUseStatementIfNecessary($relation->getTargetClassName());
+        $typeHint = !$relation->isSelfReferencing() ? $this->addUseStatementIfNecessary($relation->getTargetClassName()) : 'self';
```
weaverryan: Hmm, maybe reverse this - I think the `!` is a bit confusing:

```php
$typeHint = $relation->isSelfReferencing() ? 'self' : $this->addUseStatementIfNecessary($relation->getTargetClassName());
```

tests/Maker/FunctionalTest.php
```diff
@@ -807,6 +807,36 @@ public function getCommandEntityTests()
...
+                // the target entity
+                'User',
+                // relation type
+                'ManyToOne',
+                // nullable
+                'y',
+                // do you want to generate an inverse relation? (default to yes)
+                '',
+                // field name on opposite side
+                'wards',
```
weaverryan: This word isn't commonly used in English. Using `dependents` might be better (A parent/guardian refers to their children as "dependents")

weaverryan:
Thank you @codedmonkey!

PR #280, opened 2018-10-01
Add --no-template option for make:controller command

welcoMattic:
I had several times the case in which I did not necessarily need a template when I generated a controller. So I'm trying to add an option `--no-template` to force avoid template generation.

I think that I will need help on tests writing ;)

### DIFF
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -39,7 +39,7 @@ final class MakeController extends AbstractMaker
         $command
             ->setDescription('Creates a new controller class')
             ->addArgument('controller-class', InputArgument::OPTIONAL, sprintf('Choose a name for your controller class (e.g. <fg=yellow>%sController</>)', Str::asClassName(Str::getRandomTerm())))
-            ->addOption('no-template', 'n', InputOption::VALUE_NONE, 'Use this option to disable template generation')
+            ->addOption('no-template', null, InputOption::VALUE_NONE, 'Use this option to disable template generation')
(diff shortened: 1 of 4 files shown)


src/Maker/MakeController.php
```diff
@@ -38,6 +39,7 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
         $command
             ->setDescription('Creates a new controller class')
             ->addArgument('controller-class', InputArgument::OPTIONAL, sprintf('Choose a name for your controller class (e.g. <fg=yellow>%sController</>)', Str::asClassName(Str::getRandomTerm())))
+            ->addOption('--no-template', '-n', InputOption::VALUE_OPTIONAL)
```
codedmonkey: I think you mean `InputOption::VALUE_NONE`

weaverryan:
Hey @welcoMattic!

In general, I want to be careful about adding customizations like this. But, this one seems reasonable.

It looks like some of the existing controller tests broke. Can you check those out? I'm not sure what's going on - I'd try to run those test cases locally if you can.

Cheers!

welcoMattic:
Yes, existing tests broke. It seems to be because of `addOption` method, but it looks weird that this method broke tests 🤔 

If you can run those tests locally to help, that would be nice 😉 

Thanks

weaverryan:
Hey! Would you mind running the tests locally? I’m happy to review or debug (especially if the test suite doesn’t has any issues running on your machine), but I’d prefer you to finish this feature if you’d like it :)

weaverryan:
Sorry for the delay, but this was an excellent PR - very clean. Thank you @welcoMattic!

PR #283, opened 2018-10-03
[Voter] use ENTITY_ATTR instead of ATTR

lyrixx:
This is a bit longer, but this is much more explicite.
This is much easier to find all usage in a code base because the term is now unique


weaverryan:
Great idea! I like this much better

PR #286, opened 2018-10-10
Fix crud controller template

thlbaut:
There is a mistake in crud controller template :

Example  (`src/Resources/skeleton/crud/controller/Controller.tpl.php` - line 35) : 
`return $this->render('<?= $route_name ?>/index.html.twig', ['<?= $entity_twig_var_plural ?>' => $<?= $entity_var_plural ?>]);`

With an entity with following path :` Entity/Foo/BarBar.php` , `$route_name` will be equal to `foo_bar_bar` and does not correspond with the `index.html.twig` path (should be `foo/bar_bar`).

This pull request fix this by adding the `$templatesPath` var, used to create twig templates, for the controller generation.

:)


weaverryan:
Very well-explained and an excellent patch! Thank you @thlbaut!

PR #289, opened 2018-10-12
Class not found on make:entity --regenerate

### LINKED ISSUE #207: make:entity --regenerate on existing entity class
The command "make:entity --regenerate App" correctly generates the entities (classes) the first time, when there are not yet the classes of the entities.
```
App\Entity\Myentity:
    type: entity
    table: Myentity
    id:
        id:
            type: integer
            generator:
                strategy: AUTO`
    fields:
        existingproperty:
            type: integer
```

If I add a property "newproperty" to the file entityname.orm.yml,

```
App\Entity\Myentity:
    type: entity
    table: Myentity
    id:
        id:
            type: integer
            generator:
                strategy: AUTO`
    fields:
        existingproperty:
            type: integer
        newproperty:
            type: boolean
            nullable: true
```
 when I execute the command I get this error:
> Property App\Entity\Myentity::$newproperty does not exist

The workaround is to add property to the class:
```
<?php

namespace App\Entity;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;

class Myentity
{

    private $newproperty;
```
In this case, if i run "make:entity --regenerate App", the command correctly generate getter and setter.
It Is a bug or it is correct that it works like that?

Thanks for your work

mertoksuz:
That PR fixes #207.

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -133,7 +133,10 @@ final class DoctrineHelper
             if ($disconnected) {
                 try {
                     $loaded = $cmf->getAllMetadata();
-                } catch (PersistenceMappingException | ORMMappingException $e) {
+                } catch (\Throwable $e) {
+                    if (!$e instanceof ORMMappingException && !$e instanceof PersistenceMappingException) {
+                        throw $e;
+                    }
                     $loaded = $cmf instanceof AbstractClassMetadataFactory ? $cmf->getLoadedMetadata() : [];
                 }
 
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -133,10 +133,9 @@ final class DoctrineHelper
             if ($disconnected) {
                 try {
                     $loaded = $cmf->getAllMetadata();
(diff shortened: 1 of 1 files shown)


weaverryan:
Hey @mertoksuz!

Thanks for making this PR to fix this! But, the tests are failing because the old exception class *is* needed in the catch. Can why your change fixes things? It's quite possible that we may need to catch *both* `MappingException`.

Cheers!

mertoksuz:
@weaverryan 

Hey, i updated my codes. Is that recover both exceptions ?

weaverryan:
@mertoksuz thanks! You did well! But, we can't use the `|` syntax in catch - it breaks in 7.0. Can you manually add 2 separate catches - one for each exception?

And also, do you have any idea why the `Common` exception is the one that fixes your problem? I'll tag a release shortly - as you are working quickly, we can hopefully include this :)

mertoksuz:
@weaverryan 

i updated again :). 

I think, when doctrine tries to load metadata somehow it checks php classes but if you have only yml/xml files for entities than it throws Class not found ( MappingException ).

weaverryan:
Thanks for the quick work @mertoksuz!

PR #294, opened 2018-10-14
Adding Path option to make:Controller Command 

m0ns1f:
f.g : php bin/console make:Controller --path=Foo/bon
and adding template option fg : php bin/console make:Controller --path=Foo/bon --template

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\Bundle\MakerBundle;
 
-use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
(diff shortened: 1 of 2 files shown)


ckrack:
In part this is a duplicate of #185 and #265
Changing the templates is something that won't be merged, mainly due to having to support them in the future. For example, the maker could not change the variables in the templates without breaking things.

weaverryan:
I DID just merge a `--no-template` option, to skip the template generation. But @ckrack is right that we won't merge an option to change the controller path. Actually, this is already possible - you can pass whatever fully-qualified class name you want as the controller argument, and the bundle will use the autoloading rules to place this in the correct location.

Cheers!

PR #295, opened 2018-10-16
[Voter] Use the subject's FQCN in the support method

lyrixx:
People could replace only the `BlogPost` class. So the support method will always return `false`.

With a leading `\`, the class name will resolve "normally"


weaverryan:
Good catch - thank you Grégoire!

PR #298, opened 2018-10-19
[Serializer] Add a 'make:serializer:normalizer' command

### LINKED ISSUE #287: Added a command to create a Normalizer
> I think making custom normalizer is much more likely to happen than custom encoders though.

_Originally posted by @stof in https://github.com/symfony/maker-bundle/pull/30#issuecomment-347550483_

---

I totally agree. Of course it would be stupid to remove the `make:serializer:encoder` command. 

But a `make:serializer:normalizer` would be super useful

lyrixx:
fixes #287

### DIFF
--- a/src/Maker/MakeSerializerNormalizer.php
+++ b/src/Maker/MakeSerializerNormalizer.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Serializer\Serializer;
+
+/**
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ */
+final class MakeSerializerNormalizer extends AbstractMaker
+{
+    public static function getCommandName(): string
+    {
+        return 'make:serializer:normalizer';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConf)
+    {
+        $command
+            ->setDescription('Creates a new serializer normalizer class')
+            ->addArgument('name', InputArgument::OPTIONAL, 'Choose a class name for your normalizer (e.g. <fg=yellow>UserNormalizer</>)')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeSerializerNormalizer.txt'))
+        ;
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $normalizerClassNameDetails = $generator->createClassNameDetails(
+            $input->getArgument('name'),
+            'Serializer\\Normalizer\\',
+            'Normalizer'
+        );
+
+        $generator->generateClass(
+            $normalizerClassNameDetails->getFullName(),
+            'serializer/Normalizer.tpl.php'
+        );
+
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text([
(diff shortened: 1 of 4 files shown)


src/Resources/skeleton/serializer/Normalizer.tpl.php
```diff
@@ -0,0 +1,30 @@
...
+        $data = $this->normalizer->normalize($object, $format, $context);
+
+        // Here: add, edit, or delete some data
+
+        return $data;
+    }
+
+    public function supportsNormalization($data, $format = null): bool
+    {
+        return $data instanceof \App\Entity\BlogPost;
```
franek: I think you should update this line with $namespace/$class_name ?

lyrixx: `$namespace\$class_name` === `App\Serializer\Normalizer\TopicNormalizer` (for exemple)

So no. Did I miss something?

src/Maker/MakeSerializerNormalizer.php
```diff
@@ -0,0 +1,73 @@
...
+        ;
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $normalizerClassNameDetails = $generator->createClassNameDetails(
+            $input->getArgument('name'),
+            'Serializer\\Normalizer\\',
+            'Normalizer'
+        );
```
weaverryan: For the serializer encoder maker, we just put it into the `Serializer` directory (not a sub-directory). I'm interested in how many encoders / normalizers applications have, and whether we should put them in sub-directories, or just in Serializer (to keep things a bit smaller).

src/Resources/help/MakeSerializerNormalizer.txt
```diff
@@ -0,0 +1,5 @@
+The <info>%command.name%</info> command generates a new serializer normalizer class.
+
+<info>php %command.full_name% UserEncoder</info>
```
weaverryan: UserNormalizer?

src/Maker/MakeSerializerNormalizer.php
lyrixx: I know and I don't like it.
I have the same opinion about form type.

It really prefer to use a sub namespace to keep things clean. 

I usually have between 0 to 10 normalizers and sometimes 1 encoder. 

Btw voters are put in a sub namespace too. 

It's just an extra directory...

lyrixx:
Hello, 

> But if you're decoding, you'll need the denormalizer as well.

This is not true. de-normalization is build'in. In looked at my application, and I never had to create a de-normalizer. However, I have many normalizers.

So, I'm not sure it's a goo idea to add this option

src/Resources/help/MakeSerializerNormalizer.txt
lyrixx: good catch. Fixed. Thanks

src/Resources/skeleton/serializer/Normalizer.tpl.php
franek: No. I read your PR too quickly ;) ok for me ;)

lyrixx:
Hello, Could we merge this PR? It's really helpful during training an real life.
Regarding the deserialize option, since my last comment, I did not have to create a denormalizer ... So I'm still against this option

weaverryan:
Thank you Grégoire! If you want, you could update the docs to mention the new command :)

dunglas:
It could be nice to make the generated normalizer implementing the `CacheableSupportsMethodInterface`. Almost all normalizers will benefit from it in term of performance.

src/Resources/skeleton/serializer/Normalizer.tpl.php
OskarStark: @lyrixx why are you checking against `App\Entity\Blogpost`? 🤔

src/Resources/skeleton/serializer/Normalizer.tpl.php
```diff
@@ -0,0 +1,30 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
+
+class <?= $class_name ?> implements NormalizerInterface
```
OskarStark: Shouldn't this be `final`?

src/Resources/skeleton/serializer/Normalizer.tpl.php
lyrixx: Why not ? it's just an exemple, as in Voter

src/Resources/skeleton/serializer/Normalizer.tpl.php
lyrixx: It could indeed.Voter are not. I was consistant

PR #300, opened 2018-10-23
[CRUD] Redirect to index after edit

### LINKED ISSUE #257: maker:crud / Controller.tpl.php / redirect edit route after update
Hello,
Thanks you very much, for your "magic" bundle, very useful...
It seem redirect edit route after update is not correct, on this line : 

https://github.com/symfony/maker-bundle/blob/22a9bf33868f0fed6cab013fe5231e8485d15e20/src/Resources/skeleton/crud/controller/Controller.tpl.php#L81

On success update, redirected route should be **"_view"**  or **"_index"** page like this :

        return $this->redirectToRoute('<?= $route_name ?>_view', ['<?= $entity_identifier ?>' => $<?= $entity_var_singular ?>->get<?= ucfirst($entity_identifier) ?>()]); 

or

        return $this->redirectToRoute('<?= $route_name ?>_index');

ckrack:
Update the generated edit action.
This now redirects to the index instead of the edit form.
Fixes #257

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -78,7 +78,7 @@ class <?= $class_name ?> extends <?= $parent_class_name; ?><?= "\n" ?>
         if ($form->isSubmitted() && $form->isValid()) {
             $this->getDoctrine()->getManager()->flush();
 
-            return $this->redirectToRoute('<?= $route_name ?>_edit', ['<?= $entity_identifier ?>' => $<?= $entity_var_singular ?>->get<?= ucfirst($entity_identifier) ?>()]);
+            return $this->redirectToRoute('<?= $route_name ?>_index', ['<?= $entity_identifier ?>' => $<?= $entity_var_singular ?>->get<?= ucfirst($entity_identifier) ?>()]);
         }
 
         return $this->render('<?= $templates_path ?>/edit.html.twig', [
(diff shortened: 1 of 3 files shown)


weaverryan:
Quick question: what was wrong with redirecting back to edit? Was it just not what people are normally expecting?

sadikoff:
I disagree 👎 .

Redirect to `edit` action was here from the beginning. We can see it in symfony demo app and in SensioGeneratorBundle, and even in very old symfony 1.4 admin generator. 

After all it's not admin generator it's just crud skeleton and every user should configure it as he wants

weaverryan:
I can see it both ways. After creating a new "item", we redirect back to the index page. This PR would make that consistent. I lightly lean towards YES let's merge this.

ckrack:
> Redirect to `edit` action was here from the beginning. We can see it in symfony demo app and in SensioGeneratorBundle, and even in very old symfony 1.4 admin generator.

I don't think that it's the status quo is an argument for keeping it.

> After all it's not admin generator it's just crud skeleton and every user should configure it as he wants

The crud is often used for quick examples, especially when learning things from the docs.
The current behaviour with redirecting to the edit page after successfully editing, without any success message is confusing and people (including me) think that something is broken.

I'm against adding flash messages to the generated controller, as it's out of scope.
But redirecting back to the index could be seen as the least viable replacement for a success message.

weaverryan:
Thank you @ckrack! I agree with what you said :)

PR #301, opened 2018-10-23
[FORM] Modify make:form to allow non-entities

ckrack:
Before this modification the make:form command allowed passing a fully specified classname.
It would generate a form with the correct bound class,  but not read the fields.
It generated a form with one "field_name" field instead of something more useful.
This also affects DTOs.
This modification makes our lives easier by reading all fields of the bound class, except the id field.


src/Util/ClassDetails.php
```diff
@@ -0,0 +1,45 @@
...
+final class ClassDetails
+{
+    private $fullClassName;
+
+    public function __construct(string $fullClassName)
+    {
+        $this->fullClassName = $fullClassName;
+    }
+
+    public function getProperties():? array
```
weaverryan: Make this private and move to the bottom. At least right now, we don't need this to be public.

src/Util/ClassDetails.php
```diff
@@ -0,0 +1,45 @@
...
+        $propertiesList = [];
+
+        foreach ($props as $prop) {
+            $propertiesList[] = $prop->getName();
+        }
+
+        return $propertiesList;
+    }
+
+    public function getFormFields()
```
weaverryan: `: array` return type

src/Util/ClassDetails.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+use Symfony\Bundle\MakerBundle\Str;
+
+final class ClassDetails
```
weaverryan: Mark as `@internal` as well. We're not intending for internal classes to be used by others, until someone specifically requests that.

src/Util/ClassDetails.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+use Symfony\Bundle\MakerBundle\Str;
```
weaverryan: Not needed?

weaverryan:
Thank you @ckrack! This is an obvious, but wonderful improvement. Great work!

PR #303, opened 2018-10-24
[DTO] Add new make:dto command

### LINKED ISSUE #162: [RFC] Add a form DTO maker for entities
With the debate around the (valid/invalid) state of entities increasing since the introduction of strict typing and improved type hinting in PHP 7, I'm interested to hear if we should recommend using DTOs for forms by creating a maker for it. Adding such a command would probably mean other makers need to be semi-aware of this as well, so I'm mainly putting this question forward in an effort to see if there's demand for such a feature.

Using DTOs in forms isn't mentioned in the docs anywhere, so that might need to change first (see https://github.com/symfony/symfony-docs/issues/8893).

ckrack:
This command is added as a supplement to make:form.
To prevent problems with invalid state of entities when used in forms,
data transfer objects are a solution.
This command generates these from existing entities.
It can generate getters/setters and helper methods to fill and extract data from/to entities.

Closes #162


src/Maker/MakeDto.php
```diff
@@ -0,0 +1,250 @@
...
+            // Lookup classname for inherited properties
+            if (array_key_exists('declared', $mapping)) {
+                $fullClassName = $mapping['declared'];
+            } else {
+                $fullClassName = $boundClassDetails->getFullName();
+            }
+
+            // Property Annotations
+            $reflectionProperty = new \ReflectionProperty($fullClassName, $fieldName);
+            $propertyAnnotations = $annotationReader->getPropertyAnnotations($reflectionProperty);
```
weaverryan: So, there are a few issues with the copying of the validation constraints:

1) Validation metadata may be stored as XML or YML, and we're not covering that - or even warning the user that we don't handle that situation.
2) I believe you can use the `validator.mapping.class_metadata_factory` service (which gives you this method - https://github.com/symfony/symfony/blob/1fc577f2c7c1cff2857658537adf4f30a3ad60ac/src/Symfony/Component/Validator/Validator/RecursiveValidator.php#L81 ) to read the metadata. This will include XML and YAML metadata, but I don't know if we will know the source of the metadata (though, we could do a simple check on the class for one of the "found" constraints to see if its there - to know if annotations are used).

We probably should not try to add XML/YAML validation metadata, if that's what the user is currently using. However, we COULD do annotations (like you are now). But, if we do that, we also need to show a big warning to the user if we detect validation metadata that is XML or YAML - i.e. "The validation constraints were not copied to the new class" (not perfectly explained, but hopefully you get the idea).

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,250 @@
...
+            if (!empty($field['id'])) {
+                return false;
+            }
+
+            return true;
+        });
+
+        $DTOClassPath = $generator->generateClass(
+            $dataClassNameDetails->getFullName(),
+            __DIR__.'/../Resources/skeleton/dto/Data.tpl.php',
```
weaverryan: I'd call the template `DTO.tpl.php` - just so it's easier to recognize in the library.

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,250 @@
...
+        if (true === $assertionsImported) {
+            $io->note([
+                'The maker imported assertion annotations.',
+                'Consider removing them from the entity or make sure to keep them updated in both places.',
+            ]);
+        }
+
+        $io->text([
+            'Next: Create your form with this DTO and start using it:',
+            '$ php bin/console make:form '.$boundClassDetails->getShortName(),
```
weaverryan: This is how we've written this type of message in the past: https://github.com/symfony/maker-bundle/blob/4d2743d820004f488a3ac0861ac611ca5da5b293/src/Maker/MakeMigration.php#L99

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,250 @@
...
+            '<fg=green>Enter fully qualified data class name to bind to the form:</>',
+            '> \\'.$dataClassNameDetails->getFullName(),
+            '',
+            'Find the documentation at <fg=yellow>https://symfony.com/doc/current/forms.html</>',
```
weaverryan: Ideally, we could update this to point directly to the part of the docs that talks about DTO's

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,250 @@
...
+                'The maker imported assertion annotations.',
+                'Consider removing them from the entity or make sure to keep them updated in both places.',
+            ]);
+        }
+
+        $io->text([
+            'Next: Create your form with this DTO and start using it:',
+            '$ php bin/console make:form '.$boundClassDetails->getShortName(),
+            '<fg=green>Enter fully qualified data class name to bind to the form:</>',
+            '> \\'.$dataClassNameDetails->getFullName(),
```
weaverryan: Hmm, could we instead say something like - fake data.

```
Next: Create a form for this DTO by running:
<info>php bin/console make:form '.$boundClassDetails->getShortName()</info>
and enter <info>$dataClassNameDetails->getFullName()</info>
```

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,250 @@
...
+            Validation::class,
+            'validator',
+            // add as an optional dependency: the user *probably* wants validation
+            false
+        );
+    }
+
+    private function createClassManipulator(string $classPath, bool $omitGettersSetters = false): DTOClassSourceManipulator
+    {
+        return new DTOClassSourceManipulator(
```
weaverryan: These DTO classes aren't meant to be updateable by this command, right? I mean, you run `make:dto` to create the class. But, you would not run it later to *update* the class after adding more fields to your entity, right? If so, I'm not sure we need to use the Parser magic to add the properties, getters & setters: we should just leverage passing variables into the template. The `ClassSourceManipulator` is useful when you need to update classes (because we can't do that with a simple template). But the complexity cost is high - I'm sure you know, because implementing this `DTOClassSourceManipulator` was probably quite challenging :).

src/Resources/skeleton/dto/Data.tpl.php
```diff
@@ -0,0 +1,70 @@
...
+{
+<?php if ($addHelpers): ?>
+    /**
+     * Create DTO, optionally extracting data from a model.
+     */
+    public function __construct(?<?= $bounded_class_name ?> $<?= lcfirst($bounded_class_name) ?> = null)
+    {
+        if ($<?= lcfirst($bounded_class_name) ?> instanceof <?= $bounded_class_name ?>) {
+            $this->extract($<?= lcfirst($bounded_class_name) ?>);
+        }
```
weaverryan: Is there a case for passing something *other* than the bound entity class? Should this be:

```
if (null !== $bound_class_name) {
```

src/Resources/skeleton/dto/Data.tpl.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Fill entity with data from the DTO.
+     */
+    public function fill(<?= $bounded_class_name ?> $<?= lcfirst($bounded_class_name) ?>): <?= $bounded_class_name ?><?= PHP_EOL ?>
+    {
+<?php if (false === $omitGettersSetters): ?>
```
weaverryan: Could we reverse this - `if ($omitGettersSetters)`? Having the "not" + "omit" is a double-negative - it's a bit confusing :).

src/Resources/skeleton/dto/Data.tpl.php
```diff
@@ -0,0 +1,70 @@
...
+        $<?= lcfirst($bounded_class_name) ?><?= PHP_EOL ?>
+<?php foreach ($fields as $propertyName => $mapping): ?>
+            ->set<?= Str::asCamelCase($propertyName) ?>($this->get<?= Str::asCamelCase($propertyName) ?>())
```
weaverryan: We need to check during generation that a `setter` method exists for all of these properties. If there is no setter, we will need to either not include those properties, or add some TODO in here so the user can update them manually, and print a warning.

src/Resources/skeleton/dto/Data.tpl.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Extract data from entity into the DTO.
+     */
+    public function extract(<?= $bounded_class_name ?> $<?= lcfirst($bounded_class_name) ?>): self
+    {
+<?php if (false === $omitGettersSetters): ?>
+<?php foreach ($fields as $propertyName => $mapping): ?>
+        $this->set<?= Str::asCamelCase($propertyName) ?>($<?= lcfirst($bounded_class_name) ?>->get<?= Str::asCamelCase($propertyName) ?>());
```
weaverryan: Same down here - we need to check to make sure the getter exists. And if it doesn't, take some action of give a warning. We don't have to generate perfect code - we can't anticipate all situations, and we don't need to. But I DO like to give a warning if the user has a setup that is not supported by us.

tests/Maker/FunctionalTest.php
```diff
@@ -43,7 +44,7 @@ class FunctionalTest extends MakerTestCase
     /**
      * @var Kernel
      */
-    private $kernel;
+	private $kernel;
```
weaverryan: revert?

tests/Maker/FunctionalTest.php
```diff
@@ -66,6 +67,19 @@ public function testEntityCommands(MakerTestDetails $makerTestDetails)
...
+	 * @group functional_group3
+     * @dataProvider getCommandDtoTests
+     */
+    public function testDtoCommands(MakerTestDetails $makerTestDetails)
+    {
+        // entity tests are split into a different method so we can batch on appveyor
+        // this solves a weird issue where phpunit would die while running the tests
+
+        $this->executeMakerCommand($makerTestDetails);
+    }
```
weaverryan: No need to do this - just put your test cases into `getCommandTests()`.

weaverryan (changes requested):
Hey @ckrack!

Awesome start - a lot of complex stuff here. But, doing cool things like this is part of what makes this bundle fun. I've added a first round of changes.

Cheers!

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,322 @@
...
+         * @var ClassMetaData
+         */
+        $metaData = $this->doctrineHelper->getMetadata($boundClassDetails->getFullName());
+
+        // Get list of fields
+        $fields = $metaData->fieldMappings;
+
+        // The result is passed to the template
+        $addHelpers = $io->confirm('Add helper extract/fill methods?');
+        $omitGettersSetters = $io->confirm('Omit generation of getters/setters?');
```
Pierstoval: I'd suggest to change this feature to `Generate getters/setters? [y/N]` if possible

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,322 @@
...
+
+        // Filter identifiers from generated fields
+        $fields = array_filter($fields, function ($field) use ($metaData) {
+            return !$metaData->isIdentifier($field['fieldName']);
+        });
+
+        // See, whether there are missing methods
```
Pierstoval:
```suggestion
        // Check whether there are missing methods
```

ckrack:
I had a chat with @Pierstoval about his proposal for [DTOs in EasyAdmin](https://github.com/EasyCorp/EasyAdminBundle/pull/2403/commits) today and he raised the following issues:

> doing `$dto->fill($entity)` isn't nice IMO because it can only use the public API of the entity itself
> and this requires getters/setters
> using DTOs is here to help us get rid of setters at least

To achieve this, he uses mutators on the entity (`$entity->updateFromFoobarDto(FoobarDto $dto)`) in which he can avoid using setters.

I agree to this and we should try to enforce best practices as much as possible - while making generated DTOs usable without modifications.

An approach could be:
When generating helper methods (currently `$dto->fill()` and `$dto->extract()`), use the constructor instead of `extract()` and add a mutator to the entity itself instead of generating a `fill()` method on the DTO.
This would mean adding code to the entity, while generating the DTO.

Simple example, given an Entity `Foobar` like this:
```php
// src/Entity/Foobar.php
class Foobar {
    private $id;
    private $acme;

    // ... getters
}
```

Running the maker:
```bash
php bin/console make:dto Foobar Foobar
Add helper extract/fill methods? (yes/no) [yes]:
 > yes

 Generate getters/setters? (yes/no) [no]:
 > no
```

Would generate the following Dto:
```php
class FoobarData {
    public $acme;

    public function __construct(?Foobar $foobar = null)
    {
        if (null !== $foobar) {
            $this->acme = $foobar->getAcme();
        }
    }
}
```

And modify the Entity with this
```php
// src/Entity/Foobar.php

class Foobar {

    // ...

    public function updateFromFoobar(Foobar $foobar)
    {
        $this->acme = $foobar->acme;
    }
}
```

WDYT?

Pierstoval:
I'm totally ok with this proposal 👍 

Just as a side-note, this example:

```php
class FoobarData {
    public $acme;

    public function __construct(?Foobar $foobar = null)
    {
        if (null !== $foobar) {
            $this->acme = $foobar->getAcme();
        }
    }
}
```

Should be replaced IMO with this:

```php
class FoobarData {
    public $acme;

    public function __construct(Foobar $foobar = null)
    {
        if ($foobar) {
            $this->acme = $foobar->getAcme();
        }
    }
}
```

I'd even go further and only allow static constructors such as this:

```php
class FoobarData {
    public $acme;

    public static function createEmpty(): self
    {
        return new self();
    }

    public static function fromFoobar(Foobar $foobar): self
    {
        $dto = new self();

        $dto->acme = $foobar->getAcme();

        return $dto;
    }
}
```

mablae:
> I'd even go further and only allow static constructors such as this

Yeah named static constructors like you suggest and an scalar conctructor like this: 
    public function __construct($acme)
    {

            $this->acme = $acme;

    }

Maybe Entity Class metadata could be used to derive the correct typehints, too

Pierstoval:
Technically, a constructor can be called twice with `$object->__construct();`, while a static constructor will not affect a currently existing object, leading the update part only to proper mutators (or the Reflection API)

karser:
Just my 2 cents: I usually add the most critical properties to the DTO's constructor. It reduces the number of lines, e.g, using
 `$dto = new FoobarData('value1', 'value2')`
instead of
```
$dto = new FoobarData();
$dto->acme1 = 'value1';
$dto->acme2 = 'value2';
```
It is especial convenient in tests where a lot of DTOs instantiating is needed and they are not connected to the entities
So the @Pierstoval solution better fits to my approach.

```
class FoobarData {
    public $acme1;
    public $acme2;

    public function __construct(string $acme1 = null, string $acme2 = null)
    {
        $this->acme1 = $acme1;
        $this->acme2 = $acme2;
    }

    public static function fromFoobar(Foobar $foobar): self {}
}
```

karser:
Indeed @mablae, getters/setters are more flexible. Probably, it'd be sensible for them to be configurable in make:dto - either use getters/setters or public properties.

```
class FoobarData {
    private $acme1;
    private $acme2;

    public function __construct(string $acme1 = null, string $acme2 = null)
    {
        $this->acme1 = $acme1;
        $this->acme2 = $acme2;
    }

    public static function fromFoobar(Foobar $foobar): self {}

    public function getAcme1(): ?string {
        return $this->acme1;
    }

    public function getAcme2(): ?string {
        return $this->acme2;
    }

    //setters if object is mutable
}
```

src/Maker/MakeDto.php
```diff
@@ -0,0 +1,322 @@
...
+    {
+        return 'make:dto';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConf)
+    {
+        $command
+            ->setDescription('Creates a new "data transfer object" (DTO) class from a Doctrine entity')
+            ->addArgument('name', InputArgument::REQUIRED, sprintf('The name of the DTO class (e.g. <fg=yellow>%sData</>)', Str::asClassName(Str::getRandomTerm())))
+            ->addArgument('bound-class', InputArgument::REQUIRED, 'The name of Entity that the DTO will be bound to')
```
OskarStark: Why not entity instead of bound class?

Pierstoval:
Injecting values in the constructor is a nice idea @karser but I see here a big drawback for the Form component: it forces you to use a DataMapper. 

The Form component highly uses getters and setters, and in the end our DTOs are just here to transport data from a layer to another, so having getters and setters is mostly here to guarantee the type & provide some autocompletion features.
Injecting values in the constructor is not natively supported by the Form component, nor is used by the Serializer (in case you want to create a DTO from in API e.g. from a JSON payload), so it's actually quite useless, unless you **always** do things manually (which is not supposed to be a common case when automating such process).

And by the way, with a proper IDE, creating a constructor for a property is done in a few clicks/keyboard shortcuts, as well as getters and setters, so not having a constructor by default does not sound very harmful to me.

karser:
@Pierstoval  It's only for some properties, not for all of them, because having more than 4-5 arguments in constructor makes it useless if you want to specify only last argument, e.g rather than `new FoobarData(null, null, null, null, 'value5')` I'd prefer `$dto = new FoobarData(); $dto->acme5 = 'value5';`.
So constructor with these properties can be created manually and as you suggested, adding method fromFoobar instead of constructor makes sense a lot.

btw Serializer supports [denormalization from the constructor](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php#L389-L451)

weaverryan:
The blocker with this PR is that there's not just *one* way of doing a DTO. I see 3 main options - this is not a *full* list of possibilities, but trying to simplify into a few styles:?

1) Public properties (mutable)
2) Constructor args + getters, but not setters (immutable)
3) Getters + Setters (mutable)

(1) is so easy, and not something we should probably promote too widely (it's certainly fine in many cases - and devs that understand why vs when not will have no problem creating this very simple class).

For the other 2, I don't like having many ways to do things, but this seems like a special case to me. Could we add this question?

> Does your DTO need to be mutable (usually "yes" if using the form component)?

* If no, option (2)
* If yes, option (3)

That would add 1 question that should create DTO's that would cover *most* situations.

Btw @ckrack this PR has been delayed by this complication and my delayed review. Are you still interested and available to work on it?

Cheers!

Pierstoval:
The 3 solutions could be interesting, and solution 1 is almost identical to solution 3 with PHP 7.4 coming along, so we could instead propose the 3 ways of doing it, like:
```
Specify the type of DTO you want:
1. Mutable, with getters & setters (default)
2. Mutable, with public properties
        -> with PHP 7.4, we could introduce properties type-hints
           if DTO is based on an entity with Doctrine types or enough phpdoc
3. Immutable, with getters only
[Getters&setters by default] > 
```

weaverryan:
That makes sense to me - allow the user to choose from these 3 different "styles" of DTO's. I always like less options, but these all seem like legitimate strategies, where the "best" is subjective and dependent on the person/situation.

Pierstoval:
@ckrack what do you need to finish this PR exactly? 😄

odan:
@weaverryan What do you think about this option?

4. Immutable, with getter and wither (fluent setter)

weaverryan:
> Immutable, with getter and wither (fluent setter)

Hmm, maybe. It's really an extension to 3 (Immutable, with getters only) - it could be added now, later, or never :p. We should get at least a few options done initially - i.e. if there are other variants, those might be nice to add, but let's not make them blockers.

ckrack:
Thanks for your input.
I'm trying to free some time to finish this.

tacman:
Is it possible to have a fork of just the MakerBundle (https://github.com/symfony/maker-bundle/)?  I'd be happy to test and give feedback, I need a DTO for a complex entity, would love to give approach this a spin.  I have no opinion on where the files should go or how ->fill() should be called or what should be in the constructor.  Whatever you implement, I'll try.

maxhelias:
@tacman, you can use the fork of this PR and configure it on your `composer.json` like this : https://getcomposer.org/doc/05-repositories.md#vcs

conradkirschner:
Is this PR still in Review? What needs to be done?

weaverryan:
This PR needs a champion to carry it to the finish. @ckrack has done a great job getting it started, but he may or may not (he can correct me) have as much time as he'd like to finish it :). Anyone else can take this work and continue on a new PR if you have some time and willingness!

ckrack:
This should be ready for another review @weaverryan.

I've added the three style options.
```
 Specify the type of DTO you want::
  [1] Mutable, with getters & setters (default)
  [2] Mutable, with public properties
  [3] Immutable, with getters only
>
```

When using 1 and 2, you will get a constructor that accepts the Entity and uses the getters, setters or public properties to setup the Dto with values.

When using the immutable style, you'll get a constructor with all properties (ordered by mandatory -> optional with defaults -> optional) and a named constructor `fromFooEntity` which accepts the `FooEntity` in this case.

The mutator was moved to the Entity
```
Add mutator method to Entity (to set data from the DTO)? (yes/no) [yes]:
 > 
```

This will create `FooEntity->updateFromFooData(FooData $fooData)`.

DuboisS:
Little up/ping for this 2 years old PR just waiting to be reviewed @weaverryan 😄

seb-jean:
PR interesting :)

excitedbox:
Is there anything happening with this? Seems pretty cool, but seems to have gotten stuck right before the finish line.

PR #306, opened 2018-10-25
Checking on reserved keywords for class name validity.

### LINKED ISSUE #305: If you try to create a new entity with the name "Class" the console crashes
**Symfony version(s) affected**: x.y.z
Symfony Installer version 1.5.11

**Description**  
If you try to create a new entity with the name "Class" the console crashes

**How to reproduce**  
 ./bin/console ma:ent
Class name of the entity to create or update (e.g. GentleGnome):
 > Class
 created: src/Entity/Class.php
 created: src/Repository/ClassRepository.php

In Class.php line 10:
                                                                               
  syntax error, unexpected 'Class' (T_CLASS), expecting identifier (T_STRING)  

(and if you try again)
./bin/console ma:ent
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: syntax error, ...

**Possible Solution**  
Don't allow the name Class for new entity 
**Additional context**

SerkanYildiz:
Validator should also check if the given class name is a reserved keyword or not.
However we have to come up with a better exception message, since this one does not reflect the error. Any suggestion?

Thanks for your help @weaverryan during #SymfonyConHackDay2018

Fixes #305

### DIFF
--- a/src/Str.php
+++ b/src/Str.php
@@ -11,8 +11,8 @@
 
 namespace Symfony\Bundle\MakerBundle;
 
-use Symfony\Component\DependencyInjection\Container;
 use Doctrine\Common\Inflector\Inflector;
+use Symfony\Component\DependencyInjection\Container;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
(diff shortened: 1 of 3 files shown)


src/Str.php
```diff
@@ -197,6 +197,29 @@ public static function isValidPhpVariableName($name)
...
+            'exit', 'extends', 'final', 'for', 'foreach', 'function',
+            'global', 'goto', 'if', 'implements', 'include',
+            'include_once', 'instanceof', 'insteadof', 'interface', 'isset',
+            'list', 'namespace', 'new', 'or', 'print', 'private',
+            'protected', 'public', 'require', 'require_once', 'return',
+            'static', 'switch', 'throw', 'trait', 'try', 'unset',
+            'use', 'var', 'while', 'xor', '__CLASS__', '__DIR__', '__FILE__',
+            '__FUNCTION__', '__LINE__', '__METHOD__', '__NAMESPACE__', '__TRAIT__',
+            'int', 'float', 'bool', 'string', 'true', 'false', 'null', 'void',
+            'iterable', 'object',
```
sadikoff: > 'int', 'float', 'bool', 'string', 'true', 'false', 'null', 'void',
            'iterable', 'object',

This words are allowed to use in Namespace, anyway it will be very strange to use it :-)

src/Validator.php
```diff
@@ -29,7 +29,7 @@ public static function validateClassName(string $className, string $errorMessage
         $pieces = explode('\\', ltrim($className, '\\'));
 
         foreach ($pieces as $piece) {
-            if (!preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $piece)) {
+            if (!Str::isValidPhpClassName($piece)) {
```
sadikoff: Maybe will be better not to change this check, but add a new check with new error message, to show the user what is wrong with classname?

sadikoff:
@SerkanYildiz It's great that you have done it so fast, but I was talking about something else. I suggested not to split namespace and class validator, but leave old validation in place and add new one which will validate reserved words
ex:
```php
foreach ($pieces as $piece) {
    if (!preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $piece)) {
        $errorMessage = $errorMessage ?: sprintf('"%s" is not valid as a PHP class name (it must start with a letter or underscore, followed by any number of letters, numbers, or underscores)', $className);
        throw new RuntimeCommandException($errorMessage);
    }
    if (Str::isPhpReservedWord($piece)) {
        throw new RuntimeCommandException(sprintf('"%s" is a reserved keyword and thus cannot be used as class name in PHP.', $className));
    }
}
```

weaverryan:
Ping @SerkanYildiz! I think we need just one more small change. Does my description above make sense?

Cheers!

SerkanYildiz:
@weaverryan Pushed the requested changes. Does it look good now?

weaverryan:
Thank you @SerkanYildiz!

PR #307, opened 2018-10-26
[make:crud][experimental] Add generated tests to make:crud

### LINKED ISSUE #229: make:crud to make basic tests for each route
When running a `make:crud` command, it'd be handy if this command could scaffold tests that check for a valid 200 status for each route. e.g. `tests/Functional/Controller/FooTest` as a framework for checking the urls.

ckrack:
Add tests for the controller methods.
Add necessary dependencies.
Closes #229

Edit by @jrushlow:
- Feature is experimental - We generate a minimum WebTestCase for generated CRUD. 
- Generating tests is on an opt-in basis _if_ the user has `test-pack` already installed. Otherwise we're not advertising this feature yet.
- Entity / Form Field guessing is not _fully_ implemented. "Setter" values other than `string` will not match the Entity's property types and must be manually changed.
- Tests (other than `testIndex()`) are all generated with `markTestIncomplete()` 
- Better integration w/ Panther, Fixtures, Foundry, etc... to come in future PR's

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -197,8 +197,8 @@ final class MakeCrud extends AbstractMaker
                 'route_path' => Str::asRoutePath($controllerClassDetails->getRelativeNameWithoutSuffix()),
                 'route_name' => $routeName,
                 'class_name' => Str::getShortClassName($testClassDetails->getFullName()),
-				'namespace' => Str::getNamespace($testClassDetails->getFullName()),
-				'form_fields' => $entityDoctrineDetails->getFormFields(),
+                'namespace' => Str::getNamespace($testClassDetails->getFullName()),
+                'form_fields' => $entityDoctrineDetails->getFormFields(),
             ]
         );
 
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -33,7 +33,6 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Question\Question;
-use Symfony\Component\CssSelector\CssSelectorConverter;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -277,7 +276,7 @@ final class MakeCrud extends AbstractMaker
                 ]
             );
 
-            if (!class_exists(CssSelectorConverter::class)) {
+            if (!class_exists(WebTestCase::class)) {
                 $io->caution('You\'ll need to install the `symfony/test-pack` to execute the tests for your new controller.');
             }
         }
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -197,7 +197,8 @@ final class MakeCrud extends AbstractMaker
                 'route_path' => Str::asRoutePath($controllerClassDetails->getRelativeNameWithoutSuffix()),
                 'route_name' => $routeName,
                 'class_name' => Str::getShortClassName($testClassDetails->getFullName()),
-                'namespace' => Str::getNamespace($testClassDetails->getFullName()),
+				'namespace' => Str::getNamespace($testClassDetails->getFullName()),
+				'form_fields' => $entityDoctrineDetails->getFormFields(),
             ]
         );
 
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -248,11 +248,6 @@ final class MakeCrud extends AbstractMaker
                 'ControllerTest'
             );
 
-            $entityFields = $entityDoctrineDetails->getDisplayFields();
-            $entityFields = array_filter($entityFields, static function ($field) use ($entityDoctrineDetails) {
-                return $field['fieldName'] !== $entityDoctrineDetails->getIdentifier();
-            });
-
             $useStatements = new UseStatementGenerator([
                 $entityClassDetails->getFullName(),
                 WebTestCase::class,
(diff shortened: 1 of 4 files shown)


src/Maker/MakeCrud.php
```diff
@@ -247,5 +268,15 @@ public function configureDependencies(DependencyBuilder $dependencies)
...
+
+        $dependencies->addClassDependency(
+            CssSelectorConverter::class,
+            'css-selector'
+        );
+
+        $dependencies->addClassDependency(
+            Client::class,
+            'browser-kit'
+        );
```
weaverryan: These should not be required dependencies - the test are really an optional thing. Instead, IF these dependencies are installed, let's generate a test. If they are not installed, let's *not* generate the tests files, but print a message like:

> Skipping test generation because the test dependencies (css-select and browser-kit) are not installed.

src/Resources/skeleton/crud/test/Test.tpl.php
```diff
@@ -0,0 +1,134 @@
...
+        $this->assertContains('<?= $entity_class_name ?> index', $crawler->filter('h1')->text());
+
+        $newLink = $crawler->filter('a:contains("Create new")')->eq(0)->link();
+
+        return $newLink;
+    }
+
+    /**
+     * @TODO implement new test
+     * @depends testIndex
```
weaverryan: I'm not sure if this is a good idea. The tests should be isolated, and it's not like we *really* "need" the `testIndex` to be run first - we just need a value that it conveniently returns. I think we should remove this dependency and go directly to the index here then find that new link. This will make the `testNew()` tell a "complete story" without reading another test method.

src/Resources/skeleton/crud/test/Test.tpl.php
```diff
@@ -0,0 +1,134 @@
...
+        // $crawler = $client->followRedirect();
+        // $this->assertTrue($client->getResponse()->isSuccessful());
+        // $this->assertContains('<!DOCTYPE html>', $client->getResponse()->getContent());
+        // $this->assertContains('<?= $entity_class_name ?> index', $crawler->filter('h1')->text());
+        // $this->assertContains('<td>Lorem ipsum</td>', $client->getResponse()->getContent());
+    }
+
+    // /**
+    //  * @TODO implement edit test
+    //  * @depends testNew
```
weaverryan: Instead of depending on testNew, we should create an entity directly via the EntityManager here and persist it. Depending on `testNew` is convenient at first, but it couples the test methods together, which has a practical downside: I now need to run two full tests before running this test (testIndex, then testNew, then this one).

Btw, I'm not so sure that we need to comment out *all* of the edit tests. Because... we *know* what entity is being generated. And so, we could write some code to instantiate that object and persist it. Heck... thinking crazy, we could use faker and introspect the Doctrine annotation information to even fill in the required fields if they have Faker installed (we could recommend it).

It's a bit more work, but in theory, we could make these tests fully (or nearly fully) functional.

weaverryan:
@ckrack Do you still have an interest in finishing this? Or should we try to find another volunteer? :)

TheGarious:
Up.

Need Help for finish this PR ?

src/Maker/MakeCrud.php
ckrack: I've made this optional and added the notice.

src/Resources/skeleton/crud/test/Test.tpl.php
ckrack: Tests are now isolated.

src/Resources/skeleton/crud/test/Test.tpl.php
ckrack: I have looked into using Faker for the test generation.

What is easily achievable is using Faker inside the tests to populate Entities based on their properties. 
However, this will lead to tests never failing when properties were changed. Faker will always use an up-to-date generator, because the guesser bases on the current entity 
Tests will only fail if the `Form` is not updated and requires a change, too.

This looks like:
```php
// App\Test\Controller\ExampleControllerTest;
// inside a test* method that needs an entity in the db:
$entityPopulator = new EntityPopulator($this->entityManager->getClassMetadata(Test::class));
$entityPopulator->setColumnFormatters($entityPopulator->guessColumnFormatters($this->fakerGenerator));

/** @var App\Entity\Test */
$test = $entityPopulator->execute($this->entityManager, []);

$this->entityManager->persist($test);
$this->entityManager->flush();
```

One could argument that this is enough, because we are testing the controller - not the entity or form.
I'd prefer to generate tests that contain somewhat fixed values and will fail, when either the entity or form are changed.
So tests could either contain faker generator calls ``$faker->name()`` or the output of this. The former would require reverse-engineering Faker's name guesser, as the populators are returned as closures and do not allow retrieving their name.
This means we must keep a copy of this up to date, to allow getting the generator names as string: https://github.com/fzaninotto/Faker/blob/master/src/Faker/Guesser/Name.php
Not the best option.

This is also complicated, due to association mappings.
We can leave them out or auto-populate them in the tests.
If we leave them out, the generated tests will fail - our test of the maker command must reflect this.

WDYT?

src/Maker/MakeCrud.php
```diff
@@ -95,6 +100,10 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
             sprintf('Choose a name for your controller class (e.g. <fg=yellow>%s</>)', $defaultControllerClass),
             $defaultControllerClass
         );
+
+        if (class_exists(CssSelectorConverter::class)) {
+            $this->generateTests = $io->confirm('Do you want to generate tests for the controller? [Experimental]', false);
+        }
```
weaverryan, line 106: What would happen if we didn't check for the class? And just always asked the question? If they ran the tests, obviously they wouldn't work, but would it be obvious that they just need to install some stuff? I'm trying to limit the number of conditionals and paths we have.

src/Resources/skeleton/crud/test/Test.EntityManager.tpl.php
```diff
@@ -0,0 +1,123 @@
...
+
+        $this->client->submitForm('Save', [
+<?php foreach ($form_fields as $form_field => $typeOptions): ?>
+            '<?= $form_field_prefix; ?>[<?= $form_field; ?>]' => 'Testing',
+<?php endforeach; ?>
+        ]);
+
+        self::assertResponseRedirects('/sweet/food/');
+
+        self::assertSame(1, $this->manager->getRepository()->count([]));
```
weaverryan, line 57:
```suggestion
        self::assertSame(1, $this->repository->count([]));
```

weaverryan:
Thank you @ckrack and @jrushlow!

Issue #299, opened 2018-10-21
No directory for Entity


wow-apps:
While making PHPDocs for Maker Bundle, I found, that if I'm creating, for example, new Command and there is no `Command` directory in application, Maker creates it, but in a case of new Entity - it doesn't.

`alexeysamara: ~/projects/symfony-maker-bundle/dev (psr-fix**) $ c make:entity`

```
2018-10-21T13:54:30+00:00 [error] Error thrown while running command "make:entity". Message: "The "/Users/alexeysamara/projects/symfony-maker-bundle/dev/src/Entity/" directory does not exist."

In Finder.php line 547:

  The "/Users/alexeysamara/projects/symfony-maker-bundle/dev/src/Entity/" directory does not exist.
```

I'm not sure - this is bug or feature )

PR #310, opened 2018-10-28
Fixed issue with no entity directory in symfony 3.4

yuriy-yakubskiy:
Hey guys,

The following could fix an issue without adding extra config to maker. 

Symfony Version: **3.4**
Maker Bundle: **1.8.0** and **master**
Bug Fix:  #265, #299

P.S
1. Failed to find the way to add a test for Symfony 3.4 only 4+ tests are supported currently

2. Unfortunately can't fix as proposed in the following comment
https://github.com/symfony/maker-bundle/pull/265#issuecomment-421213663
if call $cmf->getAllMetadata() before generate entity it couldn't be loaded by calling getAllMetadata again.
See the following commit in other branch https://github.com/yuriy-yakubskiy/maker-bundle/commit/8b669991517a748878b16569452853b63c2122fb

Thanks!

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -99,20 +99,6 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
             return;
         }
 
-        $entityFinder = $this->fileManager->createFinder('src/Entity/')
-            // remove if/when we allow entities in subdirectories
-            ->depth('<1')
-            ->name('*.php');
-        $classes = [];
-        /** @var SplFileInfo $item */
-        foreach ($entityFinder as $item) {
-            if (!$item->getRelativePathname()) {
-                continue;
-            }
-
-            $classes[] = str_replace(['.php', '/'], ['', '\\'], $item->getRelativePathname());
-        }
-
         $argument = $command->getDefinition()->getArgument('name');
         $question = $this->createEntityClassQuestion($argument->getDescription());
         $value = $io->askQuestion($question);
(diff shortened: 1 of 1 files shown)


weaverryan:
Hey @yuriy-yakubskiy!

Thanks for taking this on! There is now a "competing" PR - #311. That approach looks good to me... however, the tests are failing. And I'm wondering you "hit" the same issue that the tests are hitting in #311. Specifically, can you explain your comment a bit more:

> Unfortunately can't fix as proposed in the following comment #265 (comment)
> if call $cmf->getAllMetadata() before generate entity it couldn't be loaded by calling getAllMetadata again.
> See the following commit in other branch yuriy-yakubskiy@8b66999

Thanks!

yuriy-yakubskiy:
Hey @weaverryan !

The main problem is that if use DoctrineHelper for auto-completing entity names it loads doctrine metadata 
https://github.com/symfony/maker-bundle/blob/cc5c0d9a94f66fa4844e5ad35be3780a4fc424af/src/Doctrine/DoctrineHelper.php#L135

before generate an entity class. Then after entity is generated application checks if class exists here

https://github.com/symfony/maker-bundle/blob/cc5c0d9a94f66fa4844e5ad35be3780a4fc424af/src/Maker/MakeEntity.php#L152 

but as metadata was loaded without this class it returns false and application and test accordingly fails on this stage.
I tried but failed to find a way to let know the auto-loader that there is a new generated entity class...

Hope it explains more or less the issue.

Cheers

weaverryan:
Merged #311 - thanks for starting this!

PR #311, opened 2018-10-29
Fix Entity directory does not exist

### LINKED ISSUE #299: No directory for Entity

yceruto:
Closes #299 according to https://github.com/symfony/maker-bundle/pull/185#issuecomment-389009674

Also alternative for https://github.com/symfony/maker-bundle/pull/185, https://github.com/symfony/maker-bundle/pull/265 and https://github.com/symfony/maker-bundle/pull/310

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Doctrine;
 
 use Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory;
+use Doctrine\Common\Persistence\Mapping\Driver\AnnotationDriver;
 use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Common\Persistence\Mapping\Driver\MappingDriverChain;
 use Doctrine\ORM\EntityManagerInterface;
@@ -144,6 +145,19 @@ final class DoctrineHelper
 
                 foreach ($loaded as $m) {
                     $cmf->setMetadataFor($m->getName(), $m);
+                }
+
+                // Invalidating the cached AnnotationDriver::$classNames to find new Entity classes
+                $metadataDriver = $em->getConfiguration()->getMetadataDriverImpl();
+                if ($metadataDriver instanceof MappingDriverChain) {
+                    foreach ($metadataDriver->getDrivers() as $driver) {
+                        if ($driver instanceof AnnotationDriver) {
+                            $classNames = (new \ReflectionObject($driver))->getProperty('classNames');
+                            $classNames->setAccessible(true);
+                            $classNames->setValue($driver, null);
+                            $classNames->setAccessible(false);
+                        }
+                    }
                 }
             }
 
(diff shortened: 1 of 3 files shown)


weaverryan:
OMG - finally! So happy to see this PR! And, it was even simpler than I expected!

We do have test failures, unfortunately. @yceruto could you take a look at them? They could be false failures. But it might also be an indication that using the new way may cause an unintended exception to be thrown in some cases.

Thanks!

weaverryan:
Also, see #310 - I think the other PR may have "hit" the same issue that's causing the tests to fail here 🤔

yceruto:
Yep, (in the same life time of the script) the first time that `$this->doctrineHelper->getEntitiesForAutocomplete()` is called with no entities some protected Doctrine properties are cached (e.g. `AnnotationDriver::$classNames`) and there is no way to invalidate it without force it and discover the metadatas of the new generated Entity.

Even if the last commit resolves part of the problem I guess there are still other cached properties related to the mapping type that need to be invalidated for the same reason.

So I think we should go back to the `Finder` version and use `AnnotationDriver::getPaths()` to discover the mapping Entity directory, wdyt?

yceruto:
Also we could re-think the way this Annotation Mapping requirement is validated, because now it's checked after the entity was already generated.

yceruto:
Well, I've fixed an invalid mapping from the fixture of the two remaining fails and all green again.

src/Doctrine/DoctrineHelper.php
```diff
@@ -145,6 +146,19 @@ public function getMetadata(string $classOrNamespace = null, bool $disconnected
                 foreach ($loaded as $m) {
                     $cmf->setMetadataFor($m->getName(), $m);
                 }
+
+                // Invalidating the cached AnnotationDriver::$classNames to find new Entity classes
+                $metadataDriver = $em->getConfiguration()->getMetadataDriverImpl();
+                if ($metadataDriver instanceof MappingDriverChain) {
+                    foreach ($metadataDriver->getDrivers() as $driver) {
+                        if ($driver instanceof AnnotationDriver) {
```
weaverryan: Wow! Yea, this issue is a bummer - but nice job at least finding a workaround.

Does this only affect the AnnotationDriver?

yceruto: And also for `StaticPHPDriver` but we only support annotation driver, right?

weaverryan:
Thank you very much for figuring this out @yceruto!

PR #315, opened 2018-11-03
Another way to resolve "No directory for Entity issue"

yuriy-yakubskiy:
Here is actually yet another workaround idea as alternative to #311 to resolve issue with not loaded doctrine annotation metadata.
Just requiring file after entity is dumped to the file.  

Also some tests are seems to be broken. 
That's why they failing for #311 too.

https://github.com/symfony/maker-bundle/blob/cc5c0d9a94f66fa4844e5ad35be3780a4fc424af/tests/Maker/FunctionalTest.php#L1124

and 

https://github.com/symfony/maker-bundle/blob/cc5c0d9a94f66fa4844e5ad35be3780a4fc424af/tests/Maker/FunctionalTest.php#L1148

Annotation error asserted in those test was thrown not because metadata type is wrong but because for the issue we are trying to resolve.

:)

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -35,7 +35,6 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Question\Question;
-use Symfony\Component\Finder\SplFileInfo;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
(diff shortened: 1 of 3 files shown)


weaverryan:
Thanks for taking another attempt at this! It's much more complex than I expected :). Can you explain a bit more why the tests are failing? On master, the tests ARE currently passing. So why does this change cause those tests to fail?

yuriy-yakubskiy:
1. When I first tried to use maker-bundle on Symfony 3.4 and tried to create very first entity I got Annotation unsupported mapping error. It couldn't be a problem with mapping as I'm using annotations on my project.

today I cd to the test project for testing "MakeEntityXmlMappingError" 
tests/tmp/cache/maker_app_a415fad3bbeef5284f98d1d9e50b9051 and manually debug issue with tests it appears that application throws error about not existing class relation that's why I removed the following line 

https://github.com/symfony/maker-bundle/blob/e873047b6022a0343c9354f65d5d47b9e351dd53/tests/fixtures/MakeEntityXmlMappingError/config/doctrine/User.orm.xml#L10. 

Another test tries to create UserAvatarPhoto entity and expects that annotation error for existing user xml Entity will be thrown. But current code checks only Entity which is about to be generated.

weaverryan:
Ok, hmm, this helps. Question: if you revert your changes (i.e. use the master branch of this bundle), run the test, and then move into the test project for "MakeEntityXmlMappingError", do you get the error?

weaverryan:
Merged the other pull request - thanks!

PR #316, opened 2018-11-07
Skip embedded fields in embeddables

cedricziel:
This change skips the creation of mapped fields inside embeddables.

Consider the PHP Money class:

```php
public class Invoice {
  /**
   * @var Money
   *
   * @ORM\Embedded(class="Money\Money")
   */
  private $total;
}
```

```php
final class Money implements \JsonSerializable {
   /**
    * @var Currency
    */
   private $currency;
}
```

Here, Money has the mapped field `Currency`. In doctrine Metadata,
this is correctly reflected as `total.currency`.

The EntityRegenerator declares a private field `$total.currency` afterwards.

The changeset suppresses this behaviour by skipping fields containing a
dot as is done a few lines below.


weaverryan:
Hey @cedricziel!

This is an excellent change! But could you add a test for it? There are a lot of test cases already in the `FunctionalTest` class - we just need to duplicate an existing one, use an embeddable, and make sure the generated code doesn't explode (because a property named `$total.currency` will definitely make things explode).

Thanks!

weaverryan:
Thank you Cedric!

PR #318, opened 2018-11-09
Fix fixtures test

ckrack:
The test tried to create an AppFixtures class.
However, this is included in the flex-recipe for "ormfixtures", since it's [latest release](https://github.com/symfony/recipes/commit/c447845d62ffa63eb940db89ed00a92b00350fe5)
Fix test by renaming the fixture to be generated to AcmeFixture.

### DIFF
--- a/tests/Maker/FunctionalTest.php
+++ b/tests/Maker/FunctionalTest.php
@@ -162,10 +162,10 @@ class FunctionalTest extends MakerTestCase
         yield 'fixtures' => [MakerTestDetails::createTest(
             $this->getMakerInstance(MakeFixtures::class),
             [
-                'AppFixtures',
+                'AcmeFixtures',
             ])
             ->assert(function (string $output, string $directory) {
-                $this->assertContains('created: src/DataFixtures/AppFixtures.php', $output);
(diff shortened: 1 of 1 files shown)


tests/Maker/FunctionalTest.php
```diff
@@ -162,10 +162,10 @@ public function getCommandTests()
         yield 'fixtures' => [MakerTestDetails::createTest(
             $this->getMakerInstance(MakeFixtures::class),
             [
-                'AppFixtures',
+                'AcmeFixtures',
             ])
             ->assert(function (string $output, string $directory) {
-                $this->assertContains('created: src/DataFixtures/AppFixtures.php', $output);
+                $this->assertContains('created: src/DataFixtures/AcmeFixtures.php', $output);
```
sadikoff: Don't know if we need here to check it. I think we can remove assert block here, tests automatically check if command successfully executed.

ckrack: The assertion indicates whether the correct file is created, even though it just relies on the message.
Anyway, the PR is just for fixing the error in the test.

weaverryan:
Thanks for the fix @ckrack - I apologize for merging another PR out of order (someone found the issue at the hack day). Cheers!

PR #323, opened 2018-12-03
Id integers should default to unsigned.

mcneely:
There should never be a reason to have an id a signed integer. Having your integers wrap around into the negatives is a very bad thing.


ostrolucky:
It would, because it would make this code having no effect -> dead code. Such code is dangerous, because developers are not aware it doesn't work and will rely on it. If it has no effect, it shouldn't be generated.

If this change is to be accepted, it must be more smart and generate this code only in cases default database which support this is set.

mcneely:
https://twitter.com/Ocramius/status/1070324961691267077

PR #326, opened 2018-12-08
Fix name of fixture in tests

SerkanYildiz:
Problem: Tests are failing due to the name of the fixture which is *AppFixture*. But a fixture with that name is created by [DoctrineFixtures recipe](https://github.com/symfony/recipes/commit/c447845d62ffa63eb940db89ed00a92b00350fe5).

As talked during #SymfonyConHackDay2018, the fix was to change the name of the fixture's that we want to create during the tests.


weaverryan:
Thanks Serkan!

PR #333, opened 2018-12-14
New Command! make:registration-form

weaverryan:
New command time! `make:registration-form`. It:

* Generates a completely functional controller, template & form class for registration
* Is capable of detecting or asking for the exact fields you need (e.g. email & password)
* When using Guard, you can generate code to automatically authenticate after registration.

This is one of the last steps to getting an entire authentication system in minutes without needing FOSUserBundle (`make:user`, `make:auth`, `make:registration-form`).

Feedback warmly welcomed.

Cheers!

### Example of generated code:

<details><summary>RegistrationFormType</summary>
<p>

```php
<?php

namespace App\Form;

use App\Entity\User;
use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\Extension\Core\Type\PasswordType;
use Symfony\Component\Form\FormBuilderInterface;
use Symfony\Component\OptionsResolver\OptionsResolver;
use Symfony\Component\Validator\Constraints\NotBlank;
use Symfony\Component\Validator\Constraints\Length;

class RegistrationFormType extends AbstractType
{
    public function buildForm(FormBuilderInterface $builder, array $options)
    {
        $builder
            ->add('email')
            ->add('plainPassword', PasswordType::class, [
                // instead of being set onto the object directly,
                // this is read and encoded in the controller
                'mapped' => false,
                'constraints' => [
                    new NotBlank([
                        'message' => 'Please enter a password',
                    ]),
                    new Length([
                        'min' => 6,
                        'minMessage' => 'Your password should be at least {{ limit }} characters',
                    ]),
                ],            ])
        ;
    }

    public function configureOptions(OptionsResolver $resolver)
    {
        $resolver->setDefaults([
            'data_class' => User::class,
        ]);
    }
}
```

</p>
</details>

<details><summary>RegistrationController</summary>
<p>

```php
namespace App\Controller;

use App\Entity\User;
use App\Form\RegistrationFormType;
use App\Security\StubAuthenticator;
use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Annotation\Route;
use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
use Symfony\Component\Security\Guard\GuardAuthenticatorHandler;

class RegistrationController extends AbstractController
{
    /**
     * @Route("/register", name="app_register")
     */
    public function register(Request $request, UserPasswordEncoderInterface $passwordEncoder, GuardAuthenticatorHandler $guardHandler, StubAuthenticator $authenticator): Response
    {
        $form = $this->createForm(RegistrationFormType::class);
        $form->handleRequest($request);

        if ($form->isSubmitted() && $form->isValid()) {
            /** @var User */
            $user = $form->getData();

            // encode the plain password
            $user->setPassword(
                $passwordEncoder->encodePassword(
                    $user,
                    $form->get('plainPassword')->getData()
                )
            );

            $entityManager = $this->getDoctrine()->getManager();
            $entityManager->persist($user);
            $entityManager->flush();

            // do anything else you need here, like send an email

            return $guardHandler->authenticateUserAndHandleSuccess(
                $user,
                $request,
                $authenticator,
                'main' // firewall name in security.yaml
            );
        }

        return $this->render('registration/register.html.twig', [
            'registrationForm' => $form->createView(),
        ]);
    }
}
```

</p>
</details>

<details><summary>registration/register.html.twig</summary>
<p>

```twig
{% extends 'base.html.twig' %}

{% block title %}Register{% endblock %}

{% block body %}
    <h1>Register</h1>

    {{ form_start(registrationForm) }}
        {{ form_row(registrationForm.email) }}
        {{ form_row(registrationForm.plainPassword) }}

        <button class="btn">Register</button>
    {{ form_end(registrationForm) }}
{% endblock %}
```

</p>
</details>


src/Maker/MakeRegistrationForm.php
```diff
@@ -0,0 +1,279 @@
...
+
+    public function configureCommand(Command $command, InputConfiguration $inputConf)
+    {
+        $command
+            ->setDescription('Creates a new registration form system')
+            ->addArgument('user-class', InputArgument::OPTIONAL, 'Full user class')
+            ->addArgument('username-field', InputArgument::OPTIONAL, 'Field on your User class used to login')
+            ->addArgument('password-field', InputArgument::OPTIONAL, 'Field on your User class that stores the hashed password')
+            ->addOption('auto-login-authenticator', null, InputOption::VALUE_REQUIRED, 'Authenticator class to use for logging in')
+            ->addOption('firewall-name', null, InputOption::VALUE_REQUIRED, 'Firewall key used for authentication')
```
javiereguiluz: This option name may be confusing. It's called "firewall name" ... it's described as "firewall key" ... and maybe it's the "firewall context"? (I haven't looked at the code).

tests/fixtures/MakeRegistrationFormNoGuessing/src/Security/AuthenticatorFirst.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace App\Security;
+
+class AuthenticatorFirst
+{
+}
```
javiereguiluz: Missing new line at end of file.

tests/fixtures/MakeRegistrationFormNoGuessing/src/Security/AuthenticatorSecond.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace App\Security;
+
+class AuthenticatorSecond
+{
+}
```
javiereguiluz: Missing new line at end of file.

weaverryan:
This should be ready to go! Added example generated output in the PR description.

src/Maker/MakeRegistrationForm.php
```diff
@@ -0,0 +1,328 @@
...
+
+/**
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ *
+ * @internal
+ */
+final class MakeRegistrationForm extends AbstractMaker
+{
+    private $fileManager;
+
```
hhamon: Are the empty line breaks part of the CS guidelines?

tests/fixtures/MakeRegistrationFormEntity/src/Entity/User.php
```diff
@@ -0,0 +1,104 @@
...
+    public function getEmail()
+    {
+        return $this->email;
+    }
+
+    public function setEmail(string $email): self
+    {
+        $this->email = $email;
+
+        return $this;
```
hhamon: Shall we discourage the fluent interfaces for entities, especially as they seem unused in the rest of the code?

javiereguiluz: It's true that we don't use the fluent interface ourselves. However, adding it means that end-users can choose between using it as a fluent interface or not.

If we remove this fluent interface, we remove that flexibility entirely.

lyrixx:
Hello. I'm a bit late.

Why doing that:
```php
        $form = $this->createForm(RegistrationFormType::class);
        $form->handleRequest($request);

        if ($form->isSubmitted() && $form->isValid()) {
            /** @var User */
            $user = $form->getData();
```

Instead of:
```php
        $user = new User();
        $form = $this->createForm(RegistrationFormType::class, $user);
        $form->handleRequest($request);

        if ($form->isSubmitted() && $form->isValid()) {
```

The last one is IMHO simpler to understand and less magical. Everything is explicit, and there are no need for this PHP Doc

lyrixx:
Another question: why do you put the password transformation logic in the controller and not in a form listener ?

Another question: Why do you authenticate the user? This si considered as a bad practice (the email is not yet validated) This could be let as an example, but commented with an explanation 

Finally, in your example, the `StubAuthenticator` seems useless

weaverryan:
@lyrixx

> Why doing that: ... Instead of:

I think your idea is indeed more clear. I'll open a pull request.

> Another question: why do you put the password transformation logic in the controller and not in a form listener ?

Form listeners are way too complex - I think doing a few things in the controller is easy and clear.

> Another question: Why do you authenticate the user? This is considered as a bad practice (the email is not yet validated) This could be let as an example, but commented with an explanation

We ask the user whether or not they want this - so they can choose what they want :)

> Finally, in your example, the StubAuthenticator seems useless

It is - but that's only because I messed up when copying (and simplifying) the code. It's all used correctly in your real app - if you choose to authenticate, you choose which authenticator, then we use that to authenticate the user `StubAuthenticator` is from our test suite.

lyrixx:
Thanks for your feedback 👍

PR #334, opened 2018-12-17
Remove Symfony Flex dependency from MakerBundle

danrot:
We have been confused because the documentations states that Symfony Flex is required in order to use the MakerBundle.

However, we are not using Symfony Flex in our application yet, but it was still possible to use the MakerBundle. So I would suggest to remove the part saying that Symfony Flex is a dependency.


weaverryan:
Thanks @danrot! Yea, the bundle is more flexible in this regard than it was originally - so the note was out of date. Cheers!

PR #338, opened 2018-12-20
New command - make:message for Messenger

weaverryan:
An easy one - generated for Messenger - generates an empty message class and corresponding handler:

screen shot 2018-12-20 at 2 23 54 pm

**Generated Code**

<details><summary>SendWelcomeEmailMessage</summary>
<p>

```php
<?php

namespace App\Message;

class SendWelcomeEmailMessage
{
    /*
     * Add whatever properties & methods you need to hold the
     * data for this message class.
     *
     * Note - if you're handling this message async, you may need
     * a getter & setter method for each property so that it serializes
     * and unserializes correctly.
     */

//     private $name;
//
//     public function __construct(string $name)
//     {
//         $this->name = $name;
//     }
//
//    public function getName(): string
//    {
//        return $this->name;
//    }
//
//    public function setName(string $name)
//    {
//        $this->name = $name;
//    }
}
```

</p>
</details>

<details><summary>SendWelcomeEmailMessageHandler</summary>
<p>

```php
<?php

namespace App\MessageHandler;

use App\Message\SendWelcomeEmailMessage;
use Symfony\Component\Messenger\Handler\MessageHandlerInterface;

class SendWelcomeEmailMessageHandler implements MessageHandlerInterface
{
    public function __invoke(SendWelcomeEmailMessage $message)
    {
        // do something with your message
    }
}

```

</p>
</details>

In the future... we could possibly ask which transport you want to use and update your YAML configuration (if needed) 🤔

### DIFF
--- a/src/Resources/skeleton/message/Message.tpl.php
+++ b/src/Resources/skeleton/message/Message.tpl.php
@@ -1,0 +1,32 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+class <?= $class_name."\n" ?>
+{
+    /*
+     * Add whatever properties & methods you need to hold the
+     * data for this message class.
+     *
+     * Note - if you're handling this message async, you may need
+     * a getter & setter method for each property so that it serializes
+     * and unserializes correctly.
+     */
+
+//     private $name;
+//
+//     public function __construct(string $name)
+//     {
+//         $this->name = $name;
+//     }
+//
+//    public function getName(): string
+//    {
+//        return $this->name;
+//    }
+//
+//    public function setName(string $name)
+//    {
+//        $this->name = $name;
+//    }
+}
--- a/src/Resources/skeleton/message/Message.tpl.php
+++ b/src/Resources/skeleton/message/Message.tpl.php
@@ -2,15 +2,11 @@
 
 namespace <?= $namespace; ?>;
 
-class <?= $class_name."\n" ?>
+final class <?= $class_name."\n" ?>
 {
     /*
      * Add whatever properties & methods you need to hold the
      * data for this message class.
-     *
-     * Note - if you're handling this message async, you may need
-     * a getter & setter method for each property so that it serializes
-     * and unserializes correctly.
      */
 
 //     private $name;
@@ -23,10 +19,5 @@ class <?= $class_name."\n" ?>
 //    public function getName(): string
 //    {
 //        return $this->name;
-//    }
-//
-//    public function setName(string $name)
-//    {
-//        $this->name = $name;
 //    }
 }
--- a/src/Resources/skeleton/message/MessageHandler.tpl.php
+++ b/src/Resources/skeleton/message/MessageHandler.tpl.php
@@ -1,0 +1,14 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+use <?= $message_full_class_name ?>;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+class <?= $class_name ?> implements MessageHandlerInterface
+{
+    public function __invoke(<?= $message_class_name ?> $message)
+    {
+        // do something with your message
+    }
+}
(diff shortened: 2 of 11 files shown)


src/Resources/skeleton/message/Message.tpl.php
```diff
@@ -0,0 +1,32 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+class <?= $class_name."\n" ?>
+{
+    /*
+     * Add whatever properties & methods you need to hold the
```
OskarStark: what about doing it like in the make entity command, one could nearly reuse the existing code without the annotations stuff. This would make it more handy imo

src/Resources/skeleton/message/Message.tpl.php
```diff
@@ -0,0 +1,32 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+class <?= $class_name."\n" ?>
```
OskarStark: would you mind making this `final`?

src/Resources/skeleton/message/MessageHandler.tpl.php
```diff
@@ -0,0 +1,14 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+use <?= $message_full_class_name ?>;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+class <?= $class_name ?> implements MessageHandlerInterface
```
OskarStark: `final`, too?

src/Resources/skeleton/message/Message.tpl.php
```diff
@@ -0,0 +1,32 @@
...
+//     {
+//         $this->name = $name;
+//     }
+//
+//    public function getName(): string
+//    {
+//        return $this->name;
+//    }
+//
+//    public function setName(string $name)
```
rvanlaak: Would be preferable to keep messages immutable. What about not promoting setters?

OskarStark: Good idea 💡

src/Resources/skeleton/message/Message.tpl.php
LiSinCin: Hi, can you point what part of command or entity can be reused? :)

src/Resources/skeleton/message/Message.tpl.php
```diff
@@ -0,0 +1,32 @@
...
+class <?= $class_name."\n" ?>
+{
+    /*
+     * Add whatever properties & methods you need to hold the
+     * data for this message class.
+     *
+     * Note - if you're handling this message async, you may need
+     * a getter & setter method for each property so that it serializes
+     * and unserializes correctly.
+     */
```
weaverryan: this comment can now be removed

src/Resources/skeleton/message/Message.tpl.php
weaverryan: I agree to remove the setter example

nikophil:
Hi,

according to a chat i had with @weaverryan i've completed this maker in order to allow the user to route the message to a transport

Capture du 2019-12-11 17-53-23

OskarStark:
I am routing my Messages based on an „AsyncMessage“-Marker Interface, is it worth to add such an option instead of manipulating the messenger.yaml file?

weaverryan:
It’s a very interesting idea... but I don’t think it’s standard enough yet to put into maker. Better to show them where/how routing happens

nikophil:
i kinda agree with @weaverryan. Although it is a nice trick, it may involve too much complexity and different cases to handle in the maker.

BTW, the CI is failing because of some deprecation i fixed here: https://github.com/symfony/maker-bundle/pull/518

weaverryan:
Thank you for taking this across the finish line @nikophil!

OskarStark:
Nice, when can we expect a release? 😊😎

weaverryan:
Soon! Finishing the new reset password stuff!

PR #342, opened 2018-12-21
Update index.rst

millelc:
adding informations about PHP version


weaverryan:
Hey @millelc!

Hmm, but the bundle IS compatible with PHP 7.0 - all versions of it (except maybe for a VERY early version). A few commands require PPH 7.1, but the command tells you when that is the situation.

Did you hit some issue related to this?

Cheers!

weaverryan:
Closing - but let me know if you're able to answer my question :). Thanks!

PR #346, opened 2019-01-03
Template generation take care of twig.default_path

### LINKED ISSUE #341: Template created in wrong directory
I changed **twig.default_path** to **%kernel.project_dir%/resources/templates**
maker made index.html.twig into folder **%kernel.project_dir%/templates**

LeJeanbono:
Should close #341

### DIFF
--- a/src/FileManager.php
+++ b/src/FileManager.php
@@ -13,7 +13,6 @@ namespace Symfony\Bundle\MakerBundle;
 
 use Symfony\Bundle\MakerBundle\Util\AutoloaderUtil;
 use Symfony\Component\Console\Style\SymfonyStyle;
-use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
 
@@ -30,16 +29,16 @@ class FileManager
     private $rootDirectory;
     /** @var SymfonyStyle */
     private $io;
-    private $params;
+    private $twigDefaultPath;
 
-    public function __construct(Filesystem $fs, AutoloaderUtil $autoloaderUtil, string $rootDirectory, ParameterBagInterface $params)
+    public function __construct(Filesystem $fs, AutoloaderUtil $autoloaderUtil, string $rootDirectory, string $twigDefaultPath)
     {
         // move FileManagerTest stuff
         // update EntityRegeneratorTest to mock the autoloader
         $this->fs = $fs;
         $this->autoloaderUtil = $autoloaderUtil;
         $this->rootDirectory = rtrim($this->realPath($this->normalizeSlashes($rootDirectory)), '/');
-        $this->params = $params;
+        $this->twigDefaultPath = $twigDefaultPath;
     }
 
     public function setIO(SymfonyStyle $io)
@@ -121,7 +120,7 @@ class FileManager
         return file_get_contents($this->absolutizePath($path));
     }
 
-    public function createFinder(string $in)
+    public function createFinder(string $in): Finder
     {
         $finder = new Finder();
         $finder->in($this->absolutizePath($in));
@@ -177,9 +176,9 @@ class FileManager
         return $this->rootDirectory;
     }
 
-    public function getTemplatesDir(): string
+    public function getPathForTemplate(string $filename): string
     {
-        return $this->params->get('twig.default_path').'/';
+        return $this->twigDefaultPath.'/'.$filename;
     }
 
     /**
--- a/src/FileManager.php
+++ b/src/FileManager.php
@@ -176,7 +176,7 @@ class FileManager
         return $this->rootDirectory;
     }
 
-    public function getTemplatesFolder(): string
+    public function getTemplatesDir(): string
     {
         return $this->twigDefaultPath.'/';
     }
--- a/src/FileManager.php
+++ b/src/FileManager.php
@@ -13,6 +13,7 @@ namespace Symfony\Bundle\MakerBundle;
 
 use Symfony\Bundle\MakerBundle\Util\AutoloaderUtil;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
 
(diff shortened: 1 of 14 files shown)


src/FileManager.php
```diff
@@ -174,6 +176,11 @@ public function getRootDirectory(): string
         return $this->rootDirectory;
     }
 
+    public function getTemplatesFolder(): string
```
javiereguiluz: In Symfony we always use `Dir` instead of `Folder` or `Directory`, so please rename this method to `getTemplatesDir()`. Thanks!

LeJeanbono:
@weaverryan with what can I replace parameter_bag service ? (not public)

src/Resources/config/services.xml
```diff
@@ -11,6 +11,7 @@
                 <argument type="service" id="filesystem" />
                 <argument type="service" id="maker.autoloader_util" />
                 <argument>%kernel.project_dir%</argument>
+                <argument type="service" id="parameter_bag"/>
```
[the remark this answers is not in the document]
LeJeanbono: This is my first try but tests failed to implement the service if `twig.default_path` is not defined, am I wrong ?
https://ci.appveyor.com/project/weaverryan/maker-bundle/builds/21347168

src/FileManager.php
```diff
@@ -29,14 +29,16 @@ class FileManager
...
 
-    public function __construct(Filesystem $fs, AutoloaderUtil $autoloaderUtil, string $rootDirectory)
+    public function __construct(Filesystem $fs, AutoloaderUtil $autoloaderUtil, string $rootDirectory, string $twigDefaultPath)
     {
         // move FileManagerTest stuff
         // update EntityRegeneratorTest to mock the autoloader
         $this->fs = $fs;
         $this->autoloaderUtil = $autoloaderUtil;
         $this->rootDirectory = rtrim($this->realPath($this->normalizeSlashes($rootDirectory)), '/');
+        $this->twigDefaultPath = $twigDefaultPath;
```
weaverryan: There's on problem here: the `%twig.default_path%` that is passed to this method is an absolute URL to your system, not just the string `templates/`. That's a problem (at the very least) because we use this path in some messages - e.g. `Review & adapt the login template... `

I'm not 100% sure, but I think we could just convert this absolute path to the relative path (e.g. `templates`) and set the `twigDefaultPath` property to just that string. I think the rest of the system is perfectly happy just using this relative path, because the FileManager or Generator ultimately converts it to the absolute path later before generation. We could make that conversion using the `relativizePath()` method.

weaverryan:
Ping @LeJeanbono!

I'd love to merge this, but I think there's still just one problem with the absolute Twig path that I mentioned above.

Thanks!

src/FileManager.php
LeJeanbono: Done @weaverryan !
(Sorry for the rebase...)

src/Resources/config/services.xml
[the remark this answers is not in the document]
LeJeanbono: By the way with `<argument>%twig.default_path%</argument>` you will always need twig as requirement even if your project doesn't, isn't it ?

src/Doctrine/EntityRegenerator.php
```diff
@@ -76,6 +76,10 @@ public function regenerateEntities(string $classOrNamespace)
             $embeddedClasses = [];
 
             foreach ($classMetadata->embeddedClasses as $fieldName => $mapping) {
+                if (false !== strpos($fieldName, '.')) {
+                    continue;
+                }
```
weaverryan: What's this for?

.travis.yml
```diff
@@ -22,7 +23,7 @@ env:
         - SYMFONY_PHPUNIT_DIR=.phpunit
 
 before_install:
-    - phpenv config-rm xdebug.ini
+    - phpenv config-rm xdebug.ini || echo "xdebug not available"
```
weaverryan: What's this for?

src/DependencyInjection/Configuration.php
```diff
@@ -22,7 +22,7 @@ class Configuration implements ConfigurationInterface
     public function getConfigTreeBuilder()
     {
         $treeBuilder = new TreeBuilder('maker');
-        if (\method_exists($treeBuilder, 'getRootNode')) {
+        if (method_exists($treeBuilder, 'getRootNode')) {
```
weaverryan: I think the `\` is actually part of the coding standards

src/Maker/MakeAuthenticator.php
```diff
@@ -346,5 +346,10 @@ public function configureDependencies(DependencyBuilder $dependencies, InputInte
             Yaml::class,
             'yaml'
         );
+
+        $dependencies->addClassDependency(
+            TwigBundle::class,
+            'twig-bundle'
+        );
```
weaverryan: I think this isn't right... it's more complicated than this. This command could be used to generate a blank authenticator - no Twig needed there. Twig IS needed if we generate the login form authenticator. We have a few examples in the code of doing this check manually in a command based on user input.

src/Maker/MakeController.php
```diff
@@ -91,6 +99,11 @@ public function configureDependencies(DependencyBuilder $dependencies)
             Annotation::class,
             'annotations'
         );
+
+        $dependencies->addClassDependency(
+            TwigBundle::class,
+            'twig-bundle'
+        );
```
weaverryan: It's not a dependency. IF Twig is installed, we generate a template. If not, we don't.

src/Maker/MakeController.php
```diff
@@ -29,6 +30,13 @@
  */
 final class MakeController extends AbstractMaker
 {
+    private $fileManager;
+
+    public function __construct(FileManager $fileManager)
+    {
+        $this->fileManager = $fileManager;
+    }
```
weaverryan: Is this needed?

src/Maker/MakeEntity.php
```diff
@@ -466,17 +466,21 @@ private function askRelationDetails(ConsoleStyle $io, string $generatedEntityCla
...
+            // in the Entity namespace versus just checking the full class
+            // name to avoid issues with classes like "Directory" that exist
+            // in PHP's core.
+            if (class_exists($this->getEntityNamespace().'\\'.$answeredEntityClass)) {
+                $targetEntityClass = $this->getEntityNamespace().'\\'.$answeredEntityClass;
+            } elseif (class_exists($answeredEntityClass)) {
+                $targetEntityClass = $answeredEntityClass;
+            } else {
+                $io->error(sprintf('Unknown class "%s"', $targetEntityClass));
+                $targetEntityClass = null;
```
weaverryan: What's this stuff? is it from a bad merge?

LeJeanbono:
@weaverryan I just revert my bad rebase, sorry for that...
Still have a problem with tests : 

> The service "maker.file_manager" has a dependency on a non-existent parameter "twig.default_path".

I tried `$c->setParameter('twig.default_path', 'templates');` in `FunctionalTest.php` but no way.

weaverryan:
@LeJeanbono I just took a look at this and pushed a commit. It's a tricky situation and you're the lucky person to first find it :). Basically, we cannot inject `%twig.default_path%` into `FileManager`, because MakerBundle will sometimes be used without Twig installed at all. When this parameter is missing, the service can't be instantiated and nothing works. The fix I did was to continue to inject that parameter, but add a compiler pass to remove it if it doesn't exist (and make the argument optional in `FileManager`).

Let's see if the tests pass :)

weaverryan:
Test failures are unrelated. Thank you @LeJeanbono!

PR #347, opened 2019-01-11
Exception when property not found in given user entity

zholus:
Better message when try to generate authenticator through `make:auth` command.

Before:
image

After:
image


weaverryan:
Nice improvement! Thanks @zholus!

PR #350, opened 2019-01-16
make:entity "Self" or "Parent"

adricorse:
After using ``make:entity Self`` or ``make:entity Parent``, maker-bundle build 2 files :
```
 created: src/Entity/Self.php
 created: src/Repository/SelfRepository.php
```
But, after that :
```
Fatal error: Cannot use 'Self' as class name as it is reserved in /var/www/html/demo/src/Entity/Self.php on line 10
```
 ... and others "Call Stack" to finish with a "CRITICAL".

You must remove ``src/Entity/Self.php`` and ``src/Repository/SelfRepository.php`` to continue.

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -42,8 +42,8 @@ final class Validator
             'static', 'switch', 'throw', 'trait', 'try', 'unset',
             'use', 'var', 'while', 'xor',
             'int', 'float', 'bool', 'string', 'true', 'false', 'null', 'void',
-            'iterable', 'object', '__FILE__', '__LINE__', '__DIR__', '__FUNCTION__', '__CLASS__',
-            '__METHOD__', '__NAMESPACE__', '__TRAIT__', 'self', 'parent',
+            'iterable', 'object', '__file__', '__line__', '__dir__', '__function__', '__class__',
+            '__method__', '__namespace__', '__trait__', 'self', 'parent',
         ];
 
         foreach ($pieces as $piece) {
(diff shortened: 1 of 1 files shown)


src/Validator.php
```diff
@@ -43,7 +43,7 @@ public static function validateClassName(string $className, string $errorMessage
             'use', 'var', 'while', 'xor',
             'int', 'float', 'bool', 'string', 'true', 'false', 'null', 'void',
             'iterable', 'object', '__FILE__', '__LINE__', '__DIR__', '__FUNCTION__', '__CLASS__',
-            '__METHOD__', '__NAMESPACE__', '__TRAIT__',
+            '__METHOD__', '__NAMESPACE__', '__TRAIT__', 'self', 'parent',
```
stof: btw, all these `__*__` reserved keywords are not properly forbidden, because the check uses `in_array(strtolower($shortClassName), $reservedKeywords, true)`. We need to put them in lowercase in this array. Can you also update them ?

adricorse: I tried several constants ( like`` __*__`` ) but, in ``src/Generator.php`` a ``$className`` is replaced by :
```
$className = rtrim($fullNamespacePrefix, '\\').'\\'.Str::asClassName($name, $suffix);
```
Example :
Try ``make:entity __FILE__``, the classname: ``__FILE__`` become ``file`` and is defined in ``$reservedKeywords`` and maker-bundle not build files.
But:
Try ``make:entity __METHOD__``, the classname: ``__METHOD__`` become ``method`` and is not defined in ``$reservedKeywords`` and maker-bundle build files without fatal error, call stack and critical.
I can add ``'file', 'line', 'dir', 'method'`` but I'm not sure about the result.

weaverryan:
Thank you @adricorse!

PR #353, opened 2019-01-17
make:entity __method__ (or line, dir and file)

adricorse:
After trying the latest PR #350 , I think is needed to add these controls.

However, ```'__file__', '__line__', '__dir__', '__function__', '__class__', '__file__', '__line__', '__dir__', '__function__', '__class__', '__method__', '__namespace__', '__trait__'``` can be removed from the array because in ``src/Generator.php`` the ``classname`` is replaced by :
```$className = rtrim($fullNamespacePrefix, '\\').'\\'.Str::asClassName($name, $suffix);```

For example, ``__METHOD__`` become ``method`` and ``method`` dos not exist in array and ``make:entity`` builds 2 files.
But, I don't know if ``$reservedKeywords`` is used elsewhere.

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -43,7 +43,7 @@ final class Validator
             'use', 'var', 'while', 'xor',
             'int', 'float', 'bool', 'string', 'true', 'false', 'null', 'void',
             'iterable', 'object', '__file__', '__line__', '__dir__', '__function__', '__class__',
-            '__method__', '__namespace__', '__trait__', 'self', 'parent', 'file', 'line', 'dir', 'method', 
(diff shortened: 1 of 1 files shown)


weaverryan:
Hey @adricorse!

From the conversation here: https://github.com/symfony/maker-bundle/pull/350#discussion_r248316501

If I run `make:entity __FILE__` or `make:entity __METHOD__`, I don't ever get an error (it converts the classes to FILE and METHOD). So is there any problem at this point?

Cheers!

weaverryan:
Closing - I think there might not be any problem to fix.

Cheers!

PR #355, opened 2019-01-25
MakerBundle doc : use dev file name

jdeniau:
By default, the maker bundle is only active in `dev` environment, so the configuration file should be in the `config/packages/dev/` directory, not directly in the `config/packages/`


weaverryan:
You're right! Thanks Julien!

PR #359, opened 2019-01-31
Added a "Forgotten password" maker

romaricdrigon:
Hello,

Having a password reset feature ("forgotten password") is a common feature of a web application. Maker bundle already has a User registration maker, it makes sense imo to add this too. Otherwise users are left either to code it manually (long), or to ditch Maker commands for FOSUserBundle.

I started implementing such a maker. It is heavily inspired from other Registration maker regarding style, and from FOSUserBundle process. Right now this is a work in progress, but I wanted to propose the idea & to discuss code style.

To do:
 - [x] first working POC
 - [x] add (minimal) view
 - [x] add documentation describing the command
 - [x] add tests - I got those are functional tests over the generated code?
 - [x] squash & rebase everything

Points to be discussed:
 - is the controller right now looking good? I'm trying to have something minimalistic doing the job right, and easily modifiable. Not having too many services/helpers.
 - ~~should the e-mail body be from a template, or does an heredoc looks sufficient to you?~~
 - which options would make sense for the command? for instance, redirecting or authenticating the user after having changed his password?
 - how to handle missing password setter/User class without an e-mail field? Display a warning in console?

Screenshots:

image

image

image

image

Thank you for this great bundle :)

### DIFF
--- a/src/Maker/MakeForgottenPassword.php
+++ b/src/Maker/MakeForgottenPassword.php
@@ -1,0 +1,333 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Doctrine\ORMDependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Security\InteractiveSecurityHelper;
+use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
+use Symfony\Bundle\SecurityBundle\SecurityBundle;
+use Symfony\Bundle\TwigBundle\TwigBundle;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\PasswordType;
+use Symfony\Component\Form\Extension\Core\Type\EmailType;
+use Symfony\Component\Form\Extension\Core\Type\RepeatedType;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Bundle\MakerBundle\FileManager;
+use Symfony\Bundle\MakerBundle\Renderer\FormTypeRenderer;
+use Symfony\Component\Validator\Validation;
+use Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle;
+
+/**
+ * @author Romaric Drigon <romaric.drigon@gmail.com>
+ *
+ * @internal
+ */
+final class MakeForgottenPassword extends AbstractMaker
+{
+    private $fileManager;
+
+    private $formTypeRenderer;
+
+    private $router;
+
+    public function __construct(FileManager $fileManager, FormTypeRenderer $formTypeRenderer, RouterInterface $router)
+    {
+        $this->fileManager = $fileManager;
+        $this->formTypeRenderer = $formTypeRenderer;
+        $this->router = $router;
+    }
+
+    public static function getCommandName(): string
+    {
+        return 'make:forgotten-password';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig)
+    {
+        $command
+            ->setDescription('Creates a "forgotten password" mechanism')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeForgottenPassword.txt'))
+        ;
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
+    {
+        // initialize arguments & commands that are internal (i.e. meant only to be asked)
+        $command
+            ->addArgument('user-class')
+            ->addArgument('email-field')
+            ->addArgument('email-getter')
+            ->addArgument('password-setter')
+        ;
+
+        $interactiveSecurityHelper = new InteractiveSecurityHelper();
+
+        if (!$this->fileManager->fileExists($path = 'config/packages/security.yaml')) {
+            throw new RuntimeCommandException('The file "config/packages/security.yaml" does not exist. This command needs that file to accurately build the forgotten password form.');
+        }
+
+        $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
+        $securityData = $manipulator->getData();
+        $providersData = $securityData['security']['providers'] ?? [];
+
+        $input->setArgument(
+            'user-class',
+            $userClass = $interactiveSecurityHelper->guessUserClass(
+                $io,
+                $providersData,
+                'Enter the User class that should be used with the "forgotten password" feature (e.g. <fg=yellow>App\\Entity\\User</>)'
+            )
+        );
+        $io->text(sprintf('Implementing forgotten password for <info>%s</info>', $userClass));
+
+        $input->setArgument(
+            'email-field',
+            $interactiveSecurityHelper->guessEmailField($io, $userClass)
+        );
+        $input->setArgument(
+            'email-getter',
+            $interactiveSecurityHelper->guessEmailGetter($io, $userClass)
+        );
+        $input->setArgument(
+            'password-setter',
+            $interactiveSecurityHelper->guessPasswordSetter($io, $userClass)
+        );
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies)
+    {
+        // This recipe depends upon Doctrine ORM, to save the token and update the user
+        ORMDependencyBuilder::buildDependencies($dependencies);
+
+        $dependencies->addClassDependency(
+            AbstractType::class,
+            'form'
+        );
+        $dependencies->addClassDependency(
+            Validation::class,
+            'validator'
+        );
+        $dependencies->addClassDependency(
+            TwigBundle::class,
+            'twig-bundle'
+        );
+        $dependencies->addClassDependency(
+            SecurityBundle::class,
+            'security'
+        );
+        $dependencies->addClassDependency(
+            SwiftmailerBundle::class,
+            'mail'
+        );
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $userClass = $input->getArgument('user-class');
+        $userClassNameDetails = $generator->createClassNameDetails(
+            '\\'.$userClass,
+            'Entity\\'
+        );
+        $tokenClassNameDetails = $generator->createClassNameDetails(
+            'PasswordResetToken',
+            'Entity\\'
+        );
+        $repositoryClassNameDetails = $generator->createClassNameDetails(
+            'PasswordResetTokenRepository',
+            'Repository\\'
+        );
+
+        // 1) Create a new "PasswordResetToken" entity and its repository
+        $generator->generateClass(
+            $tokenClassNameDetails->getFullName(),
+            'forgottenPassword/PasswordResetToken.tpl.php',
+            [
+                'repository_class_name' => $repositoryClassNameDetails->getFullName(),
+                'user_class_name' => $userClassNameDetails->getShortName(),
+                'user_full_class_name' => $userClassNameDetails->getFullName(),
+            ]
+        );
+        $generator->generateClass(
+            $repositoryClassNameDetails->getFullName(),
+            'forgottenPassword/PasswordResetTokenRepository.tpl.php',
+            [
+                'token_class_name' => $tokenClassNameDetails->getShortName(),
+                'token_full_class_name' => $tokenClassNameDetails->getFullName(),
+                'user_class_name' => $userClassNameDetails->getShortName(),
+                'user_full_class_name' => $userClassNameDetails->getFullName(),
+            ]
+        );
+
+        // 2) Generate the "request" (email) form class
+        $emailField = $input->getArgument('email-field');
+        $requestFormClassDetails = $this->generateRequestFormClass(
+            $generator,
+            $emailField
+        );
+
+        // 3) Generate the "new password" form class
+        $resettingFormClassDetails = $this->generateResettingFormClass($generator);
+
+        // 4) Generate the controller
+        $controllerClassNameDetails = $generator->createClassNameDetails(
+            'ForgottenPasswordController',
+            'Controller\\'
+        );
+
+        $generator->generateController(
+            $controllerClassNameDetails->getFullName(),
+            'forgottenPassword/ForgottenPasswordController.tpl.php',
+            [
+                'request_form_class_name' => $requestFormClassDetails->getShortName(),
+                'request_form_full_class_name' => $requestFormClassDetails->getFullName(),
+                'resetting_form_class_name' => $resettingFormClassDetails->getShortName(),
+                'resetting_form_full_class_name' => $resettingFormClassDetails->getFullName(),
+                'user_class_name' => $userClassNameDetails->getShortName(),
+                'user_full_class_name' => $userClassNameDetails->getFullName(),
+                'email_field' => $emailField,
+                'email_getter' => $input->getArgument('email-getter'),
+                'password_setter' => $input->getArgument('password-setter'),
+                'login_route' => 'app_login',
+                'token_class_name' => $tokenClassNameDetails->getShortName(),
+                'token_full_class_name' => $tokenClassNameDetails->getFullName(),
+            ]
+        );
+
+        // 5) Generate the "request" template
+        $generator->generateFile(
+            'templates/forgotten_password/request.html.twig',
+            'forgottenPassword/twig_request.tpl.php',
+            [
+                'email_field' => $emailField,
+            ]
+        );
+
+        // 6) Generate the reset e-mail template
+        $generator->generateFile(
+            'templates/forgotten_password/email.txt.twig',
+            'forgottenPassword/twig_email.tpl.php',
+            []
+        );
+
+        // 7) Generate the "checkEmail" template
+        $generator->generateFile(
+            'templates/forgotten_password/check_email.html.twig',
+            'forgottenPassword/twig_check_email.tpl.php',
+            []
+        );
+
+        // 8) Generate the "reset" template
+        $generator->generateFile(
+            'templates/forgotten_password/reset.html.twig',
+            'forgottenPassword/twig_reset.tpl.php',
+            []
+        );
+
+        $generator->writeChanges();
+        $this->writeSuccessMessage($io);
+
+        $io->text('Done! A new entity was added: PasswordResetToken. You should now generate a migration (make:migration) and run it to update your database.');
+        $io->text('Next: Please review ForgottenPasswordController. Then you can add a link to "app_forgotten_password_request" path anywhere you like, typically below your login form!');
+    }
+
+    private function generateRequestFormClass(Generator $generator, string $emailField)
+    {
+        $formClassDetails = $generator->createClassNameDetails(
+            'PasswordRequestFormType',
+            'Form\\'
+        );
+
+        $formFields = [
+            $emailField => [
+                'type' => EmailType::class,
+                'options_code' => <<<EOF
+                'constraints' => [
+                    new NotBlank([
+                        'message' => 'Please enter your $emailField',
+                    ]),
+                ],
+EOF
+            ],
+        ];
+
+        $this->formTypeRenderer->render(
+            $formClassDetails,
+            $formFields,
+            null,
+            [
+                'Symfony\Component\Validator\Constraints\NotBlank',
+            ]
+        );
+
+        return $formClassDetails;
+    }
+
+    private function generateResettingFormClass(Generator $generator)
+    {
+        $formClassDetails = $generator->createClassNameDetails(
+            'PasswordResettingFormType',
+            'Form\\'
+        );
+
+        $formFields = [
+            'plainPassword' => [
+                'type' => RepeatedType::class,
+                'options_code' => <<<EOF
+                'type' => PasswordType::class,
+                'first_options' => [
+                    'constraints' => [
+                        new NotBlank([
+                            'message' => 'Please enter a password',
+                        ]),
+                        new Length([
+                            'min' => 6,
+                            'minMessage' => 'Your password should be at least {{ limit }} characters',
+                            // max length allowed by Symfony for security reasons
+                            'max' => 4096,
+                        ]),
+                    ],
+                    'label' => 'New password',
+                ],
+                'second_options' => [
+                    'label' => 'Repeat Password',
+                ],
+                'invalid_message' => 'The password fields must match.',
+                // Instead of being set onto the object directly,
+                // this is read and encoded in the controller
+                'mapped' => false,
+EOF
+            ],
+        ];
+
+        $this->formTypeRenderer->render(
+            $formClassDetails,
+            $formFields,
+            null,
+            [
+                'Symfony\Component\Validator\Constraints\Length',
+                'Symfony\Component\Validator\Constraints\NotBlank',
+            ],
+            [
+                PasswordType::class,
+            ]
+        );
+
+        return $formClassDetails;
+    }
+}
--- a/src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
+++ b/src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
@@ -1,0 +1,149 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+use <?= $token_full_class_name ?>;
+use <?= $user_full_class_name ?>;
+use <?= $request_form_full_class_name ?>;
+use <?= $resetting_form_full_class_name ?>;
+use Symfony\Bundle\FrameworkBundle\Controller\<?= $parent_class_name; ?>;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+
+/**
+ * @Route("/forgotten-password")
+ */
+class <?= $class_name; ?> extends <?= $parent_class_name; ?><?= "\n" ?>
+{
+    private const SESSION_TOKEN_KEY = 'forgotten_password_token';
+    private const SESSION_CAN_CHECK_EMAIL = 'forgotten_password_check_email';
+
+    /**
+     * @Route("/request", name="app_forgotten_password_request")
+     */
+    public function request(Request $request, \Swift_Mailer $mailer): Response
+    {
+        $form = $this->createForm(<?= $request_form_class_name ?>::class);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $user = $this->getDoctrine()->getRepository(<?= $user_class_name ?>::class)->findOneBy([
+                '<?= $email_field ?>' => $form->get('<?= $email_field ?>')->getData(),
+            ]);
+
+            // Needed to be able to access next page, app_check_email
+            $request->getSession()->set(self::SESSION_CAN_CHECK_EMAIL, true);
+
+            // Do not reveal whether an user account was found or not.
+            if (!$user) {
+                return $this->redirectToRoute('app_check_email');
+            }
+
+            // If User already has a valid Token, we don't want to generate a new one.
+            // We fail silently.
+            $oldTokens = $this->getDoctrine()->getRepository(<?= $token_class_name ?>::class)->findNonExpiredForUser($user);
+            if (count($oldTokens)) {
+                return $this->redirectToRoute('app_check_email');
+            }
+
+            // Generate a reset password token, that the user could use to change their password.
+            $resetPasswordToken = new <?= $token_class_name ?>($user);
+            $this->getDoctrine()->getManager()->persist($resetPasswordToken);
+            $this->getDoctrine()->getManager()->flush();
+
+            $message = (new \Swift_Message('Your password reset request'))
+                ->setFrom(['noreply@mydomain.com' => 'Noreply'])
+                ->setTo($user-><?= $email_getter ?>())
+                ->setBody($this->renderView('forgotten_password/email.txt.twig', [
+                    'token' => $resetPasswordToken,
+                ]))
+            ;
+            $mailer->send($message);
+
+            return $this->redirectToRoute('app_check_email');
+        }
+
+        return $this->render('forgotten_password/request.html.twig', [
+            'requestForm' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @Route("/check-email", name="app_check_email")
+     */
+    public function checkEmail(SessionInterface $session)
+    {
+        // We prevent users from directly accessing this page
+        if (!$session->get(self::SESSION_CAN_CHECK_EMAIL)) {
+            return $this->redirectToRoute('app_forgotten_password_request');
+        }
+
+        $session->remove(self::SESSION_CAN_CHECK_EMAIL);
+
+        return $this->render('forgotten_password/check_email.html.twig', [
+            'tokenLifetime' => <?= $token_class_name ?>::LIFETIME_HOURS,
+        ]);
+    }
+
+    /**
+     * @Route("/reset/{tokenAndSelector}", name="app_reset_password")
+     */
+    public function reset(Request $request, UserPasswordEncoderInterface $passwordEncoder, $tokenAndSelector = null): Response
+    {
+        if ($tokenAndSelector) {
+            // We store token in session and remove it from the URL,
+            // to avoid any leak if someone get to know the URL (AJAX requests, Analytics...).
+            $request->getSession()->set(self::SESSION_TOKEN_KEY, $tokenAndSelector);
+
+            return $this->redirectToRoute('app_reset_password');
+        }
+
+        $tokenAndSelector = $request->getSession()->get(self::SESSION_TOKEN_KEY);
+        if (!$tokenAndSelector) {
+            throw $this->createNotFoundException();
+        }
+
+        $passwordResetToken = $this->getDoctrine()->getRepository(PasswordResetToken::class)->findOneBy([
+            'selector' => substr($tokenAndSelector, 0, PasswordResetToken::SELECTOR_LENGTH),
+        ]);
+
+        if (!$passwordResetToken) {
+            throw $this->createNotFoundException();
+        }
+
+        if ($passwordResetToken->isExpired() || !$passwordResetToken->isTokenEquals(substr($tokenAndSelector, PasswordResetToken::SELECTOR_LENGTH))) {
+            $this->getDoctrine()->getManager()->remove($passwordResetToken);
+            $this->getDoctrine()->getManager()->flush();
+
+            throw $this->createNotFoundException();
+        }
+
+        $form = $this->createForm(<?= $resetting_form_class_name ?>::class);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            // A PasswordResetToken should be used only once, remove it.
+            $this->getDoctrine()->getManager()->remove($passwordResetToken);
+
+            // Encode the plain password, and set it.
+            $passwordResetToken->getUser()-><?= $password_setter ?>(
+                $passwordEncoder->encodePassword(
+                    $passwordResetToken->getUser(),
+                    $form->get('plainPassword')->getData()
+                )
+            );
+
+            $this->getDoctrine()->getManager()->flush();
+
+            // TODO: please check the login route
+            return $this->redirectToRoute('<?= $login_route ?>');
+        }
+
+        return $this->render('forgotten_password/reset.html.twig', [
+            'resetForm' => $form->createView(),
+        ]);
+    }
+}
--- a/src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
+++ b/src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
@@ -1,0 +1,87 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace ?>;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity(repositoryClass="<?= $repository_class_name ?>")
+ */
+class <?= $class_name ?>
+
+{
+    const LIFETIME_HOURS = 24;
+    const SELECTOR_LENGTH = 20; // in chars
+
+    /**
+     * @ORM\Id()
+     * @ORM\GeneratedValue()
+     * @ORM\Column(type="integer")
+     */
+    private $id;
+
+    /**
+     * @ORM\Column(type="string")
+     */
+    private $selector;
+
+    /**
+     * @ORM\Column(type="string")
+     */
+    private $token;
+
+    /**
+     * @ORM\Column(type="datetime_immutable")
+     */
+    private $requestedAt;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="<?= $user_full_class_name ?>")
+     */
+    private $user;
+
+    private $plainToken;
+
+    public function __construct(User $user)
+    {
+        $this->requestedAt = new \DateTimeImmutable('now');
+        $this->selector = strtr(base64_encode(random_bytes(self::SELECTOR_LENGTH * 3 / 4)), '+/', '-_');
+        $this->plainToken = strtr(base64_encode(random_bytes(18)), '+/', '-_');
+        $this->token = password_hash($this->plainToken, PASSWORD_DEFAULT);
+        $this->user = $user;
+    }
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function getAsString(): string
+    {
+        if (!$this->selector || !$this->plainToken) {
+            throw new \Exception('You can get PasswordResetToken as a string only immediately after creation.');
+        }
+
+        return $this->selector.$this->plainToken;
+    }
+
+    public function getUser(): <?= $user_class_name ?>
+
+    {
+        return $this->user;
+    }
+
+    public function isTokenEquals(string $token): bool
+    {
+        return password_verify($token, $this->token);
+    }
+
+    public function isExpired(): bool
+    {
+        if (($this->requestedAt->getTimestamp() + self::LIFETIME_HOURS * 3600) <= time()) {
+            return true;
+        }
+
+        return false;
+    }
+}
--- a/src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
+++ b/src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
@@ -1,0 +1,6 @@
+<?= $helper->getHeadPrintCode('Check your e-mail address'); ?>
+
+{% block body %}
+    <p>An email has been sent. It contains a link you must click to reset your password. This link will expire in {{ tokenLifetime }} hours.</p>
+<p>If you don't get an email please check your spam folder or <a href="{{ path('app_forgotten_password_request') }}">try again</a>.</p>
+{% endblock %}
--- a/src/Resources/skeleton/forgottenPassword/twig_request.tpl.php
+++ b/src/Resources/skeleton/forgottenPassword/twig_request.tpl.php
@@ -1,0 +1,11 @@
+<?= $helper->getHeadPrintCode('Recover your password'); ?>
+
+{% block body %}
+    <h1>Recover your password</h1>
+
+    {{ form_start(requestForm) }}
+        {{ form_row(requestForm.<?= $email_field ?>) }}
+
+        <button class="btn btn-primary">Send e-mail</button>
+    {{ form_end(requestForm) }}
+{% endblock %}
--- a/src/Resources/skeleton/forgottenPassword/twig_reset.tpl.php
+++ b/src/Resources/skeleton/forgottenPassword/twig_reset.tpl.php
@@ -1,0 +1,11 @@
+<?= $helper->getHeadPrintCode('Reset your password'); ?>
+
+{% block body %}
+    <h1>Reset your password</h1>
+
+    {{ form_start(resetForm) }}
+        {{ form_row(resetForm.plainPassword) }}
+
+        <button class="btn btn-primary">Reset password</button>
+    {{ form_end(resetForm) }}
+{% endblock %}
--- a/src/Renderer/FormTypeRenderer.php
+++ b/src/Renderer/FormTypeRenderer.php
@@ -27,7 +27,7 @@ final class FormTypeRenderer
         $this->generator = $generator;
     }
 
-    public function render(ClassNameDetails $formClassDetails, array $formFields, ClassNameDetails $boundClassDetails = null, array $constraintClasses = [])
+    public function render(ClassNameDetails $formClassDetails, array $formFields, ClassNameDetails $boundClassDetails = null, array $constraintClasses = [], array $extraUseClasses = [])
     {
         $fieldTypeUseStatements = [];
         $fields = [];
@@ -49,7 +49,7 @@ final class FormTypeRenderer
                 'bounded_full_class_name' => $boundClassDetails ? $boundClassDetails->getFullName() : null,
                 'bounded_class_name' => $boundClassDetails ? $boundClassDetails->getShortName() : null,
                 'form_fields' => $fields,
-                'field_type_use_statements' => $fieldTypeUseStatements,
+                'field_type_use_statements' => array_merge($fieldTypeUseStatements, $extraUseClasses),
                 'constraint_use_statements' => $constraintClasses,
             ]
         );
--- a/src/Resources/config/makers.xml
+++ b/src/Resources/config/makers.xml
@@ -42,6 +42,13 @@
                 <tag name="maker.command" />
             </service>
 
+            <service id="maker.maker.make_forgotten_password" class="Symfony\Bundle\MakerBundle\Maker\MakeForgottenPassword">
+                <argument type="service" id="maker.file_manager" />
+                <argument type="service" id="maker.renderer.form_type_renderer" />
+                <argument type="service" id="router" />
+                <tag name="maker.command" />
+            </service>
+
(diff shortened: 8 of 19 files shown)


romaricdrigon:
Views were added, and the POC if fully functional (manually tested).
From now on I will stop amending my first commit, if people want to review/comment code.
I'm considering making the "request password" form (first screenshot) work only by e-mail address, as recommended [here](https://www.troyhunt.com/everything-you-ever-wanted-to-know/).

sstok:
I strongly suggest you read this 👍 https://paragonie.com/blog/2016/09/untangling-forget-me-knot-secure-account-recovery-made-simple

Almost all password reset systems are "horrible" insecure, tld;dr explanation: by using a single token (basically a random username) it's possible for an attack to guess the existing token by using a timing attack. When you provide a token, the database will use the index to look for an existing record, each character is checked against a binary-tree until it finds a record. When a character has a hit this means the time takes longer (because it continues the traversing of the tree), when there is no hit the time takes shorter because it bails out. Now using this technique an attack can use this information to try as many combinations as possible until all characters matches!

This is similar to a brute-force attack to guess a username/ID, but with a major difference. There is no password, and therefor access is directly granted.

By using the SplitToken you have selector (random username) and a verifier (random password in a hashed format), meaning that an attacker first needs to guess the selector, and then when there is a positive match, try to guess the verifier. BUT! Once you provide a wrong verifier for the selector, you simple revoke the token and no further tries are possible, even then guessing the verifier is nearly impossible as you compare the hashes in constant time (always takes the same amount of time) which leaks no timing information, making guessing nearly impossible.

I actually created a library that uses this technique https://github.com/rollerworks/split-token (MPLv2.0 licensed because it depends on HiddenString) but for most applications using the technique descried in the linked article is already a good step.

gsylvestre:
This PR is a really good idea, and would be very usefull. I tried it and I love very much the generated code. 

Some really small problems: 

- The generated twig files and folders do not follow coding stardards in their camelCased form (they should be snake_cased).
- In the controller, method `request()`, is there any reason that you inject `Twig_Environment ` instead of using the simpler approch of `$this->renderView()` ?
- The suffix Action should be removed on controller's method `checkEmailAction()`.

Also, I understand completely that you want to keep things simple and basic, and I agree. I just wonder if we could kinda incorporate some of @sstok suggestions by passing along the user email in the reset link. _That would allow us to hash the reset token in the database_, without adding much generated code and no other field in the entity. We even already use the `UserPasswordEncoderInterface ` in the `reset()` method...

gsylvestre:
In fact, maybe that those should be command options: 

The reset password system need to store random tokens for your user. Do you want to store those:
1. In your User entity
2. In a separate entity

Would you like to hash the stored tokens (more secure)?
1. Yes
2. No

Would you like to add an extra "selector" field to protect against timed attack (more secure)?
1. Yes
2. No

Obviously, that is a lot of work to imlement, but that keep things simple for the end user.

romaricdrigon:
Thank you for the review @gsylvestre 
I fixed the small problems, regarding `\Twig_Environment` at some point I wanted to render blocks independently, but I dropped this approach as it uses internal methods.

Regarding going further, I'm indecisive for now. I would like to get more feedbacks on this PR before going further regarding advanced features, about if and how it can integrates Maker bundle.

sstok:
FYI. I don't have any plans for a bundle to integrate SplitToken, in practice you can register as many SplitToken factories as you want. 

The only thing that might be interesting is a FormType to handle this format native, I actually did this for [Park-Manager](https://github.com/park-manager/park-manager/blob/333808b5d0e1eb495ab8b0be380eec432503b389/src/Module/CoreModule/Infrastructure/UserInterface/Web/Form/Type/Security/SplitTokenType.php) . But that's something for later as it's still experimental.

weaverryan:
Sorry for the delay. I really want this - and it's on my list to review - but haven't had time yet. Thanks for getting it started!

weaverryan:
@gsylvestre I was interested in getting more info from your above comments:

> In fact, maybe that those should be command options:

> The reset password system need to store random tokens for your user. Do you want to store those:
> 1. In your User entity
> 2. In a separate entity

I don't think we should do this - that's a level of customization that you can do if you want to, but I don't see any advantage to adding this flexibility (or is there a security implication?)

> Would you like to hash the stored tokens (more secure)?
> 1. Yes
> 2. No

I saw your other comment before this, but can you explain this a bit more? If we only had 1 field in the database, how is the token hashed and how does that improve security?

> Would you like to add an extra "selector" field to protect against timed attack (more secure)?
> 1. Yes
> 2. No

Maybe ;). I do think this is something to keep in mind as we get closer - it's probably not *that* much more work to implement, if indeed this is an industry-standard way of making more secure reset passwords.

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,320 @@
...
+    public static function getCommandName(): string
+    {
+        return 'make:forgotten-password';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig)
+    {
+        $command
+            ->setDescription('Creates a "forgotten password" mechanism')
+            //->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeRegistrationForm.txt'))
```
weaverryan: Don't forget to add this - I usually keep them very short :)

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,320 @@
...
+        $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
+        $securityData = $manipulator->getData();
+        $providersData = $securityData['security']['providers'] ?? [];
+
+        $input->setArgument(
+            'user-class',
+            $userClass = $interactiveSecurityHelper->guessUserClass(
+                $io,
+                $providersData,
+                'Enter the User class which should be able to use "forgotten password" feature (e.g. <fg=yellow>App\\Entity\\User</>)'
```
weaverryan: > Enter the User class that should be used with the "forgot password" feature...

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,320 @@
...
+        $io->text(sprintf('Implementing forgotten password for <info>%s</info>', $userClass));
+
+        $input->setArgument(
+            'username-field',
+            $interactiveSecurityHelper->guessUserNameField($io, $userClass, $providersData)
+        );
+        $input->setArgument(
+            'password-field',
+            $interactiveSecurityHelper->guessPasswordField($io, $userClass)
+        );
```
weaverryan: Because we're only using this to guess the setter, we could instead just guess the setter name and, if it's not found, ask the user to literally tell us the method name. For example, maybe we look for a `setPassword()` method. If that's found, we just use it. If it's not found, we ask "Enter the setter method on your User class that's used to set the encoded password (e.g. setPassword())".

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,320 @@
...
+        $generator->generateFile(
+            'templates/forgottenPassword/reset.html.twig',
+            'forgottenPassword/twig_reset.tpl.php',
+            []
+        );
+
+        $generator->writeChanges();
+        $this->writeSuccessMessage($io);
+
+        $io->text('Done! You should now generate a migration (make:mig) and run it to update your database.');
```
weaverryan: Use the full `make:migration` here

We probably also should add a line here explaining that two new fields were added to their entity:

> Two new fields were added to your User entity: resetPasswordToken and passwordResetRequestedAt.

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,320 @@
...
+        $userManipulator->setIo($io);
+        $userManipulator->addEntityField(
+            'resetPasswordToken',
+            [
+                'type' => 'string',
+                'nullable' => true,
+            ]
+        );
+        $userManipulator->addEntityField(
+            'passwordRequestedAt',
```
weaverryan: passwordResetRequestedAt

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,118 @@
...
+            // do not reveal whether an user account was found or not
+            if (!$user) {
+                return $this->redirectToRoute('app_check_email');
+            }
+
+            // generate a reset password token, the user could use to change his password
+            $resetPasswordToken =$this->generateConfirmationToken();
+
+            $user->setResetPasswordToken($resetPasswordToken);
+            $user->setPasswordRequestedAt(new \DateTime());
```
weaverryan: This is probably a spot where I'd prefer to not generate setter methods for the 2 new columns, and instead have something like: `$user->generatePasswordResetToken()` (which we would of course generate for them) that takes care of setting both of these fields.

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,320 @@
...
+        // 4) Generate the controller
+        $controllerClassNameDetails = $generator->createClassNameDetails(
+            'ForgottenPasswordController',
+            'Controller\\'
+        );
+
+        // We assume User entity has some password setter, and e-mail getter
+        // TODO: display a warning if not?
+        $passwordSetter = 'set'.ucfirst($input->getArgument('password-field'));
+        $emailGetter = 'getEmail';
```
weaverryan: I *am* kinda thinking that we should make this work for "email address" only - as you suggested. So, instead of asking the user for their "username" field, we ask them for their `email` field, then use that everywhere. We should probably do some validation that the getter & setter exist for that field, and throw an error if it does not (if people don't like that, and also want the command to ask them for a custom getter & setter, we can handle that in the future IF people ask for it - not now).

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,118 @@
...
+    public function reset(Request $request, $token, UserPasswordEncoderInterface $passwordEncoder): Response
+    {
+        $user = $this->getDoctrine()->getRepository(<?= $user_class_name ?>::class)->findOneBy([
+            'resetPasswordToken' => $token,
+        ]);
+
+        if (!$user
+            || !$user->getPasswordRequestedAt()
+            || ($user->getPasswordRequestedAt()->getTimestamp()+self::TOKEN_TTL_HOURS*3600) <= (new \DateTime())->getTimestamp()
+        ) {
```
weaverryan: I also think it might be nicer to have a method on the User here too - like `isResetPasswordTokenExpired()`

weaverryan:
Oh, I understand now about the hashed tokens: if we passed the email address in the reset URL, then we could look up the User by that email address, and then do a timing-safe check to see if the hash in the URL matches what we expect in the database.

Of course, then this *also* in theory allows for a timing attack on that "reset" URL, because you could use it to detect timing differences in return to determine if the email you entered is an email in the system :).

In general, I think we should be looking at other popular implementations (including FOSUserBundle, Laravel, Rails, whatever) and making sure we have the same safeguards at least as they do.

gsylvestre:
@weaverryan yes that's what I meant: if we hash the token in the database, we then need some kind of identifier to retrieve it. As you say, if we use the email, then an attacker can use timed attacks to find out if it's in the db... Really not a big concern imo, but still.

I was not proposing the "separate entity" option for security implication, not directly. But now that I have looked around and thought about it, I believe that we should always store the token in a new entity, instead of inside properties in the user class. See below...

Something like : 
```
class PasswordResetToken 
{
    $id
    $selector //(crypto secure string, shorter)
    $token //(crypto secure string, longer)
    $requestedAt //(datetime)
    $user //(OneToOne unidir relation)
}
```

Pros:
- We do not mess with User class by adding new properties and methods. It is often already clogged, and I feel it is intrusive.
- We can then more freely add more fields and methods to our Token entity, like the generated "selector", or ie. if the developper want, he/she can add an expireAt field easily, or maybe use this entity for email confirmation purposes.
- Indirectly more secure because we can now add the "selector" field
- Until a password reset request is made, no token are generated. And no fields are sitting there being NULL and sad in the user table.
- Better model structure, added flexibility

Cons:
- Extra entity and database table are generated
- more complicated?

In fact, is it really more complicated for the developper to understand what is going on? I believe even beginners will get this new entity easily. 
And we are doing a really secure password reset. Why settle for less? We expect Symfony to provide top secure features. 

BTW, Laravel is storing reset tokens in a new table, but use the email as the selector. 

I can implement theses changes next week, if the coding work is a problem.

romaricdrigon:
Thank you for your inputs both of you, and for your review @weaverryan, that's great to see things moving. Let's make this maker happen!

@gsylvestre I like that last idea you had, having a separate entity. It is not making things overly complex, but it both adds some security from what I got & code is a little bit cleaner in my opinion (it prevents `User` entity from growing too big...). I can work on it while fixing the other comments, but I would appreciate your review.

sstok:
> Of course, then this also in theory allows for a timing attack on that "reset" URL, because you could use it to detect timing differences in return to determine if the email you entered is an email in the system :).

See the linked article for full details ;) https://paragonie.com/blog/2016/09/untangling-forget-me-knot-secure-account-recovery-made-simple

The main focus is that you have two factors for authenticating, a selector (like a username, email address or random token) and a verifier that acts like a password (with the same security requirements for verifying). Using only a random value (token) as-is (as the FOSUserBundle is doing) makes that your token is nothing more then a random id that can be guessed, I can't make it more clear than this 👍

src/Maker/MakeForgottenPassword.php
romaricdrigon: Fixed, thanks!

src/Maker/MakeForgottenPassword.php
romaricdrigon: Ok, and it makes sense since reset password _requires_ an e-mail field. Modifying the request action to accept usernames too should be easy enough, I may add a comment in the generated code.

romaricdrigon:
I pushed commits regarding the first review comments.
On top of that, I prefixed both form classes names by `Password`, I find `PasswordRequestFormType` and `PasswordResettingFormType` to be more explicit.

Now I will work on having a separate entity, and then on testing everything.

src/Maker/MakeForgottenPassword.php
romaricdrigon: Agreed, I did that, and for e-mail getter too. I did not add `guessPasswordSetter` and `guessEmailGetter` to `InteractiveSecurityHelper`, because their re-usability would be limited.

stof:
Note that if I were building FOSUserBundle today, I would probably use a separate table to store the reset tokens instead of putting things in the User class.

@sstok the issue of using the email address as the selector is that it allows using a timing attack to identify whether the email exists in the DB, allowing to leak the user base. That's why using a random selector instead is better if you try to protect the system against timing attacks.
Another option is to use the primary key of the user table...

gsylvestre:
@stof, would you rather go with the random selector or the primary key? 
I feel like the id would _seem_ unsafe (for users and developpers to see the easily guessable user id in the url), even if it's mostly the same security level... It is one less field to generate on the other hand.

romaricdrigon:
I feel like having an extra "selector" field is not that bad, so for now I'm going with a selector separate from the ID.
One issue I'm having though: how to hash "token" without hardcoding a reference to the hash algorithm & without adding a dependency to an extra library? 
For now, I'm considering getting the "UserPasswordEncoder", it feels a little bit hacky though.

romaricdrigon:
Hello all,
I could make some good progress on the code today, the feature is finished & some unit tests were added.
I'm now working on functional tests, those are pretty hard to get started on. @weaverryan I get you bootstrapped everything, do you have any tips on how to run those locally?

romaricdrigon:
CI is green 🎉 
I added half a dozen functional tests on major cases (password reset email is sent, wrong token, expired token...). I'm waiting to hear from you now now, ping @weaverryan :)

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,311 @@
...
+            ],
+        ];
+
+        $this->formTypeRenderer->render(
+            $formClassDetails,
+            $formFields,
+            $userClassDetails,
+            [
+                'Symfony\Component\Validator\Constraints\NotBlank',
+                'Symfony\Component\Validator\Constraints\Length',
```
bocharsky-bw: Should be before `NotBlank` to be ordered alphabetically

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,311 @@
...
+    private function generateRequestFormClass(Generator $generator, string $emailField)
+    {
+        $formClassDetails = $generator->createClassNameDetails(
+            'PasswordRequestFormType',
+            'Form\\'
+        );
+
+        $formFields = [
+            $emailField => [
+                'type' => TextType::class,
```
bocharsky-bw: Why not `EmailType::class`?

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,311 @@
...
+    private function generateResettingFormClass(ClassNameDetails $userClassDetails, Generator $generator)
+    {
+        $formClassDetails = $generator->createClassNameDetails(
+            'PasswordResettingFormType',
+            'Form\\'
+        );
+
+        $formFields = [
+            'plainPassword' => [
+                'type' => PasswordType::class,
```
bocharsky-bw: Don't we want to use `RepeatedType` here?

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,112 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+use <?= $user_full_class_name ?>;
+use <?= $token_full_class_name ?>;
```
bocharsky-bw: This should be the first for the correct order

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,112 @@
...
+use <?= $request_form_full_class_name ?>;
+use <?= $resetting_form_full_class_name ?>;
+use Symfony\Bundle\FrameworkBundle\Controller\<?= $parent_class_name; ?>;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+
```
bocharsky-bw: I would add `@Route("/forgotten-password")` to prefix all the routes here

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,112 @@
...
+            'requestForm' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @Route("/check-email", name="app_check_email")
+     */
+    public function checkEmail()
+    {
+        return $this->render('forgottenPassword/check_email.html.twig', [
```
bocharsky-bw: This route is always available, so you open it even if you did not request password resetting. What about to check referer for it? Something like:
```php
        // Redirect if user does not come from request reset password page
        $refererUrl = $request->headers->get('referer');
        $requestUrl = $this->generateUrl('app_forgotten_password');
        if (!$refererUrl) {
            return $this->redirect($requestUrl);
        }
        // It's assumed we didn't have any query params on the Request page,
        // otherwise we need a more clever solution here like using parse_url()
        if ($requestUrl !== substr($refererUrl, -strlen($requestUrl))) {
            return $this->redirect($requestUrl);
        }

       return $this->render(...)
```

So, only users who were redirected from `app_forgotten_password` would be able to see this page

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,112 @@
...
+            $message = (new \Swift_Message('Your password reset request'))
+                ->setFrom(['noreply@mydomain.com' => 'Noreply'])
+                ->setTo($user-><?= $email_getter ?>())
+                ->setBody($this->renderView('forgottenPassword/email.txt.twig', [
+                    'token' => $resetPasswordToken,
+                ]))
+            ;
+            $mailer->send($message);
+
+            $this->getDoctrine()->getManager()->flush();
```
bocharsky-bw: Any reasons to flush it after the email was sent instead of doing it right after we created the token?

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,112 @@
...
+
+            throw $this->createNotFoundException();
+        }
+
+        $form = $this->createForm(<?= $resetting_form_class_name ?>::class);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            // a PasswordResetToken should be used only once, remove it
+            $this->getDoctrine()->getManager()->remove($passwordResetToken);
```
bocharsky-bw: Is it safe to use an entity further after `remove()` call?

src/Resources/skeleton/forgottenPassword/twig_request.tpl.php
```diff
@@ -0,0 +1,11 @@
+<?= $helper->getHeadPrintCode('Recover your password'); ?>
+
+{% block body %}
+    <h1>Recover your password</h1>
+
+    {{ form_start(requestForm) }}
+        {{ form_row(requestForm.<?= $email_field ?>) }}
+
+        <button class="btn">Send e-mail</button>
```
bocharsky-bw: Let's use `btn btn-primary` class, otherwise it looks weird

src/Resources/skeleton/forgottenPassword/twig_reset.tpl.php
```diff
@@ -0,0 +1,11 @@
+<?= $helper->getHeadPrintCode('Reset your password'); ?>
+
+{% block body %}
+    <h1>Reset your password</h1>
+
+    {{ form_start(resetForm) }}
+        {{ form_row(resetForm.plainPassword) }}
+
+        <button class="btn">Reset password</button>
```
bocharsky-bw: Same here, let's use both classes from TB, e.g. `btn btn-primary`

src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
```diff
@@ -0,0 +1,6 @@
+<?= $helper->getHeadPrintCode('Check your e-mail address'); ?>
+
+{% block body %}
+    <p>An email has been sent. It contains a link you must click to reset your password. This link will expires in {{ tokenLifetime }} hours.</p>
+    <p>If you don't get an email please check your spam folder or try again.</p>
```
bocharsky-bw: And we can wrap "try again" with a link to `app_forgotten_password` route for convenience

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        if (!$this->plainToken) {
+            throw new \Exception('You can get PasswordResetToken as a string only immediately after creation.');
+        }
+
+        return $this->selector.$this->plainToken;
+    }
+
+    public function getUser(): <?= $user_class_name ?>
+
```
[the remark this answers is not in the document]
romaricdrigon: You have such extra lines at many points in that file. The `PasswordResetToken` generated file is fine though, there may be an issue in the code generator & that's the way around I found. I couldn't quite dig into that, but find that hack.

src/Resources/skeleton/forgottenPassword/twig_request.tpl.php
```diff
@@ -0,0 +1,11 @@
+<?= $helper->getHeadPrintCode('Recover your password'); ?>
+
+{% block body %}
+    <h1>Recover your password</h1>
+
+    {{ form_start(requestForm) }}
+        {{ form_row(requestForm.<?= $email_field ?>) }}
```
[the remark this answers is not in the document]
romaricdrigon: I have mixed feeling there. Result will be the same, but it will be easier for developers customize the label with `form_row`. If we use `form_widget`, it may create a trap for newcomers who will modify the form label or the form theme.

src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
```diff
@@ -0,0 +1,6 @@
+<?= $helper->getHeadPrintCode('Check your e-mail address'); ?>
+
+{% block body %}
+    <p>An email has been sent. It contains a link you must click to reset your password. This link will expires in {{ tokenLifetime }} hours.</p>
```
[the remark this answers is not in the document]
romaricdrigon: 😮 fixed, thank you!

src/Resources/skeleton/forgottenPassword/twig_reset.tpl.php
```diff
@@ -0,0 +1,11 @@
+<?= $helper->getHeadPrintCode('Reset your password'); ?>
+
+{% block body %}
+    <h1>Reset your password</h1>
+
+    {{ form_start(resetForm) }}
+        {{ form_row(resetForm.plainPassword) }}
```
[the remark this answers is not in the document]
romaricdrigon: Same point, then I may change both for consistency

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    private $selector;
+
+    /**
+     * @ORM\Column(type="string")
+     */
+    private $token;
+
+    /**
+     * @ORM\Column(type="datetime")
```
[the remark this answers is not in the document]
romaricdrigon: If we use `datetime_immutable`, we need to bump `doctrine/orm` required version to `~2.6`. I thought about it, but I thought it may be too restrictive :S

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
```diff
@@ -0,0 +1,87 @@
...
+    private $user;
+
+    private $plainToken;
+
+    public function __construct(User $user)
+    {
+        $this->requestedAt = new \DateTimeImmutable('now');
+        $this->selector = strtr(base64_encode(random_bytes(self::SELECTOR_LENGTH * 3 / 4)), '+/', '-_');
+        $this->plainToken = strtr(base64_encode(random_bytes(18)), '+/', '-_');
+        $this->token = password_hash($this->plainToken, PASSWORD_BCRYPT);
```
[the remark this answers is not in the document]
romaricdrigon: I think it requires PHP 7.2, doesn't it? So we would loose compatibility with PHP 7.0 and 7.1, that's a blocker :S

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->user = $user;
+    }
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function getAsString(): string
```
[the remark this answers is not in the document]
romaricdrigon: Usually I prefer to stay away from magic method, for something more explicit, but that could do. Any third opinion, maybe?

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        if (!$this->plainToken) {
```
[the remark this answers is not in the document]
romaricdrigon: Agreed

src/Maker/MakeForgottenPassword.php
romaricdrigon: Totally, good catch

src/Maker/MakeForgottenPassword.php
romaricdrigon: `RepeatedType` is not used for Maker registration form, so I kept doing the same.  But I agree we should use it at both places.

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: Then the first would be `@Route("/request", name="app_forgotten_password_request")`, does it sound good?

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: A referer check is pretty weak. What about adding some extra (random) token both in User session & as an attribute to `checkEmail` route?

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: Not really, I will change it

src/Resources/skeleton/forgottenPassword/twig_request.tpl.php
romaricdrigon: Ok

src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
romaricdrigon: Good idea!

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: That way, I don't see any risk. I moved the `flush()` just after `remove()`, for consistency.

romaricdrigon:
Thank you @bocharsky-bw for the review! I addressed most of your comments, and I would like to discuss some more on some.

src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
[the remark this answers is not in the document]
romaricdrigon: Doing it silently sounds good, otherwise the feature could be used to reveal an user exists for this e-mail address in the app. I will work on that.

src/Resources/skeleton/forgottenPassword/twig_request.tpl.php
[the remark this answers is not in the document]
bocharsky-bw: Ah, I see... I almost always do everything in form types. Anyway, it's minor and as far as it works the same - I'm fine with it 👍

src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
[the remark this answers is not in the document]
bocharsky-bw: Awesome! Sounds good to me too 👍 Thanks!

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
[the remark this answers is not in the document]
bocharsky-bw: Ah, yes, then we can skip it for this PR 👍 But probably in a separate PR with some extra check we can figure out Doctrine version and use `datetime_immutable` if it's `2.6`

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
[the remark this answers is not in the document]
bocharsky-bw: Ah, you're right, I missed it 👍

src/Maker/MakeForgottenPassword.php
bocharsky-bw: Actually, I don't think we need it for the registration form (it gives you rapid registration), but we do need it for password reset I think. And I've noticed this strategy on most websites

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
bocharsky-bw: Perfect 👍 Well, we can leave `@Route("/", name="app_forgotten_password_request")` but `/request` is even better I think

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
bocharsky-bw: Sounds cool, I like it 👍 So, when "check email" page is rendered - we need to invalidate that token (remove it), right? And if the token is invalid - redirect to `app_forgotten_password`

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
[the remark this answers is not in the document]
romaricdrigon: Yes, it sounds like a nice enhancement!

src/Maker/MakeForgottenPassword.php
romaricdrigon: Ok, I will modify!

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: Exactly!

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,110 @@
...
+            throw $this->createNotFoundException();
+        }
+        if ($passwordResetToken->isExpired() || !$passwordResetToken->isTokenEquals(substr($tokenAndSelector, PasswordResetToken::SELECTOR_LENGTH))) {
+            $this->getDoctrine()->getManager()->remove($passwordResetToken);
+            $this->getDoctrine()->getManager()->flush();
+
+            throw $this->createNotFoundException();
+        }
+
+        $form = $this->createForm(<?= $resetting_form_class_name ?>::class);
```
bocharsky-bw: `$form = $this->createForm(<?= $resetting_form_class_name ?>::class, $passwordResetToken->getUser());`

Important to bind data on the user, otherwise the form will create a new user :p

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,110 @@
...
+            $this->getDoctrine()->getManager()->remove($passwordResetToken);
+            $this->getDoctrine()->getManager()->flush();
+
+            // encode the plain password, and set it
+            $passwordResetToken->getUser()-><?= $password_setter ?>(
+                $passwordEncoder->encodePassword(
+                    $passwordResetToken->getUser(),
+                    $form->get('plainPassword')->getData()
+                )
+            );
```
bocharsky-bw: And then we need one more `flush()` then for the new password... or move token's `remove()` and `flush()` after `setPasswrod()` call?

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
bocharsky-bw: Actually, that's complex case, because we bind form to `User::class`, and on `isValid()` call it says that user object is invalid if you have validation constraints on `User`. I think we should either keep for binding to `User` and pass the user object to the form as the 2nd arg, or do not bind the form on User object at all

romaricdrigon:
Hmm, that's problematic indeed. Definitely we should have no limits there, because a user can request password several times (well, within discussed "retry TTL" limit). And there's no way not to leak any account information with such a feature. So I will modify the relationship to `ManytoOne`, and consider a test covering that scenario.

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: Really? The form has no `data_class`, I was expecting it to be bound to an array.

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: Correct, my modification broke it! So actually I would rather keep the `flush()` at the previous position.

romaricdrigon: (so everything is done in one "DB" operation)

bocharsky-bw:
Yeah, looks like so... sounds good tome, but probably we need `ManyToOne` in this case, but probably you meant it 👍

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: I deleted my previous comment, I found where the binding was done. 
I prefer not to bind the form to anything, but to an array, as I don't see any scenario where we may need to access/validate `user` from the form, we just want to be able to do `getData()`

romaricdrigon:
Exactly, I pushed the modification ;)

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: Modification done!

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: Modification done

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public function checkEmail()
+    {
+        return $this->render('forgottenPassword/check_email.html.twig', [
+            'tokenLifetime' => <?= $token_class_name ?>::TTL_HOURS,
+        ]);
+    }
+
+    /**
+     * @Route("/reset-password/{tokenAndSelector}", name="app_reset_password")
```
bocharsky-bw: And the last but not least important feature I would like to see in this PR - store token in session and hide it from the URL. It means we should make `tokenAndSelector` optional, and if the `tokenAndSelector` is passed - we store it in session and redirect users to just `/reset-password/`, but if there's no token in the URL - we read it from the session. If there's no token in both places - `createNotFoundException()` then. Wdyt?

This is a pretty important feature btw, because some JS scripts may reveal token to other third-party services when sending AJAX requests from this page, so this is important for security reasons.

romaricdrigon:
Thanks for the recap, I will try to work on those next week.
Once those are done, are we getting closer to a merge? This PR has been going for a while already, with quite some time of "sleep".

bocharsky-bw:
I bet yes! @weaverryan is _really_ want to merge this - just busy enough right now! Well, at least if we're talking about the "Forgotten password" feature in general, but probably he would request more minor code tweaks after all these new features, but it's going to be merged anyway.

weaverryan:
Yes, Victor has been keeping me up to date :). I should have time to play with this later in the week!

stof:
I'm wondering whether this should actually be implemented as a maker generating the whole feature in the project. If we ever find a flaw in the generated code, we will have no way to ship a patched code to all the projects which used the maker before the fix. At least I don't see a way for that.

weaverryan:
I’ve had that same thought... and of course you can try to think “does this contain sensitive code paths”, but trying to predict security flaws is of course impossible.

I had discussed this with @bocharsky-be originally as a bundle. We decided generation was better for DX - needing to override templates and configure stuff vs just changing the generated code is much harder.

Anyways, some more opinions on this would be great.

stof:
Well, maybe the email template itself should be configurable (with a maker to generate one you can customize). But I'm not sure other stuff needs to be overwritten (or even should be)

romaricdrigon:
As a developer, I actually prefer generated code versus a bundle. I prefer to deal with code rather than with configuration.
In a few years (2,3?) FOSUserBundle password reset feature did not change at all. I have the impression changes are unlikely. And in case we happen to have a fix to ship to all generated code, I wonder if something like an "updater" would be doable, given that most developers won't rename anything.

bocharsky-bw:
"Updater" sounds interesting, then along with `git diff` it would be possible to review changes. But how to know that you have to run the updater in case you don't watch changes in MakerBundle?

romaricdrigon:
That raises a good question: if we discover a security issue in code generated by MakerBundle, I guess we should open a CVE for MakerBundle? So security-advisory could warn users if they did not update.
For the case where they update regularly, but a symfony.com blogpost, I don't quite see any way to warn them.

stof:
@romaricdrigon no, that would not work. Updating MakerBundle will not fix the vulnerability in their own project if they already generated the code before.

A vulnerability in the generated code would actually require a CVE for MakerBundle (so that projects stop generating insecure code) and a CVE for each project containing the existing insecure code. That's not something we can do for the second part. And that's why I think we should be careful about bring this whole password-recovery feature as a maker.

romaricdrigon:
@stof agreed, but how many times did you have such a vulnerability in FOSUserBundle reset password feature? My point is not that it is impossible, but that it is unlikely, plus the current code is already to some pretty high standards compared to what developers would do on their own. That's a (relative) progress?

I'm back from holidays, should I resume working on that? 
It is not clear to me whether the features is OK with everybody or if it is on hold.

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
```diff
@@ -0,0 +1,87 @@
...
+    private $user;
+
+    private $plainToken;
+
+    public function __construct(User $user)
+    {
+        $this->requestedAt = new \DateTimeImmutable('now');
+        $this->selector = strtr(base64_encode(random_bytes(self::SELECTOR_LENGTH * 3 / 4)), '+/', '-_');
+        $this->plainToken = strtr(base64_encode(random_bytes(18)), '+/', '-_');
+        $this->token = password_hash($this->plainToken, PASSWORD_BCRYPT);
```
weaverryan: Can we use `PASSWORD_DEFAULT` here? I think that means it would continue improving over time as the app's php versions, etc, increased. Or could we use the new `NativePasswordEncoder` class? I would be ok requiring Symfony 4.3 for this feature if it means we're able to eliminate sensitive code (things that generate and check password) from the generated code

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
```diff
@@ -0,0 +1,87 @@
...
+    public function getUser(): <?= $user_class_name ?>
+
+    {
+        return $this->user;
+    }
+
+    public function isTokenEquals(string $token): bool
+    {
+        return password_verify($token, $this->token);
+    }
```
weaverryan: And I think I'd like to use `NativePasswordEncoder` here too. Well, not here of course :). But probably do this check in the controller.

romaricdrigon:
Thank you for the feedback @weaverryan 
I will work on all your comments very soon to finish this PR :)

bocharsky-bw:
Hey @romaricdrigon , will you have time to finish this soon?

romaricdrigon:
I will! I just rebased my branch.

Pierstoval:
@romaricdrigon Do you need some help? I might fix the reviews and PR them to your repo if you like, it's less work for you 😄

romaricdrigon:
Thanks for your offer, I've been kinda swamped last days due to personal reasons + flight issues which got me wandering in the airport instead of being at home :D

Right now, the biggest blocker is Travis failing. I don't understand why. And it's kinda a blocker. If you can help me with that, that would be immensely helpful!

romaricdrigon:
I opened another PR (#469), and it looks like Travis is just crashing on every MakerBundle build (even on `master`). So I will mostly ignore it :(

Pierstoval:
Let's try to figure it out, I'll open a PR and work on that tonight :smiley:

Pierstoval:
Well I spent a part of the night, and... All I could find was that the PHP commands take very very long to execute, so I think linking their output to PHPUnit's one could be a good start, and separate some tests in "groups" in order to execute phpunit once for each group could be a second start, but it might be more than that :confused:

romaricdrigon:
Damn, that's complex :(
Functional tests take a very long time to start. I suspect loading "fixtures" to disk is too heavy or suboptimal.

Pierstoval:
I investigated, it's not the fixtures, it's the maker commands. A symfony/skeleton project is created, and functional tests consists in copy/pasting the default skeleton, alter it a little for the sake of the test, and execute maker commands in order to check the generated output is the right one.

I wonder whether we could make this "enhanced unit tests" instead to make them a bit faster...

weaverryan:
Yea, the tests are super slow... I wish we could make them faster (and maybe we can), but they're meant to be realistic - creating a fresh Flex project, running `composer require` on dependencies, and actually running the commands. Indeed, better output on failure might help debugging this stuff easier. Btw, I usually run `./vendor/bin/simple-phpunit --filter=` and then the name of the test case in `FunctionalTest` that I specifically want to run.

> Travis is just crashing on every MakerBundle build (even on master)

That's another issue (feature!) about the test suite. Because they're so realistic, if something changes upstream (e.g. a recipe change) it could break our tests. That's actually what we *want*... because it means that some assumption that MakerBundle was making might now be incorrect... but it also causes more frequent "sudden" test failures.

Pierstoval:
A solution to make the tests faster would be to execute them in parallel instead. Since each test relies on a specific directory, it could be achieved by using the Process component or something similar. Implementation might be a bit complex, but to me it's probably the best option to make it faster.

romaricdrigon:
About making tests faster, we could install a "full" Symfony app with Flex only at the beginning of the test suite, and then copy/reset it after each it. It should be faster than having as many `composer require` / Flex install as there are tests (those 2 take about 50% of test time, and a lot of memory).
Anyhow, that requires further investigation, and likely an issue/a PR to discuss it.

I'm resuming my work on this PR, and I will finish it this afternoon. Sorry for the long time it took!
Since I implemented several similar tokens systems, *à la* @sstok, I may also improve a little bit interfaces.

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
romaricdrigon: I just modified it. And I forgot to react to your comment, but actually, that's a nice catch and a smart feature 👍

src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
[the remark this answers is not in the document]
stof: well, if you do it non-silently, you would have to always remember that a reset was requested for that email address, even if there is no user.

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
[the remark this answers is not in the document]
stof: versions of Doctrine older than 2.6 are not maintained anymore, so that's fine IMO.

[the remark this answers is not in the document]
romaricdrigon: Ok, definitely I prefer `datetime_immutable` :)

src/Resources/skeleton/forgottenPassword/twig_check_email.tpl.php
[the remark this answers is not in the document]
romaricdrigon: I added "retry TTL" feature (with silent failure). For simplicity, I'm re-using the same lifetime than token: you can't request a new `passwordResetToken` if you already have a non-expired one.

src/Resources/skeleton/forgottenPassword/PasswordResetToken.tpl.php
romaricdrigon: I changed to `PASSWORD_DEFAULT`. I think moving to `NativePasswordEncoder` should be done later, at least after 4.4 LTS release. Right now it sounds too much of a restriction, Symfony 4.3 is relatively new.

romaricdrigon:
The PR is now good regarding feedback from @bocharsky-bw and @weaverryan !
I added a few tests regarding new features (retry TTL...), test are green on Travis / 7.3 and AppVeyor. I tested the Maker on a fresh Symfony project, I like the generated code, and I believe it will improve many applications :)

Also thank you @stof for your comments and @Pierstoval for your help.
There are still 2 comments open for discussion, which are either personal preference either to be done later I think.

Pierstoval:
🚢 ❕ 

😄

src/Resources/skeleton/forgottenPassword/ForgottenPasswordController.tpl.php
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            // If User already has a valid Token, we don't want to generate a new one
+            // We fail silently.
+            $oldTokens = $this->getDoctrine()->getRepository(<?= $token_class_name ?>::class)->findNonExpiredForUser($user);
+            if (count($oldTokens)) {
+                return $this->redirectToRoute('app_check_email');
+            }
+
+            // Generate a reset password token, that the user could use to change his password.
```
sstok:
```suggestion
            // Generate a reset password token, that the user could use to change their password.
```

romaricdrigon: Thanks for noticing, I'm fixing it!

romaricdrigon:
@bocharsky-bw and @weaverryan  could you please check how I implemented modifications you requested?
I still hope it can get merged soon :)

FBL-dev:
Do you know when your changes will be available? I would like to use them

romaricdrigon:
> Do you know when your changes will be available? I would like to use them

@FBL-dev we had a chat over the feature today at SymfonyCon hack day, I rebased and fixed some small issues, hopefully @bocharsky-bw should review it soon, and I believe we are good for release :)

src/Maker/MakeForgottenPassword.php
```diff
@@ -0,0 +1,333 @@
...
+        $dependencies->addClassDependency(
+            TwigBundle::class,
+            'twig-bundle'
+        );
+        $dependencies->addClassDependency(
+            SecurityBundle::class,
+            'security'
+        );
+        $dependencies->addClassDependency(
+            SwiftmailerBundle::class,
```
rbaarsma: should be the new Symfony Mailer component instead of Swiftmailer.

I tried to install this bundle, but even though I installed mail, the dependency was swiftmailer, this error kept popping up

romaricdrigon:
@rbaarsma you have a point, but right now we don't quite have the infrastructure in MakerBundle to handle that (ie., have 2 different templates, switching according to installed dependency, test that...). I think SF5 is a topic we should treat "globally" in Maker bundle, maybe with some infrastructure refactor.

So I would prefer that PR to be merged, as @bocharsky-bw says, and then I will help with setting up that infrastructure + supporting Mailer. It does not degrade Maker as a whole, Maker won't work on SF5, but `make:functional-test` does not work too.

weaverryan:
Sorry for the slow merge on this. THANK you Romaric for your HUGE work here and also to Victor for helping to review and give feedback in my absence. I'm very excited about this feature :)

darraghenright:
Hi all. I just came across this. I want to implement this forgotten password functionality and I see it is merged, but I cannot find any documentation for it, nor is it in the  list of `make:` commands available in my project. 

I'm using version `1.21.1` of the maker bundle.

bocharsky-bw:
@darraghenright see `make:reset-password` command, it should be there

Issue #322, opened 2018-11-27
While running command "'make:e'". Message: "Could not find namespace node"


nic33wild:
Hi All,
does anyone know how to solve this problem with make: entity?

Error comes out when I try to create relationships with "bin/console make:entity" command.

```
$ bin/console make:e

 Class name of the entity to create or update (e.g. TinyKangaroo):
 > File

 Your entity already exists! So let's add some new fields!

 New property name (press <return> to stop adding fields):
 > directory

 Field type (enter ? to see all types) [string]:
 > relation

 What class should this entity be related to?:
 > Directory

What type of relationship is this?
 ------------ ---------------------------------------------------------------- 
  Type         Description                                                     
 ------------ ---------------------------------------------------------------- 
  ManyToOne    Each File relates to (has) one Directory.                       
               Each Directory can relate/has to (have) many File objects       
                                                                               
  OneToMany    Each File relates can relate to (have) many Directory objects.  
               Each Directory relates to (has) one File                        
                                                                               
  ManyToMany   Each File relates can relate to (have) many Directory objects.  
               Each Directory can also relate to (have) many File objects      
                                                                               
  OneToOne     Each File relates to (has) exactly one Directory.               
               Each Directory also relates to (has) exactly one File.          
 ------------ ---------------------------------------------------------------- 

 Relation type? [ManyToOne, OneToMany, ManyToMany, OneToOne]:
 > ManyToOne

 Is the File.directory property allowed to be null (nullable)? (yes/no) [yes]:
 > yes

 Do you want to add a new property to Directory so that you can access/update File objects from it - e.g. $directory->getFiles()? (yes/no) [yes]:
 > yes

 A new property will also be added to the Directory class so that you can access the related File objects from it.

 New field name inside Directory [files]:
 > 

2018-11-27T15:44:24+01:00 [error] Error thrown while running command "'make:e'". Message: "Could not find namespace node"

In ClassSourceManipulator.php line 810:
                                 
  Could not find namespace node  
                                 

make:entity [-a|--api-resource] [--regenerate] [--overwrite] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<name>]
```

antalaron:
[...]

PR #360, opened 2019-02-04
Look for the target entity in the namespace first

antalaron:
As in #322 the entity maker cannot make a relation, where the target entity is also presented in the root namespace, beacuse `Symfony\Bundle\MakerBundle\Maker\MakeEntity::askRelationDetails()` looks for it in the root namespace at first.

I don't think this is the good behaviour.

This PR makes it look for first in the namepace defined in the config, and after that in other places.


antalaron:
I don't know if the tests is enought, I've only created one for a self-referencing ManyToOne

weaverryan:
Thank you Antal - excellent job fixing this!

PR #367, opened 2019-03-04
Question helper return zero as string

### LINKED ISSUE #366: make:entity. Attribute "scale" on property expects a(n) integer, but got string.
Automatically generates this code:
```
/**
 * @ORM\Column(type="decimal", precision=6, scale="0")
 */
private $delivery_price;
```

And there should be such code:
```
/**
 * @ORM\Column(type="decimal", precision=6, scale=0)
 */
private $delivery_price;
```

Doctrine throws an error:
_In AnnotationException.php line x:
[Type Error] Attribute "scale" of `@ORM\Column` declared on property App\Entity\Order::$delivery_price expects a(n) integer, but got string._

P.S. on v1.11.3

LeJeanbono:
Fix #366

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -404,7 +404,7 @@ final class ClassSourceManipulator
             return 'null';
         }
 
-        if (\is_int($value)) {
+        if (\is_int($value) || '0' === $value) {
             return $value;
         }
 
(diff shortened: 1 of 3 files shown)


src/Util/ClassSourceManipulator.php
```diff
@@ -404,7 +404,7 @@ private function quoteAnnotationValue($value)
             return 'null';
         }
 
-        if (\is_int($value)) {
+        if (\is_int($value) || '0' === $value) {
```
weaverryan: What if we use a scale of "1", wouldn't that still print as `"1"`?

LeJeanbono: You mean type `"1"` in the command console ?

In question helper, all numbers typed are ok and concedered as number, except 0 which is '0' (so not a number)

weaverryan:
Thank you @LeJeanbono!

PR #370, opened 2019-03-11
[make:validator] Prevent an error in PHPStorm

dunglas:
The currently generated DocBlock prevent auto-completion to work properly, and triggers a warning. This patch fixes it.


weaverryan:
Thank you!

PR #371, opened 2019-03-12
[Serializer] Normalizer now implements CacheableSupportsMethodInterface

lyrixx:
related to https://github.com/api-platform/core/issues/2593


weaverryan:
This is a homerun... except that the bundle is useable with symfony/serializer 3.4, where this interface doesn't exist. We either need to:

I believe this code needs to be conditional, based on whether or not this interface exists in the user's project.

weaverryan:
See #399 for continuation

lyrixx:
I forgot this PR. Thanks @jojotjebaby :)

Issue #296, opened 2018-10-18
Own make apicrud


nicraMarcin:
I wrote my own maker apicrud in `src/Maker` which creates Controller, Form and Service.
`addOperation` looks only in vendor ... directory

**vendor/symfony/maker-bundle/src/Generator.php**
```
private function addOperation(string $targetPath, string $templateName, array $variables)
    {
        if ($this->fileManager->fileExists($targetPath)) {
            throw new RuntimeCommandException(sprintf(
                'The file "%s" can\'t be generated because it already exists.',
                $this->fileManager->relativizePath($targetPath)
            ));
        }

        $variables['relative_path'] = $this->fileManager->relativizePath($targetPath);

        $templatePath = $templateName;
        if (!file_exists($templatePath)) {
            $templatePath = __DIR__.'/Resources/skeleton/'.$templateName;

            if (!file_exists($templatePath)) {
                throw new \Exception(sprintf('Cannot find template "%s"', $templateName));
            }
        }

        $this->pendingOperations[$targetPath] = [
            'template' => $templatePath,
            'variables' => $variables,
        ];
    }
```

I think that AddOperation in Generator Class should look for templates in `src/Maker` directory
Now I solved this by symlink from my src/Maker/Resources/skeleton to vendor directory but to be compatible with symfony Generator should seek in `src/Maker` directory first.

jrushlow:
Closing this for now. See pull #375 for additional context.

PR #375, opened 2019-04-01
Generator search templates in 'src/Maker/Repository/skeleton'. Accord…

nicraMarcin:
…int to issue #296

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -170,10 +170,14 @@ class Generator
 
         $templatePath = $templateName;
         if (!file_exists($templatePath)) {
-            $templatePath = __DIR__.'/Resources/skeleton/'.$templateName;
+            $templatePath = $this->getRootDirectory()."/src/Maker/Resources/skeleton/".$templateName;
 
             if (!file_exists($templatePath)) {
-                throw new \Exception(sprintf('Cannot find template "%s"', $templateName));
+                $templatePath = __DIR__.'/Resources/skeleton/'.$templateName;
+                
+                if (!file_exists($templatePath)) {
+                    throw new \Exception(sprintf('Cannot find template "%s"', $templateName));
+                }
             }
         }
 


src/Generator.php
```diff
@@ -170,10 +170,14 @@ private function addOperation(string $targetPath, string $templateName, array $v
 
         $templatePath = $templateName;
         if (!file_exists($templatePath)) {
-            $templatePath = __DIR__.'/Resources/skeleton/'.$templateName;
+            $templatePath = dirname(__DIR__, 4)."/src/Maker/Resources/skeleton/".$templateName;
```
yceruto: The value of the parameter `%kernel.project_dir%` should  be used  instead of `dirname(__DIR__, 4)`.

Also, it will be great if we find (if possible) a way to not hardcode this path `/src/Maker/Resources/skeleton/`.

nicraMarcin: When I tried with `%kernel.project_dir%` I got syntax errors, I tried with double quoted `PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: syntax error, unexpected '%' in /media/sf_sskeleton/vendor/symfony/maker-bundle/src/Generator.php:176`

Why not hardcode `/src/Maker/Resources/skeleton/`? if path doesn't exists then will be vendor directory searched.

nicraMarcin: I changed `dirname(__DIR__, 4)`. there is `getRootDirectoty()` function few lines above :)

yceruto: Yes, `getRootDirectory()` is what I meant with `The value of the parameter %kernel.project_dir%`.

> Why not hardcode /src/Maker/Resources/skeleton/? if path doesn't exists then will be vendor directory searched. This is normal behavior in other bundles, for example in Twig. You can overwrite each template in vendor directory creating the same folder structure in src directory

Right, but it is hardcoding a convention, what if I want to store my skeleton templates in `src/Resources/MakerBundle/skeleton/` instead? or in `templates/maker/skeleton/`?

Configuration over convention, so I'm asking for a general solution like "a paths config/option" where to looking for skeleton templates.

nicraMarcin: > Right, but it is hardcoding a convention, what if I want to store my skeleton templates in src/Resources/MakerBundle/skeleton/ instead? or in templates/maker/skeleton/?

Because of convetnion of symfony? even now this path `$templatePath = __DIR__.'/Resources/skeleton/'.$templateName;` is fully hardcoded and users creating own makers can't use own path in symfony way.

> Configuration over convention, so I'm asking for a general solution like "a paths config/option" where to looking for skeleton templates.

I also think that It would be better to store this path in some variable, but Maker-Bundle doesn't provide any configuration file or doesn't add any lines into .env file. So the simplest way to achevie oportunity to store maker templates in own project, not vendor directory.

nicraMarcin: I tried to add variable in config but i get error ` Unrecognized option "template_path" under "maker". Available option is "root_namespace". `

nicraMarcin: > For that reason, in general, the generated code cannot be modified. In many cases, adding your own maker command is so easy, that we recommend that.

I admit, adding own maker command is so easy, but not adding own templates without coding in vendor dir  :/

yceruto: > I also think that It would be better to store this path in some variable, but Maker-Bundle doesn't provide any configuration file ...

Look at this file https://github.com/symfony/maker-bundle/blob/master/src/DependencyInjection/Configuration.php

The `root_namespace` is already a config option for `Generator` service, what about add a new option `skeleton_path` (or better) to configure the default skeleton path at project level?

nicraMarcin:
I'm not main symfony developer to know each vendor file. For hints I use
IDE not study whole code.
I've tried to make small improvement and You asking me to do more. I'm
sorry, I'm not fluent with PHP and if you think that this change is not
enough so close it or add changes you think are better.
ps. sorry for my English.

pon., 1 kwi 2019, 22:20 użytkownik Yonel Ceruto <notifications@github.com>
napisał:

yceruto:
Hi, @nicraMarcin I think your solution is good, but we may have to improve it a bit, so everyone can benefit from it.

It is perfectly fine if you don't want to continue improving it, I am nobody in this bundle, so my opinion has the same value as yours. Then, let's wait to see what other think.

I just try to help.

nicraMarcin:
@yceruto so if you aren't main developer so I have to thak you for engagement for project. I thik thak Maker does great job for users.
Could you please tell me how to provide variable in to `src/Generator.php`? Do I have to inject some service if yes which?

weaverryan:
Hey @nicraMarcin!

Thanks for this. I understand what you're trying to do - adding *new* "templates" so that you can use them in your customer maker. That should be allowed. What should *not* be allowed is to override existing "core" templates (we just don't want to promise backwards compatibility for these). Those are two separate ideas, but by making this directory configurable, it would make overriding possible.

What if we allow `$templateName` to be an absolute file path? And `if (file_exists($templateName))`, then simply use that template instead of trying to find it in the normal directory. In your maker, you would do things like:

```php
$generator->addClass('SomeClass', __DIR__.'/templates/CustomClassTemplate.tpl.php, ['someVar' => 'foo'])
```

What do you think?

nicraMarcin:
> In your maker, you would do things like:
> 
> ```
> $generator->addClass('SomeClass', __DIR__.'/templates/CustomClassTemplate.tpl.php, ['someVar' => 'foo'])
> ```
> 
> What do you think?

I think that is better than now.



> What should _not_ be allowed is to override existing "core" templates

Could you tell me why? Are there any sensible reasons?
For example, I use "make:entity" to generate Entities but I had to add some function at end of each Repository. 
```
    public function transformAll()
    {
        $<?= $entity_alias; ?>s = $this->findAll();
        $<?= $entity_alias; ?>sArray = [];
    
        foreach ($<?= $entity_alias; ?>s as $<?= $entity_alias; ?>) {
            $<?= $entity_alias; ?>sArray[] = $this->transform($<?= $entity_alias; ?>);
        }

        return $<?= $entity_alias; ?>sArray;
    }
```
so I copied **Repository.tpl.php** to my src/Maker ... and add what I wanted. This way I can overwrite core files. 
Nobody do this without knowledge how it works so I think that _Maker_ should allow to overwrite files.

But if you think that solution you proposed is also good and users can use their own templates for maker it could be good also.

weaverryan:
> What should not be allowed is to override existing "core" templates

> Could you tell me why? Are there any sensible reasons?

The simple reason is that we don't want to guarantee backwards compatibility of the variables those templates are passed. If we *did* allow you to override them, then it makes it harder to make big changes to the generated code - it slows down how fast this bundle can improve. It's a trade-off.

Cheers!

nicraMarcin:
> If we _did_ allow you to override them, then it makes it harder to make big changes to the generated code - it slows down how fast this bundle can improve. It's a trade-off.

I don't think so but I'm not developer. 
The truth is that user should have possibility to use own templates and add something to core templates like I mentioned before.

Cheers!

weaverryan:
Closing - we may consider allowing templates to be overridden in the future, but this PR isn't enough to do that - so we'll discuss elsewhere.

Cheers!

PR #376, opened 2019-04-02
[make:validator] Custom constraints should ignore null and empty values

voronkovich:
See https://symfony.com/doc/current/validation/custom_constraint.html#creating-the-validator-itself

### DIFF
--- a/src/Resources/skeleton/validator/Validator.tpl.php
+++ b/src/Resources/skeleton/validator/Validator.tpl.php
@@ -15,6 +15,7 @@ class <?= $class_name ?> extends ConstraintValidator
             return;
         }
 
+        // TODO: implement the validation here
         $this->context->buildViolation($constraint->message)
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/validator/Validator.tpl.php
```diff
@@ -11,6 +11,10 @@ public function validate($value, Constraint $constraint)
     {
         /* @var $constraint \<?= $constraint_class_name ?> */
 
+        if (null === $value || '' === $value) {
+            return;
+        }
+
```
stof: should we also add a `// TODO implement the validation here` so that they know where to put it ?

voronkovich: @stof, I think you're right! I've added a TODO entry.

weaverryan:
Thank you @voronkovich!

PR #378, opened 2019-04-07
Configuration for custom namespaces

natepage:
Hi there,

I'm working on projects which have the convention to put all the entities and repositories under `App\Database\Entity` and `App\Database\Repository`.

I've had a quick look at the issues and the PRs and I've seen other people asking for the same thing so I've decided to give it a go :)

Here is a simple solution to implement a helper to hold all the different namespaces, give it to the generator so it can be used in all makers.

This solution changes the constructor signatures of:
- `Symfony\Bundle\MakerBundle\Generator`
- `Symfony\Bundle\MakerBundle\Doctrine\DoctrineHelper`

Those objects are used with DI so it doesn't impact the existing makers but if anyone has extended them in their project it would then be a BC.

I leave this PR here and I'm looking forward to hearing from you.

PR #379, opened 2019-04-16
More robust autoloader detection

### LINKED ISSUE #196: Maker Bundle does not find App namespace
I have a very strange problem with maker bundle in my symphony 4.1 project. I am using the standard App namespace. The project works fine, no issues with composer update or composer install.

When I try to run bin/console make:entity I get the following warning:

`! [NOTE] It looks like your app may be using a namespace other than "App".
 !
 !        To configure this and make your life easier, see:
 !        https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html#configuration.`

My autoload keys in composer.json are:

> "autoload": {
        "psr-4": {
            "App\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "App\\Tests\\": "tests/"
        }
    },

At this point I am not sure where to start the trouble shooting. I have seen some posts to make sure that the services.yaml file is set up correctly, but I cannot find an error there (App related entries below...):

> 
    # makes classes in src/ available to be used as services
    # this creates a service per class whose id is the fully-qualified class name
    App\:
        resource: '../../src/*'
        exclude: '../../src/{Entity,Migrations,Tests}'

    # controllers are imported separately to make sure services can be injected
    # as action arguments even if you don't extend any base controller class
    App\Controller\:
        resource: '../../src/Controller'
        tags: ['controller.service_arguments']

    # add more service definitions when explicit configuration is needed
    # please note that last definitions always *replace* previous ones

    # create aliases for bundle services to make them compliant with symfony 4

    # this listener is required by App\EventListener\MyUserAccessListener
    ApiPlatform\Core\EventListener\ReadListener: ~

    # this service needs to be public as it replaces the
    # API Platform PUT controller for MedicalSchool resources
    App\Action\MedicalSchoolPut:
        public: true

    # this service needs to be public as it replaces the
    # API Platform POST controller for File resources
    App\Action\FilePost:
        public: true`

Any thoughts? Thanks!

### LINKED ISSUE #231: "make:command" leads into "Could not determine where to locate the new class"
I'm trying to get maker working with a new Symfony 4.1 installation. I've created it using composer, changed the "App"-Namespace to "MakerTest" (see composer.json) and configures maker to use "MakerTest" as root-namespace.

composer.json:
```
    "autoload": {
        "psr-4": {
            "MakerTest\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "MakerTest\\Tests\\": "tests/"
        }
    },
```

config/packages/maker.yaml:
```
maker:
  root_namespace: MakerTest
```

Output:
```
root@xyz:/var/www/# bc make:command

 ! [NOTE] It looks like your app may be using a namespace other than "MakerTest".
 !
 !        To configure this and make your life easier, see:
 !        https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html#configuration.

 Choose a command name (e.g. app:delicious-chef):
 > create-user


In Generator.php line 43:

  Could not determine where to locate the new class "MakerTest\Command\CreateUserCommand", maybe try with a full namespace like "\My\Full\Namespace\CreateUserCommand"


make:command [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<name>]
```

Tried it with latest stable and also dev-master - same issue with both.

### LINKED ISSUE #313: Maker Bundle does not find App namespace
Hi, if i require [phpstan/phpstan-shim](https://github.com/phpstan/phpstan-shim) in my project Maker Bundle does not find App namespace, after remove this library every thing work fine. 
Is it possible that Maker Bundle will work with phpstan-shim together?

GawainLynch:
When the project, or dependent package, adds a custom autoloader, e.g. PhpStan, the added autoloader will be returned instead of the one containing the project's autoloader when `ComposerAutoloaderFinder` iterates the autoloader list leading to the mysterious error:

```
Could not determine where to locate the new class "App\Controller\DeliciousPuppyController", maybe try with a full namespace like "\My\Full\Namespace\DeliciousPuppyController"
```

Fixes #196
Fixes #231
Fixes #313

Related #378

### DIFF
--- a/src/Util/ComposerAutoloaderFinder.php
+++ b/src/Util/ComposerAutoloaderFinder.php
@@ -37,7 +37,7 @@ class ComposerAutoloaderFinder
     public function getClassLoader(): ClassLoader
     {
         if (null === $this->classLoader) {
-            $this->findComposerClassLoader();
+            $this->classLoader = $this->findComposerClassLoader();
         }
 
         if (null === $this->classLoader) {
@@ -47,16 +47,26 @@ class ComposerAutoloaderFinder
         return $this->classLoader;
     }
 
+    /**
+     * @return ClassLoader|null
+     */
     private function findComposerClassLoader()
     {
         $autoloadFunctions = spl_autoload_functions();
 
         foreach ($autoloadFunctions as $autoloader) {
             $classLoader = $this->extractComposerClassLoader($autoloader);
-            if ($classLoader && $this->locateMatchingClassLoader($classLoader)) {
-                return;
+            if (null === $classLoader) {
+                continue;
+            }
+
+            $finalClassLoader = $this->locateMatchingClassLoader($classLoader);
+            if (null !== $finalClassLoader) {
+                return $finalClassLoader;
             }
         }
+
+        return null;
     }
 
     /**
@@ -78,28 +88,28 @@ class ComposerAutoloaderFinder
         return null;
     }
 
-    private function locateMatchingClassLoader(ClassLoader $classLoader): bool
+    /**
+     * @return ClassLoader|null
+     */
+    private function locateMatchingClassLoader(ClassLoader $classLoader)
     {
+        $makerClassLoader = null;
         foreach ($classLoader->getPrefixesPsr4() as $prefix => $paths) {
-            // We can default to using the autoloader containing this component if none are matching.
             if ('Symfony\\Bundle\\MakerBundle\\' === $prefix) {
-                $this->classLoader = $classLoader;
+                $makerClassLoader = $classLoader;
             }
             if (0 === strpos($this->rootNamespace['psr4'], $prefix)) {
-                $this->classLoader = $classLoader;
-
-                return true;
+                return $classLoader;
             }
         }
 
         foreach ($classLoader->getPrefixes() as $prefix => $paths) {
             if (0 === strpos($this->rootNamespace['psr0'], $prefix)) {
-                $this->classLoader = $classLoader;
-
-                return true;
+                return $classLoader;
             }
         }
 
-        return false;
+        // We can default to using the autoloader containing this component if none are matching.
+        return $makerClassLoader ?: null;
     }
 }
--- a/src/Util/ComposerAutoloaderFinder.php
+++ b/src/Util/ComposerAutoloaderFinder.php
@@ -19,9 +19,6 @@ use Symfony\Component\Debug\DebugClassLoader;
  */
 class ComposerAutoloaderFinder
 {
-    /**
-     * @var string
-     */
     private $rootNamespace;
 
     /**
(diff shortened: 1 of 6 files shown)


src/Util/ComposerAutoloaderFinder.php
```diff
@@ -19,11 +19,21 @@
  */
 class ComposerAutoloaderFinder
 {
+    /**
+     * @var string
+     */
+    private $rootNamespace;
```
weaverryan: phpdoc is not needed (thanks to type-hint in constructor)

src/Util/ComposerAutoloaderFinder.php
```diff
@@ -19,11 +19,21 @@
...
+
     /**
      * @var ClassLoader|null
      */
     private $classLoader = null;
 
+    public function __construct(string $rootNamespace = 'App\\')
```
weaverryan: No need to have a default value.

src/Util/ComposerAutoloaderFinder.php
```diff
@@ -46,13 +56,15 @@ private function findComposerClassLoader()
 
         foreach ($autoloadFunctions as $autoloader) {
             if (\is_array($autoloader) && isset($autoloader[0]) && \is_object($autoloader[0])) {
-                if ($autoloader[0] instanceof ClassLoader) {
+                if ($autoloader[0] instanceof ClassLoader
```
weaverryan: Maybe we should improve the `Composer ClassLoader not found ` message above?

> Could not find a Composer autoloader that autoloads from "{$this->rootNamespace}"

What about PSR-0?

And also, there is still an edge case: I have `App\\` configured as my default, but I actually always generate things using a full class name for some reason - e.g. `MyCompany\Entity\FooEntity`. And, I don't have `App` configured in my autoloader. In that case, I would get the error EVEN if `MyCompany` *is* mapped in a valid way.

Perhaps, to at least help reduce this, we should only apply this extra check that you've added IF we find *multiple* class loaders? If we only find 1, it's safe to use that, even if it doesn't make the root namespace.

GawainLynch:
OK, so while I am not happy with my choice of method naming nor elegance (ideas very welcome!). 

I started with a baseline of a fresh `composer create-project symfony/website-skeleton`, which already has two autoloaders. So I figured the best approach now is to "remember" the autoloader that contains the component, and should the match still not found, using that as the fallback.

Also, I've slightly expanded the tests to cover PSR-0, PSR-4 & fallback.

src/Util/ComposerAutoloaderFinder.php
```diff
@@ -19,45 +19,87 @@
...
+
+        foreach ($classLoader->getPrefixes() as $prefix => $paths) {
+            if (0 === strpos($this->rootNamespace['psr0'], $prefix)) {
+                $this->classLoader = $classLoader;
+
+                return true;
+            }
+        }
+
+        return false;
```
weaverryan: Instead of setting the property, let's return the class loader or null.

src/Util/ComposerAutoloaderFinder.php
```diff
@@ -19,45 +19,87 @@
...
-
-                if ($autoloader[0] instanceof DebugClassLoader
-                    && \is_array($autoloader[0]->getClassLoader())
-                    && $autoloader[0]->getClassLoader()[0] instanceof ClassLoader) {
-                    return $autoloader[0]->getClassLoader()[0];
-                }
+            $classLoader = $this->extractComposerClassLoader($autoloader);
+            if ($classLoader && $this->locateMatchingClassLoader($classLoader)) {
+                return;
+            }
```
weaverryan: Here, if we make `locateMatchingClassLoader()` return a value, we can simplify a bit (at least, simpler for how I think!)

```php
if (null === classLoader) {
    continue;
}

$finalClassLoader = $this->locateMatchingClassLoader($classLoader);

if (null !== $finalClassLoader) {
    return $finalClassLoader;
}
```

src/Util/ComposerAutoloaderFinder.php
```diff
@@ -19,45 +19,87 @@
...
+            'psr0' => rtrim($rootNamespace, '\\'),
+            'psr4' => rtrim($rootNamespace, '\\').'\\',
+        ];
+    }
+
     public function getClassLoader(): ClassLoader
     {
         if (null === $this->classLoader) {
-            $this->classLoader = $this->findComposerClassLoader();
+            $this->findComposerClassLoader();
```
weaverryan: Then we can change this back - I prefer to have all the other methods return a value, and then we set it here :)

src/Util/ComposerAutoloaderFinder.php
```diff
@@ -19,45 +19,87 @@
...
         return null;
     }
+
+    private function locateMatchingClassLoader(ClassLoader $classLoader): bool
+    {
+        foreach ($classLoader->getPrefixesPsr4() as $prefix => $paths) {
+            // We can default to using the autoloader containing this component if none are matching.
+            if ('Symfony\\Bundle\\MakerBundle\\' === $prefix) {
+                $this->classLoader = $classLoader;
+            }
```
weaverryan: I'm not sure about this. Well, specifically, it makes sense. But, what edge case are we coding for? It definitely complicates the algorithm (all my thoughts about returning from these methods instead of setting the property directly fall apart).

GawainLynch:
> 1. What's the edge case that we're trying to solve by falling back to the autoloader that contains this bundle's namespace? I'd prefer to not do this, but I might be missing some legitimate use-case.

Specifically for the fallback option, and _assuming_ `symfony/maker-bundle` is installed via Composer:
  - It is a namespace we know will be in the Composer autoloader
  - The end-project root namespace is a config setting, so could be wrong, missing, or not 'dumped' by Composer yet
  - The end-project `composer.json` is not guaranteed to have its own `psr-4` or `psr-0` defined and loading

> 2. If we don't need that, we can return from all the methods, instead of setting the property directly.

Done :grinning:

weaverryan:
This is very nice work @GawainLynch! Thanks for the final changes - the flow was easy for me to understand :). Cheers!

GawainLynch:
A pleasure mate, thank you too.

Kocal:
Hi, thanks for this patch! I had the phpstan-shim issue and now it's working properly! :)

PR #381, opened 2019-04-17
[make:*-test] Use the new WebTestAssertionsTrait methods in the generated functional tests

adrienlucas:
Hi, this PR is to fix the way of calling the assertions static methods of PHPunit in the generated unit and functional tests.

[First commit](https://github.com/symfony/maker-bundle/commit/8f4af140015f6e30fe871fbacbc8c066901e0a25) : As pointed out in [this article](https://arkadiuszkondas.com/the-right-way-to-call-assertion-in-phpunit/), we should not be calling the assertions using `$this->assert*` but instead with `static::assert*`.

[Second commit](https://github.com/symfony/maker-bundle/commit/cfb0715dc99fb6cac45f08a6c3daefc3bafba542) : To go further, we can use [the new assertion API](https://github.com/symfony/symfony/pull/30813) provided by the `WebTestCase`.

I don't kown if this could be considered a BC break... but i hope not.

### DIFF
--- a/src/Maker/MakeFunctionalTest.php
+++ b/src/Maker/MakeFunctionalTest.php
@@ -20,6 +20,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\CssSelector\CssSelectorConverter;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestAssertionsTrait;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
@@ -52,7 +53,9 @@ class MakeFunctionalTest extends AbstractMaker
         $generator->generateClass(
             $testClassNameDetails->getFullName(),
             'test/Functional.tpl.php',
-            []
+            [
+                'web_assertions_are_available' => class_exists(WebTestAssertionsTrait::class),
+            ]
         );
 
         $generator->writeChanges();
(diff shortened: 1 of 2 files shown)


weaverryan:
Hey @adrienlucas!

Thank you for this! So, let's look at the 2 commits independently:

1) `self::` vs `$this->` for the assert functions. For this, we should be consistent with Symfony's core and what's recommended in Symfony's docs, which is `$this->` currently. We could discuss that on an issue on `symfony/symfony-docs`. I don't feel too strongly, but MakerBundle will follow that lead.

2) About using the new assertion API in WebTestCase. I <3 this idea. However, as this feature is only available in Symfony 4.3 or higher, we need to make the generated code smart: it needs to use the *old* code if the user doesn't have the new `WebTestAssertions` trait and *new* code if they do. We could determine that in the maker and pass it in as a variable. Could you make that change?

Thanks!

adrienlucas:
@weaverryan yes I will

1. I get it, while being a little disappointed about it... Is there an explanation for using a non-static call on a static method here ? I will fix it anyway, to keep consistency. 
2 . Will do !

To go further : should i try to assess Panther installation too, and make the generated test extends the `PantherTestCase` when possible ?

B-Galati:
For the 1st one, see https://github.com/sebastianbergmann/phpunit/issues/1914

adrienlucas:
@B-Galati thank you !

@weaverryan does the new version of this PR suits more the needs ?

should we try to use travis to test against the `symfony/framework-bundle:dev-master` dependency, or something else to ensure the two versions of the possibly generated functional tests are working ?

(i renamed the PR to reflect the new intent)

weaverryan:
Thank you Adrien!

PR #384, opened 2019-04-22
Add column attribute "unique" to MakeEntity #244

### LINKED ISSUE #244: make:entity unique property
Console command make:entity doesn't ask if I want the property to be unique and if I add it to the annotation manually, then continue adding new properties, it removes it... :/
I think it should either ask if I wanted unique, or at least keep my previous properties untouched.

jschaedl:
This fixes #244.

Todo:

- [x] adding unique attribute question to `MakeEntity` class
- [ ] add a corresponding unit test

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -363,6 +363,10 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
             $data['nullable'] = true;
         }
 
+        if ($io->confirm('Is this field unique in the database (unique)', false)) {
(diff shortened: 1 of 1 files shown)


weaverryan:
Hi @jschaedl!

Thanks for the PR :). However, I'm -1 on this idea - unique columns just aren't that common. If we add the question, then it will ask this question on *every* single field, even though it's rarely needed. I think this is something that users should add by hand if they need it.

Cheers!

jschaedl:
Hi @weaverryan Ok, I see. Makes sense.

PR #385, opened 2019-04-26
Fix various typos

fabpot:


src/Validator.php
```diff
@@ -194,15 +194,15 @@ public static function entityExists(string $className = null, array $entities =
         self::notBlank($className);
 
         if (empty($entities)) {
-            throw new RuntimeCommandException('There is no registered entities. Please create entity before use this command');
+            throw new RuntimeCommandException('There is no registered entities; please create entity before use this command.');
```
przemyslaw-bogusz: I think this one still needs some fixing. How about:
'There **are** no registered entities; please create **an** entity before **using** this command.'?

weaverryan:
Thanks Fabien!

src/Validator.php
weaverryan: 💯 fixed at sha: 6732a15f37e254eb074bba19bcebc0ec13642f64

Issue #249, opened 2018-08-23
Warning: DOMElement::setAttribute(): string is not in UTF-8


pochy-ja:
After create a Controller this first error appear, but the file 

InventarioUbicaci�onController

in the controller places do not exist and in the console can not clear the cache, I try to erase the cache but the other error appears...

**FIRST**

WEBROWSER
```
(1/1)                                     ErrorException                                                                  Warning: DOMElement::setAttribute(): string is not in UTF-8
--
in XmlDumper.php line 101
at XmlDumper->addService(object(Definition), 'App\\Controller\\InventarioUbicaci�onController', object(DOMElement))in XmlDumper.php line 241
at XmlDumper->addServices(object(DOMElement))in XmlDumper.php line 53
at XmlDumper->dump()in ContainerBuilderDebugDumpPass.php line 32
at ContainerBuilderDebugDumpPass->process(object(ContainerBuilder))in Compiler.php line 95
at Compiler->compile(object(ContainerBuilder))in ContainerBuilder.php line 736
at ContainerBuilder->compile()in Kernel.php line 519
at Kernel->initializeContainer()in Kernel.php line 123
at Kernel->boot()in Kernel.php line 183
at Kernel->handle(object(Request))in index.php line 37
```

**SECOND**

_CONSOLE_
```
In XmlDumper.php line 101:

  Warning: DOMElement::setAttribute(): string is not in UTF-8
```

pochy-ja:
BTW, is a project with Symfony 4.1.3 in Ubuntu 16.04 with PHP 7.2.9-1+ubuntu16.04.1+deb.sury.org+1

stof:
what is the command you ran exactly ?

pochy-ja:
After

_./bin/console make:controller InventarioUbicacion_

that create a problem, then any command that i try to use, respond with the error

In XmlDumper.php line 101:

  Warning: DOMElement::setAttribute(): string is not in UTF-8

weaverryan:
Hmm. After you ran the command, it did *not* create the controller class? It looks to me like the new controller maybe does exist, but Symfony is having a problem processing it because the class name has a special character that is causing issues. What files do you have in your controller directory? And, if you manually delete all the files in var/cache/dev/, does the error change?

Cheers!

pochy-ja:
[...]

PR #387, opened 2019-05-07
[Validator] Add encoding test to validateClassName

przemyslaw-bogusz:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets    | #249
| Deprecations? | no
| License       | MIT

This checks if the input string is a UTF-8-encoded, and throws an exception if not.

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -47,6 +47,12 @@ final class Validator
         ];
 
         foreach ($pieces as $piece) {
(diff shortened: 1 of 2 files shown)


weaverryan:
Thank you Przemysław! This seems like a very practical step +1

PR #398, opened 2019-05-21
Deprecate Argon2i encoder used in make:user

### LINKED ISSUE #386: [4.3] make:user should suggest "auto" instead of Argon2i/BCrypt
See symfony/symfony#31140 - the make:user command will likely need to use a new password hash algorithm.
Related to https://github.com/symfony/recipes/pull/576 et al.

nicolas-grekas:
`auto` is the only future-proof option, no need to allow ppl to choose.

Fixes #386


weaverryan:
@pluseg if you're interested, I'm sure Nicolas would love you if you grabbed his commit, updated the tests, and finished the PR :)

weaverryan:
Thank you Nicolas!

Issue #400, opened 2019-05-22
Update MakeSubscriber for 4.3 event name changes


weaverryan:
In Symfony 4.3, many of the event names changes to be class names. The EventRegistry class - https://github.com/symfony/maker-bundle/blob/master/src/EventRegistry.php - as a short list of the most common events in Symfony. We should update these to use the latest event names, IF those new event classes are available (i.e. use the new event names if they are available in the user's app, otherwise use the old event names): https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching#updated-httpkernel-event-classes

javiereguiluz:
Closing and fixed by #403.

PR #403, opened 2019-05-23
Added conditions to the constructor in order to use the new events cl…

jojotjebaby:
related to #400 

I am not sure this is the best way to do this. Would love some input / suggestions if there is a better way ?


jojotjebaby:
You are right, I will do an update tomorrow with the following changes: 

Remove the changed events from the default array and add them in the constructor by checking wether the new class exists. If not assign the old one. That could be done with a ternary comparator. 

Do you think that would be a good way to approach this ?

src/EventRegistry.php
```diff
@@ -57,6 +64,35 @@ class EventRegistry
     public function __construct(EventDispatcherInterface $eventDispatcher)
     {
         $this->eventDispatcher = $eventDispatcher;
+
+        // Check if the new event classes exist, if so replace the old one with the new.
+        if (class_exists(ControllerArgumentsEvent::class)) {
+            self::$eventsMap['kernel.controller_arguments'] = ControllerArgumentsEvent::class;
```
weaverryan: In the new system (with these new event classes), the event class and the event name are the same! So, you'd actually need to `unset()` the original key, then add a new item to the array like this:

`ControllerArgumentsEvent::class => ControllerArgumentsEvent::class`.

I might even create a new static property with a map of the old event names to the new event class, so that you could just loop in here:

```php
private static $newEventsMap = [
    'kernel.controller_arguments' => ControllerArgumentsEvent::class
];

// .. and then in the constructor
foreach (self::$newEventsMap as $oldEventName => $newEventClass) {
    // unset old key
    // add new key
}
```

Nice solution though - it's a surprisingly tricky thing to get right.

weaverryan:
> @jojotjebaby Won't there be an error with non-found old classes as they were renamed?

I don't see the problem? It *is* ok to reference `NonExistentClassName::class`, even if that class doesn't exist.

src/EventRegistry.php
jojotjebaby: Thank you for the feedback ! I updated the PR with the solution you recommended. I was first thinking of using the new events by default (because in the future they will be the default ones) but I didn't find an easy solution to do this. I prefer to do oldX / X rather then X / newX because in the future you would just need to delete the oldX in order to update the code.

src/EventRegistry.php
```diff
@@ -52,11 +59,30 @@ class EventRegistry
...
         $this->eventDispatcher = $eventDispatcher;
+
+        // Loop through the new event classes
+        foreach (self::$newEventsMap as $oldEventName => $newEventClass) {
+            //Check if the new event classes exist, if so replace the old one with the new.
+            if (class_exists($newEventClass)) {
+                unset(self::$eventsMap[$oldEventName]);
+                self::$eventsMap[$newEventClass] = $newEventClass;
+            }
+        }
```
weaverryan: Let's add a test for this :). Actually, we have one: `EventRegistryTest::testGetEventClassNameFromStandardList()`. But now, I think it needs to have a different "assert" value based on whether or not the `ExceptionEvent` class existss.

jojotjebaby: I changed this bit in order to display just the Classname and not the FQCN. This fixes the first part of the bug we encounter in the subscriber.tpl.php and displays the list of events in the CLI much better.

weaverryan: Did you push those changes? Or are you saying you already did something previously, and what I was saying aren't issues?

src/Resources/skeleton/event/Subscriber.tpl.php
```diff
@@ -15,7 +15,7 @@ public function <?= $method_name ?>(<?= $event_arg ?>)
     public static function getSubscribedEvents()
     {
         return [
-            '<?= $event ?>' => '<?= $method_name ?>',
+           <?= $event ?> => '<?= $method_name ?>',
```
weaverryan: If $event is a class name, we'll also need a use statement for it.

jojotjebaby: The variable $event_full_class_name will have the full class name so a use statement will be inserted on the line 6.

weaverryan: Of course :)

weaverryan:
Thank you Joris! Excellent work!

PR #404, opened 2019-05-23
Removed duplicate array from getAuthenticatorTests dataProvider

St0iK:


weaverryan:
Thanks @St0iK! Our test suite just got a *tiny* bit faster :)

Issue #397, opened 2019-05-21
make:auth login form: add logout support and help for logged in user


weaverryan:
After talking with @nicolas-grekas, we're proposing three changes to `make:auth` for the login form:

1) Add a question: "Do you want to generate a /logout URL?" and add the needed security.yaml and route/controller code for this if yes

2) On the login form, if the user is already logged in, display a message (in addition to the login form) like:

> you're logged in as XXX [Logout Liink]

3) In the controller, add commented-out code that the user could uncomment to redirect to the "target path" if the user is already logged in.

maxhelias:
This feature is now present. The issue can be closed or there are still things to do?

PR #406, opened 2019-05-25
[make:auth] Added logout support and help for logged in user

St0iK:
Issue: https://github.com/symfony/maker-bundle/issues/397


src/Security/SecurityControllerBuilder.php
```diff
@@ -55,4 +63,19 @@ public function addLoginMethod(ClassSourceManipulator $manipulator)
...
+    public function addLogoutMethod(ClassSourceManipulator $manipulator)
+    {
+        $logoutMethodBuilder = $manipulator->createMethodBuilder('logout', null, false, ['@Route("/logout", name="app_logout" , methods={"GET"})']);
+
+        $manipulator->addUseStatementIfNecessary(Response::class);
+        $manipulator->addUseStatementIfNecessary(Route::class);
+        $manipulator->addUseStatementIfNecessary(AuthenticationUtils::class);
+        $manipulator->addMethodBody($logoutMethodBuilder, <<<'CODE'
+<?php
+// controller can be blank: it will never be executed!
```
weaverryan: Let's throw an exception here with a similar message inside:

```
throw new \Exception('This method can be blank - it will be intercepted by the logout key on your firewall');
```

src/Security/SecurityControllerBuilder.php
```diff
@@ -55,4 +63,19 @@ public function addLoginMethod(ClassSourceManipulator $manipulator)
...
         $manipulator->addMethodBuilder($loginMethodBuilder);
     }
+
+    public function addLogoutMethod(ClassSourceManipulator $manipulator)
+    {
+        $logoutMethodBuilder = $manipulator->createMethodBuilder('logout', null, false, ['@Route("/logout", name="app_logout" , methods={"GET"})']);
+
+        $manipulator->addUseStatementIfNecessary(Response::class);
+        $manipulator->addUseStatementIfNecessary(Route::class);
+        $manipulator->addUseStatementIfNecessary(AuthenticationUtils::class);
```
weaverryan: I think only `Route` is needed, right?

src/Security/SecurityControllerBuilder.php
```diff
@@ -55,4 +63,19 @@ public function addLoginMethod(ClassSourceManipulator $manipulator)
         );
         $manipulator->addMethodBuilder($loginMethodBuilder);
     }
+
+    public function addLogoutMethod(ClassSourceManipulator $manipulator)
+    {
+        $logoutMethodBuilder = $manipulator->createMethodBuilder('logout', null, false, ['@Route("/logout", name="app_logout" , methods={"GET"})']);
```
weaverryan: `methods=""` part is not needed. In fact, I think the `LogoutListener` will allow any HTTP method, so better to stay consistent with that.

src/Security/SecurityConfigUpdater.php
```diff
@@ -80,6 +80,16 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
             $firewall['guard']['entry_point'] = $chosenEntryPoint ?? current($firewall['guard']['authenticators']);
         }
 
+        if (!isset($firewall['logout']) && $logoutSetup) {
+            $firewall['logout'] = ['path' => 'app_logout'];
+            $firewall['logout'][1] = $this->manipulator->createCommentLine(
```
weaverryan: Just use `$firewall['logout'][]` here and below - it'll append without needing the index.

weaverryan:
This is an absolute pleasure to merge - wonderful work. Thank you @St0iK!

St0iK:
🙌 Thank you very much for your help and for taking the time to review this @weaverryan . 😃

PR #407, opened 2019-05-27
Fixed error message when class is unknown on creating relation

eiannone:
When creating a new relation, if the requested class is not found, the error message should refer to the answered class, not to the $targetEntityClass variable, as it is now, because this variable is always null at this point in the code, as per while condition, and it always prints an empty string in the message for the class.


weaverryan:
Excellent catch and fix! Thank you Emanuele!

PR #409, opened 2019-05-29
Update PHPDoc in Repository.tpl

Tarasovych:
Add missed `@param` directive for `$value` parameter


weaverryan:
Hi @Tarasovych!

Thanks for the PR :). The `@param` in this case it's actually *not* missing. Coding styles are subjective, so we follow Symfony's coding standards as often as possible. In this case, adding `@param $value` doesn't add any information. Basically, phpdoc is only added if it adds *new* information that can't be communicated in code, which basically comes down to these two situations:

1) Situations where the return type/argument type can't be expressed in code - this isn't very common, it's usually when something is an `array` but you want to communicate (with phpdoc) that, for example, it's an array of some object (`MyEntity[]`). That's why we have the `@return` on this method.

2) Situations where the return type/argument type IS expressed in code, but you want to add further, human documentation about that argument.

Cheers!

PR #410, opened 2019-06-01
Travis: testing high and low deps

weaverryan:
I realized that we're not testing against low deps. I've taken most of this from WebpackEncoreBundle.

Cheers!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,46 @@
 language: php
-
-dist: trusty
 sudo: false
-
-git:
-    depth: 1
-
-php:
-  - 7.0
-  - 7.1
-  - 7.2
-  - 7.3
-
 cache:
     directories:
         - $HOME/.composer/cache/files
-        - .phpunit
-        - vendor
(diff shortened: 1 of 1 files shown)


weaverryan:
Tests waiting on https://github.com/symfony/recipes/pull/602

pluseg:
@weaverryan Ryan, the symfony/recipes#602 is merged, but tests here aren't passed. We should fix it further, right? Or do you have another plan already?

weaverryan:
@pluseg Even after that merge, the tests are broken for at least one (but probably more) reason that's outside of our control. For example, this PR (https://github.com/symfony/symfony/pull/31988) fixes a bug that's causing some of our generated code to throw 500 errors because if a missing translation contract in 4.3.1. I don't want to fix that in MakerBundle (by adding translation as a dependency to some maker commands) because that's unnecessary and we would be working around that bug.

So, right now, I'm prepping and merging PR's here on MakerBundle and then we can look at the tests and figure out what fixes are needed on our end, and what things we're waiting for upstream :).

PR #411, opened 2019-06-01
Fixing bad check for trait

weaverryan:
Fixes one little detail from #381


tests/Maker/FunctionalTest.php
```diff
@@ -239,7 +239,7 @@ public function getCommandTests()
             ->setFixtureFilesPath(__DIR__.'/../fixtures/MakeFormEmbedableEntity')
         ];
 
-        yield 'functional' => [MakerTestDetails::createTest(
+        yield 'functional_maker' => [MakerTestDetails::createTest(
```
weaverryan: makes the name more unique - so that this test can be run in isolation when needed.

PR #413, opened 2019-06-04
[CRUD] Fix redirect to index after edit

Totof6942:
Remove useless parameter in the edit redirect like as suggested here : https://github.com/symfony/maker-bundle/issues/257#issuecomment-475234045.


weaverryan:
Thank you @Totof6942!

Issue #395, opened 2019-05-21
"BlogPost" hardcoded in templates for Normalizer and Voter


stefan2:
Shouldn't "BlogPost" be replaced by an entity name?

**Normalizer**

https://github.com/symfony/maker-bundle/blob/master/src/Resources/skeleton/serializer/Normalizer.tpl.php#L28
```php
public function supportsNormalization($data, $format = null): bool 
{
    return $data instanceof \App\Entity\BlogPost;
}
```
**Voter** 

https://github.com/symfony/maker-bundle/blob/master/src/Resources/skeleton/security/Voter.tpl.php#L16
```php
protected function supports($attribute, $subject)
{
    // replace with your own logic
    // https://symfony.com/doc/current/security/voters.html
    return in_array($attribute, ['POST_EDIT', 'POST_VIEW'])
        && $subject instanceof \App\Entity\BlogPost;
}
```

pluseg:
@stefan2 @nicolas-grekas I propose adding more comments and turning off generated Normalizer/Voter by default. Please approve and I will create a PR.
```php
public function supportsNormalization($data, $format = null): bool 
{
    // Replace with your own logic
    // See https://symfony.com/doc/current/serializer/custom_normalizer.html
    // return $data instanceof \App\Entity\SomeEntity;
    return false; 
}
```

```php
protected function supports($attribute, $subject)
{
    // Replace with your own logic
    // See https://symfony.com/doc/current/security/voters.html
    // return in_array($attribute, ['POST_EDIT', 'POST_VIEW'])
    //    && $subject instanceof \App\Entity\BlogPost;
    return false;
}
```

weaverryan:
That sounds sensible to me 👍

stefan2:
I am new in this topic and created a normalizer for an entity. In this case the sample code was absolutely conclusive (except the name of the entity). I didn't realize (means didn't read the docs), that "$data" isn't necessarily an (my) entity. Now i know.

So this is fine for me too, thanks for the lesson :-)

PR #414, opened 2019-06-06
Turn off generated Voter and Normalizer by default

pluseg:
It fixes https://github.com/symfony/maker-bundle/issues/395. 
There are hardcoded non-existent classes for Voter and Normalizer.


lyrixx:
I saw the issue, But with this patch, I will have to edit more code. I don't see the benefit here.

Instead I would try to guess the entity name base on the Voter FQCN

pluseg:
@lyrixx I think that FQCN doesn't require to mention an entity name. 
> I will have to edit more code
It's the same amount, just delete comments and place your own logic. Or uncomment the example.

ps. I've noticed that I forgot a `return false;` statement. My bad.

PR #416, opened 2019-06-10
[Voter] [Normalizer] Turn off generated Voter and Normalizer by default, also fix failed test

### LINKED ISSUE #395: "BlogPost" hardcoded in templates for Normalizer and Voter

pluseg:
It fixes #395.
There are hardcoded non-existent classes for Voter and Normalizer.
Also, it fixes one failed test of Validator.

@lyrixx Sorry, I had to recreate the previous PR — https://github.com/symfony/maker-bundle/pull/414.

### DIFF
--- a/src/Resources/skeleton/security/Voter.tpl.php
+++ b/src/Resources/skeleton/security/Voter.tpl.php
@@ -10,10 +10,13 @@ class <?= $class_name ?> extends Voter
 {
     protected function supports($attribute, $subject)
     {
-        // replace with your own logic
-        // https://symfony.com/doc/current/security/voters.html
-        return in_array($attribute, ['POST_EDIT', 'POST_VIEW'])
-            && $subject instanceof \App\Entity\BlogPost;
+        // Replace with your own logic
+        // See https://symfony.com/doc/current/security/voters.html
+        //
+        // return in_array($attribute, ['POST_EDIT', 'POST_VIEW'])
+        //     && $subject instanceof \App\Entity\YourEntity;
+
+        return false;
     }
 
     protected function voteOnAttribute($attribute, $subject, TokenInterface $token)
--- a/src/Resources/skeleton/serializer/Normalizer.tpl.php
+++ b/src/Resources/skeleton/serializer/Normalizer.tpl.php
@@ -25,6 +25,11 @@ class <?= $class_name ?> implements NormalizerInterface
 
     public function supportsNormalization($data, $format = null): bool
     {
-        return $data instanceof \App\Entity\BlogPost;
+        // Replace with your own logic
+        // See https://symfony.com/doc/current/serializer/custom_normalizer.html
+        //
+        // return $data instanceof \App\Entity\YourEntity;
+
+        return false;
     }
 }
--- a/tests/ValidatorTest.php
+++ b/tests/ValidatorTest.php
@@ -66,7 +66,9 @@ class ValidatorTest extends TestCase
     public function testInvalidEncodingInClassName()
     {
         $this->expectException(RuntimeCommandException::class);
-        $this->expectExceptionMessage('"�Controller" is not a UTF-8-encoded string.');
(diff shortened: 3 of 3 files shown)


src/Resources/skeleton/security/Voter.tpl.php
```diff
@@ -10,10 +10,13 @@ class <?= $class_name ?> extends Voter
...
     protected function supports($attribute, $subject)
     {
-        // replace with your own logic
-        // https://symfony.com/doc/current/security/voters.html
-        return in_array($attribute, ['POST_EDIT', 'POST_VIEW'])
-            && $subject instanceof \App\Entity\BlogPost;
+        // Replace with your own logic
+        // See https://symfony.com/doc/current/security/voters.html
+        //
+        // return in_array($attribute, ['POST_EDIT', 'POST_VIEW'])
```
nicolas-grekas: Maybe replace POST by something else? It means "after" also, this can be confusing.

src/Resources/skeleton/security/Voter.tpl.php
```diff
@@ -10,10 +10,13 @@ class <?= $class_name ?> extends Voter
...
+        //     && $subject instanceof \App\Entity\YourEntity;
+
+        return false;
```
nicolas-grekas: I'm not sure this is better: the now commented line were fine uncommented, no? Because there is this line to remove now, this might confuse some.

src/Resources/skeleton/serializer/Normalizer.tpl.php
```diff
@@ -25,6 +25,11 @@ public function normalize($object, $format = null, array $context = array()): ar
 
     public function supportsNormalization($data, $format = null): bool
     {
-        return $data instanceof \App\Entity\BlogPost;
+        // Replace with your own logic
+        // See https://symfony.com/doc/current/serializer/custom_normalizer.html
+        //
+        // return $data instanceof \App\Entity\YourEntity;
```
nicolas-grekas: Same note as above, it'd make sense uncommenting this to me.

lyrixx: Totally agree, That's what I said in the previous PR.

weaverryan: Hmm. I agree. The current example code seems pretty clear, and it's better than needing to know to remove that `return false`. I think we should leave as-is or add a second question to each maker:

> What class will this voter/normalizer handle? (e.g. BlogPost or leave blank)

(if a non-FQN were passed, we'd look for it in `Entity`). If a class is passed, we use that in the generated code. If none is passed, we leave the class-checking part off of the voter and, for the normalizer, probably put a big TODO in `supportsNormalization()`.

WDYT?

src/Resources/skeleton/security/Voter.tpl.php
weaverryan: Good catch. At the very least, we could change this to `BLOG_POST_EDIT`... though there is also no rule that says that things need to be upper case...

If we add a question about the entity/class, then we would derive this from the entity class name.

maxhelias:
Feature merge in #672, this one should be closed

PR #417, opened 2019-06-12
[make:functional-test] Use Panther when available.

adrienlucas:
Hope this will be useful for others.

Let me know if there are some mistakes to fix !

### DIFF
--- a/src/Maker/MakeFunctionalTest.php
+++ b/src/Maker/MakeFunctionalTest.php
@@ -21,6 +21,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\CssSelector\CssSelectorConverter;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestAssertionsTrait;
+use Symfony\Component\Panther\PantherTestCaseTrait;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
@@ -55,6 +56,7 @@ class MakeFunctionalTest extends AbstractMaker
             'test/Functional.tpl.php',
             [
                 'web_assertions_are_available' => trait_exists(WebTestAssertionsTrait::class),
+                'panther_is_available' => trait_exists(PantherTestCaseTrait::class),
             ]
         );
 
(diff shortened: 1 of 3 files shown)


src/Resources/skeleton/test/Functional.tpl.php
```diff
@@ -2,9 +2,13 @@
 
 namespace <?= $namespace; ?>;
 
+<?php if($panther_is_available) { ?>
+use Symfony\Component\Panther\PantherTestCase;
+<?php } else { ?>
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+<?php } ?>
```
weaverryan: Restructure the if statement like this: https://github.com/symfony/maker-bundle/blob/master/src/Resources/skeleton/form/Type.tpl.php#L5-L7

So, `<?php if($panther_is_available): ?>` and similar changes on all the lines

adrienlucas: done !  
[i also fixed](https://github.com/symfony/maker-bundle/pull/417/commits/367d604345c45b1ae1faee0b83b573c52cd11783) the other `if` statement to follow the same style.

src/Maker/MakeFunctionalTest.php
```diff
@@ -55,6 +56,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             'test/Functional.tpl.php',
             [
                 'web_assertions_are_available' => trait_exists(WebTestAssertionsTrait::class),
+                'panther_is_available' => class_exists(PantherTestCaseTrait::class),
```
weaverryan: Ah! one more thing: this needs to use `trait_exists()`.

src/Maker/MakeFunctionalTest.php
```diff
@@ -82,5 +84,11 @@ public function configureDependencies(DependencyBuilder $dependencies)
             true,
             true
         );
+        $dependencies->addClassDependency(
```
weaverryan: And, I realized that we need a small tweak in `DependencyBuilder::calculateMissingDependencies()` - it should now also check for `trait_exists()` in there. Currently, it does not, so it would ALWAYS think this dependency wasn't installed.

adrienlucas: [Fixed there !](https://github.com/symfony/maker-bundle/pull/426)

src/Maker/MakeFunctionalTest.php
adrienlucas: resolved !

weaverryan:
Thank you @adrienlucas!

PR #418, opened 2019-06-12
Fix the 'Invalid encoding' class name validation test.

adrienlucas:
I don't know why this test was written like that, and I did not find other way to fix it...  
Hope I'm not missing something.

### DIFF
--- a/tests/ValidatorTest.php
+++ b/tests/ValidatorTest.php
@@ -66,7 +66,7 @@ class ValidatorTest extends TestCase
     public function testInvalidEncodingInClassName()
     {
         $this->expectException(RuntimeCommandException::class);
-        $this->expectExceptionMessage('"�Controller" is not a UTF-8-encoded string.');
-        Validator::validateClassName(mb_convert_encoding('Ś', 'ISO-8859-2', 'UTF-8'));
+        $this->expectExceptionMessage(sprintf('"%sController" is not a UTF-8-encoded string.', chr(0xA6)));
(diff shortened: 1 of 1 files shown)


pluseg:
@adrienlucas Please be aware of code style rule of Symfony about concatination in exceptions:
https://symfony.com/doc/current/contributing/code/standards.html#structure
> Exception and error message strings must be concatenated using sprintf;

You could check it in my PR — https://github.com/symfony/maker-bundle/pull/416/files#diff-a85bb5d7b405aafe9836ee24c6aee6f7R70

adrienlucas:
@pluseg thanks for your review !

I did not see that your PR was solving this issue as its title was not mentioning it at all.

About the coding style, I think the standard applies only for exception construction in production code, not for `expectExceptionMessage` in test code, am I right ?

pluseg:
@adrienlucas Added that into PR title. Thanks for the comment.

Regarding the standard — it relates to all code that goes to public, I guess. Imho, it should be implemented in a unified way for all code. It eases finding code line by exception message at least.

Cheers!

adrienlucas:
@pluseg ok about searching the message (while it's not really relevant for full text search, as the concatenation was used in the variant part of the message, not the "hard coded" part of it, ie. `is not a UTF-8-encoded string.`), I fixed the code to follow your advice.

@weaverryan should I close this PR and let the fix be embedded in @pluseg 's ?

weaverryan:
Thank you Adrien! I'm going to merge this now to get the tests passing - and allow for some tweaks on that other PR :). Cheers!

PR #419, opened 2019-06-13
Fix test suite : missing dependency symfony/translation for auth tests

adrienlucas:
Because of a change merged in the `symfony/twig-bridge` 13 days ago, the test-suite is KO (see https://github.com/symfony/twig-bridge/commit/52aa76480b775be0f6465b90ca9e3c2dccc8f3cd).

This solve the issue, let me know if there is a better way of doing so.

### DIFF
--- a/src/Test/MakerTestEnvironment.php
+++ b/src/Test/MakerTestEnvironment.php
@@ -133,7 +133,7 @@ final class MakerTestEnvironment
                     \dirname($this->flexPath),
                     [
                         'FLEX_PATH' => $this->flexPath,
-                        'APP_PATH' => $this->path
+                        'APP_PATH' => $this->path,
                     ]
                 )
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks for the PR @adrienlucas. Indeed, this is a tricky one :). After checking into it, I think the bug isn't that we/users now need to install an extra dependency, but that there was a bug in Symfony itself. This was fixed at symfony/symfony#31988 - but it's not been released yet.

To help catch these things faster in the future... and to prove that our tests WILL pass once that is tagged, I've made Travis (in #421) run our test suite against the next, unrelased patch version of Symfony (stable-dev) and the next unreleased minor version (dev). These tests *are* now passing on stable-dev (which, right now means that it's testing Symfony's 4.3 branch).

Cheers!

PR #424, opened 2019-06-19
[CRUD] Prefix route names

iljamulders:
See https://symfony.com/blog/new-in-symfony-4-1-prefix-imported-route-names

> In Symfony 3.4 and 4.0 we added the possibility of prefixing the names of all the routes defined in a controller class with the name option in the main @Route annotation. In the following example, the route names will be blog_index and blog_post:
```php
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;

/** @Route("/blog", name="blog_") */
class BlogController extends Controller
{
    /** @Route("/", name="index") */
    public function indexAction() { ... }

    /** @Route("/posts/{slug}", name="post") */
    public function showAction(Post $post) { ... }
}
```

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -13,12 +13,12 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 
 /**
- * @Route("<?= $route_path ?>")
+ * @Route("<?= $route_path ?>", name="<?= $route_name ?>_")
  */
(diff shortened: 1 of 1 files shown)


weaverryan:
Hi @iljamulders!

This is definitely a valid feature. I personally find that this is harder to read than the more explicit version (and with some, but not a huge benefit). If other people disagree with me, they can say so here :).

Cheers!

TheGarious:
Hi @iljamulders  and @weaverryan, 

I think for this feature is not a good idea. Beginner Symfony can't know maybe we can have got a prefix name in annotation's controller and the rest into annotation's function.

This but for makerbundle is create a simplier structure of Symfony Project.

weaverryan:
Thanks for the suggestion! Closing this one.

PR #425, opened 2019-06-19
Take care of migrations argument

### LINKED ISSUE #393: [Maker] generate migration with many databases / entityManager fails
**Symfony version(s) affected**: 4.2.5

**Description**  
make:migrations argument to select the db and entity are not filled to the doctrine:migrations:diff command , always the default db is used.

**How to reproduce**  
Have a doctrine config with  2 dbal / orm in the same project.
The first one is the default and the second one is another manager as described in the official doc here: https://symfony.com/doc/current/doctrine/multiple_entity_managers.html
I use yaml config.

When using the make:migration command, even if I fill the argument --em=XXX, the default one is always evaluated, the migration is always performed against the default DB.

If I launch the doctrine:migrations:diff command with the same argument em=XXX, the migration is correct.

**Possible Solution**  
Change the argvInput in the Symfony\Bundle\MakerBundle\Maker\makeMigration or something similar to fill maker argument to doctrine arguments

**Additional context**

LeJeanbono:
Fix #393

### DIFF
--- a/src/Maker/MakeMigration.php
+++ b/src/Maker/MakeMigration.php
@@ -65,13 +65,17 @@ final class MakeMigration extends AbstractMaker implements ApplicationAwareMaker
 
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
     {
-        $options = [
-            'db' => $input->getOption('db'),
-            'em' => $input->getOption('em'),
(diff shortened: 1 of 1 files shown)


weaverryan:
Awesome! Thanks @LeJeanbono!

PR #430, opened 2019-06-27
TypeHint getter for doctrine with bigint

### LINKED ISSUE #429: make:entity bigint has int getter/setter types.
This results in all sorts of issues. And a field that can never pass validation if new auto_mapping feature is enabled.

Doctrine maps bigint to string. So the parameter type and return types should also be string.

See this issue for more detail:
https://github.com/symfony/symfony/issues/32151

LeJeanbono:
Fix #429
According to : https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/basic-mapping.html#doctrine-mapping-types

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -1002,11 +1002,13 @@ final class ClassSourceManipulator
             case 'string':
             case 'text':
             case 'guid':
+            case 'bigint':
                 return 'string';
 
             case 'array':
             case 'simple_array':
             case 'json':
+            case 'json_array':
                 return 'array';
 
             case 'boolean':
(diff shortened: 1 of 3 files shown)


src/Util/ClassSourceManipulator.php
```diff
@@ -1002,19 +1002,20 @@ private function getEntityTypeHint($doctrineType)
             case 'string':
             case 'text':
             case 'guid':
+            case 'bigint':
```
xabbuh: I think `decimal` should be moved here too

LeJeanbono: True ! Fixed

weaverryan:
Nice catch - thanks @LeJeanbono!

PR #433, opened 2019-07-02
fix typo of 'bundles'

gdc5:


lyrixx:
Good catch, thanks @gdc5.

PR #438, opened 2019-07-05
Remove php-parser workaround for adding interfaces

acrobat:
The bug where this workaround was needed for was fixed, so let's remove this workaround. This will make sure adding an interface to a class with another interface will also work now.

See: https://github.com/nikic/PHP-Parser/pull/527


weaverryan:
This is awesome! Thanks so much for this @acrobat - excellent PR!

PR #443, opened 2019-07-13
Fix unclosed div on login form skeleton

matthieumota:
A little fix when generate guard auth in login form skeleton. "checkbox" class is also useless.


lyrixx:
Good catch, thanks @matthieumota.

PR #445, opened 2019-07-14
Require doctrine/annotations instead of sensio/framework-extra-bundle where sufficient

### LINKED ISSUE #442: doctrine/annotations dependency tells to require sensio/framework-extra-bundle
When I tried to use `make:controller` I got

> [ERROR] Missing package: to use the make:controller command, run:              
>                                                                                
>         composer require annotations 

which I ran, before realizing it was installing sensio/framework-extra-bundle.

That surprised me because you're no longer supposed to require this bundle to annotate routes. Then I saw

> we only need doctrine/annotations

in https://github.com/symfony/maker-bundle/blob/master/src/Maker/MakeController.php#L89

So why does it ask to install sensio/framework-extra-bundle?

This also concerns `MakeRegistrationForm`.

MatTheCat:
Fixes #442

### DIFF
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -86,10 +86,8 @@ final class MakeController extends AbstractMaker
     public function configureDependencies(DependencyBuilder $dependencies)
     {
         $dependencies->addClassDependency(
-            // we only need doctrine/annotations, which contains
-            // the recipe that loads annotation routes
             Annotation::class,
-            'annotations'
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks @MatTheCat!

MatTheCat:
:ok_hand: 

Do you think https://github.com/symfony/symfony-docs/issues/10426 should be reopened?

PR #446, opened 2019-07-17
The closing div is missing on the logout fragment

Jir4:
just a small oversight I guess


Jir4:
Didn't dig enough in closed PR

PR #448, opened 2019-07-23
For autocomplete json_array type must be after json

### LINKED ISSUE #447: Do not replace json by json_array when making an Entity
When I create an entity with make:entity and type json for the type of a field, it is replaced by json_array, which is deprecated since Doctrine 2.6 if I remember well.

There already has been a pull request to remove the support to json_array, yet I have this behaviour. When I look at the code, the only place (inside the MakerBundle) I can found the json_array string is in src/GeneratorTwigHelper.php but I don't know if this is related, because if there is a possibility of extending the types managed by the maker bundle (didn't search deep enough to know), this behaviour could come from another bundle.

As a side note, the json_array is displayed among the "Other Types" section.

LeJeanbono:
Fix #447 
image
Maybe a little ugly, I'm open to suggestions

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -373,6 +373,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         unset($input[Type::JSON_ARRAY]);
         // Put JSON_ARRAY at the end
         $input[Type::JSON_ARRAY] = $json_array;
+
         return $input;
     }
 
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -319,7 +319,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
 
         $type = null;
         $allValidTypes = array_merge(
-            array_keys(Type::getTypesMap()),
+            array_keys($this->reOrderJsonArray(Type::getTypesMap())),
             EntityRelation::getValidRelationTypes(),
             ['relation']
         );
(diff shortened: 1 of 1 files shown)


src/Maker/MakeEntity.php
```diff
@@ -366,6 +366,17 @@ private function askForNextField(ConsoleStyle $io, array $fields, string $entity
...
+    private function reOrderJsonArray($input)
+    {
+        // JSON must before JSON_ARRAY for autocomplete
+        $json_array = $input[Type::JSON_ARRAY];
+        unset($input[Type::JSON_ARRAY]);
+        // Put JSON_ARRAY at the end
+        $input[Type::JSON_ARRAY] = $json_array;
+
+        return $input;
+    }
```
weaverryan: Hmm. Could we use https://www.php.net/manual/en/function.asort.php

The idea would be that, I think, `json` comes after `json_array` alphabetically... so that should sort it like we want it.

LeJeanbono: `arsort` do the job, thanks @weaverryan

javiereguiluz:
Instead of this ... which I'm not sure I like much as a solution ... couldn't we instead filter the list of Doctrine types by hand and remove the `json_array` type because it's deprecated since Doctrine 2.6 and shouldn't be used anymore in new entities?

romaricdrigon:
I agree with @javiereguiluz suggestion of removing entirely `json_array`.
Maker bundle is moving towards supporting only doctrine 2.6+, released in 2017. We should remove deprecated types.

@LeJeanbono would you like to work on that? I will close this PR since it does not make sense anymore.

PR #450, opened 2019-08-01
[make:entity] Change arguments of adder and remover to variadic and iterate over them

hermann8u:
Hi,

The goal of this PR is to use the powerful variadic operator with the arguments of generated adders and removers in entities.

Example :
``` php
public function addTag(Tag ...$tags): self
{
    foreach ($tags as $tag) {
        if (!$this->tags->contains($tag)) {
            $this->tags[] = $tag;
        }
    }

    return $this;
}
```

This is something we already do in my company and I really like this approach. It allows to quickly add an array of entity in the collection without iterating over it.

Something really great with this change is that it allows to continue using them the old way :

``` php
// This will still works
$article->addTag($tag);

// Or in order to add an array like before
foreach ($tags as $tag) {
    $article->addTag($tag);
}

// Add an array without iterate over items
$article->addTag(...$tags);

// Or even
$article->addTag($firstTag, $secondTag);
```

The generated code is already ok but I didn't write test yet. This is something I think is necessary but 
I'm not used to doing. If someone could lead me in the right file to do it or even contribute, it will be great.

I tried to edit the "tests/Util/fixtures/" files to pass the tests but it didn't change anything. Is it necessary?

Thanks you in advance!

---

**Edit :**
I changed the files used to compare the generated code. Now tests pass, but I should  also add test in order to check if we can use the variadic operator features combine with array unpacking.

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -545,12 +545,14 @@ final class ClassSourceManipulator
             [sprintf('@return %s|%s[]', $collectionTypeHint, $typeHint)]
         );
 
-        $argName = Str::pluralCamelCaseToSingular($relation->getPropertyName());
+        $argName = $relation->getPropertyName();
+        $foreachItemName = Str::pluralCamelCaseToSingular($relation->getPropertyName());
 
         // adder method
         $adderNodeBuilder = (new Builder\Method($relation->getAdderMethodName()))->makePublic();
 
         $paramBuilder = new Builder\Param($argName);
+        $paramBuilder->makeVariadic();
         $paramBuilder->setTypeHint($typeHint);
         $adderNodeBuilder->addParam($paramBuilder->getNode());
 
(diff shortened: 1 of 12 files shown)


weaverryan:
Very cool work @hermann8u!

> but I should also add test in order to check if we can use the variadic operator features combine with array unpacking.

Do you mean, add a test to actually verify that the generated code will work as expected? If so, yea, I agree. Check out something like `tests/fixtures/MakeEntityOneToMany/tests/GeneratedEntityTest.php`. That is a test file that's executed inside a project after running the `make:entity` command. You could enhance that test to show calling `$user->addPhoto()` in different ways.

ismail1432:
Sorry but I’m not convinced by this change and I don’t see a real benefit, now we have an addTag method that allows to add many tagS (same for removeTagg). It’s a bit confusing plus It’s introduce a logical custom as a standard, I mean from my experience I never had to foreach on many objects to add them one by one calling addObject...

weaverryan:
I also asked the Twitter world what they thought about this. A lot of people liked it, but most said then it should be called `addTags()` (and we can't make it plural, as that wouldn't integrate well with the form and serializer component). Also, this would allow people to accidentally call the method with no arguments.

So, while I think it's an interesting approach, I also think it's not a good idea at this time.

Thanks @hermann8u for the suggestion and really nice PR. I hope we'll see you again :)

PR #456, opened 2019-08-13
Use prefix "is" for getters on boolean fields

MaximePinot:
This is subject to discussion as there is no standard about this in the PHP world. There is a ["Methods and Functions"](https://www.php-fig.org/psr/psr-12/#44-methods-and-functions) section in PSR-12 but nothing specific to *getters* and *setters*. However, I believe this is a widely used naming convention...


weaverryan:
Hey @MaximePinot!

In general, I like this idea. The question is, will it integrate with the other parts of Symfony - Twig (`product.published`), serializer & form component. And I all of these *do* support issers. So, I think we can totally do this :).

Could you write a test for this in `ClassSourceManipulatorTest`?

MaximePinot:
Hi @weaverryan,

I'm glad you like the idea!

I added a unit test for this change.

src/Util/ClassSourceManipulator.php
```diff
@@ -185,7 +185,13 @@ public function addAccessorMethod(string $propertyName, string $methodName, $ret
 
     public function addGetter(string $propertyName, $returnType, bool $isReturnTypeNullable, array $commentLines = [])
     {
-        $methodName = 'get'.Str::asCamelCase($propertyName);
+        if ('bool' === $returnType) {
+            $snakeCasedPropertyName = Str::asSnakeCase($propertyName);
+            $propertyNameWithoutPrefix = (0 === strpos($snakeCasedPropertyName, 'is_')) ? substr($snakeCasedPropertyName, 3) : $propertyName;
```
stof: This logic seems wrong to me. We must always add either `get` or `is` as prefix on the property name. Otherwise, we are making this code incompatible with PropertyAccess (causing issues with the usage of Form and other components).

javiereguiluz: That's sad ... because, if I'm right, it's generally recommended to add `is` to methods but never to property names. Couldn't we improve PropertyAccess instead? Thanks.

stof: Well, that case is trying to handle the case where the property itself starts with `is`, by skipping the prefix in such case. That's precisely what PropertyAccess does not support. If you insist on having `is` in the property name, PropertyAccess will expect `isIsFoo` or `getIsFoo` (or `hasIsFoo`).

MakerBundle can choose one of them, but should not generate `isFoo()` as getter.

If your naming follows your recommendation, you are precisely *not* relying on the weird magic that I'm asking to remove.

MaximePinot: As a matter of fact, I agree with @stof as it is more consistent. This is why the handling of `is` in property names is in [a second commit](https://github.com/symfony/maker-bundle/pull/456/commits/505d4fd2a19d209822e424435f0aa75bc2fea7bb). 

I chose to handle property names prefixed with `is` when I saw these two lines:
https://github.com/symfony/maker-bundle/blob/c4388410e2fb6321e77c5dd6e3cb2dba821f9fe6/src/Maker/MakeEntity.php#L314-L315

I thought that the `MakerBundle` would _always_ expect people to prefix boolean properties with `is`. Therefore, it made sense to me to handle this case.

MaximePinot:
Hi,

The Travis CI build failed, but it has nothing to do with my changes. Almost all other opened pull requests encounter the same issue...

Is this pull request ready to be merged? @stof, do you approve these changes? :)

romaricdrigon:
That's correct regarding tests, even master is red. We will be looking into that next Saturday, and I hope after that we can move forward all blocked PRs.

BatsaxIV:
Hi,

This looks good, I think than you can rebase to master @MaximePinot and @weaverryan this can probably be merged WDYT?

MaximePinot:
Hi @BatsaxIV, thanks a lot for your feedback! 

I rebased to master and all checks have passed! (cc @weaverryan)

MaximePinot:
Friendly ping @weaverryan and @stof. One year later, is this PR still relevant? 

(The branch was updated from `main` and the test that I added passed. The two failing checks are not related to this PR. They are also failing on `main`.)

PR #459, opened 2019-08-20
Add return types to generated guard authenticators

LeJeanbono:


stof:
The title of the PR does not describe at all what it does...

What it does is adding return types to the generated guard authenticators (for forward-compatibility with Symfony 5)

romaricdrigon:
Thank you @stof for the explanation.
I have mixed feelings about what to do. 

Purely regarding Symfony, return types won't be enforced in 5.0 (see https://github.com/symfony/symfony/pull/33236) and deprecation warnings may start in 5.1, but I'm not totally sure if they will match this PR (for instance, `UserInterface` does not have any return types at the moment). So there's no rush there.

If we look at this purely from a CS perspective, having return types is nice - as long as Symfony interfaces does not break those later (6.x ?).

What do you think?
Anyhow, @LeJeanbono could you please edit PR title (something like "Added return types"), get rid of the merge commit and squash it?

src/Security/UserClassBuilder.php
```diff
@@ -223,27 +223,33 @@ private function addGetPassword(ClassSourceManipulator $manipulator, UserClassCo
...
-            null,
+            '?string',
             false,
             ['@see UserInterface']
         );
         if ($userClassConfig->hasPassword()) {
-            $builder->addStmt(
+            $getSaltMethodBuilder->addStmt(
                 $manipulator->createMethodLevelCommentNode(
                     'not needed when using the "bcrypt" algorithm in security.yaml'
```
nicolas-grekas: `not needed when using modern algorithms in security.yaml` + corresponding changes in fixtures

LeJeanbono: Done, thanks

LeJeanbono:
Are we ok @romaricdrigon  ?

LeJeanbono:
You're welcome !!

Issue #396, opened 2019-05-21
make:auth should add access_control for the login path


nicolas-grekas:
That would be consistent with https://symfony.com/doc/current/security/form_login_setup.html
```yaml
    access_control:
        - { path: ^/login$, roles: IS_AUTHENTICATED_ANONYMOUSLY }
```

JJarrie:
Hello Nicolas,

Fix this issue look like we need to change depth's rules for switching to inline struct into Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator::convertToYaml() but doing this create some unexpected changes on others keys, for example:
`yaml
security:
    encoders:
        App\Security\User: { algorithm: argon2i }
`
instead of

`yaml
security:
    encoders:
        App\Security\User:
            algorithm: argon2i
`
I'm looking to track the type of value when we adding new key in file and do case by case rule (must be inline when value are array and parent are sequential array, imho), this is potentially Symfony\Component\Yaml\Dumper::dump() issue.

PR #460, opened 2019-08-21
Add access_control rule for form login auth

GaryPEGEOT:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #396 
| License       | MIT

Add `access_control` rule when `make:auth` command is used with login form authenticator

### DIFF
--- a/src/Security/SecurityConfigUpdater.php
+++ b/src/Security/SecurityConfigUpdater.php
@@ -137,7 +137,7 @@ final class SecurityConfigUpdater
         $accessControlRules = $newData['security']['access_control'] ?? [];
 
         foreach ($accessControlRules as $rule) {
-            if ('^/login$' === $rule['path']) {
+            if (0 === strpos($rule['path'], '^/login')) {
                 $addAccessControlRule = false;
                 break;
             }
--- a/src/Security/SecurityConfigUpdater.php
+++ b/src/Security/SecurityConfigUpdater.php
@@ -59,7 +59,7 @@ final class SecurityConfigUpdater
         return $contents;
     }
 
-    public function updateForAuthenticator(string $yamlSource, string $firewallName, $chosenEntryPoint, string $authenticatorClass, bool $logoutSetup, bool $useSecurity52): string
+    public function updateForAuthenticator(string $yamlSource, string $firewallName, $chosenEntryPoint, string $authenticatorClass, bool $logoutSetup, bool $useSecurity52, bool $addAccessControlRule = false): string
     {
         $this->manipulator = new YamlSourceManipulator($yamlSource);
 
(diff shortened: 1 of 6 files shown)


romaricdrigon:
Hello,
Thank you for this PR, sorry for the delay in reviewing it.
The feature sounds nice. Does it support merging rules with already existing acces_control rules? We should not break user config.

GaryPEGEOT:
> 
> 
> Hello,
> Thank you for this PR, sorry for the delay in reviewing it.
> Does it support merging rules with already existing acces_control rules? We should not break user config.

Hi! Yes, it's [done here](https://github.com/symfony/maker-bundle/pull/460/files#diff-f6787c726ea39bb082882861822a762bR98-R103): If an existing config is detected, the line is not added. The test case name is `simple_security_with_access_control`

src/Security/SecurityConfigUpdater.php
```diff
@@ -92,6 +92,21 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
         }
 
         $newData['security']['firewalls'][$firewallName] = $firewall;
+
+        $accessControlRules = $newData['security']['access_control'] ?? [];
+
+        foreach ($accessControlRules as $rule) {
+            if ('^/login$' === $rule['path']) {
```
romaricdrigon: I think that check is a little bit too strict: I think we should check for `^/login`, it's the syntax from the (old) documentation: https://symfony.com/doc/4.0/security/form_login_setup.html#be-sure-the-login-page-isn-t-secure-redirect-loop

tests/Security/yaml_fixtures/expected_authenticator/simple_security_with_added_access_control.yaml
```diff
@@ -0,0 +1,16 @@
...
+        dev: ~
+        main:
+            anonymous: true
+            guard:
+                authenticators:
+                    - App\Security\AppCustomAuthenticator
+    access_control:
+        -
+            path: ^/login$
+            roles: IS_AUTHENTICATED_ANONYMOUSLY
```
romaricdrigon: Is there any way to use the inline syntax (as in `simple_security_with_access_control.yaml`) instead of that one?

romaricdrigon (changes requested):
Great, we are almost there!

src/Security/SecurityConfigUpdater.php
GaryPEGEOT: Fixed!

tests/Security/yaml_fixtures/expected_authenticator/simple_security_with_added_access_control.yaml
GaryPEGEOT: Not that I'm aware of. It was pointed out in the [related issue](https://github.com/symfony/maker-bundle/issues/396#issuecomment-496578320) that this might happen

romaricdrigon: Ok, thank you for the clarification!

romaricdrigon (approved):

weaverryan:
Hiya!

This generally look very good (including the tests). But, I think I need to do a bit of work on `YamlSourceManipulator` to allow for the `access_control` to be inline as we really want. Also, I find a few edge-cases with it - like if the `security.yaml` file looks like the recipe - https://github.com/symfony/recipes/blob/master/symfony/security-bundle/4.4/config/packages/security.yaml - where it has `access_control` set to no value, then with 2 lines of comments, those comments end up in a screwy place. I'd like to get that done, then this PR (which is already perfect for what it does) should be ready

maxhelias (approved):

maxhelias:
A rebase is required

GaryPEGEOT:
> A rebase is required

Rebase done !

Issue #463, opened 2019-08-22
Wrong path in Twig with make:crud


Soonz:
Hey,

Not a big issue but I just noticed that when you were doing **make:crud Community\Comment** the Twig files are created like so : 

- community/comment/edit.html.twig
- community/comment/index.html.twig
- community/comment/new.html.twig
- community/comment/show.html.twig

But the include inside those files are like : 

{{ include('community_community/edit.html.twig') }}

Instead of : 

{{ include('community/comment/edit.html.twig') }}

TheGarious:
Hi,

I look your issue, i find where is a problem.

After that i tag your issue into pr.

Thanks

jrushlow:
fixed in #870

PR #464, opened 2019-08-26
Fix Replace Route Name variable by Templates Path variable for include path into twig template 

TheGarious:
For this [issue](https://github.com/symfony/maker-bundle/issues/463)

When i use command make:crud and my entity is under a directory (Admin/Article for example), lot of twig generate by this command don't have got a good path for include file.

Example from [issue](https://github.com/symfony/maker-bundle/issues/463)  :  

But the includes inside those files are like :

{{ include('community_comment/edit.html.twig') }}

Instead of :

{{ include('community/comment/edit.html.twig') }}


romaricdrigon:
Thank you, good catch.
To prevent a regression later, could you please add a test? You have to edit `FunctionalTest` to add a corresponding scenario, and maybe you will need some fixtures in `tests/fixtures` too.

TheGarious:
@romaricdrigon No Problem,
I work this pr :D 

Thanks

PR #466, opened 2019-08-30
[Entity] Add getReference to generated Repository to give correct type hint

pbowyer:
When using Psalm I need to implement this in every Repository, otherwise an `ArgumentTypeCoercion` Error is reported when I use `$entityManger->getReference()`, as "expects App\Entity\Foo|null, parent type object|null provided".

I'd originally generated my entities and repositories using the Maker bundle, so thought this would be helpful to other developers.

### DIFF
--- a/src/Resources/skeleton/doctrine/Repository.tpl.php
+++ b/src/Resources/skeleton/doctrine/Repository.tpl.php
@@ -22,6 +22,19 @@ class <?= $class_name; ?> extends ServiceEntityRepository<?= $with_password_upgr
         parent::__construct($registry, <?= $entity_class_name; ?>::class);
     }
 
+    /**
+     * Obtain a reference to an entity for which the identifier is known, 
+     * without loading that entity from the database.
+     * @see https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/advanced-configuration.html#reference-proxies
+     * @param mixed $id
+     * @return <?= $class_name; ?>
+     * @throws \Doctrine\ORM\ORMException
+     */
+    public function getReference($id)
+    {
+        return $this->getEntityManager()->getReference(<?= $entity_class_name; ?>::class, $id);
+    }
+
 <?php if ($with_password_upgrade): ?>
     /**
      * Used to upgrade (rehash) the user's password automatically over time.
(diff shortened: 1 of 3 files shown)


pbowyer:
Sorry, I can't work out why Travis CI is failing. The code formatting looks OK when I run `php-cs` on the project or on a generated repository.

weaverryan:
> Sorry, I can't work out why Travis CI is failing

Yea, they can be a pain... even when the failures are correct. We need to continue to improve how those CS errors are reported on failure - I've also had issues debugging this.

> When using Psalm I need to implement this in every Repository, otherwise an ArgumentTypeCoercion Error is reported when I use $entityManger->getReference(), as "expects App\Entity\Foo|null, parent type object|null provided".

That makes sense... but I've actually never used this method before. I understand the use-case, but I think it might be too specific to generate into everyone's repository class. Also, we try not to add more questions to the generators whenever possible. However, we *do* allow for global-level options. So, I would consider this if we had a global-level option to enable it - e.g.

```yaml
maker:
    doctrine:
        include_get_reference_method: true
```

javiereguiluz:
I'm 👎 on this. The reason is that I feel this is just "noise" to make some static analyzer happy. We should focus on average Symfony developers, which is what this bundle target audience is.

Don't get me wrong: I like static analyzers and I think we should follow *some* of their recommendations, but not all. Thanks.

stof:
Thus, in 99% of the cases, using `getReference` is a mistake.
If you cannot guarantee with 100% certainty that the id exists in the DB, you should not use `getReference`. Doctrine will trust you it exists, as the whole point of this method is to skip the DB query.
In most cases, you cannot have that certainty and so using `find` is necessary.

romaricdrigon:
Hello,

I also -1 this.
Technically, you are correct. But there, in the PHPdoc, we reference only a handful of methods to help discovery, we don't aim for exhaustivity (there are so many more methods in Doctrine repositories). I would recommend most users to avoid `getReference()`, as it can lead to brittle code (it does not check existence of object, etc.), and I have the impression it is the consensus there.

So I'm closing this PR, more advanced users can still manually edit PHPDoc to include more `@method`.

PR #467, opened 2019-09-02
Missing return statement in example

snebes:
Missing return statement in example SecurityController.  Added a missing space as well.


weaverryan:
Thank you for catching this and fixing it @snebes! :)

PR #468, opened 2019-09-03
[make:command] Use the initialize method to instantiate SymfonyStyle

hermann8u:
Hi, I'm back with a new PR. Hopefully this one will be more lucky!

The goal of this PR is to modify the command skeleton in order to use the initialize method to instantiate SymfonyStyle.

This is something I discovered in the symfony demo repository and I was surprised to not see it here.

https://github.com/symfony/demo/blob/master/src/Command/AddUserCommand.php#L95

Moreover, the $io variable is now a class property making it easier to use it in others methods added by the developer.

Once again, I didn't write/modify tests yet, but it will come if you tell me it is necessary.

Open to feedback!

### DIFF
--- a/src/Resources/skeleton/command/Command.tpl.php
+++ b/src/Resources/skeleton/command/Command.tpl.php
@@ -13,6 +13,11 @@ class <?= $class_name; ?> extends Command
 {
     protected static $defaultName = '<?= $command_name; ?>';
 
+    /**
+    * @var SymfonyStyle
+    */
+    private $io;
+
     protected function configure()
     {
         $this
@@ -22,19 +27,23 @@ class <?= $class_name; ?> extends Command
         ;
     }
 
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $this->io = new SymfonyStyle($input, $output);
+    }
+
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $io = new SymfonyStyle($input, $output);
         $arg1 = $input->getArgument('arg1');
 
         if ($arg1) {
-            $io->note(sprintf('You passed an argument: %s', $arg1));
+            $this->io->note(sprintf('You passed an argument: %s', $arg1));
         }
 
         if ($input->getOption('option1')) {
             // ...
         }
 
-        $io->success('You have a new command! Now make it your own! Pass --help to see your options.');
+        $this->io->success('You have a new command! Now make it your own! Pass --help to see your options.');
     }
 }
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/command/Command.tpl.php
```diff
@@ -13,6 +13,11 @@ class <?= $class_name; ?> extends Command
 {
     protected static $defaultName = '<?= $command_name; ?>';
 
+    /**
```
weaverryan: I think this might have one too many spaces

javiereguiluz:
Personally I prefer the `$io` as a class property ... but I dislike the `initialize()` method. I don't have strong arguments to defend this. I know this method was created exactly for that ... it's just that I think in most cases it's just "noise" and shouldn't be used.

stof:
I tend to agree with @javiereguiluz. I never use `initialize()` in my commands.

hermann8u:
I understand this point of view. That's why I tend to create an AbstractCommand that does thies initialization for all my command in my project. It also allows me to define a title property and to format all my commands the same way. It could be an alternative to that PR in the Symfony repo.

romaricdrigon:
Hello,

I like the idea of helping new users discover a Symfony "nugget", in the sense that `SymfonyStyle` is a lesser known feature, and it helps having a more concise, more expressive syntax, and a better output consistency.
Though some users, as I'm getting from @javiereguiluz and @stof, may not like it and then they will be forced to edit Maker bundle output everytime.

Then, as a compromise, what about making it configurable?
We could ask user `Do you want to use Symfony Style? Y/n`, and accordingly set it or not.

javiereguiluz:
@romaricdrigon I think it's OK to keep the `$io` variable like it is, so I wouldn't add the question. Also, we should try to minimize the question to not force the user make too many decisions. Also, when possible, we should try to avoid questions where the user needs some knowledge about the question being asked:

`Do you want to use Symfony Style?` What's Symfony Style? Are you going to change the directory structure of my application to match Symfony's Flex directory style? Are you going to install Symfony's CSS styles in my application? Are you going to format the syntax of my PHP code to match Symfony's coding style? etc.

The thing we were discussing in this proposal if we should really introduce the `initialize()` method. It's technically correct of course ... but still.

romaricdrigon:
Right, my bad.
So, I don't have any strong opinion regarding using `initialize` or not. I guess we need more feedback, or maybe a small poll around.

src/Resources/skeleton/command/Command.tpl.php
```diff
@@ -13,6 +13,11 @@ class <?= $class_name; ?> extends Command
...
+    * @var SymfonyStyle
+    */
+    private $io;
```
bbuehrle: Should this property be private?  If a user extends from the command, will they still have access to the `$this->io` property?

sadikoff:
According to time passed since last messages and all messages I think we can close this one. `initialize` method will be really like a noise in the code, it will be ok to use inside some abstract class, not in each generated command so it up to user how it will be organised, and for skeleton generated command it's ok to have just an `$io` variable.

PR #469, opened 2019-09-04
Fixed make:auth when using a custom repository method as UserProvider

romaricdrigon:
Hello,

It is possible in vanilla Symfony to change how the User is retrieved by the `entity` user provider, so you use your own repository - [https://symfony.com/doc/current/security/user_provider.html#using-a-custom-query-to-load-the-user](https://symfony.com/doc/current/security/user_provider.html#using-a-custom-query-to-load-the-user). Then you remove `providers.xxxx.entity.property` key.

Unfortunately, it causes `make:auth` command to crash:
image

This PR patches `InteractiveSecurityHelper`, so it checks `property` key exists before reading from it. No BC break.
I added a test for that very specific use case.

### DIFF
--- a/src/Security/InteractiveSecurityHelper.php
+++ b/src/Security/InteractiveSecurityHelper.php
@@ -114,7 +114,7 @@ authenticators will be ignored, and can be blank.',
 
     public function guessUserNameField(SymfonyStyle $io, string $userClass, array $providers): string
     {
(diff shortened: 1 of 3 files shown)


weaverryan:
That makes perfect sense - was a clear oversight on my part. Thanks for fixing Romaric!

PR #474, opened 2019-09-24
Use a better exception when using make:auth with a form

lyrixx:


romaricdrigon:
Hello,

I have some mixed feelings about using a `BadMethodCallException`.
From PHP manual, `Exception thrown if a callback refers to an undefined method or if some arguments are missing.`, I feel like we don't quite have the same semantic.
Is there any reason why this exception is better? 
Personally, I would prefer a `LogicException`.

lyrixx:
Anyway, just exception is not cool

romaricdrigon:
Sure. Then could you please modify with `LogicException`, and fix failing test?
Thank you

lyrixx:
I fixed it, but it seems that tests are broken

lyrixx:
ping @romaricdrigon

weaverryan:
Thanks Grégoire!

PR #479, opened 2019-10-18
[make:crud] Fix Replace Route Name variable by Templates Path variable for include path into twig template

TheGarious:
For this [issue](https://github.com/symfony/maker-bundle/issues/463)

When i use command make:crud and my entity is under a directory (Admin/Article for example), lot of twig generate by this command don't have got a good path for include file.

Example from [issue](https://github.com/symfony/maker-bundle/issues/463)  :  

But the includes inside those files are like :

{{ include('community_comment/edit.html.twig') }}

Instead of :

{{ include('community/comment/edit.html.twig') }}

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## [v1.41.0](https://github.com/symfony/maker-bundle/releases/tag/v1.41.0)
+
+*May 4th, 2022*
+
+### Feature
+
+- [#1110](https://github.com/symfony/maker-bundle/pull/1110) - [make:user] Don't add to passwork_hashers if default recipe is in use - *@nicolas-grekas*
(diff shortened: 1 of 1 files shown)


ckrack:
@TheGarious Hi, can you fix the php-cs problem (trailing comma in array).
And I think there should be a test, which must fail with the current version.
Take a look at `tests/Maker/MakeCrudTest.php`, you'll probably just need to add a new yield and copy & modify the `crud_basic` fixture to have the entity in a subdirectory.

TheGarious:
Hi @ckrack, yes i work this week, i have got a time this weekend.

jrushlow:
Thanks @TheGarious for getting this going, and I apologize for the delay in getting it merged. I'm going to add some tests to this, and we'll have it in the next release..

PR #480, opened 2019-10-20
Allow Symfony 5 + many cleanups

nicolas-grekas:
Note that having php-cs-fixer in the deps blocks installing Symfony 5 (did I already mention it's a bad practice to add tools as deps? ;) )
But locally when I remove php-cs-fixer (and patch the code to use my local version), all is green, with SF5 deps.


src/Test/MakerTestEnvironment.php
```diff
@@ -350,13 +350,20 @@ private function buildFlexSkeleton()
         // fetch a few packages needed for testing
         MakerTestProcess::create('composer require phpunit browser-kit symfony/css-selector --prefer-dist --no-progress --no-suggest', $this->flexPath)
                         ->run();
+        $this->fs->remove($this->flexPath.'/vendor/symfony/phpunit-bridge');
+
+        if ('\\' !== \DIRECTORY_SEPARATOR) {
+            $this->fs->symlink('../../../../../../vendor/symfony/phpunit-bridge', './vendor/symfony/phpunit-bridge');
+        } else {
+            $this->fs->mirror(\dirname(__DIR__, 2).'/vendor/symfony/phpunit-bridge', $this->flexPath.'/vendor/symfony/phpunit-bridge');
+        }
```
weaverryan: Using the required phpunit bridge was now a problem?

src/Util/ClassSourceManipulator.php
```diff
@@ -1137,22 +1137,22 @@ private function addNodesToSetOtherSideOfOneToOne(RelationOneToOne $relation, Bu
...
+                        $this->createNullConstant(),
+                        new Node\Expr\Variable($relation->getPropertyName())
                     ),
                     $this->createNullConstant(),
                     new Node\Expr\Variable('this')
                 )
             ))
         );
 
-        // if ($newUserProfile !== $user->getUserProfile()) {
```
weaverryan: What prompted needing to flip all of these?

src/Test/MakerTestEnvironment.php
nicolas-grekas: The bridge used for the main test suite and for functional tests must be at exactly the same because they share the same runtime code. This leads to a fatal error right now when main is 4.4 or 5.0 but functional stays at 4.3.

src/Util/ClassSourceManipulator.php
nicolas-grekas: Running php-cs-fixer

weaverryan:
Thanks Nicolas!

src/Test/MakerTestEnvironment.php
GromNaN: I think this is no longer necessary, all the tests inside the temporary project are run in a distinct php process: https://github.com/symfony/maker-bundle/pull/1779

PR #481, opened 2019-10-23
Change edit method from POST to PUT

LaurentMarquet:
According to https://www.keycdn.com/support/put-vs-post and uses throughout of the web, when editing something with HTTP methods we may(should) use PUT instead of POST

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -74,11 +74,13 @@ class <?= $class_name ?> extends <?= $parent_class_name; ?><?= "\n" ?>
     }
 
     /**
-     * @Route("/{<?= $entity_identifier ?>}/edit", name="<?= $route_name ?>_edit", methods={"GET","POST"})
+     * @Route("/{<?= $entity_identifier ?>}/edit", name="<?= $route_name ?>_edit", methods={"GET","PUT"})
      */
     public function edit(Request $request, <?= $entity_class_name ?> $<?= $entity_var_singular ?>): Response
     {
-        $form = $this->createForm(<?= $form_class_name ?>::class, $<?= $entity_var_singular ?>);
+        $form = $this->createForm(<?= $form_class_name ?>::class, $<?= $entity_var_singular ?>, array(
+            'method' => 'PUT',
+        ));
         $form->handleRequest($request);
 
         if ($form->isSubmitted() && $form->isValid()) {


sadikoff:
APIs are more complex I think, and CRUD is just a simple implementation for the standard web. But lets wait for more opinions 😃

LaurentMarquet:
You're right PUT is not used in the html world ;) but neither DELETE which is set in the Route `XXX_delete` (below in the controller template). Both of them are converted to POST.
So to be coherent I would suggest to use the PUT (or remove DELETE), but not really sure that this PR should go further...

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -74,11 +74,13 @@ public function show(<?= $entity_class_name ?> $<?= $entity_var_singular ?>): Re
     }
 
     /**
-     * @Route("/{<?= $entity_identifier ?>}/edit", name="<?= $route_name ?>_edit", methods={"GET","POST"})
+     * @Route("/{<?= $entity_identifier ?>}/edit", name="<?= $route_name ?>_edit", methods={"GET","PUT"})
      */
     public function edit(Request $request, <?= $entity_class_name ?> $<?= $entity_var_singular ?>): Response
     {
-        $form = $this->createForm(<?= $form_class_name ?>::class, $<?= $entity_var_singular ?>);
+        $form = $this->createForm(<?= $form_class_name ?>::class, $<?= $entity_var_singular ?>, array(
```
romaricdrigon: We should use short array syntax in Symfony Style :)

LaurentMarquet: You're right, my old habit... Also because I find it more readable with array() ;)
I'll modify if PR goes further

romaricdrigon:
Hello,

I'm not quite convinced by this change. I will details a few points as food for thoughts.

**On the plus side**: it looks cleaner :)

**Regarding semantic:**

I have the impression HTTP verbs semantic is extremely important over API ; in an UI, it depends. For instance, validation error return an HTML page, not a `400 Bad Request`.
 
Here we don't know for sure:
 -  form will be to the _full_ entity representation (otherwise it should be PATCH)
 - whether it will be idempotent. 

**Regarding HTML**

While researching the topic, I found that HTML `<form>` supports only `GET` and `POST` natively ([MDN link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form)). So it would be a big -1. 

**Regarding DX:**

`PUT` is still not that commonly used. Will it be easy to understand by newcomers?
What are the other pitfalls we could create?
What are other frameworks scaffolding tools doing?

weaverryan:
I also consider someone doing this as a “more advanced feature” that many people shouldn’t need to know or care about. HTTP verbs just aren’t the same in a browser as an API, and there’s no tangible benefit for the user.

-1 from me but thanks for the proposal!

sadikoff:
And about `xxx_delete` route, probably we should think about removing DELETE method from it, and use more simple post with `<entity>/<id>/delete` POST route

weaverryan:
Hmm, I missed the inconsistent with DELETE that was mention - apologies.

The DELETE trick is more hidden from the user iirc - I don’t love that we need to show PUT in the controller. It’s totally inconsistent, but I’d still like to keep it as it is now.

LaurentMarquet:
I agree with comments above. **The big thing is that only GET and POST are supported by html.** (What I didn't suspect before this discussion.)
I found this because I'm working on making a course on Symfony to deploy an API but I also wanted to show Twig part, so using the `make:crud` and finding `DELETE` verb, I was thinking to "correct" a "mistake" by adding the `PUT ` verb, but we can take rid of this PR, no problem.

weaverryan:
Thanks for the clarification - cheers!

PR #486, opened 2019-10-26
Changed placeholder to real label

ThomasLandauer:
Although common, placing the label into the placeholder is bad practice and reduces usability and accessibility. If people wanna do it this way, they still can; but it shouldn't be suggested :-)

If you merge it, I'll update the docs at https://symfony.com/doc/current/security/form_login_setup.html#generating-the-login-form too.

### DIFF
--- a/src/Resources/skeleton/authenticator/login_form.tpl.php
+++ b/src/Resources/skeleton/authenticator/login_form.tpl.php
@@ -17,10 +17,10 @@
 <?php endif; ?>
 
     <h1 class="h3 mb-3 font-weight-normal">Please sign in</h1>
-    <label for="input<?= ucfirst($username_field); ?>" class="sr-only"><?= $username_label; ?></label>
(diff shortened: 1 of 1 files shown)


romaricdrigon:
Hello,
Thanks for the feedback and the PR. 
I'm having mixed feelings, because we still had a `<label>` for screen reader, so it was IMO it was OK regarding accessibility. Only thing is that `sr-only` class require Bootstrap ; so maybe we should make default theme less Bootstrap dependant? Really, I don't know.

ThomasLandauer:
Just some quick links why placeholders should be avoided:

https://uxdesign.cc/alternatives-to-placeholder-text-13f430abc56f
https://joshuawinn.com/ux-input-placeholders-are-not-labels/
https://uxplanet.org/10-rules-for-efficient-form-design-e13dc1fb0e03

weaverryan:
This looks reasonable - and we're still using Bootstrap markup, so sure - let's do it :). Thanks Thomas!

weaverryan:
Just tagged a patch release - so could you please update the docs @ThomasLandauer? Thanks!

ThomasLandauer:
Update the docs at 5.0 master https://symfony.com/doc/master/security/form_login_setup.html or which branch?

PR #487, opened 2019-11-03
Yaml source manipulator - fix access control

weaverryan:
This helps #460 by fixing a few bugs with YamlSourceManipulator. But, it doesn't yet allow for the `access_control` array structure to be printed in an inline format (https://github.com/symfony/maker-bundle/pull/460#issuecomment-548923526). 

Cheers!

### DIFF
--- a/src/Util/YamlSourceManipulator.php
+++ b/src/Util/YamlSourceManipulator.php
@@ -461,11 +461,29 @@ class YamlSourceManipulator
             // empty space between key & value
             $newYamlValue = ' '.$newYamlValue;
         }
-        $newContents = substr($this->contents, 0, $this->currentPosition)
-            .$newYamlValue
-            .substr($this->contents, $endValuePosition);
 
         $newPosition = $this->currentPosition + \strlen($newYamlValue);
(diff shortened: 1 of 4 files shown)


src/Util/YamlSourceManipulator.php
```diff
@@ -461,12 +461,30 @@ private function changeValueInYaml($value)
             // empty space between key & value
             $newYamlValue = ' '.$newYamlValue;
         }
+
+        $newPosition = $this->currentPosition + \strlen($newYamlValue);
+        $isNextContentComment = $this->isPreviousLineComment($newPosition);
+        if ($isNextContentComment) {
+            $newPosition++;
+        }
```
GaryPEGEOT: Shouldn't it be a `while` in case of multi-line comment?

weaverryan: The `$newPosition++` here is just because of the code below that adds *1* new character (`\n`) if the next line is a comment. I'm moving this pointer/position forward 1 for that 1 `\n`. It's confusing code :). The test case actually does have 2 lines of comment.

PR #497, opened 2019-11-23
[sfhackday2019] Fixing tests

sadikoff:
Improve symlinking phpunit-bridge inside testing flex project.


weaverryan:
Thank you for your detailed work on this Vladimir and Romaric - really happy to have the tests in a better state (even if we're waiting for some Symfony5-related tthings still)

PR #514, opened 2019-12-05
[WIP] adding DisableAutoMapping code to registration form

weaverryan:
If/when we merge the auto-validation recipe, the `make:registration-form` should disable validation on the `password` field. We could/should also try to add the annotation in `make:user` (but it's only possible if you already have the validator installed).

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -252,7 +252,8 @@ final class MakeRegistrationForm extends AbstractMaker
             );
             $userManipulator->setIo($io);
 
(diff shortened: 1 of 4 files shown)


weaverryan:
This should be more-or-less done, but it's blocked by symfony/symfony#34694 and I also opened symfony/symfony#34860

dunglas:
Can't this one be merged now?

weaverryan:
I think this isn't "really" needed, as the auto-validation is "opt in" and I think most people are not enabling it globally anyways.

dunglas:
We cannot enable it by default in the recipe because this PR isn't merge. Everything else is ready.

Issue #513, opened 2019-12-05
Could not find a Composer autoloader that autoloads from ...


michaljusiega:
Hello: I tried to use make:entity command from this bundle. I got 

`Could not find a Composer autoloader that autoloads from 'App\'`

my command:

`php bin/console make:entity --regenerate`

Some changes was approved ? I saw the last one 4 months ago.

michaljusiega:
I found a solution, because the `ComposerClassLoader` will be always expected on second level. In my example `ComposerClassLoader` was on fourth level, so for current version of bundle, the loader is not found.

PR #515, opened 2019-12-06
[Bug] Potential error of Composer Autoloader Finder

michaljusiega:
Hi,
I am very pleased to help you develop this component.
An error occurred while generating additional methods for the entity in #513 
Loader has always been searched strictly at the second level. My patch searches for this `ClassLoader` class through a recursive function.

### DIFF
--- a/src/Util/ComposerAutoloaderFinder.php
+++ b/src/Util/ComposerAutoloaderFinder.php
@@ -77,18 +77,13 @@ class ComposerAutoloaderFinder
     /**
      * @return ClassLoader|null
      */
-    private function extractComposerClassLoader(array $autoloader)
+    private function extractComposerClassLoader(array $autoloaders)
     {
-        if (isset($autoloader[0]) && \is_object($autoloader[0])) {
-            if ($autoloader[0] instanceof ClassLoader) {
-                return $autoloader[0];
-            }
-            if (
-                ($autoloader[0] instanceof DebugClassLoader
-                    || $autoloader[0] instanceof ErrorHandlerDebugClassLoader)
-                && \is_array($autoloader[0]->getClassLoader())
-                && $autoloader[0]->getClassLoader()[0] instanceof ClassLoader) {
-                return $autoloader[0]->getClassLoader()[0];
+        foreach ($autoloaders as $autoloader) {
+            if (($autoloader instanceof DebugClassLoader || $autoloader instanceof ErrorHandlerDebugClassLoader) && \is_array($autoloader->getClassLoader())) {
+                return $this->extractComposerClassLoader($autoloader->getClassLoader());
+            } elseif ($autoloader instanceof ClassLoader) {
+                return $autoloader;
             }
         }
 
--- a/src/Util/ComposerAutoloaderFinder.php
+++ b/src/Util/ComposerAutoloaderFinder.php
@@ -82,7 +82,9 @@ class ComposerAutoloaderFinder
         foreach ($autoloaders as $autoloader) {
             if (($autoloader instanceof DebugClassLoader || $autoloader instanceof ErrorHandlerDebugClassLoader) && \is_array($autoloader->getClassLoader())) {
                 return $this->extractComposerClassLoader($autoloader->getClassLoader());
-            } elseif ($autoloader instanceof ClassLoader) {
+            }
+            
+            if ($autoloader instanceof ClassLoader) {
                 return $autoloader;
             }
         }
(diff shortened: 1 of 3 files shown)


michaljusiega:
cc: @weaverryan , can you look ASAP ? : >

weaverryan:
Hi! Could you add a test for this? Thanks for the PR!

michaljusiega:
It will be hard, but I will try.

michaljusiega:
In general, this situation may occur that if someone has Symfony 4.4 and DebugComponent with the new ErrorHandler component at the same time, the autoloader is not in the first place according to the earlier code.

So the use of foreach with recursive function is very flexible in this case. If someone does not have this component - there is only one ClassLoader - in this case it will be loaded. If there is a Debug or ErrorHandler component then the component has this ComposerClassLoader in it.

If there are two components, the situation looks like one component has information about the other and this component has information about the ClassLoader being searched for.

michaljusiega:
CI/CD not passed but errors not related with my changes :(

src/Util/ComposerAutoloaderFinder.php
```diff
@@ -77,18 +77,13 @@ private function findComposerClassLoader()
...
-            if (
-                ($autoloader[0] instanceof DebugClassLoader
-                    || $autoloader[0] instanceof ErrorHandlerDebugClassLoader)
-                && \is_array($autoloader[0]->getClassLoader())
-                && $autoloader[0]->getClassLoader()[0] instanceof ClassLoader) {
-                return $autoloader[0]->getClassLoader()[0];
+        foreach ($autoloaders as $autoloader) {
+            if (($autoloader instanceof DebugClassLoader || $autoloader instanceof ErrorHandlerDebugClassLoader) && \is_array($autoloader->getClassLoader())) {
+                return $this->extractComposerClassLoader($autoloader->getClassLoader());
+            } elseif ($autoloader instanceof ClassLoader) {
```
weaverryan:
```suggestion
            }
            
            if ($autoloader instanceof ClassLoader) {
```

no `else` needed because of the `return`.

tests/Util/ComposerAutoloaderFinderTest.php
```diff
@@ -59,6 +59,21 @@ public function testGetClassLoaderWhenItIsEmpty()
         (new ComposerAutoloaderFinder(static::$rootNamespace))->getClassLoader();
     }
 
+    public function testGetClassLoaderAfterEnabledSymfonyDebug()
+    {
+        if (\class_exists('Symfony\Component\Debug\Debug')) {
+            \Symfony\Component\Debug\Debug::enable();
```
weaverryan: we need to put this into `require-dev` so that class IS available

tests/Util/ComposerAutoloaderFinderTest.php
```diff
@@ -59,6 +59,21 @@ public function testGetClassLoaderWhenItIsEmpty()
         (new ComposerAutoloaderFinder(static::$rootNamespace))->getClassLoader();
     }
 
+    public function testGetClassLoaderAfterEnabledSymfonyDebug()
```
weaverryan: Add `@group legacy` to avoid deprecation warnings

tests/Util/ComposerAutoloaderFinderTest.php
```diff
@@ -59,6 +59,21 @@ public function testGetClassLoaderWhenItIsEmpty()
...
+    {
+        if (\class_exists('Symfony\Component\Debug\Debug')) {
+            \Symfony\Component\Debug\Debug::enable();
+        }
+
+        if (\class_exists('Symfony\Component\ErrorHandler\DebugClassLoader')) {
+            \Symfony\Component\ErrorHandler\DebugClassLoader::enable();
+        }
+
+        $loader = (new ComposerAutoloaderFinder(static::$rootNamespace))->getClassLoader();
```
weaverryan: After running the tests, I see (in red) `THE ERROR HANDLER HAS CHANGED`. We need to "replace" the error handlers at the end of the test - whether it fails or not. Probably at the start we need something like:

```
$handler = set_error_handler('var_dump');
$handler = \is_array($handler) ? $handler[0] : null;
restore_error_handler();
```

That is (I believe) the sort of "hacky" way to get a reference to the current error handler. Then, at the end, we need to `set_error_handler($handler)` to "replace" it. We need to do that if the test passes or is successful.

michaljusiega: Waiting for green from CI

michaljusiega:
Still CI/CD not passed, but I think the command is not related to my changes.
Let's finish this PR please.

PR #516, opened 2019-12-08
[make:migration] add ability to specify a configuration file

### LINKED ISSUE #508: Support configuration file when making migrations
The current project I'm working on has two sets of migration systems which have completely different migration files for different tables. 

The doctrine commands accept parameters, such as --em, --connection, --configuration, which makes it possible to support multiple sets of migration systems.

When using maker to make migrations, I also want to change the configurations such as 'migrations_directory' or 'migrations_namespace'. Is that possible by using the maker command?

LeJeanbono:
Fix #508


LeJeanbono:
@weaverryan Can we go for it ?

jrushlow:
@LeJeanbono can you rebase please? Thanks in advance!

LeJeanbono:
Thanks Jesse !

PR #518, opened 2019-12-12
Fixing deprecation

nikophil:
Hi,

just fixing a deprecation from doctrine`Doctrine\Common\Persistence` to `Doctrine\Persistence`

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -11,16 +11,20 @@
 
 namespace Symfony\Bundle\MakerBundle\Doctrine;
 
-use Doctrine\Common\Persistence\ManagerRegistry;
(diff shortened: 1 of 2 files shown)


nikophil:
Actually the problem was a little more tricky: we need to fix the deprecation to have the CI green, but we still need to support the version the bundles pretends to support. So, the object `\Symfony\Bundle\MakerBundle\Doctrine\DoctrineHelper` needs to work with both new and deprecated dependencies

LeJeanbono:
Hi @nikophil , can you had `Doctrine\Common\Persistence\ObjectManager` please ?
From #528 
Thanks

weaverryan:
Sorry for the delay on merging this! Thanks @nikophil! I think there may be at least one other spot in the code (maybe more) that need to be update, but let's merge this and then take a look.

PR #520, opened 2019-12-16
Allow user to create Custom Skeleton Dir 

MagnusDot:
This is linked to the issue #33

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -131,24 +131,25 @@ final class MakeCrud extends AbstractMaker
 
         $routeName = Str::asRouteName($controllerClassDetails->getRelativeNameWithoutSuffix());
         $templatesPath = Str::asFilePath($controllerClassDetails->getRelativeNameWithoutSuffix());
+        $customSkeletonDir=(file_exists(getcwd()."/templates/bundles/MakerBundle/skeleton/")?getcwd()."/templates/bundles/MakerBundle/skeleton/":"");
 
         $generator->generateController(
             $controllerClassDetails->getFullName(),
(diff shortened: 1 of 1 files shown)


sadikoff:
I totally disagree with such changes! It should be implemented in another way, to allow changing all skeleton templates. 

I'll advise opening an issue about it, and start a discussion to find out how it should be implemented or even should it be implemented or no.

MagnusDot:
> I totally disagree with such changes! It should be implemented in another way, to allow changing all skeleton templates.
> 
> I'll advise opening an issue about it, and start a discussion to find out how it should be implemented or even should it be implemented or no.

An issue is already opened #33

sadikoff:
@MagnusDot Whoops missed it, thanks for the link!

But it doesn't change my opinion on it. I think it shouldn't be implemented only for the CRUD generator.

MagnusDot:
Owever for the moment nobody done the possibility to have our own skeleton files, 
I propose my solution but if someone wants to do it in another way, free to him

johnpancoast:
Not to knock anyone's work, but for what it's worth, I don't think allowing devs to change the directory is the best approach. See my [new comment](https://github.com/symfony/maker-bundle/issues/33). I think this is/was a perfect use case for Twig.

jrushlow:
Thanks for the PR, but I'm going to close this - I do not think this is the direction we should go. Yes, it would be nice to be able to override the templates we have in MakerBundle for something like `make:crud`. However, in doing so, I'm worried it would increase our maintenance burden. By shifting that responsibility to the consumers of MakerBundle - via the usage `AbstractMaker` - we can keep focusing on iterating over new features/fixes instead of working around potential BC to support customizable templates.

PR #524, opened 2020-01-01
Do not use (deprecated) class, to be removed in future releases.

vaibhavpandeyvpz:
Since doctrine itself recommends to use `Doctrine\Persistence\ManagerRegistry`, lets not use `Doctrine\Common\Persistence\ManagerRegistry`. This should be an easy merge as they seem to be signature compatible.


weaverryan:
Fully fixed in #570 :). Thanks!

PR #526, opened 2020-01-09
Update success message in MakeMigration

PGLongo:
I've always found annoying and frustrating that running `make:entity` the output is 

>  Next: When you're ready, create a migration with `make:migration `

instead of 

>  Next: When you're ready, create a migration with `php bin/console make:migration `

As ` make:migration` does

>  Then: Run the migration with `php bin/console doctrine:migrations:migrate`


weaverryan:
Sorry for the slow merge! This is very sensible - thank you @PGLongo!

PR #529, opened 2020-01-15
Fix user class name on upgradePassword method of user repository

pyrech:


lyrixx:
Good catch, thanks Loïck.

PR #532, opened 2020-01-21
Support class inheritance in UserProvider::supportsClass by default

MisatoTremor:
I propose to add the check if given class name is inherited from the User class (like the EntityUserProvider implementation does) to the UserProvider template.

Also I would follow up with a documentation PR reflecting that change in https://github.com/symfony/symfony-docs/blob/master/security/user_provider.rst

Or would it better to just add a info about this additional check to the documentation?

### DIFF
--- a/src/Resources/skeleton/security/UserProvider.tpl.php
+++ b/src/Resources/skeleton/security/UserProvider.tpl.php
@@ -59,7 +59,7 @@ class <?= $class_name ?> implements UserProviderInterface<?= $password_upgrader 
      */
     public function supportsClass($class)
     {
-        return <?= $user_short_name ?>::class === $class;
+        return <?= $user_short_name ?>::class === $class || is_subclass_of($class, <?= $user_short_name ?>::class);
     }
 <?php if ($password_upgrader): ?>
 


wouterj:
There is nothing directly wrong with doing `AppUser::class === $class`. There is something wrong if you have multiple user classes extending from `AppUser` or if you do `UserInterface::class === $class`.

So I'm leaning towards keeping the current `AppUser::class === $class` template: It's what most people need, doesn't involve a function call and making this call more strict is better than making it too loose.

MisatoTremor:
And it's wrong if e.g. using an ORM to load the user or other cases where you can end up getting a proxy object (which seems to be the main issue lately).

As for the strictness, that's indeed a valid point ...

My alternate proposition was to add a note to the Custom User Provider documentation (and maybe add a link to that in the nextSteps message). How about that?

wouterj:
> And it's wrong if e.g. using an ORM to load the user or other cases where you can end up getting a proxy object (which seems to be the main issue lately).

Oh, I missed this. Then definitely :+1: (99.999% of the applications uses some sort of Doctrine library to save the user I think)

javiereguiluz:
I also suffered this problem. Every page load resulted in this error: `There is no user provider for user "Proxies\__CG__\App\Entity\User.`  After a page reload, the error disappears ... but that means that I must load each page twice to use my app.

@linaori kindly pointed me to this PR, so I applied the proposed fix and everything works now as it should. Please, consider merging this as soon as possible. Thanks!

joshlopes:
Same issue as @javiereguiluz mentioned, it was pointed to me by someone from support sf channel. It appeared with the new sf minor update -- tests started failing on my app. 

This fixed the issue, thanks.

weaverryan:
Thanks @MisatoTremor! Sorry for the delay - I don't know how I missed this PR!

PR #534, opened 2020-01-30
Replace hard coded PasswordResetToken with class name variable

ker0x:


weaverryan:
Thanks Romain! FYI, I'll be doing some refactoring of this feature before we make a release that uses it. Cheers!

PR #536, opened 2020-01-31
[csm] Adding test case and fix for syntax error when updating __construct()

### LINKED ISSUE #535: Maker Bundle fails to create ManyToOne relation
_Originally posted by @Aracon in https://github.com/symfony/maker-bundle/issues/329#issuecomment-580379873
(Creating new issue because can't reopen existing one.)_

Here is the console log:

```
>php bin/console make:entity

 Class name of the entity to create or update (e.g. GentleChef):
 > TransportCostRule

 > settingEntity

 > relation

 What class should this entity be related to?:
 > Setting

What type of relationship is this?
 ------------ --------------------------------------------------------------------------------
  Type         Description

 ------------ --------------------------------------------------------------------------------
  ManyToOne    Each TransportCostRule relates to (has) one Setting.

               Each Setting can relate to (can have) many TransportCostRule objects


  OneToMany    Each TransportCostRule can relate to (can have) many Setting objects.
               Each Setting relates to (has) one TransportCostRule



  ManyToMany   Each TransportCostRule can relate to (can have) many Setting objects.
               Each Setting can also relate to (can also have) many TransportCostRule objects


  OneToOne     Each TransportCostRule relates to (has) exactly one Setting.

               Each Setting also relates to (has) exactly one TransportCostRule.

 ------------ --------------------------------------------------------------------------------

 Relation type? [ManyToOne, OneToMany, ManyToMany, OneToOne]:
 > ManyToOne

 Is the TransportCostRule.settingEntity property allowed to be null (nullable)? (yes/no) [yes]:
 > no

 Do you want to add a new property to Setting so that you can access/update TransportCostRule objects from it - e.g. $setting->getTransportCostRules()? (yes/no) [yes]:
 >

 A new property will also be added to the Setting class so that you can access the related TransportCostRule objects from it.

 New field name inside Setting [transportCostRules]:
 >

 Do you want to activate orphanRemoval on your relationship?
 A TransportCostRule is "orphaned" when it is removed from its related Setting.
 e.g. $setting->removeTransportCostRule($transportCostRule)

 NOTE: If a TransportCostRule may *change* from one Setting to another, answer "no".

 Do you want to automatically delete orphaned App\Entity\TransportCostRule objects (orphanRemoval)? (yes/no) [no]:
 > yes


In ParserAbstract.php line 315:

  Syntax error, unexpected T_VARIABLE, expecting '(' on line 61


make:entity [-a|--api-resource] [--regenerate] [--overwrite] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<name>]
```

Entity: https://gist.github.com/Aracon/418af4c405bee420cb6aafaf5fcd87ff
Related entity: https://gist.github.com/Aracon/9c1d15391041fc4e6010a88af6b8b727
composer.lock: https://gist.github.com/Aracon/31c77c7ffce6fb6928e48952f6afd4ef

PHP  7.2.10

weaverryan:
Fixes #535

The php-parser seems to have problems updating a method (adding a new statement)
when the only existing statement is a nop. This works around that problem.

This may be a bug in php-parser - i'll try to create a failing test case there.

Cheers!


weaverryan:
See nikic/PHP-Parser#649

If the issue can be fixed there, that's *awesome*. But I have no problems patching the issue here first - the workaround is fine.

jrushlow (changes requested):
Bringing this back to our radar - do we still need this fix? It looks like the issue in `php-parser` has been resolved, but i could be wrong.

PR #538, opened 2020-01-31
temp. disable make forgotten password

jrushlow:
In reference to issue #537

PR #539, opened 2020-01-31
fix security logout fixture

jrushlow:
#474 changed implemented \LogicException instead of \Exception for make:auth. Fixture still produced \Exception for assertion.


jrushlow:
PR #538 fixes(skips) failing test in this PR

weaverryan:
My bad! Thanks for the fix @jrushlow!

PR #544, opened 2020-02-12
Make Crud - Change field name appearance

### LINKED ISSUE #543: (minor) make:crud should be consistent with capital letters
In the generated index.html.twig from make:crud, each field starts with a capital letter, except the "actions" (last column). It would be a minor, but pleasing change, if Actions could be written starting with a capital letter :)

LeJeanbono:
Fix #543
I Adapted `asHumanWords` method to be more clever

### DIFF
--- a/src/Resources/skeleton/crud/templates/index.tpl.php
+++ b/src/Resources/skeleton/crud/templates/index.tpl.php
@@ -1,3 +1,8 @@
+<?php
+
+use Symfony\Bundle\MakerBundle\Str;
+
+?>
 <?= $helper->getHeadPrintCode($entity_class_name.' index'); ?>
 
 {% block body %}
@@ -7,9 +12,9 @@
         <thead>
             <tr>
 <?php foreach ($entity_fields as $field): ?>
-                <th><?= ucfirst($field['fieldName']) ?></th>
+                <th><?= Str::asHumanWords($field['fieldName']) ?></th>
 <?php endforeach; ?>
-                <th>actions</th>
+                <th>Actions</th>
             </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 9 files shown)


src/Resources/skeleton/crud/templates/index.tpl.php
```diff
@@ -19,13 +24,13 @@
                 <td>{{ <?= $helper->getEntityFieldPrintCode($entity_twig_var_singular, $field) ?> }}</td>
 <?php endforeach; ?>
                 <td>
-                    <a href="{{ path('<?= $route_name ?>_show', {'<?= $entity_identifier ?>': <?= $entity_twig_var_singular ?>.<?= $entity_identifier ?>}) }}">show</a>
-                    <a href="{{ path('<?= $route_name ?>_edit', {'<?= $entity_identifier ?>': <?= $entity_twig_var_singular ?>.<?= $entity_identifier ?>}) }}">edit</a>
+                    <a href="{{ path('<?= $route_name ?>_show', {'<?= $entity_identifier ?>': <?= $entity_twig_var_singular ?>.<?= $entity_identifier ?>}) }}">Show</a>
```
rbaarsma, line 27: I guess it's debatable whether or not these should be capitalized. Personally I see <th> as headings and think they should have a capital letter, but <td> as values (similar to bullet list) and don't give them a capital letter, unless it's a name or some such. 

A quick search on Google reveals there's no universal standard for table cells. In my experience many values in <td> cells are database values that could be wrongly interpreted when given a capital letter, therefore I suggest to not give them a capital letter.

rbaarsma (approved):
I see no problems, just one comment that I wondered about

src/Resources/skeleton/crud/templates/index.tpl.php
LeJeanbono: @weaverryan wdyt ?

Issue #473, opened 2019-09-23
Move make:entity reserved validation until later & skip for relations


weaverryan:
When you use `make:entity`, we validate property names against the reserved keyword list. If a property name is a keyword, we fail validation and don't allow this. On some systems, this (for example) doesn't allow you to create a `user` property, which is kind of convenient :).

And... if this is a relation property, it really shouldn't go through this validation, and a `user` property would become a `user_id` column - and the addition of the `id` part will probably mean that this will *never* be a reserved word.

To fix this, we would need to move the validation until *later* when we know the "type" of the field. Then, if the field *does* fail validation, we can ask the user for a different "column" name - e.g. they might have a `group` field name, but call it a `group_name` column.

johnpancoast:
Was just going to write the same thing about the resultant `user_id` column not being reserved. I usually just name `user` something else to get by (really love the rest of `make:entity` features) and then I rename it after, but would be great to have the check later like you said.

johnpancoast:
Another solution that was mentioned in #485 would be to allow the `name` values for tables and columns to be specified, i.e., the name values for `@ORM\Table()`, `@ORM\Column()`, and `@ORM\JoinColumn()` annotations.

PR #545, opened 2020-02-14
Make User: availability to chose table's name

### LINKED ISSUE #485: `make:user` default table name not compatible with PostgreSQL
The `make:user` command defaults to creating a user entity with the table name `user`. This is a reserved word in PostgreSQL, so out of the box, Doctrine queries don't work.

I'm using:
* Maker bundle v1.14.0
* Doctrine ORM v2.6.4
* PostgreSQL v12.0

As soon as I try something like `$userRepository->findOneByUsername()` it fails:

```
An exception occurred while executing 'SELECT t0.id AS id_1, t0.username AS username_2, t0.roles AS roles_3, t0.password AS password_4 FROM user t0 WHERE t0.username = ? LIMIT 1' with params ["bob"]:

SQLSTATE[42703]: Undefined column: 7 ERROR: column t0.id does not exist
LINE 1: SELECT t0.id AS id_1, t0.username AS username_2, t0.roles AS...
```

If I change the name of the table to `users` in the `User` entity, queries work fine. This seems like a confluence of bugs from different projects, but it seems to me this would be the easiest place to fix it.

Possible fixes:
* Automatically pluralise the table name
* Automatically add a prefix to the table name
* Ask for a table name in the console command, and validate that it doesn't equal `user`. I'm not sure if this affects other databases - if it's a PostgreSQL specific issue then perhaps it could selectively apply the validation rule.

nikophil:
Hello,

this fixes #485

i may provide a fix for #473 in another PR

### DIFF
--- a/src/Maker/MakeUser.php
+++ b/src/Maker/MakeUser.php
@@ -102,15 +102,6 @@ final class MakeUser extends AbstractMaker
             if ($missingPackagesMessage) {
                 throw new RuntimeCommandException($missingPackagesMessage);
             }
-
-            $question = 'Enter the entity\'s table name. (<comment>It will be auto-escaped.</comment>)';
-            $entityTableName = $io->ask(
-                $question, 'user', function ($name) {
-                return Validator::validateDoctrineTableName($name, $this->doctrineHelper->getRegistry());
-            });
-
-            $command->addOption('entity-table-name', $entityTableName);
-            $input->setOption('entity-table-name', $entityTableName);
         }
         $input->setOption('is-entity', $userIsEntity);
 
@@ -148,7 +139,7 @@ final class MakeUser extends AbstractMaker
 
         // A) Generate the User class
         if ($userClassConfiguration->isEntity()) {
-            $entityClassGenerator = new EntityClassGenerator($generator);
+            $entityClassGenerator = new EntityClassGenerator($generator, $this->doctrineHelper);
             $classPath = $entityClassGenerator->generateEntityClass(
                 $userClassNameDetails,
                 false, // api resource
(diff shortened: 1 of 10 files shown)


napestershine:
Any chance of it to get merged ?

src/Maker/MakeUser.php
```diff
@@ -98,6 +102,15 @@ class_exists(DoctrineBundle::class)
...
             }
+
+            $question = 'Enter the entity\'s table name. (<comment>It will be auto-escaped.</comment>)';
+            $entityTableName = $io->ask(
+                $question, 'user', function ($name) {
+                return Validator::validateDoctrineTableName($name, $this->doctrineHelper->getRegistry());
+            });
+
+            $command->addOption('entity-table-name', $entityTableName);
+            $input->setOption('entity-table-name', $entityTableName);
```
weaverryan: I'd prefer not to ask a new question to fix an edge-case bug. In most cases, I don't want the user to think about their table name. Suggestion:

1) Move this logic to `EntityClassGenerator::generateEntityClass()`

2) Inside that method, use the `doctrine.orm.configuration` service to fetch the naming strategy and check of the *potential* table name is invalid. Here's fake code:

```php
        /** @var NamingStrategy $namingStrategy */
        $namingStrategy = $this->doctrineConfiguration
            ->getNamingStrategy();
        $entityClassDetails->getShortName();

        $potentialTableName = $namingStrategy
            ->classToTableName($entityClassDetails->getFullName());
        if ($this->connection->getDatabasePlatform()->getReservedKeywordsList()->isKeyword($potentialTableName)) {
            // add ORM\Table
        }
```

nikophil: I didn't perceived this as an edge case: a lot of projects use prefixes for their tables, so this feature prevent the need for the user to add the annotation by itself. 
From my point of view, this adds more granularity for the maker.

nikophil: Of course, if you still think it's not a good idea, i'll amend the code with your suggestion ;)

weaverryan: Yea, I think we should remove the question and just add the `@ORM\Table` automatically *when* needed. The topic whether we should ask this question everywhere that we create entities is a separate thing. And really, if people are prefixing all their database tables (which sometimes makes sense - I have done that before), we should make that easier to do on a global level in Doctrine (so that you don't need the `@ORM\Table` everywhere).

nikophil: i'm working on that, but i'm facing a problem: `doctrine.orm.configuration` is abstract, thus i need to inject `doctrine.orm.default_configuration`. that means this feature will only work on default connection.

And if none exists (which *can* happen, i think) ==> :boom:

nikophil: hmm actually `ManagerRegistry` does have a method `getManager()`, but it only returns a `ObjectManager` and not an `EntityManagerInterface`, so we don't have access to method `getConfiguration()`

weaverryan: Yep, let's cheat. But check with an `instanceof` and if it's not an `EntityManager`, let's throw an exception.

nikophil: ok it's done... but now, aren't we lacking a test on postgreSQL for instance?

weaverryan: That's very true. I think we should live with that for now.

weaverryan:
Love this! Thank you @nikophil!

blowski:
Nice! Thanks for fixing this.

PR #546, opened 2020-02-17
Fix SecurityConfigUpdater

nikophil:
this fixes the bug which occurs presently on master

since we remove memory provider in "make:user" if it is the only
provider, we need to remove this memory provider from firewalls.

And the tests were failing now because the security bundle's recipe has been updated recently, and the memory provider has been added to main firewall

### DIFF
--- a/src/Security/SecurityConfigUpdater.php
+++ b/src/Security/SecurityConfigUpdater.php
@@ -103,12 +103,7 @@ final class SecurityConfigUpdater
 
     private function updateProviders(UserClassConfiguration $userConfig, string $userClass)
     {
-        if ($this->isSingleInMemoryProviderConfigured()) {
-            // empty the providers if the generic "in_memory" is the only one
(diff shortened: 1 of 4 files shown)


src/Security/SecurityConfigUpdater.php
```diff
@@ -149,6 +144,27 @@ private function updateEncoders(UserClassConfiguration $userConfig, string $user
...
+
+        $newData = $this->manipulator->getData();
+
+        $memoryProviderName = array_keys($newData['security']['providers'])[0];
+
+        $newData['security']['providers'] = [];
+
+        foreach ($newData['security']['firewalls'] as &$firewall) {
+            if (($firewall['provider'] ?? null) === $memoryProviderName) {
+                $firewall['provider'] = 'app_user_provider';
```
chalasr: 👍 For the record, we prefer setting the provider explicitly. See https://github.com/symfony/recipes/pull/741

PR #547, opened 2020-02-19
Use UserProvider in LoginFormAuthenticator::getUser

thled:
We should use the UserProvider already injected in the getUser method to receive the user.
I do not see why we make a difference if the User entity is used or not.

### DIFF
--- a/src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
+++ b/src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
@@ -2,14 +2,11 @@
 
 namespace <?= $namespace ?>;
 
-<?= $user_is_entity ? "use $user_fully_qualified_class_name;\n" : null ?>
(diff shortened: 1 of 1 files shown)


kbond:
I was curious about this as well. @nikophil, what was the reason for doing it this way?

weaverryan:
The reasoning is that I don't like the UserProvider :). 

The UserProvider exists so that third-party systems have a systematic way to find your User. Inside an authenticator, *you* are in control. You can, of course, choose to use your user provider. But if we make this change, and someone needs to do something more complex than just `loadUserByUsername()`, it may not be obvious that they are *allowed* to just grab a repository and make whatever query they want :). 

So, you're definitely free to change the code and use the user provider. But generating code that uses it - while it will simplify the code a bit - has a big downside.

kbond:
Ok, thanks for the explanation!

thled:
Thanks for the explanation.

PR #549, opened 2020-03-02
MakeForm - Remove hard dependency on Doctrine

GaryPEGEOT:
Currently, if you try to generate a Form for a DTO without Doctine installed, you get the exception message:
`Somehow the doctrine service is missing. Is DoctrineBundle installed?`

But Doctine is not supposed to be mandatory. This PR add a note to inform that doctine is missing, without falling hard if the FQCN is valid

### DIFF
--- a/src/Maker/MakeForm.php
+++ b/src/Maker/MakeForm.php
@@ -53,7 +53,7 @@ final class MakeForm extends AbstractMaker
         $command
             ->setDescription('Creates a new form class')
             ->addArgument('name', InputArgument::OPTIONAL, sprintf('The name of the form class (e.g. <fg=yellow>%sType</>)', Str::asClassName(Str::getRandomTerm())))
-            ->addArgument('bound-class', InputArgument::OPTIONAL, 'The name of Entity or fully qualified model class name that the new form will be bound to (empty for none)')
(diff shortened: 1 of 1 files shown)


src/Maker/MakeForm.php
```diff
@@ -95,7 +95,14 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
 
-            $doctrineEntityDetails = $this->entityHelper->createDoctrineDetails($boundClassDetails->getFullName());
+            try {
+                $doctrineEntityDetails = $this->entityHelper->createDoctrineDetails($boundClassDetails->getFullName());
+            } catch (\Exception $e) {
+                // DoctrineBundle isn't installed
+                $io->comment($e->getMessage());
+
+                $doctrineEntityDetails = null;
+            }
```
BatsaxIV, line 105: Not sure that keeping the io output (_Somehow the doctrine service is missing. Is DoctrineBundle installed?_) is usefull. 
You probably know that doctrine is not installed and if not, this may not be the most appropriate moment to indicate it.

```suggestion
            } catch (\Exception $e) {
                $doctrineEntityDetails = null;
            }
```

BatsaxIV (commented):
Hi,

I've made a small review and also maybe this need a test? 
There seems to be a test here https://github.com/symfony/maker-bundle/blob/master/tests/Maker/MakeFormTest.php#L43, but to my understanding it is launch with doctrine/orm.

PR #550, opened 2020-03-02
"Variables" param in file and template generators should be optional

jrushlow:
- file & template generator $variables param is now optional

Not all file and templates require variables to be passed for creation.


weaverryan:
Thanks Jesse!

PR #511, opened 2019-12-05
Tests structure optimization

sadikoff:
Hi!!! It's again me with some crazy test stuff 😄 Hope it will improve debugging! 

ping @romaricdrigon and @weaverryan 

closes #498

PR #551, opened 2020-03-03
make:subscriber: fix list active event

maxhelias:
With an addtionnal bundle or a component like "security-bundle", we have an exception
```
Notice: Undefined index: debug.security.authorization.vote
```

Now the mapping is good : 
Capture d’écran 2020-03-03 à 15 23 00


fabpot:
This bug has been introduced in #511 and is part of the latest release.

weaverryan:
Thank you Maxime!

PR #556, opened 2020-03-06
Git clone command wrong work in windows. Replace to Xcopy command

uginroot:


LeJeanbono:
Works for me !

uginroot:
The reason was in the global gitignore file.
It has a /vendor directory.
Not sure what to use git clone correctly, as the operation of this command depends on user settings. It may be worth rolling back to simple copying if the user has a global .gitignore file

PR #557, opened 2020-03-08
Set table name to eliminate database keyword error

JakubSzczesniak:
Is really problematic for begginers to find and eliminate this problem. When we set default table name to `users` instead od `user` everything will be working.

### DIFF
--- a/src/Maker/MakeUser.php
+++ b/src/Maker/MakeUser.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Maker;
 
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
+use Doctrine\ORM\Mapping\Table;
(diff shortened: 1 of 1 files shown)


jrushlow:
I believe, correct me if I am wrong, this is related to issue #485, with a fix in PR #545 (Yet to be merged).

Is your issue covered under either the referenced issue or pending PR?

JakubSzczesniak:
@jrushlow yes, you're right. I did this proposition, because I thought this fix will be easier and faster to merge instead of #545

JakubSzczesniak:
@jrushlow is any change to merge this PR? Many people have problems without this. We don't know when discussion in #545 will be over...

weaverryan:
Sorry for the delay! I'm working to merge #545 for the next release. Thanks for your attention on this detail!

PR #559, opened 2020-03-12
Allow to display links on created files

l-vo:
This PR aims to display terminal links on the created file paths when the `framework.ide` configuration param (or `debug.file_link_format` container parameter) is set (only displayed if the terminal support it).

### DIFF
--- a/src/Util/MakerFileLinkFormatter.php
+++ b/src/Util/MakerFileLinkFormatter.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+use Symfony\Component\HttpKernel\Debug\FileLinkFormatter;
+
+/**
+ * @internal
+ */
+final class MakerFileLinkFormatter
+{
+    private $fileLinkFormatter;
+
+    public function __construct(FileLinkFormatter $fileLinkFormatter = null)
+    {
+        // Since nullable types are not available in 7.0; can be removed when php >= 7.1 required
+        if (0 == \func_num_args()) {
+            throw new \LogicException('$fileLinkFormatter argument is required');
+        }
+
+        $this->fileLinkFormatter = $fileLinkFormatter;
+    }
+
+    public function makeLinkedPath(string $absolutePath, string $relativePath): string
+    {
+        if (!$this->fileLinkFormatter) {
+            return $relativePath;
+        }
+
+        if (!$formatted = $this->fileLinkFormatter->format($absolutePath, 1)) {
+            return $relativePath;
+        }
+
+        return $this->createLink(
+            $relativePath,
+            $formatted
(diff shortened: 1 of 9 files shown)


l-vo:
Hi @weaverryan 

> Sorry for the slow review

No problem :) 

I created a new `MakerFileLinkFormatter` class and I fixed nullable typed arguments (`?string`) since the bundle is compatible with php 7.0 applications.

src/Util/MakerFileLinkFormatter.php
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            return $this->createLink(
+                $relativePath,
+                strtr($this->fileLinkFormat, ['%f' => $absolutePath, '%l' => 1])
+            );
+        }
+
+        return $relativePath;
+    }
```
weaverryan, line 49: I think we can reverse a bunch of this logic to exit earlier and avoid nesting. I also think we can probably do NOTHING if there is no `fileLinkFormatter` - it would be a super edge case to have a `fileLinkFormat` but *not* a `fileLinkFormatter`:

```php
if (null === $this->fileLinkFormatter || null === $this->fileLinkFormat) {
    return $relativePath;
}

return $this->createLink(
     $relativePath,
     $this->fileLinkFormatter->format($absolutePath, 1)
);
```

src/Util/MakerFileLinkFormatter.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+use Symfony\Component\HttpKernel\Debug\FileLinkFormatter;
+
+final class MakerFileLinkFormatter
```
weaverryan, line 16: Add `@internal` above also please :)

l-vo:
Updated 🙂

weaverryan:
Well-done! A very nice new feature - thank you @l-vo!

ThomasLandauer:
I was just about to suggest such a feature ;-) However, this kinda means that it's not working for me :-( What's the expected behavior? Since you're saying "only displayed if the terminal support it", which link syntax are you using?

What I would have expected is a clickable link in this line after `php bin/console make:migration`:

> Next: Review the new migration "src/Migrations/Version20211030130836.php"

I committed a similar feature to [phpstan](https://github.com/phpstan/phpstan-src/pull/515) and [Codeception](https://github.com/Codeception/Codeception/pull/6261).

l-vo:
@ThomasLandauer It's because `make:migration` rely on `doctrine:migrations:diff`. So to have a clickable link, the feature must be bring to the doctrine command.

ThomasLandauer:
@l-vo thanks! However, I just found out that the line I'm talking about is generated here in this bundle at https://github.com/symfony/maker-bundle/blob/main/src/Maker/MakeMigration.php#L129

So what does your PR do then? I tried it with `bin/console make:controller`, but still getting no link.

l-vo:
No, look here: https://github.com/symfony/maker-bundle/blob/f76abb1e4facd712a5cdbf35aff998e87c416801/src/Maker/MakeMigration.php#L111, it's generated by the `doctrine:migrations:diff` command. Apart if your terminal is not compatible, it's working for `make:controller`.

PR #560, opened 2020-03-18
Doctrine\Common\Persistence\ObjectManager is deprecated

ahmedali5530:
Make implimentation compatible with `Doctrine\Common\DataFixtures\FixtureInterface::load`

### DIFF
--- a/src/Resources/skeleton/doctrine/Fixtures.tpl.php
+++ b/src/Resources/skeleton/doctrine/Fixtures.tpl.php
@@ -3,7 +3,7 @@
 namespace <?= $namespace; ?>;
 
 use Doctrine\Bundle\FixturesBundle\Fixture;
-use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/doctrine/Fixtures.tpl.php
```diff
@@ -3,7 +3,7 @@
 namespace <?= $namespace; ?>;
 
 use Doctrine\Bundle\FixturesBundle\Fixture;
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\Persistence\ObjectManager;
```
weaverryan: Hmm, I think it might be a bit trickier than this. If someone is still using, for example, `doctrine/orm` 2.6, then they will not have the `doctrine/persistence` package. And the new code will break. We actually have this problem all over the place in the codebase. I'm going to take a look at it right now.

weaverryan:
Fully fixed in #570 :). Thanks!

PR #562, opened 2020-03-25
function to change the name path of the template Template.tpl.php

Jonathanlight:


weaverryan:
Hey @jonathankablan!

Thanks for the PR :). This topic is being discussed in #560. For the record, I'm against making the templates overrideable. Well, at least, making them *easily* overrideable. The reason is that we don't guarantee backward-compatibility with the templates: we may (in a minor version) change the variables that we pass into a template, which could break a custom template.

That's not to say that I'm against it completely, but it needs to be clearly and "advanced" thing to do, and should possibly be a little difficult. 

About your PR, I definitely don't like the `templateNameEntity` and `templateNameRepository`. Whatever solution we choose, should be a global solution - not something that affects just one maker command. Your `rootTemplateName` (which should really be called `rootTemplateDirectory`) would solve this, as then you have a hook point for *any* template.

If you're willing to put in some work and push on this, let me know. I can make some suggestions about how we could make this possible, but still "hard enough" that it's for advanced users ;).

Cheers!

jonathankablan:
Thanks for this first reading, I had wanted to have the possibility to change the TemplateName but this is not really essential for the blow. I would like to allow the change of the RootTemplateDirectory in the Generator class simply which won't change anything to the current operation of the Generator class, and discuss the other methods further.

Thanks

src/MakerTemplateRendererInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @author Jonathan Kablan <jonathan.kablan@gmail.com>
+ */
+interface MakerTemplateRendererInterface
+{
+    /**
+     * @param string $templateName
+     * @return string
+     */
+    public function supports(string $templateName): string;
```
weaverryan, line 15:
```suggestion
    public function supports(string $templateName): bool;
```

This will return a boolean. The idea is that `MakerRenderer` will look a bit different than you have. It will look like this:

```php
class MakerRenderer
{
    /**
     * @var MakerTemplateRendererInterface[]
     */
    private $templateRenderers;

    public function __construct(array $templateRenderers)
    {
        $this->templateRenderers = $templateRenderers;
    }

    public function renderTemplate(string $templateName, array $variables)
    {
        foreach ($this->templateRenderers as $templateRenderer) {
            if ($templateRenderer->supports($templateName)) {
                return $templateRenderer->render($templateName, $variables);
            }
        }
        
        throw new \Exception(sprintf('No template renderers for template "%s"', $templateName);
    }
}
```

There would be ONE class in MakerBundle that implements `MakerTemplateRendererInterface` (and users could add more by using a dependency injection tag). It would probably look like this:

```php
class MakerDefaultTemplateRenderer implements MakerTemplateRendererInterface
{
    public function supports(string $templateName): string
    {
        $templatePath = __DIR__.'/Resources/skeleton/'.$templateName;

        return file_exists($templatePath);
    }

    public function render(string $templateName, array $variables)
    {
        $templatePath = __DIR__.'/Resources/skeleton/'.$templateName;

        // this is basically the current logic for rendering templates
        $contents = $this->fileManager->parseTemplate($templatePath, $variables);
        $this->fileManager->dumpFile($templatePath, $contents);
    }
}
```

src/MakerTemplateRendererInterface.php
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * @param string $templateName
+     * @return string
+     */
+    public function render(string $templateName = null): string;
```
weaverryan, line 21:
```suggestion
    public function render(string $templateName, array $variables): string;
```

src/Generator.php
```diff
@@ -64,6 +77,11 @@ public function generateClass(string $className, string $templateName, array $va
 
     /**
      * Generate a normal file from a template.
+     *
+     * @param string $targetPath
```
maxhelias, line 81: This information does not add value because we already have the type in the signature. You should delete them.
BTW, PHP-CS-Fixer should remove it 🤔

jonathankablan: okay Max

PR #565, opened 2020-03-27
explicitly set framework.router.utf8 value in test kernel

jrushlow:
fixes `Since symfony/framework-bundle 5.1: Not setting the "framework.router.utf8" configuration option is deprecated, it will default to "true" in version 6.0.` deprecation warning in tests / ci

not intended to address any other deprecation's or fix/refactor other logic


weaverryan:
Thanks Jesse!

PR #567, opened 2020-03-27
create maker for symfony casts reset password bundle

jrushlow:
Adds a `make:reset-password` maker to create the classes, entities, and twig templates required by Symfony Casts Reset Password Bundle. This complements the bundle by doing the leg work to provide a true out-of-the-box solution for handling forgotten passwords in traditional Symfony applications.

- [x] test suite depends on PR #565
- [x] depends on flex recipe https://github.com/symfony/recipes/pull/748

### DIFF
--- a/src/Maker/MakeResetPassword.php
+++ b/src/Maker/MakeResetPassword.php
@@ -1,0 +1,315 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\FileManager;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Security\InteractiveSecurityHelper;
+use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
+use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Yaml\Yaml;
+use SymfonyCasts\Bundle\ResetPassword\SymfonyCastsResetPasswordBundle;
+
+/**
+ * @author Romaric Drigon <romaric.drigon@gmail.com>
+ * @author Jesse Rushlow  <jr@rushlow.dev>
+ * @author Ryan Weaver    <ryan@symfonycasts.com>
+ *
+ * @internal
+ * @final
+ */
+class MakeResetPassword extends AbstractMaker
+{
+    private $fileManager;
+
+    public function __construct(FileManager $fileManager)
+    {
+        $this->fileManager = $fileManager;
+    }
+
+    public static function getCommandName(): string
+    {
+        return 'make:reset-password';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig)
+    {
+        $command
+            ->setDescription('Create controller, entity, and repositories for use with symfonycasts/reset-password-bundle.')
+        ;
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies)
+    {
+        $dependencies->addClassDependency(SymfonyCastsResetPasswordBundle::class, 'symfonycasts/reset-password-bundle');
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
+    {
+        $io->title('Let\'s make a password reset feature!');
+
+        // initialize arguments & commands that are internal (i.e. meant only to be asked)
+        $command
+            ->addArgument('from-email-address', InputArgument::REQUIRED)
+            ->addArgument('from-email-name', InputArgument::REQUIRED)
+            ->addArgument('controller-reset-success-redirect', InputArgument::REQUIRED)
+            ->addArgument('user-class')
+            ->addArgument('email-property-name')
+            ->addArgument('email-getter')
+            ->addArgument('password-setter')
+        ;
+
+        $interactiveSecurityHelper = new InteractiveSecurityHelper();
+
+        if (!$this->fileManager->fileExists($path = 'config/packages/security.yaml')) {
+            throw new RuntimeCommandException('The file "config/packages/security.yaml" does not exist. This command needs that file to accurately build the reset password form.');
+        }
+
+        $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
+        $securityData = $manipulator->getData();
+        $providersData = $securityData['security']['providers'] ?? [];
+
+        $input->setArgument(
+            'user-class',
+            $userClass = $interactiveSecurityHelper->guessUserClass(
+                $io,
+                $providersData,
+                'What is the User entity that should be used with the "forgotten password" feature? (e.g. <fg=yellow>App\\Entity\\User</>)'
+            )
+        );
+
+        $input->setArgument(
+            'email-property-name',
+            $interactiveSecurityHelper->guessEmailField($io, $userClass)
+        );
+        $input->setArgument(
+            'email-getter',
+            $interactiveSecurityHelper->guessEmailGetter($io, $userClass)
+        );
+        $input->setArgument(
+            'password-setter',
+            $interactiveSecurityHelper->guessPasswordSetter($io, $userClass)
+        );
+
+        $io->text(sprintf('Implementing reset password for <info>%s</info>', $userClass));
+
+        $io->section('- ResetPasswordController -');
+        $io->text('A named route is used for redirecting after a successful reset. Even a route that does not exist yet can be used here.');
+        $input->setArgument('controller-reset-success-redirect', $io->ask(
+            'What route should users be redirected to after their password has been successfully reset?',
+            'app_home',
+            [Validator::class, 'notBlank']
+        )
+        );
+
+        $io->section('- Email Templates -');
+        $emailText[] = 'These are used to generate the email code. Don\'t worry, you can change them in the code later!';
+        $io->text($emailText);
+
+        $input->setArgument('from-email-address', $io->ask(
+            'What email address will be used to send reset confirmations? e.g. admin@your-domain.com',
+            null,
+            [Validator::class, 'validateEmailAddress']
+        ));
+
+        $input->setArgument('from-email-name', $io->ask(
+            'What "name" should be associated with that email address? e.g. "Acme Mail Bot"',
+            null,
+            [Validator::class, 'notBlank']
+        )
+        );
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $userClass = $input->getArgument('user-class');
+        $userClassNameDetails = $generator->createClassNameDetails(
+            '\\'.$userClass,
+            'Entity\\'
+        );
+
+        $controllerClassNameDetails = $generator->createClassNameDetails(
+            'ResetPasswordController',
+            'Controller\\'
+        );
+
+        $requestClassNameDetails = $generator->createClassNameDetails(
+            'ResetPasswordRequest',
+            'Entity\\'
+        );
+
+        $repositoryClassNameDetails = $generator->createClassNameDetails(
+            'ResetPasswordRequestRepository',
+            'Repository\\'
+        );
+
+        $requestFormTypeClassNameDetails = $generator->createClassNameDetails(
+            'ResetPasswordRequestFormType',
+            'Form\\'
+        );
+
+        $changePasswordFormTypeClassNameDetails = $generator->createClassNameDetails(
+            'ChangePasswordFormType',
+            'Form\\'
+        );
+
+        $generator->generateController(
+            $controllerClassNameDetails->getFullName(),
+            'resetPassword/ResetPasswordController.tpl.php',
+            [
+                'user_full_class_name' => $userClassNameDetails->getFullName(),
+                'user_class_name' => $userClassNameDetails->getShortName(),
+                'request_form_type_full_class_name' => $requestFormTypeClassNameDetails->getFullName(),
+                'request_form_type_class_name' => $requestFormTypeClassNameDetails->getShortName(),
+                'reset_form_type_full_class_name' => $changePasswordFormTypeClassNameDetails->getFullName(),
+                'reset_form_type_class_name' => $changePasswordFormTypeClassNameDetails->getShortName(),
+                'password_setter' => $input->getArgument('password-setter'),
+                'success_redirect_route' => $input->getArgument('controller-reset-success-redirect'),
+                'from_email' => $input->getArgument('from-email-address'),
+                'from_email_name' => $input->getArgument('from-email-name'),
+                'email_getter' => $input->getArgument('email-getter'),
+            ]
+        );
+
+        $generator->generateClass(
+            $requestClassNameDetails->getFullName(),
+            'resetPassword/ResetPasswordRequest.tpl.php',
+            [
+                'repository_class_name' => $repositoryClassNameDetails->getFullName(),
+                'user_full_class_name' => $userClassNameDetails->getFullName(),
+            ]
+        );
+
+        $generator->generateClass(
+            $repositoryClassNameDetails->getFullName(),
+            'resetPassword/ResetPasswordRequestRepository.tpl.php',
+            [
+                'request_class_full_name' => $requestClassNameDetails->getFullName(),
+                'request_class_name' => $requestClassNameDetails->getShortName(),
+            ]
+        );
+
+        $this->setBundleConfig($io, $generator, $repositoryClassNameDetails->getFullName());
+
+        $generator->generateClass(
+            $requestFormTypeClassNameDetails->getFullName(),
+            'resetPassword/ResetPasswordRequestFormType.tpl.php',
+            [
+                'email_field' => $input->getArgument('email-property-name'),
+            ]
+        );
+
+        $generator->generateClass(
+            $changePasswordFormTypeClassNameDetails->getFullName(),
+            'resetPassword/ChangePasswordFormType.tpl.php'
+        );
+
+        $generator->generateTemplate(
+            'reset_password/check_email.html.twig',
+            'resetPassword/twig_check_email.tpl.php'
+        );
+
+        $generator->generateTemplate(
+            'reset_password/email.html.twig',
+            'resetPassword/twig_email.tpl.php'
+        );
+
+        $generator->generateTemplate(
+            'reset_password/request.html.twig',
+            'resetPassword/twig_request.tpl.php',
+            [
+                'email_field' => $input->getArgument('email-property-name'),
+            ]
+        );
+
+        $generator->generateTemplate(
+            'reset_password/reset.html.twig',
+            'resetPassword/twig_reset.tpl.php'
+        );
+
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+        $this->successMessage($input, $io, $requestClassNameDetails->getFullName());
+    }
+
+    private function setBundleConfig(ConsoleStyle $io, Generator $generator, string $repositoryClassFullName)
+    {
+        $configFileExists = $this->fileManager->fileExists($path = 'config/packages/reset_password.yaml');
+
+        /*
+         * reset_password.yaml does not exist, we assume flex was present when
+         * the bundle was installed & a customized configuration is in use.
+         * Remind the developer to set the repository class accordingly.
+         */
+        if (!$configFileExists) {
+            $io->text(sprintf('We can\'t find %s. That\'s ok, you probably have a customized configuration.', $path));
+            $io->text('Just remember to set the <fg=yellow>request_password_repository</> in your configuration.');
+            $io->newLine();
+
+            return;
+        }
+
+        $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
+        $data = $manipulator->getData();
+
+        $symfonyCastsKey = 'symfonycasts_reset_password';
+
+        /*
+         * reset_password.yaml exists, and was probably created by flex;
+         * Let's replace it with a "clean" file.
+         */
+        if (1 >= \count($data[$symfonyCastsKey])) {
+            $yaml = [
+                $symfonyCastsKey => [
+                    'request_password_repository' => $repositoryClassFullName,
+                ],
+            ];
+
+            $generator->dumpFile($path, Yaml::dump($yaml));
+
+            return;
+        }
+
+        /*
+         * reset_password.yaml exists and appears to have been customized
+         * before running make:reset-password. Let's just change the repository
+         * value and preserve everything else.
+         */
+        $data[$symfonyCastsKey]['request_password_repository'] = $repositoryClassFullName;
+
+        $manipulator->setData($data);
+
+        $generator->dumpFile($path, $manipulator->getContents());
+    }
+
+    private function successMessage(InputInterface $input, ConsoleStyle $io, string $requestClassName)
+    {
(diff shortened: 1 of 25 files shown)


src/Validator.php
```diff
@@ -167,6 +167,15 @@ public static function validateDoctrineFieldName(string $name, ManagerRegistry $
         return $name;
     }
 
+    public static function validateEmailAddress(string $email): string
```
jrushlow: method name could be as is or just `emailAddress`. As a whole, the class has functions like `validate..` and `notBlank`. Although `Validator::class` naming conventions are outside the scope of this PR, may be worth considering which name would be appropriate for this method.

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Security\InteractiveSecurityHelper;
+use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
+use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use SymfonyCasts\Bundle\ResetPassword\SymfonyCastsResetPasswordBundle;
+
+class MakeResetPassword extends AbstractMaker
```
jrushlow: @romaricdrigon  - I haven't added the authors header for this class yet, but as you were the author of the original `MakeForgottenPassword` class that this is based on, would you like your name included on this maker?

weaverryan: I think we definitely should :) - I don't think he'll mind

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
+    {
+        $io->title('Reset Password Bundle Requires:');
+        $requirements[] = '1) A user entity has been created.';
+        $requirements[] = '2) The user entity contains an email property with a getter method.';
+        $requirements[] = '3) A user repository exists for the user entity.'."\n";
+        $requirements[] = '<fg=yellow>bin/console make:user</> will generate the user entity and it\'s repository...'."\n";
+        $io->text($requirements);
```
weaverryan, line 66: You're doing some nice work with the formatting!

But I'm not sure we need this entire section. What if we:

A) Move the "user" questions to be the *first* questions
B) Removed this section

Then the user should be able to naturally discover that they're missing the `User` class (or are missing an email field somehow) very early.

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+            ->addArgument('user-class')
+            ->addArgument('email-property-name')
+            ->addArgument('email-getter')
+            ->addArgument('password-setter')
+        ;
+
+        $io->section('- Email Templates -');
+        $emailText[] = 'Please answer the following questions that will be used to generate the email templates.';
+        $emailText[] = 'If you are unsure of what these answers should be, that\'s ok.';
+        $emailText[] = 'You can change these later after the templates have been generated.';
```
weaverryan, line 82: How about just this:

> These are used to generate the email code. Don't worry, you can change them in the code later!

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+            ->addArgument('from-email-address', InputArgument::REQUIRED)
+            ->addArgument('from-email-name', InputArgument::REQUIRED)
+            ->addArgument('controller-reset-success-redirect', InputArgument::REQUIRED)
+            ->addArgument('user-class')
+            ->addArgument('email-property-name')
+            ->addArgument('email-getter')
+            ->addArgument('password-setter')
+        ;
+
+        $io->section('- Email Templates -');
```
weaverryan, line 79: My initial instinct is that this is cool :)

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+        $emailText[] = 'Please answer the following questions that will be used to generate the email templates.';
+        $emailText[] = 'If you are unsure of what these answers should be, that\'s ok.';
+        $emailText[] = 'You can change these later after the templates have been generated.';
+        $io->text($emailText);
+
+        $input->setArgument('from-email-address', $io->ask(
+            'What email address will be used to send reset confirmations? I.e. admin@your-domain.com',
```
weaverryan, line 86: Use e.g. - there's a subtle difference between e.g. and i.e. ... and I think e.g. is the correct one

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+            null,
+            [Validator::class, 'validateEmailAddress']
+        ));
+
+        $input->setArgument('from-email-name', $io->ask(
+            'What name will be associated with the email address used to send password reset confirmations? I.e. John Doe or Your Company, LLC.',
```
weaverryan, line 92: > What "name" should be associated with that email address? e.g. "Acme Mail Bot"

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+            'user-class',
+            $userClass = $interactiveSecurityHelper->guessUserClass(
+                $io,
+                $providersData,
+                'Enter the User class that should be used with the "forgotten password" feature (e.g. <fg=yellow>App\\Entity\\User</>)'
+            )
+        );
+
+        $io->section('- Controller Template -');
+        $io->comment('<fg=yellow>A named route is required for redirection after a successful reset. Even routes that do not yet exist can be used here.</>');
```
weaverryan, line 118: No `comment` here - use `text()`. Comment makes it yellow - same color as the `->section()` above.

And:

> A named route is used for redirecting after a successful reset. Even a route that does not exist yet can be used here.

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+        $input->setArgument(
+            'user-class',
+            $userClass = $interactiveSecurityHelper->guessUserClass(
+                $io,
+                $providersData,
+                'Enter the User class that should be used with the "forgotten password" feature (e.g. <fg=yellow>App\\Entity\\User</>)'
+            )
+        );
+
+        $io->section('- Controller Template -');
```
weaverryan, line 117: Let's take out the word "template" in these sections unless the code is *really* going to be used in a "template". In this case, the code will be used in the controller. Heck we could even put `ResetPasswordController` in the section title as we now this will be the class name.

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+            $interactiveSecurityHelper->guessEmailField($io, $userClass)
+        );
+        $input->setArgument(
+            'email-getter',
+            $interactiveSecurityHelper->guessEmailGetter($io, $userClass)
+        );
+        $input->setArgument(
+            'password-setter',
+            $interactiveSecurityHelper->guessPasswordSetter($io, $userClass)
+        );
```
weaverryan, line 139: These 3 should be done immediately after guessing/asking the User class

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+                'request_class_full_name' => $requestClassNameDetails->getFullName(),
+                'request_class_name' => $requestClassNameDetails->getShortName(),
+            ]
+        );
+
+        if (!$this->fileManager->fileExists($path = 'config/packages/reset_password.yaml')) {
+            throw new RuntimeCommandException(sprintf('The file "%s" does not exist. This command needs that file to accurately build the reset password config.', $path));
+        }
+
+        $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
```
weaverryan, line 218: This is a "funny" situation. We sorta don't want to "update" this file (and leave the old comments). We probably want to completely *override* the file contents with the new ones. Specifically, we don't want our "Replace symfonycasts.reset_password.fake_request_repository with the full..." text to be in this file when we finish.

Perhaps we could check if `request_password_repository: symfonycasts.reset_password.fake_request_repository` is the only config in that file. And:

A) If it is, use the normal `Yaml` file to completely dump new contents
B) If not for some reasons, then use the manipulator to update the file more carefully

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+        );
+
+        $generator->generateTemplate(
+            'reset_password/reset.html.twig',
+            'resetPassword/twig_reset.tpl.php'
+        );
+
+        $generator->writeChanges();
+
+        $this->successMessage($input, $io, $controllerClassNameDetails->getFullName());
```
weaverryan, line 265: We have a semi-standardized format for the success message - you can see an example here: https://github.com/symfony/maker-bundle/blob/master/src/Maker/MakeCrud.php#L206-L208

It should include something like.

```
Next:
  1) Run "php bin/console make:migration" to generate a migration for the new "ResetPasswordRequest" entity
  2) Review and customize the templates in `templates/reset_password`.
  3) Make sure your MAILER_DSN env var has the correct settings

Then go to "/reset-password" and enjoy!
```

I don't think we need any more details (that I can think of)

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -0,0 +1,166 @@
...
+
+    private $resetPasswordHelper;
+
+    public function __construct(ResetPasswordHelperInterface $resetPasswordHelper)
+    {
+        $this->resetPasswordHelper = $resetPasswordHelper;
+    }
+
+    /**
+     * @Route(name="app_forgot_password_request")
```
weaverryan, line 36: It just occurred to me that it might be useful to have a short description above each method:

> Display & process form to request a password reset.

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -0,0 +1,166 @@
...
+            );
+        }
+
+        return $this->render('reset_password/request.html.twig', [
+            'requestForm' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @Route("/check-email", name="app_check_email")
```
weaverryan, line 56: > Confirmation page after a user has requested a password reset.

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -0,0 +1,166 @@
...
+            return $this->redirectToRoute('app_forgot_password_request');
+        }
+
+        return $this->render('reset_password/check_email.html.twig', [
+            'tokenLifetime' => $this->resetPasswordHelper->getTokenLifetime(),
+        ]);
+    }
+
+    /**
+     * @Route("/reset/{token}", name="app_reset_password")
```
weaverryan, line 71: > Validates and process the reset URL that the user clicked in their email.

src/Resources/skeleton/resetPassword/twig_check_email.tpl.php
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html.twig' %}
+
+{% block title %}Password Reset Email Sent{% endblock %}
+
+{% block body %}
+    <p>An email has been sent that contains a link you can click to reset your password. This link will expire in {{ tokenLifetime|date('g') }} hour.</p>
```
weaverryan, line 6: > ... a link *that* you can click...

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: hmm, off the top of my head, all of these args are "required". Do we need `InputArgument::REQUIRED` for the last 4 even though the interactive security guesser sort of "enforces" this..

src/Maker/MakeResetPassword.php
jrushlow: WIP

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
jrushlow: done

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
jrushlow: done

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
jrushlow: done

src/Resources/skeleton/resetPassword/twig_email.tpl.php
```diff
@@ -0,0 +1,11 @@
+<h1>Hi!</h1>
+
+<p>
+    To reset your password, please visit
+    <a href="{{ url('app_reset_password', {token: resetToken.token}) }}">here</a>
+    This link will expire in {{ resetToken.ExpiresAt|date('Y-m-d') }}TODO - convert to hours! hours.
```
weaverryan, line 6: Don't forget about this TODO! :)

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -0,0 +1,166 @@
...
+            // We store the token in session and remove it from the URL, to avoid the URL being
+            // loaded in a browser and potentially leaking the token to 3rd party JavaScript.
+            $this->storeTokenInSession($token);
+
+            return $this->redirectToRoute('app_reset_password');
+        }
+
+        $token = $this->getTokenFromSession();
+        if (null === $token) {
+            throw $this->createNotFoundException();
```
weaverryan, line 91: We should add a message to the exception:

> No reset password token found in the URL or in the session.

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+        );
+
+        $generator->generateClass(
+            $requestClassNameDetails->getFullName(),
+            'resetPassword/ResetPasswordRequest.tpl.php',
+            [
+                'repository_class_name' => $repositoryClassNameDetails->getFullName(),
+                'user_full_class_name' => $userClassNameDetails->getFullName(),
+            ]
+        );
```
weaverryan, line 209: Let's save this for later (cause I *think* it will take some work), but:

It would be ideal if we could use `EntityClassGenerator` to generate the entity and repository classes. That's our centralized entry point for creating these classes and making sure they all look uniform. We would use that class and *then* use the `ClassSourceManipulator` to add in our custom code (that's the tricky part - writing PHP Parser code to add methods, etc take a bit of work to get right).

src/Resources/skeleton/resetPassword/twig_request.tpl.php
```diff
@@ -0,0 +1,22 @@
...
+{% block title %}Reset your password{% endblock %}
+
+{% block body %}
+    {% for flashError in app.flashes('reset_password_error') %}
+        <div class="alert alert-danger" role="alert">{{ flashError }}</div>
+    {% endfor %}
+    <h1>Reset your password</h1>
+
+    {{ form_start(requestForm) }}
+    {{ form_row(requestForm.<?= $email_field ?>) }}
```
weaverryan, line 12: Can we indent this 4 more spaces so it's inside the `{{ form_start() }}`?

src/Resources/skeleton/resetPassword/twig_reset.tpl.php
```diff
@@ -0,0 +1,12 @@
+{% extends 'base.html.twig' %}
+
+{% block title %}Reset your password{% endblock %}
+
+{% block body %}
+    <h1>Reset your password</h1>
+
+    {{ form_start(resetForm) }}
+    {{ form_row(resetForm.plainPassword) }}
```
weaverryan, line 9: Indent for more spaces here too

src/Resources/skeleton/resetPassword/twig_check_email.tpl.php
weaverryan: Also, I guess we should say "hour(s)" on the end. I don't think we should try to introduce the translator so that we can "translate" hour versus hours. Or we could add the proper logic, though it's a little ugly:

```twig
{{ tokenLifetime|date('g') }} hour{{ tokenLifetime|date('g') != 1 ? 's' }} hour
```

If we did that, I might also pass `tokenLifetimeHours` into the template for convenience.

src/Maker/MakeResetPassword.php
weaverryan: Add `final` and `@internal` for good measure.

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+
+    public static function getCommandName(): string
+    {
+        return 'make:reset-password';
+    }
+
+    public function configureCommand(
+        Command $command,
+        InputConfiguration $inputConfig
+    ) {
```
weaverryan, line 56: Single line

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+        $command
+            ->setDescription('Create controller, entity, and repositories for use with SymfonyCasts Reset Password Bundle.')
```
weaverryan, line 58:
```suggestion
            ->setDescription('Create controller, entity, and repositories for use with symfonycasts/reset-password-bundle.')
```

src/Maker/MakeResetPassword.php
```diff
@@ -0,0 +1,277 @@
...
+            $repositoryClassNameDetails->getFullName(),
+            'resetPassword/ResetPasswordRequestRepository.tpl.php',
+            [
+                'request_class_full_name' => $requestClassNameDetails->getFullName(),
+                'request_class_name' => $requestClassNameDetails->getShortName(),
+            ]
+        );
+
+        if (!$this->fileManager->fileExists($path = 'config/packages/reset_password.yaml')) {
+            throw new RuntimeCommandException(sprintf('The file "%s" does not exist. This command needs that file to accurately build the reset password config.', $path));
```
weaverryan, line 221: Per my comment below, this is not true. It IS unexpected for sure. I guess in this case I would... still create the file from "scratch".

src/Maker/MakeResetPassword.php
jrushlow: done and done

tests/fixtures/MakeResetPassword/config/bundles.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+return [
+    Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => true],
+    Symfony\Bundle\MakerBundle\MakerBundle::class => ['dev' => true],
+    Symfony\Bundle\SecurityBundle\SecurityBundle::class => ['all' => true],
+    SymfonyCasts\Bundle\ResetPassword\SymfonyCastsResetPasswordBundle::class => ['all' => true],
+];
```
weaverryan, line 8: This shouldn't be needed now: the test will install the bundle, and its recipe will update this.

Same below for the `reset_password.yaml` file - the recipe will add that.

tests/fixtures/MakeResetPassword/src/Entity/User.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace App\Entity;
+
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class User implements UserInterface
+{
+
```
weaverryan, line 9: extra line

src/Maker/MakeResetPassword.php
jrushlow: done

tests/fixtures/MakeResetPassword/src/Entity/User.php
```diff
@@ -0,0 +1,39 @@
...
+    private $email;
```
weaverryan, line 10: No setter for this?

tests/Maker/MakeResetPasswordTest.php
```diff
@@ -0,0 +1,34 @@
...
+        yield 'reset_password' => [MakerTestDetails::createTest(
+            $this->getMakerInstance(MakeResetPassword::class),
+            [
+                'jr@rushlow.dev',
+                'SymfonyCasts',
+                'App\Entity\User',
+                'app_home',
+            ])
+            ->addExtraDependencies('security-bundle')
+            ->setFixtureFilesPath(__DIR__.'/../fixtures/MakeResetPassword'),
```
weaverryan, line 31: We should be a functional test inside the skeleton. For simplicity, it could *just* go to `/reset-password` and check for a 200 status code. The gold standard would be to actually fill out the form, capture the reset password link somehow, then go to THAT URL and reset the password :). We could do that "gold standard" never or later ;)

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
jrushlow: done

src/Resources/skeleton/resetPassword/twig_request.tpl.php
jrushlow: done

src/Resources/skeleton/resetPassword/twig_reset.tpl.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done

src/Resources/skeleton/resetPassword/twig_check_email.tpl.php
jrushlow: I went with `hours(s)`.

From a technical stand point, using the proper logic makes sense. But, I suspect most people will be changing the format of this template to suite their needs (match UI/business themes, etc..). So perhaps we would introduce the "proper logic" at a later time if it is requested..

jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: made a note on the board to implement in the near future..

jrushlow:
TODO - I keep deleting `tests/tmp/.gitignore` locally between tests. Storm auto magically adds that change on `ctrl+k`, not to self: remember to put that back when all is said and done....

tests/fixtures/MakeResetPassword/src/Entity/User.php
jrushlow: done

src/Resources/skeleton/resetPassword/twig_email.tpl.php
jrushlow: done

src/Maker/MakeResetPassword.php
jrushlow: done,

1) if the config file doesn't exist, we assume dev is using a customized config. Remind dev to set correct repository and move on.

2) if the config exists with only 1 key, assume config was created by flex; create a new config file.

3) if config exists with more than 1 key, assume the dev modified the config before running maker. Replace repository value and preserve everything else.

src/Maker/MakeResetPassword.php
jrushlow: logic removed - see `MakeResetPassword::setBundleConfig()`

weaverryan:
Merged! A HUGE thanks to @jrushlow for his MASSIVE work on this feature and also @romaricdrigon who kick-started this a long time ago and kept pushing on it. I'm super happy to not only have this feature in Maker, but for it to be *really*, really nice. ❤️

PR #568, opened 2020-03-28
ignoring external deprecations

weaverryan:
For example, there is nothing we can do about doctrine/orm using doctrine/common persistence classes. Reference: https://symfony.com/doc/current/components/phpunit_bridge.html#internal-deprecations

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -12,6 +12,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="TEST_DATABASE_DSN" value="mysql://root:@127.0.0.1:3306/test_maker" />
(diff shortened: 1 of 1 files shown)


weaverryan:
Once we get this all handled, we could use an approach like Symfony's: https://github.com/symfony/symfony/blob/eb0e47b40c6524b51b1d33338f19bc2ac1355538/phpunit#L25-L32

Basically, we might isolate the tests that have these `Doctrine\Common` deprecations into a separate test suite (or give them all a tag) and then run the tests in two passes. For one of the passes, we can change the `SYMFONY_DEPRECATIONS_HELPER` to some version of `max[indirect]=7`

PR #569, opened 2020-03-28
remove forgotten password maker

jrushlow:
maker replaced with reset password maker


weaverryan:
Thank you Jesse! For the updated version, see #567

PR #572, opened 2020-03-28
created email validator

jrushlow:
validated an email address using PHP's `filter_var` method else throw a `RuntimeCommandException`

use case: reset-password-bundle

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -167,6 +167,15 @@ final class Validator
         return $name;
     }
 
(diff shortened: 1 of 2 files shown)


weaverryan:
@jrushlow Can you just put this into #567 if we need it? That way we can see it in context there :)

jrushlow:
sounds good to me.. i think we talked about that before too.. 

will close PR after updating 567

PR #573, opened 2020-03-28
targetEntity short syntax, fixes #561

### LINKED ISSUE #561: [Question] targetEntity syntax
Hi, I've a question about the generated relationship annotation, that is using the fqcn as string instead of `Target::class` (forcing the import of the class).

The maker-bundle is able to receive an PR with this changes or have some reason to keep the string?

**Example:**

Generated annotation:

```php
/**
 * @ORM\ManyToOne(targetEntity="App\Entity\User")
 * @ORM\JoinColumn(nullable=false)
 */
private $user;
```

Instead of force to import target class:

```php
/**
 * @ORM\ManyToOne(targetEntity=User::class)
 * @ORM\JoinColumn(nullable=false)
 */
private $user;
```

ps: Currently the maker already use import statement to reference the class in setter/getter methods.

rogeriolino:
Relation annotation targetEntity value as short class name: `Target::class`

### DIFF
--- a/src/Util/ClassNameValue.php
+++ b/src/Util/ClassNameValue.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+use Symfony\Bundle\MakerBundle\Str;
+
+/**
+ * @internal
+ */
+final class ClassNameValue
+{
+    private $typeHint;
+    private $fullClassName;
+
+    public function __construct(string $typeHint, string $fullClassName)
+    {
+        $this->typeHint = $typeHint;
+        $this->fullClassName = $fullClassName;
+    }
+
+    public function getShortName(): string
+    {
+        if ('self' === $this->typeHint) {
+            return Str::getShortClassName($this->fullClassName);
+        }
+
+        return $this->typeHint;
+    }
+
+    public function __toString()
+    {
+        return $this->getShortName();
+    }
+}
--- a/tests/fixtures/MakeFormForEntity/src/Entity/SourFood.php
+++ b/tests/fixtures/MakeFormForEntity/src/Entity/SourFood.php
@@ -23,7 +23,7 @@ class SourFood
     private $title;
 
     /**
-     * @ORM\OneToMany(targetEntity="App\Entity\Property", mappedBy="sourFood")
+     * @ORM\OneToMany(targetEntity=Property::class, mappedBy="sourFood")
      */
     private $properties;
 
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -46,6 +46,7 @@ final class EntityClassGenerator
             'doctrine/Entity.tpl.php',
             [
                 'repository_full_class_name' => $repoClassDetails->getFullName(),
+                'repository_class_name' => $repoClassDetails->getShortName(),
                 'api_resource' => $apiResource,
                 'should_escape_table_name' => $this->doctrineHelper->isKeyword($tableName),
                 'table_name' => $tableName,
--- a/src/Resources/skeleton/doctrine/Entity.tpl.php
+++ b/src/Resources/skeleton/doctrine/Entity.tpl.php
@@ -4,12 +4,13 @@ namespace <?= $namespace ?>;
 
 <?php if ($api_resource): ?>use ApiPlatform\Core\Annotation\ApiResource;
 <?php endif ?>
+use <?= $repository_full_class_name ?>;
 use Doctrine\ORM\Mapping as ORM;
 
 /**
 <?php if ($api_resource): ?> * @ApiResource()
 <?php endif ?>
- * @ORM\Entity(repositoryClass="<?= $repository_full_class_name ?>")
+ * @ORM\Entity(repositoryClass=<?= $repository_class_name ?>::class)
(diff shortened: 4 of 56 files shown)


rogeriolino:
It's not so easy (sorry). I'll fix it.

weaverryan:
Haha, no problem. Yes, it's tricky. And if the target class is in a different namespace, we also need a `use` statement. If it's in the *same* namespace, then we should not add a `use` statement.

rogeriolino:
@weaverryan what yout think about create a new class to wrap the string value? So we can use it in the `Embedded` annotation too.

```php
// ClassSourceManipulator::quoteAnnotationValue

if ($value instanceof TargetClassName) {
    // TargetClassName::__tostring
    return sprintf('%s::class', $value);
}
```

This approach will prevent to change the `ClassSourceManipulator::quoteAnnotationValue` signature to check the prop/option name to know if parse it or not as `::class`.

ps: If you agree with that, please suggest me a name for this new class (and the namespace).

src/Util/ClassNameValue.php
```diff
@@ -0,0 +1,43 @@
...
+    public function getShortName(): string
+    {
+        if ('self' === $this->typeHint) {
+            return Str::getShortClassName($this->fullClassName);
+        }
+
+        return $this->typeHint;
+    }
+
+    public function __toString()
```
weaverryan, line 39: Can we avoid having a `__toString()`? I'd prefer to be explicit and call `getShortName()`

src/Util/ClassNameValue.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function __construct(string $fullClassName, string $typeHint)
+    {
+        $this->fullClassName = $fullClassName;
+        $this->typeHint = $typeHint;
+    }
+
+    public function getShortName(): string
+    {
+        if ('self' === $this->typeHint) {
```
weaverryan, line 32: The `self` case is (I believe) for cases when you're adding a relation that is from one class to *itself*. In that case, why would we not just use the short class name like normal? I think I'm missing something :).

tests/fixtures/MakeFormForEntity/src/Entity/SourFood.php
```diff
@@ -23,7 +23,7 @@ class SourFood
     private $title;
 
     /**
-     * @ORM\OneToMany(targetEntity="App\Entity\Property", mappedBy="sourFood")
+     * @ORM\OneToMany(targetEntity=Property::class, mappedBy="sourFood")
      */
     private $properties;
```
weaverryan, line 29: Could we add a test case for when the 2 entities are *not* in the same namespace?

rogeriolino: In this case, add another relationship with `Property` from another namespace?

- App\Entity\SourFood
- App\Entity\Property
- Foo\Bar\Property

src/Util/ClassNameValue.php
rogeriolino: I was afraid to use `self::class`, then I'd it.

tests/fixtures/MakeFormForEntity/src/Entity/SourFood.php
weaverryan: That's exactly what I was thinking. Or, I think it would be fine to create another entity in a sub-directory - e.g. `App\Entity\SubDirectory\SomeOtherClass` - and them make a relationship from `SourFood` to *that* entity. That would still require a `use` statement (so it would be a good example) but having it inside of the `Entity/` directory might be easier to get working :)

src/Util/ClassNameValue.php
weaverryan: Does `self::class()` work inside the relation annotations? If it does, I guess, yea, let's use it!

rogeriolino: I've tried to use `self` inside annotation but it doesn't work.

tests/fixtures/MakeFormForEntity/src/Entity/SourFood.php
rogeriolino: Added test to add new relation with same short class name: https://github.com/symfony/maker-bundle/pull/573/commits/5f2a8b968a6368af66f006572f4a3c5a40744f4d#diff-55d2be7d85da27a6541ac1a80e425d1c

src/Util/ClassNameValue.php
weaverryan, line 39: We still need to remove this method :)

src/Util/ClassNameValue.php
weaverryan: Ok, no problem :).

So... is this `if` statement needed? I don't see how `self` is a special case. We pass the `$typeHint` as the second arg to the constructor. This comes from `ClassSourceManipulator`, where it has already done the logic to figure out the correct "short name" that's needed... including (I believe) when a relation is being made it itself. Can we remove this whole if block?

rogeriolino: Right, the `ClassSourceManipulator` resolve the `typeHint`, that can be `self` (when self relation `[1]`), class shortname (normal use `[2]`), or fqcn (when conflict `[3]`). Only in the case that the `typeHint` value is `self` we can't use it, we need to specify the class name (shortname for the self class).

```php
class Product {
    /** [1]
     * @var self
     * @ORM\ManyToOne(targetEntity=Product::class)
     */
    private $parent;

    /** [2]
     * @var Category
     * @ORM\ManyToOne(targetEntity=Category::class)
     */
    private $category;

    /** [3]
     * @var \App\Entity\Subdir\Category
     * @ORM\ManyToOne(targetEntity=\App\Entity\Subdir\Category::class)
     */
    private $subCategory;
}
```

BTW I can't figure out a better way to resolve it. :confused:

weaverryan:
Thank you @rogeriolino! This is awesome - great work!

PR #574, opened 2020-03-29
New Messenger Middleware Maker command

zairigimad:
New Command to generate Middleware

capture

generated code : 

<details>
  <summary>CustomMiddleware</summary>
  
```php
<?php
namespace App\Middleware;

use Symfony\Component\Messenger\Envelope;
use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
use Symfony\Component\Messenger\Middleware\StackInterface;

final class CustomMiddleware implements MiddlewareInterface
{
     public function handle(Envelope $envelope, StackInterface $stack): Envelope
     {
         // ...
         return $stack->next()->handle($envelope, $stack);
     }
}

```
</details>

### DIFF
--- a/src/Maker/MakeMessengerMiddleware.php
+++ b/src/Maker/MakeMessengerMiddleware.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+/**
+ * @author Imad ZAIRIG <imadzairig@gmail.com>
+ *
+ * @internal
+ */
+final class MakeMessengerMiddleware extends AbstractMaker
+{
+    public static function getCommandName(): string
+    {
+        return 'make:messenger-middleware';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig)
+    {
+        $command
+            ->setDescription('Creates a new messenger middleware')
+            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the middleware class (e.g. <fg=yellow>CustomMiddleware</>)')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeMessage.txt'));
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $middlewareClassNameDetails = $generator->createClassNameDetails(
+            $input->getArgument('name'),
+            'Middleware\\',
+            'Middleware'
+        );
+
+        $generator->generateClass(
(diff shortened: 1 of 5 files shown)


src/Maker/MakeMiddleware.php
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * @author Imad ZAIRIG <imadzairig@gmail.com>
+ *
+ * @internal
+ */
+final class MakeMiddleware extends AbstractMaker
+{
+    public static function getCommandName(): string
+    {
+        return 'make:middleware';
```
weaverryan: I think it needs to be `make:messenger-middleware`. While the only concept in Symfony of a "middleware" is the Messenger middleware, the concept, in general, is generic.

src/Maker/MakeMiddleware.php
```diff
@@ -0,0 +1,73 @@
...
+            'middleware/Middleware.tpl.php'
+        );
+
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text([
+            'Next: Open your new middleware class and add the code you need.',
+            'Find the documentation at <fg=yellow>https://symfony.com/doc/current/messenger.html</>',
```
weaverryan: I guess we could link straight to: https://symfony.com/doc/current/messenger.html#middleware

weaverryan: And there is a second step we need to mention: you need to add this middleware to your `messenger.yaml` file. But I don't think we should attempt to add this automatically in this command - there are just too many use-cases. But we should mention (in this message) that you need to add it.

src/Maker/MakeMessengerMiddleware.php
```diff
@@ -59,7 +59,8 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
 
         $io->text([
             'Next: Open your new middleware class and add the code you need.',
-            'Find the documentation at <fg=yellow>https://symfony.com/doc/current/messenger.html</>',
+            '      you need to add this middleware to your <info>messenger.yaml</info> file.',
+            'Find the documentation at <fg=yellow>https://symfony.com/doc/current/messenger.html#middleware</>',
```
weaverryan, line 63: I'd format it like this:

```
Next:
- Open the <info>src/Messenger/.../RealPath.php</> class and add the code you need
- Add the middleware to your <info>config/packages/messenger.yaml</> file

Find the documentation at <fg=yellow>https://symfony.com/doc/current/messenger.html#middleware</>
```

tests/fixtures/MakeMessengerMiddleware/src/Middleware/CustomMiddleware.php
```diff
@@ -0,0 +1,16 @@
...
+use Symfony\Component\Messenger\Middleware\StackInterface;
+
+final class CustomMiddleware implements MiddlewareInterface
+{
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        // ...
+        return $stack->next()->handle($envelope, $stack);
+    }
+}
```
weaverryan, line 16: This file (and directory) aren't needed. fixtures directories are used when you need to *start* the test project with files *before* running the maker commands. It's not used as an "assert" that the created file matches this (or anything like that).

tests/Maker/MakeMessengerMiddlewareTest.php
```diff
@@ -0,0 +1,30 @@
...
+class MakeMessengerMiddlewareTest extends MakerTestCase
+{
+    public function getTestDetails()
+    {
+        yield 'messenger_middleware' => [MakerTestDetails::createTest(
+            $this->getMakerInstance(MakeMessengerMiddleware::class),
+            [
+                // middleware name
+                'CustomMiddleware',
+            ]),
```
weaverryan, line 27: Add this so we can assert the file was created:

```php
                $this->assertFileExists($directory.'/src/Middleware/CustomMiddleware.php');
            }),
```

weaverryan:
Great work! Thank you very much for this @zairigimad!

zairigimad:
> Great work! Thank you very much for this @zairigimad!

Thank you for your help @weaverryan :)

PR #577, opened 2020-04-01
Make Migration - Fix db/shard options value forward

cvergne:
Hi,

A very small fix noticed when playing a little with the `make:migration` command :)


weaverryan:
What an embarrassing but that I'm pretty sure I made :). Thanks for fixing that Christophe!

PR #578, opened 2020-04-01
Move login route to const

JakubSzczesniak:
To eliminate long debug time and bug fixing in case of change login route.  Connected with https://github.com/symfony/symfony-docs/pull/13363

### DIFF
--- a/src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
+++ b/src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
@@ -24,6 +24,8 @@ class <?= $class_name; ?> extends AbstractFormLoginAuthenticator<?= $password_au
 {
     use TargetPathTrait;
 
+    public const LOGIN_ROUTE = 'app_login';
+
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
```diff
@@ -24,6 +24,8 @@ class <?= $class_name; ?> extends AbstractFormLoginAuthenticator<?= $password_au
 {
     use TargetPathTrait;
 
+    public const LOGIN_ROUTE = 'app_login';
```
weaverryan, line 27: This could be private, correct? We're only using it from inside this class?

JakubSzczesniak: I set it as public if someone want to use this const in other place. Personally, I use it one action.

weaverryan: Sure :). Seems reasonable

weaverryan:
Thank you Jakub!

PR #584, opened 2020-04-09
Add also sodium in the comment

mercuryseries:
Add also sodium in the comment

### DIFF
--- a/src/Security/UserClassBuilder.php
+++ b/src/Security/UserClassBuilder.php
@@ -232,7 +232,7 @@ final class UserClassBuilder
         if ($userClassConfig->hasPassword()) {
             $builder->addStmt(
                 $manipulator->createMethodLevelCommentNode(
-                    'not needed when using the "bcrypt" algorithm in security.yaml'
+                    'not needed when using the "bcrypt/sodium" algorithm in security.yaml'
                 )
             );
         } else {


src/Security/UserClassBuilder.php
```diff
@@ -232,7 +232,7 @@ private function addGetSalt(ClassSourceManipulator $manipulator, UserClassConfig
         if ($userClassConfig->hasPassword()) {
             $builder->addStmt(
                 $manipulator->createMethodLevelCommentNode(
-                    'not needed when using the "bcrypt" algorithm in security.yaml'
+                    'not needed when using the "bcrypt/sodium" algorithm in security.yaml'
```
weaverryan, line 235: Also not needed for `auto` or `argon2i`. Maybe we should just say:

> not needed because we are using a modern encoder algorithm in `security.yaml`

BatsaxIV: > not needed when using a modern encoder algorithm in security.yaml ?

Maybe? _Because_ presume that a modern encoder is used, but it's not necessarly the case right?

jrushlow, line 235:
```suggestion
                    'not needed when using a modern encoder algorithm in security.yaml'
```

@BatsaxIV I think it's safe to go with your suggestion. I'll take the heat if @weaverryan disagrees :P

chalasr:
This has been addressed in #805. Thank you for the PR.

PR #587, opened 2020-04-12
change from-email-address default value to empty string

IndraGunawan:
when create a reset-password at question `What email address will be used to send reset confirmations? e.g. mailer@your-domain.com` and we just enter without filling the answer it will thrown an exception 
```
Argument 1 passed to Symfony\Bundle\MakerBundle\Validator::validateEmailAddress() must be of the type string, null given, called in /path/vendor/symfony/console/Helper/QuestionHelper.php on line 468
```

there are 2 possible solution. 1. change the default value to empty string (this PR), or 2. make the `validateEmailAddress` argument nullable and check the value inside the method.

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -167,7 +167,7 @@ final class Validator
         return $name;
     }
 
-    public static function validateEmailAddress(string $email): string
+    public static function validateEmailAddress(?string $email): string
     {
         if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
             throw new RuntimeCommandException(sprintf('"%s" is not a valid email address.', $email));
--- a/tests/ValidatorTest.php
+++ b/tests/ValidatorTest.php
@@ -68,10 +68,17 @@ class ValidatorTest extends TestCase
     public function testValidateEmailAddress()
     {
         $this->assertSame('jr@rushlow.dev', Validator::validateEmailAddress('jr@rushlow.dev'));
+    }
 
+    public function testInvalidateEmailAddress()
+    {
         $this->expectException(RuntimeCommandException::class);
         $this->expectExceptionMessage('"badEmail" is not a valid email address.');
         Validator::validateEmailAddress('badEmail');
+
+        $this->expectException(RuntimeCommandException::class);
+        $this->expectExceptionMessage('"" is not a valid email address.');
+        Validator::validateEmailAddress('');
     }
 
     public function testInvalidClassName()


jrushlow:
I'll look deeper into this tonight, but as I recall, the email address & name questions are not intended to accept `null` or be `empty` as those values are required to properly generate the mailer template. 
https://github.com/symfony/maker-bundle/blob/a9909bc23c89c3d29a4319d050bd27047cc6d435/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php#L158

the actual values provided to those questions can be changed later in the controller.

src/Maker/MakeResetPassword.php
```diff
@@ -125,7 +125,7 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
 
         $input->setArgument('from-email-address', $io->ask(
             'What email address will be used to send reset confirmations? e.g. mailer@your-domain.com',
-            null,
+            '',
             [Validator::class, 'validateEmailAddress']
         ));
```
jrushlow, line 130: Allowing an empty value `''` would cause the email address validator to fail resulting in the question being repeated until a valid email address was provided.

weaverryan: That IS the behavior we want (they should be forced to enter a value). So are you saying the patch is +1 for you?

jrushlow:
> 2. make the validateEmailAddress argument nullable

I personally am in favor of option 2. Setting the default value to `''` results in - 
`What email address will be used to send reset confirmations? e.g. mailer@your-domain.com []:` 
That could be interpreted as the value _could_ be null. 

But keeping the default `null` and allowing `validateEmailAddress(?string $email)` keeps the question formatting the same (no `... []:`) while throwing the validator error, as intended. Thoughts?

@IndraGunawan thanks for catching this, I must has looked at that invalid argument error a dozen + times and it never clicked that I was seeing the wrong error message.

src/Maker/MakeResetPassword.php
jrushlow: @weaverryan see my comment below. I think that better conveys what I was trying to say here.

IndraGunawan:
I also push option 2, which option will be selected later then i will revert another commit

weaverryan:
Yes, we want option 2. We want the prompt to not include the `[]` at the end. And if the user enters nothing, we want it to continue asking the question. Can you revert the extra code from option (1)? 

Thanks!

src/Validator.php
```diff
@@ -167,8 +167,12 @@ public static function validateDoctrineFieldName(string $name, ManagerRegistry $
         return $name;
     }
 
-    public static function validateEmailAddress(string $email): string
+    public static function validateEmailAddress(?string $email): string
     {
+        if (empty($email)) {
```
weaverryan, line 172:
```suggestion
        if (null === $email) {
```

IndraGunawan:
done. @weaverryan

src/Validator.php
```diff
@@ -167,8 +167,12 @@ public static function validateDoctrineFieldName(string $name, ManagerRegistry $
         return $name;
     }
 
-    public static function validateEmailAddress(string $email): string
+    public static function validateEmailAddress(?string $email): string
     {
+        if (null === $email) {
+            throw new RuntimeCommandException('Email should not be null.');
+        }
+
```
jrushlow, line 175: I don't think this is needed as the `filter_var()` call just below is doing the same thing.

IndraGunawan:
nullable types is in php 7.1 (https://www.php.net/manual/en/migration71.new-features.php) but this bundle require php 7.0. is it ok? cc: @weaverryan

weaverryan:
> nullable types is in php 7.1 (https://www.php.net/manual/en/migration71.new-features.php) but this bundle require php 7.0. is it ok? cc: @weaverryan

Good catch! It's actually not ok. We should make the argument `string $email = null` to allow a null value to be passed in PHP 7.0. You could add a check like this - https://github.com/symfony/maker-bundle/blob/10200754ce88a795c6933d4be7ac0c936c443e86/src/Util/MakerFileLinkFormatter.php#L25-L28 - so that we could easily make the argument nullable later without breaking backwards compatibility. 

Thanks!

IndraGunawan:
this PR doesn't need any change because of this #598 has been merged. ready for merge

weaverryan:
Ha! You're right! In fact, it was your comment that led me to revisit what version of php we should support today. So, a double thanks to you!

PR #588, opened 2020-04-15
fixed missing doctrine dependency

jrushlow:
To use the classes generated by `make:reset-password` as is, doctrine is required.


jrushlow:
- [ ] require annotations

weaverryan:
Thanks Jesse!

PR #589, opened 2020-04-16
replaced Generator by EntityClassGenerator in MakerResetPassword

archeoprog:
Q | A
-- | --
Branch? | master
Bug fix? | no
New feature? | - yes (internal)
Deprecations? | - no
Tickets | - no
License | MIT
Doc PR | - no need

At the request of Ryan Weaver.

I replaced the use of `Generator` by `EntityClassGenerator`. 

We then needed to use `ClassSourceManipulator` and i propose to add 3 methods to this class, tests included.

1) `$manipulator->clearClassNodeStmts()`
Very short method to remove all `Node\Stmts` in the class and update the source code.

2) `$manipulator->addTrait()`
Add a `use Trait;` at first in the class.

3) `$manipulator->addConstructor()`
Is a shortcut to add a constructor in the right place.

I also propose to add two optional parameters to the `$manipulator->addMethodBuilder()` method: `array $params = [], string $methodBody = null;`

Allowing to build methods more directly when needed.


.env
```diff
@@ -0,0 +1,14 @@
+# In all environments, the following files are loaded if they exist,
```
jrushlow, line 1: I think you may have accidentally committed this `.env` file.

jrushlow:
PHP-CS-Fixer is failing in CI, perhaps run `vendor/bin/php-cs-fixer fix --diff --diff-format=udiff --dry-run` locally to inspect the proposed changes, then run that command again without `--dry-run`

composer.json
```diff
@@ -30,7 +32,7 @@
         "friendsofphp/php-cs-fixer": "^2.8",
         "friendsoftwig/twigcs": "^3.1.2",
         "symfony/http-client": "^4.3|^5.0",
-        "symfony/phpunit-bridge": "^4.3|^5.0",
+        "symfony/phpunit-bridge": "^5.1@dev",
```
jrushlow, line 35: Why the change in version constraints? We may want to leave this at `^4.3|^5.0`. `minimum-stability: dev` will use the latest dev releases.

composer.json
```diff
@@ -14,13 +14,15 @@
     "minimum-stability": "dev",
     "require": {
         "php": "^7.0.8",
+        "doctrine/annotations": "^1.11@dev",
```
jrushlow, line 17: I think this may have been added by accident. But I could be mistaken. Either way, requiring `@dev` would be wise here, as app in production would then be forced to use unstable branches.  Same for `symfony/flex`

archeoprog: Yes, sorry. i am looking for fixing all theses errors, it is my first contribution to a collectif project i am a bit lost...

jrushlow: hey no problem, if you have any questions feel free to DM on slack `jrushlow`

archeoprog:
First experience

archeoprog:
Mishandling, major conflicts.

PR #590, opened 2020-04-17
(Update #589) Replace Generator by EntityClassGenerator in MakerResetPassword

archeoprog:
**UPDATE PULL REQUEST #589**
Q | A
-- | --
Branch? | master
Bug fix? | no
New feature? | - yes (internal)
Deprecations? | - no
Tickets | - no
License | MIT
Doc PR | - no need

At the request of Ryan Weaver.

I replaced the use of Generator by EntityClassGenerator.

We then needed to use ClassSourceManipulator and i propose to add 3 methods to this class, tests included.

    $manipulator->clearClassNodeStmts()
    Very short method to remove all Node\Stmts in the class and update the source code.

    $manipulator->addTrait()
    Add a use Trait; at first in the class.

    $manipulator->addConstructor()
    Is a shortcut to add a constructor in the right place.

I also propose to add two optional parameters to the $manipulator->addMethodBuilder() method: array $params = [], string $methodBody = null;

Allowing to build methods more directly when needed.


src/Maker/MakeResetPassword.php
```diff
@@ -194,23 +203,94 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
-            ]
+        $manipulator->addUseStatementIfNecessary(ResetPasswordRequestInterface::class);
+
+        $manipulator->addUseStatementIfNecessary(ResetPasswordRequestTrait::class);
+
+        $manipulator->addInterface('ResetPasswordRequestInterface');
+
+        $manipulator->clearClassNodeStmts();
+
+        $manipulator->addTrait('ResetPasswordRequestTrait');
```
weaverryan, line 224: Inside this method, IT should handle adding the use statement. So we should not need the `addUseStatementIfNecessary` above this.

The same is true for `addInterface()`: THAT method should add the `use` statement *and* add implement the interface.

src/Maker/MakeResetPassword.php
```diff
@@ -194,23 +203,94 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+        $manipulator->addTrait('ResetPasswordRequestRepositoryTrait');
+
+        $manipulator->addConstructor([
+                ['registry', null, 'ManagerRegistry'],
+            ], <<<'CODE'
+<?php
+parent::__construct($registry, ResetPasswordRequest::class);
+CODE
+        );
+        $methodBuilder = $manipulator->createMethodBuilder('createResetPasswordRequest', 'ResetPasswordRequestInterface', false);
```
weaverryan, line 281: I wrote this method originally (I see), but now that I look at it, I think the 2nd argument to this method should be the full class/interface name of the return type (for cases where it's a class/interface) - so `ResetPasswordRequestInterface::class`. THEN, the `createMethodBuilder()` internally should check to see if the passed returnType is a class or interface (check both `class_exists()` and `interface_exists()`) and add the `use` statement automatically IF it is a class or interface.

src/Maker/MakeResetPassword.php
archeoprog: Done

src/Maker/MakeResetPassword.php
archeoprog: Done

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +233,36 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
...
-    public function addMethodBuilder(Builder\Method $methodBuilder)
+    public function addConstructor(array $params = [], string $methodBody = null)
     {
+        if ($this->getConstructorNode()) {
+            throw new \LogicException('Constructor already exists.', 1);
+        }
+
+        $methodBuilder = $this->createMethodBuilder('__construct', null, false, []);
+
+        if (!empty($params)) {
```
weaverryan, line 244: This check isn't needed - if `$params` is empty, `addMethodParams()` simply won't loop over anything.

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +233,36 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
         $this->addMethod($builder->getNode());
     }
 
-    public function addMethodBuilder(Builder\Method $methodBuilder)
+    public function addConstructor(array $params = [], string $methodBody = null)
     {
+        if ($this->getConstructorNode()) {
+            throw new \LogicException('Constructor already exists.', 1);
```
weaverryan, line 239: No 2nd argument is needed - we rarely specify an exception code.

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +233,36 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
...
+        }
+
+        $methodBuilder = $this->createMethodBuilder('__construct', null, false, []);
```
weaverryan, line 242: Last argument is optional, so it's not needed here.

src/Util/ClassSourceManipulator.php
```diff
@@ -208,13 +272,29 @@ public function addMethodBody(Builder\Method $methodBuilder, string $methodBody)
         $methodBuilder->addStmts($nodes);
     }
 
+    public function addMethodParams(Builder\Method $methodBuilder, array $params)
```
weaverryan, line 275: This should be private (and so, it should be at the bottom of this class)

src/Maker/MakeResetPassword.php
```diff
@@ -194,23 +203,84 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+
+        $manipulator->clearClassNodeStmts();
+
+        $manipulator->addTrait(ResetPasswordRequestTrait::class);
+
+        $manipulator->addConstructor([
+                ['user', null, 'object'],
+                ['expiresAt', null, '\DateTimeInterface'],
+                ['selector', null, 'string'],
+                ['hashedToken', null, 'string'],
```
weaverryan, line 226: Instead of `$params` being an array of arrays, could we use the `PhpParser\Builder\Param` class instead? I think it would look like this:

```php
$manipulator->addConstructor([
    (new Param('user')->setType('object')),
    (new Param('expiresAt')->setType(\DateTimeInterface::class)),
],
// ...
```

Then we would just need to update the `ClassSourceManipulator` to call `->getNode()` on each of these Param "Builder" classes to convert it into a real `Param` node.

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +233,36 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
         $this->addMethod($builder->getNode());
     }
 
-    public function addMethodBuilder(Builder\Method $methodBuilder)
+    public function addConstructor(array $params = [], string $methodBody = null)
```
weaverryan, line 236: I think we could make both arguments required - I don't see a big use-case for an empty constructor body.

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +233,36 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
...
     {
+        if ($this->getConstructorNode()) {
```
weaverryan, line 238:
```suggestion
        if (null !== $this->getConstructorNode()) {
```

This is a type of stronger coding standard we use in Symfony. I don't always do this in my *own* code, but I do use this here.

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +233,36 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
...
+            throw new \LogicException('Constructor already exists.', 1);
+        }
+
+        $methodBuilder = $this->createMethodBuilder('__construct', null, false, []);
+
+        if (!empty($params)) {
+            $this->addMethodParams($methodBuilder, $params);
+        }
+
+        if ($methodBody) {
```
weaverryan, line 248: Can we remove this if? If, for some reason, no methodBody were passed, do we get an error?

src/Util/ClassSourceManipulator.php
```diff
@@ -178,6 +180,40 @@ public function addInterface(string $interfaceName)
         $this->updateSourceCodeFromNewStmts();
     }
 
+    public function addTrait($classTraitName)
+    {
+        $this->addUseStatementIfNecessary($classTraitName);
```
weaverryan, line 185: I think this should be:

```suggestion
        $name = $this->addUseStatementIfNecessary($classTraitName);
```

The `addUseStatementIfNecessary` returns the "short name" that is now available to be used through the rest of the class. Then, the next 2 lines should not be needed.

This catches edge-case situations where the `use` statement is already on the class, but maybe is aliased to something.

src/Util/ClassSourceManipulator.php
```diff
@@ -178,6 +180,40 @@ public function addInterface(string $interfaceName)
...
+
+        $classNode = $this->getClassNode();
+        $name = Str::getShortClassName($classTraitName);
+
+        if (empty($classNode->stmts)) {
+            $classNode->stmts[0] = new Node\Name('use '.$name.';');
```
weaverryan, line 191: Do we need this `if` statement? If there are no `stmts` yet on the class, wouldn't the `array_unshift` on line 211 work like normal? This is not a spot where we need to optimize performance, so I don't mind to the work below, even if the stmts are empty.

src/Maker/MakeResetPassword.php
```diff
@@ -14,21 +14,29 @@
...
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Yaml\Yaml;
+use SymfonyCasts\Bundle\ResetPassword\Model\ResetPasswordRequestInterface;
+use SymfonyCasts\Bundle\ResetPassword\Model\ResetPasswordRequestTrait;
+use SymfonyCasts\Bundle\ResetPassword\Persistence\Repository\ResetPasswordRequestRepositoryTrait;
+use SymfonyCasts\Bundle\ResetPassword\Persistence\ResetPasswordRequestRepositoryInterface;
 use SymfonyCasts\Bundle\ResetPassword\SymfonyCastsResetPasswordBundle;
 
 /**
+ * @author Antoine Michelet <thearcheoprogressive@gmail.com>
```
weaverryan, line 39: This is ok - but usually, subsequent authors go at the bottom of the list :)

src/Util/ClassSourceManipulator.php
```diff
@@ -178,6 +180,40 @@ public function addInterface(string $interfaceName)
...
+        } else {
+            $traitNode = $this->findFirstNode(function ($node) {
+                return $node instanceof Node\Stmt\Class_
+                    && !empty($node->stmts[0])
+                    && $node->stmts[0] instanceof Node\Stmt\TraitUse;
+            });
```
weaverryan, line 197: This isn't actually finding a "trait node". This is kind of just checking that the `Class_` node has a `Node\Stmt\TraitUse` as its first statement. This callback will be called for *every* node in the class. And so, I think you can simply do:

```php
return $node instanceof Node\Stmt\TraitUse;
```

Then `$traitNode` will actually be an instance of `TraitUse`.

weaverryan: Hmm, but overall, I think we can simplify this and the code below by adding a new `findAllNodes(callable $filterCallback)` method that would return `Node[]`. Then we could find *all* of the `TraitUse` nodes with it, loop over *all* of them, and check if any of them match the one that we're trying to add.

src/Util/ClassSourceManipulator.php
```diff
@@ -178,6 +180,40 @@ public function addInterface(string $interfaceName)
...
+
+            if ($traitNode) {
+                foreach ($classNode->stmts as $value) {
+                    if (!$value instanceof Node\Stmt\TraitUse) {
+                        break;
+                    }
+                    // if trait already exists
+                    if (strstr($value->traits[0]->parts[0], $name)) {
```
weaverryan, line 205: We typically have used something like `$value->traits[0]->toString()`. Basically, when working with a `PhpParser\Node\Name` instance (which is what `$value->traits[0]` is here) we try to call a method on it, instead of working with the `->parts` array.

src/Maker/MakeResetPassword.php
```diff
@@ -194,23 +203,84 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+                ['user', null, 'object'],
+                ['expiresAt', null, '\DateTimeInterface'],
+                ['selector', null, 'string'],
+                ['hashedToken', null, 'string'],
+            ], <<<'CODE'
+<?php
+return new ResetPasswordRequest($user, $expiresAt, $selector, $hashedToken);
+CODE
         );
+        $this->fileManager->dumpFile($pathRequestRepository, $manipulator->getSourceCode());
```
weaverryan, line 283: I think putting all of this new code into a private method in this class might be nice. Or, its own class, which could then be unit tested a bit easier - like `UserClassBuilder`

src/Maker/MakeResetPassword.php
```diff
@@ -194,23 +203,84 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+<?php
+$this->user = $user;
+$this->initialize($expiresAt, $selector, $hashedToken);
+CODE
+        );
+        $manipulator->addProperty('id', [
+            '@ORM\Id()',
+            '@ORM\GeneratedValue()',
+            '@ORM\Column(type="integer")',
+        ]);
```
weaverryan, line 237: Because you've done such a nice job with all this code, is this code and `$manipulator->clearClassNodeStmts();` still needed on top? Or can we just leave the original class statements and avoid needing to re-add the id?

src/Maker/MakeResetPassword.php
```diff
@@ -194,23 +203,84 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+
+        $manipulator->addProperty('user', [
+            '@ORM\ManyToOne(targetEntity="App\Entity\User")',
+        ]);
+
+        $manipulator->addGetter('user', 'object', false);
```
weaverryan, line 243: Could we use `$manipulator->addSingularRelation()` ? Maybe with a new option to avoid the setter?

src/Maker/MakeResetPassword.php
archeoprog: We can add directly nodes :

archeoprog: We can pass directly nodes, this is what i wanted to do in the beginning

src/Maker/MakeResetPassword.php
archeoprog: I don't understand, all the code or a specific part ? You just select one line

src/Maker/MakeResetPassword.php
archeoprog: addSingularRelation is a private method wich is indirectly called. It would make the methods less "clean" and I don't think that's necessary, maybe we can look at this later if we have other cases ? but if you insist I can looking for .

src/Util/ClassSourceManipulator.php
archeoprog: Done, and i also add a test : testAddTraitAlReadyExists()

src/Util/ClassSourceManipulator.php
archeoprog: Done.

src/Util/ClassSourceManipulator.php
archeoprog: Done

src/Util/ClassSourceManipulator.php
archeoprog: Okay

src/Util/ClassSourceManipulator.php
archeoprog: Okay

src/Util/ClassSourceManipulator.php
archeoprog: Done

src/Maker/MakeResetPassword.php
archeoprog: I think no, if you are okay for having a getId() method wich it was not on the skeleton template. But we already need it for repository.

src/Maker/MakeResetPassword.php
weaverryan: You're right, we should just pass nodes directly. But, our code, let's *use* the "builder" - it's more convenient, or at least more expressive :) 

```php
$manipulator->addConstructor([
    (new Param('user')->setType('object'))->getNode(),
    (new Param('expiresAt')->setType(\DateTimeInterface::class))->getNode(),
],
```

src/Maker/MakeResetPassword.php
weaverryan: Ah yes, a `getId()` seems fine to me

src/Maker/MakeResetPassword.php
weaverryan: The problem is that we need to centralize the `@ORM\ManyToOne` code - I don't want that duplicated. In fact, recently, I merged a PR that updates the relationship methods on `ClassSourceManipulator` so that they generate like this:

```
@ORM\ManyToOne(targetEntity=User::class)
```

By centralizing the logic, we'll keep the way we generate things consistent. I didn't realize that `addSingularRelation` was private. So yes, we will need to make this public or think of a new public method we could add that would call (it just depends on what's cleanest - I'm not looking deeply at the code at the moment) :)

src/Maker/MakeResetPassword.php
archeoprog: Obviously, I wasn't thinking

src/Maker/MakeResetPassword.php
archeoprog: Okay, done.

src/Util/ClassSourceManipulator.php
archeoprog: You right, done. 
At the beginning it was optional because i worked on a another strategy.

src/Util/ClassSourceManipulator.php
archeoprog: Okay, good to know

src/Util/ClassSourceManipulator.php
archeoprog: Yes, done. Yes, an error will be caught

src/Util/ClassSourceManipulator.php
archeoprog: Okay, done.

src/Util/ClassSourceManipulator.php
archeoprog: The problem is not about array_unshift or performance optimization but about the presence or not of nodes. If there is nodes, we need to add a "\n" for create blankline after our first use Trait but if stmts is empty. 

`  

        if ($traitNode) {
            foreach ($traitNode as $node) {
                if ($node->traits[0]->toString() === $name) {
                    return;
                }
            }
            array_unshift($classNode->stmts, new Node\Stmt\TraitUse([
                new Node\Name($name)
            ]));
        } else {
            array_unshift($classNode->stmts,
                new Node\Name('use '.$name.(empty($classNode->stmts) ? ";" : ";\n"))
            );
        } `

src/Maker/MakeResetPassword.php
archeoprog: Ha okay, i did'nt know that we can use builder like this. I used instance of builder like that. Great, it is very clear, you right.

src/Maker/MakeResetPassword.php
archeoprog: If you are okay we can create the relation with the setter. Is it a problem ? Usually it's the classic behavior when maker creates the relations.

archeoprog: I can add this in BaseRelation

archeoprog: `

    public function isAvoidSetter() {
        return $this->avoidSetter;
    }

    public function avoidSetter(bool $avoidSetter = true) {
        $this->avoidSetter = $avoidSetter;

        return $this;
    }`

archeoprog: And avoid the setter in addSingularRelation. I just tried and it works.

archeoprog: In makeResetPassword
`

        $manipulator->addManyToOneRelation((new RelationManyToOne())
            ->setPropertyName('user')
            ->setTargetClassName("App\Entity\User")
            ->setIsNullable(true)
            ->setMapInverseRelation(false)
            ->avoidSetter()
        );`

src/Maker/MakeResetPassword.php
archeoprog: When we are okay for this issue https://github.com/symfony/maker-bundle/pull/590#discussion_r422145767
I start to think about.

src/Maker/MakeResetPassword.php
archeoprog: But i have a problem with the interface of the ResetPasswordRequestInterface
I need to type return with an object, by default it returns ?User

interface ResetPasswordRequestInterface
{
    /**
     * Get the user whom requested a password reset.
     */
    public function getUser(): object;

archeoprog: https://github.com/SymfonyCasts/reset-password-bundle/blob/96391ea830a24bc2dcfc4437edd6648cc55ebcfd/src/Model/ResetPasswordRequestInterface.php#L38

archeoprog: I can add this in BaseRelation
`

    public function getReturnType() {
        return $this->returnType;
    }

    public function setReturnType(string $returnType) {
        $this->returnType = $returnType;

        return $this;
    }`

archeoprog: And tests works

archeoprog: In makeResetPassword
`$manipulator->addManyToOneRelation((new RelationManyToOne()) ->setPropertyName('user') ->setTargetClassName("App\Entity\User") ->setIsNullable(true) ->setMapInverseRelation(false) ->setReturnType('object', false) ->avoidSetter() );`

weaverryan: Let's do this!

src/Maker/MakeResetPassword.php
archeoprog: I added a new private like you said in first because i think that this stuff is really specific to this command, not really a UserClass wich is built. See if you like :)

src/Maker/MakeResetPassword.php
archeoprog: Done.

src/Doctrine/BaseRelation.php
```diff
@@ -19,8 +19,11 @@ abstract class BaseRelation
     private $propertyName;
     private $targetClassName;
     private $targetPropertyName;
+    private $returnType;
+    private $returnTypeIsNullable = true;
```
weaverryan, line 23:
```suggestion
    private $returnTypeIsNullable;
```

I just don't want this to have a default value... so let's leave it null.

weaverryan: And maybe we should call these properties:`overriddenReturnType` and `isOverriddenReturnTypeNullable`... because that's what they *really* are. If they are not set, then it means that the "default" return type of the related class would be used.

src/Util/ClassSourceManipulator.php
```diff
@@ -178,6 +178,35 @@ public function addInterface(string $interfaceName)
...
+                if ($node->traits[0]->toString() === $name) {
+                    return;
+                }
+            }
+            array_unshift($classNode->stmts, new Node\Stmt\TraitUse([
+                new Node\Name($name),
+            ]));
+        } else {
+            array_unshift($classNode->stmts,
+                new Node\Name('use '.$name.(empty($classNode->stmts) ? ';' : ";\n"))
```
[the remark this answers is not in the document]
archeoprog: Look if you like, i done my best.

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +226,30 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
         $this->addMethod($builder->getNode());
     }
 
-    public function addMethodBuilder(Builder\Method $methodBuilder)
+    public function addConstructor(array $params, string $methodBody)
```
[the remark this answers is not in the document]
archeoprog: Done.

src/Util/ClassSourceManipulator.php
```diff
@@ -215,6 +266,10 @@ public function createMethodBuilder(string $methodName, $returnType, bool $isRet
         ;
 
         if (null !== $returnType) {
+            if (class_exists($returnType) || interface_exists($returnType)) {
+                $this->addUseStatementIfNecessary($returnType);
+                $returnType = Str::getShortClassName($returnType);
```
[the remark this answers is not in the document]
archeoprog: Node but i need to verify if class or interface exists

src/Maker/MakeResetPassword.php
```diff
@@ -320,4 +316,86 @@ private function successMessage(InputInterface $input, ConsoleStyle $io, string
         $io->text('Then open your browser, go to "/reset-password" and enjoy!');
         $io->newLine();
     }
+
+    private function generateRequestEntity(Generator $generator, $requestClassNameDetails, $repositoryClassNameDetails): void
```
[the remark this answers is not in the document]
archeoprog: Done.

src/Util/ClassSourceManipulator.php
```diff
@@ -197,8 +226,30 @@ public function addSetter(string $propertyName, $type, bool $isNullable, array $
...
+
+        $this->addMethodParams($methodBuilder, $params);
+
+        $this->addMethodBody($methodBuilder, $methodBody);
+
+        $this->addNodeAfterProperties($methodBuilder->getNode());
+        $this->updateSourceCodeFromNewStmts();
+    }
+
+    public function addMethodBuilder(Builder\Method $methodBuilder, array $params = [], string $methodBody = null)
```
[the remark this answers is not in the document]
archeoprog: Done.

src/Maker/MakeResetPassword.php
```diff
@@ -320,4 +316,86 @@ private function successMessage(InputInterface $input, ConsoleStyle $io, string
...
+    {
+        $entityClassGenerator = new EntityClassGenerator($generator, $this->doctrineHelper);
+
+        $pathRequestEntity = $entityClassGenerator->generateEntityClass($requestClassNameDetails, false);
```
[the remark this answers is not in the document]
archeoprog: Done.

src/Maker/MakeResetPassword.php
```diff
@@ -320,4 +316,86 @@ private function successMessage(InputInterface $input, ConsoleStyle $io, string
...
+<?php
+$this->user = $user;
+$this->initialize($expiresAt, $selector, $hashedToken);
+CODE
+        );
+        $manipulator->addProperty('id', [
+            '@ORM\Id()',
+            '@ORM\GeneratedValue()',
+            '@ORM\Column(type="integer")',
+        ]);
```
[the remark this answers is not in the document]
archeoprog: Done.

src/Maker/MakeResetPassword.php
```diff
@@ -320,4 +316,86 @@ private function successMessage(InputInterface $input, ConsoleStyle $io, string
...
+
+        $manipulator->addManyToOneRelation((new RelationManyToOne())
+            ->setPropertyName('user')
+            ->setTargetClassName("App\Entity\User")
```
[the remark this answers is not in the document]
archeoprog: Done.

src/Maker/MakeResetPassword.php
```diff
@@ -320,4 +316,86 @@ private function successMessage(InputInterface $input, ConsoleStyle $io, string
...
+            ->setIsNullable(true)
```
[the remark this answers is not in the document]
archeoprog: Done.

src/Doctrine/BaseRelation.php
```diff
@@ -83,4 +86,34 @@ public function setMapInverseRelation(bool $mapInverseRelation)
...
+
+        return $this;
+    }
+
+    public function getReturnType()
+    {
+        return $this->returnType;
+    }
+
+    public function setReturnType(string $returnType, $returnTypeIsNullable = true)
```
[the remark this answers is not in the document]
archeoprog: Done.

composer.json
```diff
@@ -1,53 +1,53 @@
...
+    "psr-4": { "Symfony\\Bundle\\MakerBundle\\": "src/" }
+  },
+  "autoload-dev": {
+    "psr-4": { "Symfony\\Bundle\\MakerBundle\\Tests\\": "tests/" }
+  },
+  "extra": {
+    "branch-alias": {
+      "dev-master": "1.0-dev"
+    }
+  }
```
weaverryan, line 52: This needs to be reverted - your editor may have done this automatically

src/Doctrine/BaseRelation.php
```diff
@@ -83,4 +86,34 @@ public function setMapInverseRelation(bool $mapInverseRelation)
...
+        $this->returnType = $returnType;
+        $this->overriddenReturnType = $overriddenReturnType;
+
+        return $this;
+    }
+
+    public function isOverriddenReturnTypeNullable(): bool
+    {
+        return $this->overriddenReturnType;
+    }
```
weaverryan, line 118: I think I may have confused you. Here is how I would expect the new parts of this class to look (I'm also changing the name "overridden" to "custom"):

```php
private $customReturnType;
private $isCustomReturnTypeNullable;

public function getCustomReturnType(): ?string
{
    return $this->customReturnType;
}

public function isCustomReturnTypeNullable(): bool
{
    return $this->isCustomReturnTypeNullable;
}

public function setCustomReturnType(string $customReturnType, bool $isNullable): void
{
    $this->customReturnType = $customReturnType;
    $this->isCustomReturnTypeNullable = $isNullable;
}
```

Then, in `ClassSourceManipulator`, when we use this, the code would be:

```php
        $this->addGetter(
            $relation->getPropertyName(),
            $typeHint,
            $relation->getCustomReturnType() ?: $typeHint,
            // getter methods always have nullable return values
            // unless this has been customized explicitly
            $relation->getCustomReturnType() ? $relation->isCustomReturnTypeNullable() : $typeHint,
            $relation->isOverriddenReturnTypeNullable()
        );
```

src/Doctrine/BaseSingleRelation.php
```diff
@@ -20,7 +20,11 @@ abstract class BaseSingleRelation extends BaseRelation
 
     public function isNullable(): bool
     {
-        return $this->isNullable;
+        if ($this->isNullable) {
+            return $this->isNullable;
+        }
+
+        return false;
```
weaverryan, line 27: I think (?) this isn't needed. It's true that `isNullable` could be null, but the `:bool` return type will cast null into a `false` boolean.

src/Maker/MakeResetPassword.php
```diff
@@ -320,4 +317,84 @@ private function successMessage(InputInterface $input, ConsoleStyle $io, string
...
+        ], <<<'CODE'
+<?php
+$this->user = $user;
+$this->initialize($expiresAt, $selector, $hashedToken);
+CODE
+        );
+
+        $manipulator->addManyToOneRelation((new RelationManyToOne())
+            ->setPropertyName('user')
+            ->setTargetClassName("App\Entity\User") // customization fail in php 7.3
```
weaverryan, line 351: What's the problem in PHP 7.3? We need to be explicit. And if there IS a problem currently in 7.3, then that's a block for the PR and we should be clear that this is still a "TODO" (like by posting a comment about it in the PR or in the PR description)

src/Maker/MakeResetPassword.php
```diff
@@ -320,4 +316,86 @@ private function successMessage(InputInterface $input, ConsoleStyle $io, string
...
+
+        $manipulator->addTrait(ResetPasswordRequestRepositoryTrait::class);
+
+        $manipulator->addConstructor([
+            (new Param('registry'))->setType('ManagerRegistry')->getNode(),
+        ], <<<'CODE'
+<?php
+parent::__construct($registry, ResetPasswordRequest::class);
+CODE
+        );
```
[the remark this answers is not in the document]
weaverryan: As we talked about, if we want to avoid the comments in the repository class, let's pass an argument to `EntityClassGenerator::generateEntityClass()` to make it not generate the comments. That method name is getting a little long, but that's ok - it's an internal class.

src/Util/ClassSourceManipulator.php
```diff
@@ -293,6 +363,12 @@ public function addAnnotationToClass(string $annotationClass, array $options)
         $this->updateSourceCodeFromNewStmts();
     }
 
+    public function clearClassNodeStmts()
+    {
+        $this->getClassNode()->stmts = [];
+        $this->updateSourceCodeFromNewStmts();
+    }
```
weaverryan, line 370: Hopefully this won't be needed at the end :)

src/Doctrine/BaseRelation.php
archeoprog: For now, without modifying other parts of the MakerBundle, i think we should do things like this in `BaseRelation`
```php

    private $customReturnType;
    private $isCustomReturnTypeNullable = true;

    public function isCustomReturnTypeNullable(): bool
    {
        return $this->isCustomReturnTypeNullable;
    }


    public function setCustomReturnType(string $customReturnType, bool $isNullable)
    {
        $this->customReturnType = $customReturnType;
        $this->isCustomReturnTypeNullable = $isNullable;

        return $this;
    }

Comment (archeoprog) [reply src/Doctrine/BaseRelation.php:118]: And this in `ClassSourceManipulator`

File: src/Doctrine/BaseSingleRelation.php (hunk 32 above)
Comment (archeoprog) [reply src/Doctrine/BaseSingleRelation.php:27]: Done.

File: src/Util/ClassSourceManipulator.php (hunk 35 above)
Comment (archeoprog) [reply src/Util/ClassSourceManipulator.php:370]: Yes :)

File: src/Maker/MakeResetPassword.php (hunk 33 above)
Comment (archeoprog) [reply src/Maker/MakeResetPassword.php]: Not php 7.3, i wrote that because only the test for php 7.3 failed in travis.
https://github.com/symfony/maker-bundle/blob/bea8c3c959e48a2c952cc7c4f4f32964be8b8874/tests/Maker/MakeResetPasswordTest.php#L173

Comment (archeoprog) [reply src/Maker/MakeResetPassword.php]: jrushlow added new code. Our targetClass is generaated like a fetch class `UserCustom::class`  not in FQCN `"App\Entity\UserCustom"`

File: src/Doctrine/EntityClassGenerator.php [hunk 36]
```diff
@@ -56,13 +56,14 @@ public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $
         $this->generateRepositoryClass(
             $repoClassDetails->getFullName(),
             $entityClassDetails->getFullName(),
-            $withPasswordUpgrade)
+            $withPasswordUpgrade,
+            $avoidDefaultComments)
```
weaverryan, line 60: The `)` should be on the next line, right? It looks like that was wrong before this PR :). Can you fix it?

src/Doctrine/EntityClassGenerator.php
```diff
@@ -31,7 +31,7 @@ public function __construct(Generator $generator, DoctrineHelper $doctrineHelper
         $this->doctrineHelper = $doctrineHelper;
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $avoidDefaultComments = false): string
```
weaverryan, line 34: Now that I see this, it might be better to have the last argument be `bool $generateRepositoryClass = true`. Then, we would only generate the repository class automatically if this is true. In `make:reset-password`, we would set this to `false`. Then we would call `generateRepositoryClass()` directly. It just feels more clear to do the 2 things independently... and maybe I should have always made this method *only* generate the entity class and not also the repository.

src/Doctrine/EntityClassGenerator.php
```diff
@@ -56,13 +56,14 @@ public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $
...
-            $withPasswordUpgrade)
+            $withPasswordUpgrade,
+            $avoidDefaultComments)
         ;
 
         return $entityPath;
     }
 
-    public function generateRepositoryClass(string $repositoryClass, string $entityClass, bool $withPasswordUpgrade)
+    public function generateRepositoryClass(string $repositoryClass, string $entityClass, bool $withPasswordUpgrade, bool $avoidDefaultComments = false)
```
weaverryan, line 66:
```suggestion
    public function generateRepositoryClass(string $repositoryClass, string $entityClass, bool $withPasswordUpgrade, bool $includeExampleComments = true)
```

I renamed the argument. But the REAL important thing here is that I'm reversing the argument to be "positive" instead of negative. Doing this isn't always possible, but when it *is* possible, it's good. Negatives are hard to understand:

```php
if ($includeExampleComments) {
    // add comments
}

// vs
if (!$avoidDefaultComments) {
    // add comments
}
```

src/Doctrine/EntityClassGenerator.php
```diff
@@ -75,6 +76,7 @@ public function generateRepositoryClass(string $repositoryClass, string $entityC
                 'entity_alias' => $entityAlias,
                 'with_password_upgrade' => $withPasswordUpgrade,
                 'doctrine_registry_class' => $this->managerRegistryClassName,
+                'avoid_default_comments' => $avoidDefaultComments,
```
weaverryan, line 79:
```suggestion
                'include_example_comments' => $includeExampleComments,
```

src/Resources/skeleton/doctrine/Repository.tpl.php
```diff
@@ -21,7 +21,9 @@ public function __construct(ManagerRegistry $registry)
     {
         parent::__construct($registry, <?= $entity_class_name; ?>::class);
     }
+<?php if (false === $avoid_default_comments) { // When adding a new method without existing default comments, the blank line is automatically added.?>
```
weaverryan, line 24: 2 minor things:

1) When you update this to be `include_example_comments,` just make the if statement ` if ($include_example_comments)` - no need to check `=== true`. I admit that this is something I've been going back and forth on recently, but I've gotten advice from a smart friend to keep these if statements simpler.

2) When you have if statements in a template like this, use this syntax:

```php
<?php if ($include_example_comments): ?>

<?php endif; ?>
```

src/Util/ClassSourceManipulator.php
```diff
@@ -438,16 +508,21 @@ private function addSingularRelation(BaseRelation $relation)
...
 
         $this->addGetter(
             $relation->getPropertyName(),
-            $typeHint,
+            $relation->getCustomReturnType() ?: $typeHint,
             // getter methods always have nullable return values
-            // because even though these are required in the db, they may not be set
-            true
+            // unless this has been customized explicitly
+            $relation->isCustomReturnTypeNullable()
```
weaverryan, line 519: I still think this should be:

```php
$relation->getCustomReturnType() ? $relation-> isCustomReturnTypeNullable() : true
```

It, of course, makes no technical difference. The difference is in how the code "reads". Your code says:

> Always read the custom return type nullable value... which happens to default to true

Mine says:

> If the user has customized the return type, use their preference for nullable vs not nullable. If they have
> *not* customized the return type, then we know that, in this situation, the return type should default to `true`.

src/Doctrine/EntityClassGenerator.php
archeoprog: Done

src/Doctrine/EntityClassGenerator.php
archeoprog: Done

src/Doctrine/EntityClassGenerator.php
archeoprog: Done.

src/Doctrine/EntityClassGenerator.php
archeoprog: Done.

src/Resources/skeleton/doctrine/Repository.tpl.php
archeoprog: Done.

src/Util/ClassSourceManipulator.php
archeoprog: Done.

archeoprog:
I'll reopen a new PR. Conflict due to a use of two differents account.

weaverryan:
Thank you @archeoprog!

PR #593, opened 2020-04-21
Added UUID and GUID default type to entity maker

thomas-miceli:
Hello,

When you want to create a field, for an entity, representing an uuid, if you name it `uuid` the maker will set the default type to `uuid`. Same for `guid`.
image
image


weaverryan:
Very sensible idea! Thanks Thomas!

PR #599, opened 2020-05-05
Fix class name

fabpot:
Fix in #596 looks wrong to me as we are looking for the Doctrine Migration bundle here.


weaverryan:
Thank you Fabien!

PR #601, opened 2020-05-07
Prevent open redirect issues when generating LoginFormAuthenticator

franmomu:
This adds `HttpUtils` to create the redirect `onAuthenticationSuccess` and prevent open redirect issues when creating a LoginFormAuthenticator.

Ref: 
- https://symfony.com/blog/cve-2017-16652-open-redirect-vulnerability-on-security-handlers
- https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationSuccessHandler.php#L56


weaverryan:
Hi @franmomu!

I understand the general concept of an open redirect, but I don't see the vector here. With something like `DefaultAuthenticationSuccessHandler`, the URI to redirect can come from several places that could, in theory, be not safe - like the Referer or a `target_path_parameter` POST value, if that's enabled for the authentication listener.

But in this case, `$this->getTargetPath()` reads *only* from the session. And there is only *one* place in core where this is set - `ExceptionListener` where it is set to the *current* URL on your site when trying to access a protected page. Unless the user manually calls `$this->saveTargetPath()` somewhere and sets it to a spoofable URL, I don't see how it's possible for this to happen.

Cheers!

franmomu:
Hi @weaverryan,

The thing is that some time ago I started with the generated code and then later I added the ability of passing a URL in the request and I didn't pay attention to this.

But you're right! As it is, that's the only place where the URL is stored in the Session.

Thanks!

PR #602, opened 2020-05-08
Wrong class in deprecation message in MakeEntity

7ochem:
The deprecation message hints the Generator class, while the fifth argument is the EntityClassGenerator class


weaverryan:
Thanks @7ochem!

PR #606, opened 2020-05-13
Update return typehint

Nyholm:
I want to do this change
```diff

    /**
-    * @return Collection|Team[]
+    * @return Collection&Team[]
     */
    public function getTeams(): Collection
    {
        return $this->teams;
    }
```

The return type will not be an array of Teams or a Collection. It will be both =)

Reference: https://psalm.dev/docs/annotating_code/type_syntax/intersection_types/


Nyholm:
oops, nope. I was wrong

Issue #608, opened 2020-05-15
YamlSourceManipulator cannot find original value for multline string


jon-ht:
I'm making my own maker command and I need to update a config file.

The original file looks like this

```yaml
nelmio_api_doc:
    documentation:
        info:
            title: API Title
            description: |
                Symfony 5 project

                Some comments with markdown like [/api/doc/users](/api/doc/users)
                or `/api/doc/{area}`
        securityDefinitions:
            Bearer:
                type: apiKey
                description: 'Value: Bearer {jwt}'
                name: Authorization
                in: header
        security:
            - Bearer: []
    areas: # to filter documented areas
        default:
            path_patterns:
                - ^/api(?!/doc$) # Accepts routes under /api except /api/doc
```
In my maker, I have `$manipulator = new YamlSourceManipulator(file_get_contents($path));` which returns me the content above.

When I edit manipulator data, I'm getting this error

```
In YamlSourceManipulator.php line 793:
                                                                     
  Cannot find the original value "Symfony 5 project                  
                                                                     
  Some comments with markdown like [/api/doc/users](/api/doc/users)  
  or `/api/doc/{area}`                                               
  "    
```

If I remove multiline string to have `description: Symfony 5 project`, another error is thrown

```
In YamlSourceManipulator.php line 549:
                                
  Cannot find the key "Bearer"  
                                
```

Do you know if there is a workaround for this ?

weaverryan:
Hi!

Yep, sounds like a bug - the `YamlSourceManipulator` is COMPLEX. Could you create a PR with, at least, a failing test case? You can see how to create a new test case here: https://github.com/symfony/maker-bundle/tree/master/tests/Util/yaml_fixtures

I'd start with your full YAML source, but then try to remove as much as you can to focus on the *exact* requirements for a bug which I'm sure is related to that multi-line string syntax :).

Cheers!

jon-ht:
Sure, I'll create a test case :)

Should I also create a test case for the second error, caused by `- Bearer: []` ?

weaverryan:
Ah yes! I read too quickly and didn't notice the two problems. Yes, a second test case (you could put it in the same PR) for that would be awesome :)

jon-ht:
I've created a second PR instead, because these bugs seem really different and could be treated separately

jon-ht:
Hi @weaverryan 

#610 and #612 should fix both related issues. 

Would you mind give me some feedback, please ? :)

PR #610, opened 2020-05-15
Add test case for multiline string

jon-ht:
Test case for #608

```
In YamlSourceManipulator.php line 793:
                                                                     
  Cannot find the original value "Symfony 5 project                  
                                                                     
  Some comments with markdown like [/api/doc/users](/api/doc/users)  
  or `/api/doc/{area}`                                               
  "
```


weaverryan:
This test case doesn't actually expose the error you're talking about. It *does* expose a formatting bug:

```
 'nelmio_api_doc:
     documentation:
         info:
-            description: |
-                Multiline string'
+            description: "Multiline string\n"'
```

But it does not show the error you're seeing.

jon-ht:
It's weird, I could reproduce the error from CI but not always. It seems to be related to file ending (CRLF)

Sometime having one line of description give the expected error, sometimes not, can't understand why or what is happening

tests/Util/YamlSourceManipulatorTest.php
```diff
@@ -58,7 +58,8 @@ private function getYamlDataTests()
         foreach ($finder as $file) {
             list($source, $changeCode, $expected) = explode('===', $file->getContents());
 
-            $data = Yaml::parse($source);
+            // Multiline string ends with an \n
+            $data = Yaml::parse(trim($source, "\n"));
```
weaverryan, line 62: We actually trim this below already. So it looks like we just need to trim it earlier. I'm thinking the whole inside of the `foreach` would be this:

```php
list($source, $changeCode, $expected) = explode('===', $file->getContents());

$source = rtrim($source, "\n");
$expected = ltrim($expected, "\n");

$data = Yaml::parse($source);
eval($changeCode);

yield $file->getFilename() => [
    'source' => $source,
    'newData' => $data,
    'expectedSource' => $expected,
];
```

tests/Util/yaml_fixtures/multiline_string.test
```diff
@@ -0,0 +1,16 @@
...
+===
+$data['nelmio_api_doc']['documentation']['info']['description'] = "Multiline string\nSecond line\nThird line";
+===
+nelmio_api_doc:
+    documentation:
+        info:
+            description: |
+                Multiline string
+                Second line
+                Third line
```
weaverryan, line 16: beautiful test cases

src/Util/YamlSourceManipulator.php
```diff
@@ -771,16 +792,26 @@ private function findEndPositionOfValue($value, $offset = null)
...
             } else {
-                $pattern = sprintf('\'?"?%s\'?"?', preg_quote($value, '#'));
-            }
+                // Multiline value ends with \n.
+                // If we remove this character, the next property will ne merged with this value
+                $quotedValue = preg_quote(rtrim($value, "\n"), '#');
+                $patternValue = $quotedValue;
+
+                // Iterates until we find a new line char or we reach end of file
+                if ($this->findPositionOfMultilineCharInLine($offset)) {
```
weaverryan, line 809:
```suggestion
                if ($this->findPositionOfMultilineCharInLine($offset) !== null) {
```

That will help clarify that we're looking for IF there is a multiline char here, we're not (in this case) actually caring about the exact position returned.

src/Util/YamlSourceManipulator.php
```diff
@@ -216,6 +216,14 @@ private function updateData(array $newData)
                 continue;
             }
 
+            if (
+                $this->isMultilineString($newVal)
+                && rtrim($currentVal, "\n") !== $currentVal
+                && rtrim($newVal, "\n") === $newVal
+            ) {
+                $newVal .= "\n";
```
weaverryan, line 224: Could you add a `$this->log()` line here that explains what you're doing? There's a little-known feature where, in `YamlSourceManipulatorTest.php`, you can uncomment a `setLogger()` line to see all these logs entries. And it helps document what each part does, since it's SO complex. For instance, in this case, I'm not entirely sure what edge-case this is fixing.

src/Util/YamlSourceManipulator.php
```diff
@@ -449,8 +457,15 @@ private function changeValueInYaml($value)
 
         $endValuePosition = $this->findEndPositionOfValue($originalVal);
 
-        $newYamlValue = $this->convertToYaml($value);
-        if (!\is_array($originalVal) && \is_array($value)) {
+        $isMultilineValue = null !== $this->findPositionOfMultilineCharInLine($this->currentPosition);
+        $originalValEndsWithNewLine = $isMultilineValue ? rtrim($originalVal, "\n") !== $originalVal : false;
```
weaverryan, line 461: I think the `rtrim($originalVal, "\n") !== $originalVal` is confusing me. What about this instead?

`str_ends_with($originalValue, "\n")`

That's a PHP 8 function, but if you require `"symfony/polyfill-php80": "^1.15"`, then we can use it.

src/Util/YamlSourceManipulator.php
```diff
@@ -449,8 +457,15 @@ private function changeValueInYaml($value)
...
+
+        // In case of multiline, $value is converted as plain string like "Foo\nBar"
+        // We need to keep it "as is"
+        $newYamlValue = $isMultilineValue ? rtrim($value, "\n") : $this->convertToYaml($value);
```
weaverryan, line 465: Why do we need to `rtrim()` here? I can see that if I remove it the tests fail, but I can't figure out why it's needed.

tests/Util/YamlSourceManipulatorTest.php
jon-ht: Good catch, I missed this !

src/Util/YamlSourceManipulator.php
jon-ht: Are you OK to require `symfony/polyfill-php80` then ? I'll made apprioriate changes if so (there are some other lines like https://github.com/symfony/maker-bundle/pull/610/files#diff-a07d297f64d0bf4a1e9c902b787ecf20R219-R223)
```
if (
    $this->isMultilineString($newVal)
    && rtrim($currentVal, "\n") !== $currentVal
    && rtrim($newVal, "\n") === $newVal
)
```

src/Util/YamlSourceManipulator.php
jon-ht: One thing I noticed is, in this test case [multiline_string_before_property.test](https://github.com/symfony/maker-bundle/pull/610/files#diff-c137fbbe9d94d4a5f4bc29116ea1273e), `$value` ends with an `\n`

image

This will cause `indentMultilineYamlArray` will add an extra line that's not wanted.

src/Util/YamlSourceManipulator.php
jon-ht: Well, it seems not required. I think I went crazy and wanted to add/remove `\n` everywhere 😄.
I will remove completely

src/Util/YamlSourceManipulator.php
weaverryan: Ah, I understand. I think the implementation is what's confusing me. All multiline arrays must end in a line break, by their very nature. If anything, in this spot, (just thinking about the logical flow of the code) I would expect you to be guaranteeing that the `$newYamlValue` *does* end in `\n`, instead of stripping it. After all, if we made sure that `$newYamlValue` ended in 1 line break... then that should work perfectly with YAML, which *also* wants the line break :). I'm wondering if we're trimming and adding `\n` in so many places that we're doing and undoing extra work.

There *is* the edge-case of not needing the "\n" if the multiline string is at the end of the file. That should be handled (somewhere) with an `if` statement specifically for it. Something like:

```php
// maybe down before the `$newContents = substr(...)` line
if (false === $this->findNextLineBreak($newPosition) && $isMultilineValue) {
    $newYamlValue = trim($newYamlValue, "\n");
}
```

weaverryan:
Thanks for taking this on - I knew that I didn't cover this case when I built `YamlSourceManipulator`, and I wasn't looking forward to trying to do it - cause it's hard ;).

src/Util/YamlSourceManipulator.php
jon-ht: I won't lie to you, I also think there is a lot of trimming to handle multiline string...

Actually, it seems that re-adding original \n (after `$newYamlValue = "\n".$this->indentMultilineYamlArray($newYamlValue);`) breaks test case because, as mentionned above, YamlDumper will add an extra line

image

Maybe this bug is related to Yaml component and not `YamlSourceManipulator`

jon-ht:
I was able to remove some useless trimming after a second pass, maybe this is less confusing now.

PR #612, opened 2020-05-15
Add test case for keyed array item

jon-ht:
Not sure about test case naming, tell me if you have a better one :)

Part of #608

```
In YamlSourceManipulator.php line 549:
                                
  Cannot find the key "Bearer"  
                   
```

### DIFF
--- a/src/Util/YamlSourceManipulator.php
+++ b/src/Util/YamlSourceManipulator.php
@@ -546,6 +546,16 @@ class YamlSourceManipulator
                 return $this->currentPosition;
             }
 
+            $cursor = $this->currentPosition;
+
+            while ('-' !== substr($this->contents, $cursor - 1, 1) && -1 !== $cursor) {
+                --$cursor;
+            }
+
+            if ($cursor >= 0) {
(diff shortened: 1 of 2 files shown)


src/Util/YamlSourceManipulator.php
```diff
@@ -623,7 +643,7 @@ private function getEndOfPreviousKeyPosition($key): int
 
     private function getKeyRegex($key)
     {
-        return sprintf('#%s( )*:#', preg_quote($key));
+        return sprintf('#(- )?%s( )*:#', preg_quote($key));
```
weaverryan, line 646: We either have too many changes in this PR, or we're missing some more tests :).

If I revert this change AND revert the changes from lines 530-538, the new test PASSES. It seems that only the code from 559-568 is actually needed. If the other code fixes some other bug, we need a test case for that.

jon-ht: You're right. I think this line was a "first resolution try" and I forgot it :)

weaverryan:
Thanks @jon-ht!

Issue #522, opened 2019-12-20
Let's use the String component everywhere


javiereguiluz:
This bundle makes lots of string manipulations. Originally we created some utils (e.g. https://github.com/symfony/maker-bundle/blob/master/src/Util/ClassNameDetails.php) and a `Str` utility class (https://github.com/symfony/maker-bundle/blob/master/src/Str.php).

The only problem for us is that `String` requires using PHP 7.2 (and this project requires *the legacy and unmaintained* PHP 7.0). Can we update the PHP requirement in the next stable version?

For more info:

* [New in Symfony 5.0: String Component](https://symfony.com/blog/new-in-symfony-5-0-string-component) blog post
* [String component docs](https://symfony.com/doc/current/components/string.html)
* [Migration to String component in Symfony Demo app](https://github.com/symfony/demo/pull/1055/files)

stof:
ClassNameDetails will still need to exist. And the one call to `str_replace` in it won't become more readable when using the String component.

And in the `Str` util, the String component could replace the implementation of `asCamelCase` and `asSnakeCase` (well, here, the weird thing is that the main implementation is in `asTwigVariable` for the conversion to snake case, while the rule is that Twig variables are using snake-cased, not that snake-case is following the Twig rules, so we might want to swap the implementation).
Most other helpers will still be needed:

- some of there are simple aliases of the `asSnakeCase` method, but they have a domain meaning (they encode in a single place that route names are using snake case for instance). Removing them would not simplify the bundle but would duplicate the domain rule of what a route name uses.
- some of them are too specific to be moved to the String component itself (`getShortClassName` or `singularCamelCaseToPluralCamelCase` for instance)

javiereguiluz:
You are right @stof but I just want to replace the implementation details, not the methods used in the bundle. As you said, they have "domain meaning" so we must keep them. Let's just replace all `str*` or `mb_*` calls (if possible and if it makes sense). Thanks!

LeJeanbono:
Min version is now 7.1, I think we can begin to work on it :)

javiereguiluz:
Closing this old issue because it's already being tackled in PRs like #1100. Thanks.

PR #613, opened 2020-05-16
Add Symfony String

LeJeanbono:
#522

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "minimum-stability": "dev",
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.2.5",
         "doctrine/inflector": "^1.2",
         "nikic/php-parser": "^4.0",
         "symfony/config": "^3.4|^4.0|^5.0",
@@ -22,7 +22,8 @@
         "symfony/filesystem": "^3.4|^4.0|^5.0",
         "symfony/finder": "^3.4|^4.0|^5.0",
         "symfony/framework-bundle": "^3.4|^4.0|^5.0",
-        "symfony/http-kernel": "^3.4|^4.0|^5.0"
+        "symfony/http-kernel": "^3.4|^4.0|^5.0",
+        "symfony/string": "^5.0"
     },
     "require-dev": {
         "doctrine/doctrine-bundle": "^1.8|^2.0",
--- a/src/Str.php
+++ b/src/Str.php
@@ -64,15 +64,7 @@ final class Str
      */
     public static function asClassName(string $value, string $suffix = ''): string
     {
-        $value = u($value)
-            ->trim()
-            ->replace('-', ' ')
-            ->replace('_', ' ')
-            ->replace('.', ' ')
-            ->replace(':', ' ')
-            ->title(true)
-            ->replace(' ', '')
-            ->title();
+        $value = u($value)->camel()->title();
         $value = self::addSuffix($value, $suffix);
 
         return $value;
@@ -84,12 +76,7 @@ final class Str
      */
     public static function asTwigVariable(string $value): string
     {
-        return u($value)
-            ->trim()
-            ->replaceMatches('/[^a-zA-Z0-9_]/', '_')
-            ->replaceMatches('/(?<=\\w)([A-Z])/', '_$1')
-            ->replaceMatches('/_{2,}/', '_')
-            ->lower();
+        return u($value)->snake();
     }
 
     public static function asLowerCamelCase(string $str): string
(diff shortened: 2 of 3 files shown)


src/Str.php
```diff
@@ -40,7 +44,7 @@ public static function hasSuffix(string $value, string $suffix): bool
      */
     public static function addSuffix(string $value, string $suffix): string
     {
-        return self::removeSuffix($value, $suffix).$suffix;
+        return self::removeSuffix($value, $suffix) . $suffix;
```
javiereguiluz, line 47: This can be updated as `return u($value)->ensureEnd($suffix);`

src/Str.php
```diff
@@ -60,11 +64,15 @@ public static function removeSuffix(string $value, string $suffix): string
      */
     public static function asClassName(string $value, string $suffix = ''): string
     {
-        $value = trim($value);
-        $value = str_replace(['-', '_', '.', ':'], ' ', $value);
-        $value = ucwords($value);
-        $value = str_replace(' ', '', $value);
-        $value = ucfirst($value);
+        $value = u($value)
```
javiereguiluz, line 67: I'm not 100% sure, but I think this can be refactored as:

```php
return u($value)->camel()->title()->ensureEnd($suffix);
```

At least in the Symfony Docs show this:

```php
u('Foo: Bar-baz.')->camel()->title(); // 'FooBarBaz'
```

src/Str.php
```diff
@@ -76,28 +84,27 @@ public static function asClassName(string $value, string $suffix = ''): string
...
     public static function asTwigVariable(string $value): string
     {
-        $value = trim($value);
-        $value = preg_replace('/[^a-zA-Z0-9_]/', '_', $value);
-        $value = preg_replace('/(?<=\\w)([A-Z])/', '_$1', $value);
-        $value = preg_replace('/_{2,}/', '_', $value);
-        $value = strtolower($value);
-
-        return $value;
+        return u($value)
```
javiereguiluz, line 87: This can be simplified as `return u($value)->snake();`

LeJeanbono: Yep

src/Str.php
LeJeanbono: test failed => `addSuffix('Generatecommand', 'Command')` produce `GeneratecommandCommand` instead of `GenerateCommand`

src/Str.php
LeJeanbono: seems ok

composer.json
```diff
@@ -22,7 +22,8 @@
         "symfony/filesystem": "^3.4|^4.0|^5.0",
         "symfony/finder": "^3.4|^4.0|^5.0",
         "symfony/framework-bundle": "^3.4|^4.0|^5.0",
-        "symfony/http-kernel": "^3.4|^4.0|^5.0"
+        "symfony/http-kernel": "^3.4|^4.0|^5.0",
+        "symfony/string": "^5.0"
```
weaverryan, line 26: There's a practical problem that I don't know if we can overcome. If a project uses Symfony 4.4, and has `extra.symfony.require` set to `4.4.*`, I don't think they will be able to install Maker. They would get an error that no `symfony/string` version was found, because Flex filters it to only match 4.4 (but we should actually try this to be sure).

LeJeanbono: Btw, can we drop 3.4 support for a min version with 4.4 ?

OskarStark: Just for the record, looks like I can install `symfony/string:^5.1` with `symfony/maker-bundle:^1.21` + `symfony.extra.require:4.4.*`. But `symfony/string` is a direct dependency 🤔  Will it work with an indirect dependency by `maker-bundle`, too?

CleanShot 2020-10-23 at 20 34 11

nicolas-grekas: This specific combinations should work because String is not in 4.4, this is not bounded by the extra require constraint.

composer.json
```diff
@@ -13,7 +13,7 @@
     ],
     "minimum-stability": "dev",
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.2.5",
```
jrushlow, line 16: Playing devils advocate - How would this affect support for Symfony 3.4 && why not bump to 7.3.x as 7.2 ends security support at the end of November?

OskarStark: Its not needed, 7.2.5 is the minimum version for symfony string 👍

composer.json
OskarStark: > Btw, can we drop 3.4 support for a min version with 4.4 ?

@LeJeanbono IMO this is not needed, as you can install `3.4` when using `7.2.5` php afterwards 👍

PR #617, opened 2020-05-20
Update namespace for Authenticator

seb-jean:


weaverryan:
Hi @seb-jean!

Thanks for the PR. The location where this class should be generated is totally subjective. As I usually don't have a lot of "security" things, and because authenticators are my most important security classes even when I have multiple, I prefer to put them directly into `Security` and avoid extra directories. So, I'm going to close this - which will be great for some people, but probably not for you - can't please everyone, unfortunately.

Cheers!

PR #619, opened 2020-05-20
Make entity --api-resources allow to add ApiPlatform filters

archeoprog:
Branch? | master
-- | --
Bug fix? | no
New feature? | - yes
Deprecations? | - no
Tickets | - no
License | MIT
Doc PR | - in progress

Notice to Core Developers. Supervised development on Slack with Ryan Weaver and Kevin Dunglas.

Development still in progress.

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -41,6 +41,7 @@ use Symfony\Component\Console\Question\Question;
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Kévin Dunglas <dunglas@gmail.com>
+ * @author Antoine Michelet <jean.marcel.michelet@gmail.com>
  */
 final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
 {
@@ -159,6 +160,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
 
         if ($classExists) {
             $entityPath = $this->getPathOfClass($entityClassDetails->getFullName());
+
             $io->text([
                 'Your entity already exists! So let\'s add some new fields!',
             ]);
@@ -175,7 +177,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
 
         $isFirstField = true;
         while (true) {
-            $newField = $this->askForNextField($io, $currentFields, $entityClassDetails->getFullName(), $isFirstField);
+            $newField = $this->askForNextField($io, $currentFields, $entityClassDetails->getFullName(), $isFirstField, $input->getOption('api-resource'));
             $isFirstField = false;
 
             if (null === $newField) {
@@ -187,8 +189,18 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
 
             if (\is_array($newField)) {
                 $annotationOptions = $newField;
-                unset($annotationOptions['fieldName']);
-                $manipulator->addEntityField($newField['fieldName'], $annotationOptions);
+                unset($annotationOptions['fieldName'], $annotationOptions['apiFilter']);
+                if (null !== $newField['apiFilter']) {
+                    $manipulator->addUseStatementIfNecessary('ApiPlatform\\Core\\Annotation\\ApiFilter');
+                    if ('TermFilter' === $newField['apiFilter'] || 'MatchFilter' === $newField['apiFilter']) {
+                        $manipulator->addUseStatementIfNecessary('ApiPlatform\Core\Bridge\Elasticsearch\DataProvider\Filter\\'.$newField['apiFilter']);
+                    } else {
+                        $manipulator->addUseStatementIfNecessary('ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\\'.$newField['apiFilter']);
+                    }
+                    $manipulator->addEntityField($newField['fieldName'], $annotationOptions, ['@ApiFilter('.$newField['apiFilter'].'::class)']);
+                } else {
+                    $manipulator->addEntityField($newField['fieldName'], $annotationOptions);
+                }
 
                 $currentFields[] = $newField['fieldName'];
             } elseif ($newField instanceof EntityRelation) {
@@ -280,7 +292,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         ORMDependencyBuilder::buildDependencies($dependencies);
     }
 
-    private function askForNextField(ConsoleStyle $io, array $fields, string $entityClass, bool $isFirstField)
+    private function askForNextField(ConsoleStyle $io, array $fields, string $entityClass, bool $isFirstField, bool $apiOptions)
     {
         $io->writeln('');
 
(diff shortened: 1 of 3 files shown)


tests/Maker/MakeEntityTest.php
```diff
@@ -566,5 +566,181 @@ public function getTestDetails()
             ->configureDatabase()
             ->updateSchemaAfterCommand(),
         ];
+
```
archeoprog, line 569: I need to add more tests. Maybe a functionnal test in a new fixtures (maybe MakeApiResource) wich request the api with correspondig filters. Also we probably need to test the cases where the filters are not supposed to match the type assigned to the current field. I have another problem, the TermFilter and MatchFilter filters need Elasticsearch

src/Maker/MakeEntity.php
```diff
@@ -376,6 +382,66 @@ private function askForNextField(ConsoleStyle $io, array $fields, string $entity
             $data['nullable'] = true;
         }
 
+        $data['apiFilter'] = null;
+
+        if (true === $apiOptions) {
+            $availableFilters = [
```
dunglas, line 388: Maybe could you store this list in a `public static` property, so users can add their own filters if needed. Also, I would use FQCN here (for the same reason).

src/Maker/MakeEntity.php
```diff
@@ -376,6 +382,66 @@ private function askForNextField(ConsoleStyle $io, array $fields, string $entity
...
+                'TermFilter',
+            ];
+
+            $apiFilter = null;
+            while (null === $apiFilter) {
+                $question = new Question('Do you want to add a filter for your api resource? (enter <comment>?</comment> to see all filters)', 'No');
+                $question->setAutocompleterValues($availableFilters);
+                $apiFilter = $io->askQuestion($question);
+
+                if ('No' === $apiFilter) {
```
dunglas, line 406: Maybe could we just hint the user to type `enter` to skip (as when you're done adding fields).

src/Maker/MakeEntity.php
```diff
@@ -376,6 +382,66 @@ private function askForNextField(ConsoleStyle $io, array $fields, string $entity
...
+                'ExistsFilter',
+                'OrderFilter',
+                'MatchFilter',
+                'TermFilter',
+            ];
+
+            $apiFilter = null;
+            while (null === $apiFilter) {
+                $question = new Question('Do you want to add a filter for your api resource? (enter <comment>?</comment> to see all filters)', 'No');
+                $question->setAutocompleterValues($availableFilters);
```
dunglas, line 403: This list should be filtered to include only filters compatible with the type of the current property.

src/Maker/MakeEntity.php
```diff
@@ -814,4 +880,72 @@ private function getEntityNamespace(): string
...
+            'date_immutable',
+            'time',
+            'time_immutable',
+            'dateinterval',
+        ];
+
+        $type = $data['type'];
+        switch ($apiFilter) {
+            case 'SearchFilter':
+                return \in_array($type, $numerics) || 'string' === $type || 'text' ? true : false;
```
dunglas, line 921:
```suggestion
                return \in_array($type, $numerics) || 'string' === $type || 'text';
```

Same everywhere.

src/Maker/MakeEntity.php
```diff
@@ -814,4 +880,72 @@ private function getEntityNamespace(): string
...
+        } else {
+            $manipulator->addUseStatementIfNecessary('ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\\'.$filter);
+        }
+
+        $manipulator->addEntityField($filter, $annotationOptions, ['@ApiFilter('.$filter.'::class)']);
+    }
+
+    private function isTypeCompatibleWithApiFilter($data, $apiFilter): bool
+    {
+        $numerics = [
```
dunglas, line 899: could be stored in a class constant, same for `$dates`.

src/Maker/MakeEntity.php
archeoprog: Done. But i also added features for managing the short/full className because autocompletion, we probably need to display shortClassName but we want works with FQCN.

src/Maker/MakeEntity.php
archeoprog: Done. Then see: https://github.com/symfony/maker-bundle/pull/619#discussion_r428528750

src/Maker/MakeEntity.php
archeoprog: Done.

src/Maker/MakeEntity.php
archeoprog: Damn it, stupid mistake. Thanks!

src/Maker/MakeEntity.php
archeoprog: Done.

archeoprog:
Actually, i need to add several options. Allow to add filters to mapped properties
```php
/**
     * @ORM\OneToMany(targetEntity="DummyCarColor", mappedBy="car")
     * @ApiFilter(SearchFilter::class, properties={"colors.prop": "ipartial"})
     */
    public $colors;

weaverryan:
Hi!

This is a big undertaking :). And the tricky part is getting the user experience correct, especially because it can be awkward to use a console command to configure so many things on multiple levels (configuring ApiResource options on the class, ORM\Entity options on the properties, and API-related options on the properties). Getting that correct will be an iterative process.

Here are my initial thoughts, which will be imperfect, but hopefully closer:

1) I think creating an entity and adding entity properties should stay in `make:entity`. In other words, if you want to add a new Doctrine entity class or add new properties to your Doctrine entity, use `make:entity`

2) I think that if you pass a *new* class name to `make:api:resource` (i.e. you do not want to modify an existing class), then it should create a class that is NOT an entity (and create it with probably zero properties). So, this would create an empty class, but also a data provider and data persisted that have some basic starting code to support this class. That's a very powerful thing on its own.

3) As I mentioned in (2), we won't ever add fields in `make:api:resource`. Once you enter the class name, probably we would do the same thing as you are doing currently:

> First, configure your api resource
> 
>  [0] collection/item operations
>  [1] normalization/denormalization groups
>  [2] pagination
> ...
> [6] add a new field
> [7] I'm done!

If you selected `[6] add a new field`, then:

A) If this is an entity, we give them an error and say "run `make:entity BlogPost` to add more fields"
B) If this is not an entity, then we ask them what property name they want and we generate the getter and the setter. What may even automatically add the property to the groups specified in normalization/denormalization groups, but I'm not sure yet. Maybe that is one other question we ask.

I actually *don't* want to ask about a "filter" when adding the new field. I think that can be done in a different way - by adding an new option to the list above - e.g.

>  [0] collection/item operations
>  [1] normalization/denormalization groups
> ...
> [4] Add a filter
> ...
> [6] add a new field
> [7] I'm done!

If you select "Add a filter", then we ask you what type you want & what fields. If they select a single field, the new can generate it above that property. If they select multiple, we put it
above the class.

So... that's my current thinking :)

archeoprog:
I am a bit confused. Probably because i never used data providers. So we can add all the configuration in this kind of class? https://api-platform.com/docs/core/data-providers/

archeoprog:
@weaverryan, @dunglas 
Sorry, I've been really busy and a little off my game lately. I realize I've been explaining my point of view and the work I've been doing quite badly.

I'm making my point at this point.

1) Entities or not?

a) I understand the need to divide the maker-bundle features as well as possible, we would like to have one command that takes care of entities and one command that takes care of API resources. If the make:api:resource command repeat in part the behavior of make-entity, it can also be considered as an overhead, it's not a bad thing. 
Obviously, I thought it was a shame to repeat code between make-entity and make:api:resource. That's why I created two Helpers
[MakeEntityHelper](https://github.com/symfony/maker-bundle/pull/619/files#diff-41b6e0ddb33cc9195142f67e41bdc122) and [MakeApiResourceHelper](https://github.com/symfony/maker-bundle/pull/619/files#diff-58f6db08fece0bfc35c730aa45688527). This gives us the opportunity to test these classes [unitarily](https://github.com/symfony/maker-bundle/pull/619/files#diff-9a180b3f23cfe38a80f1237704f5870c) and improve, I think, the test coverage and scalability of these features.

b) It's much simpler and more obvious, imho, to generate a configured entity and allow to add the fields later. Especially since I think this kind of features will be very "noob-friendly", expert companies surely prefer to configure everything by hand in detail.
After this difficult mission, couldn't we create two "sister" Commands such as make:api:provider and make:api:persister?
This could generate base code for developers wishing to use these features if they wish. I'm quite ready to work on it with pleasure.

2) The filters

> If you select "Add a filter", then we ask you what type you want & what fields. If they select a single field, the new can generate it above that property. If they select multiple, we put it
above the class.

a) We originally decided to add filters to the fields because they allow us to check the type of the field present. We can as well "trust" it and configure according to its choices without any verification. Putting everything above the class is cleaner and simpler.

b) I think we should choose between one of the two options, and can't have to choose based on the number of fields chosen, which implies parsing the entity in order to add configuration, and which can confuse the user more than help him. 

c) Currently, the developer can add as many filters (with strategy) as he wants to the new fields he generates, as long as these filters are compatible with the type of fields, and the strategies are compatible with the chosen filter. This accompaniment with autocompletion and errors in the console allows the developer to educate the developer on the functioning of the filters.

archeoprog:
@weaverryan @dunglas 
If I don't have an answer within a week, I'll stop the development.

archeoprog:
And because I'm tired of having the morality of leftists imposed on me: Development largely carried out by a supporter of Donald Trump, opposed to the BlackLivesMatters movement, strongly opposed to the white genocide in South Africa.

Issue #618, opened 2020-05-20
YamlSourceManipulator: insert new element into indexed array


7ochem:
I have a original Yaml file as seen below. I want to add a new element `App\Entity\Page\NewsPage` but keep an alphabetical order in the elements.

```yaml
page_types:
    - App\Entity\Page\ContentPage
    - App\Entity\Page\ErrorPage
    - App\Entity\Page\HomePage
    - App\Entity\Page\OverviewPage
```
I am able to load and manipulate the Yaml file through the `YamlSourceManipulator` through the code below. If I'm adding the new element at the end of the array, then this goes just fine.

```php
<?php
$manipulator = new YamlSourceManipulator(file_get_contents('page_types.yaml'));
$data = $manipulator->getData();
$data['page_types'] = 'App\\Entity\\Page\\NewsPage';
$manipulator->setData($data);
```

Now I have tried two methods to get this in alphabetical order.
The first is by determining the position where to insert and use array slice:
```php
// $pos = position determination magic here 🦄...
$data['page_types'] = array_merge(array_slice($data['page_types'], 0, $pos), ['App\\Entity\\Page\\NewsPage'], array_slice($data['page_types'], $pos));
```
Then, at `$manipulator->setData($data);`, I get:

> ```
> In YamlSourceManipulator.php line 641:
> 
>   [Symfony\Bundle\MakerBundle\Util\YamlManipulationFailedException]
>   Content was updated, but updated content does not match expected data. Original source: "page_types:
>       - App\Entity\Page\ContentPage
>       - App\Entity\Page\ErrorPage
>       - App\Entity\Page\HomePage
>       - App\Entity\Page\OverviewPage
>   ", updated source: "page_types:
>       - App\Entity\Page\ContentPage
>       - App\Entity\Page\ErrorPage
>       - App\Entity\Page\HomePage App\Entity\Page\NewsPage
>   ", updated data: array (
>     'page_types' =>
>     array (
>       0 => 'App\\Entity\\Page\\ContentPage',
>       1 => 'App\\Entity\\Page\\ErrorPage',
>       2 => 'App\\Entity\\Page\\HomePage',
>       3 => 'App\\Entity\\Page\\NewsPage',
>     ),
>   )
> ```

So apparently there's a mismatch and looking at the "updated data" it shows `NewsPage` in the new position, but `OverviewPage` got cut off.

[...]

PR #620, opened 2020-05-22
Add Failing Test

LeJeanbono:
Test case for #618


weaverryan:
Yo @LeJeanbono!

Thanks for getting this started for me! It turned out, the bug was much simpler - I used your test case to find a smaller test case, which led me to the true issue. After fixing that issue, I was pleasantly surprised that the original test case now passed :).

Cheers!

LeJeanbono:
Thanks !
I tried to look at `YamlSourceManipulator` but it's really hard...

weaverryan:
> I tried to look at YamlSourceManipulator but it's really hard...

Yea, it's a real mess :)

PR #626, opened 2020-06-01
Replace Command arguments with MakerParams

jrushlow:
Currently every maker abuses the intended usage of `Command` arguments - which were meant to be values that are passed to a command. E.g.

```
$> some-command --arg=value
```
`Command` arguments can either be required or optional. But, due to the way all current Makers use `Command` arguments as questions - they are all optional in regards to their intended usage.

This PR introduces a new `MakerParam` object that is stored in `AbstractMaker::makerParams` with a `getMakerParamValue()` and `setMakerParam()` *open to suggestions on naming conventions 

As a side affect of declaring arguments first, then setting them. Makers rely heavily on what is essentially a `null|mixed` return type for `$input->getArgument()`. An example would be passing arguments to a template - then the template has a conditional for that argument. In the example below, a value for `$my-var` may not exist depending on the flow of `interact()`

```
// MakeX

public function generate(..Generator $generator): void
{
    $generator->generateController('Some\\Name\\', 'template.tpl.php', [
        'my-var' => $input->getArgument('x')
    ]);
}

// template.tpl.php
<? if ($some-other-arg): ?> 
    ...
<? else: ?>
    <?= $my-var ?>
<? endif; ?>
```

While we would ideally like to get away from having to declare an argument, set a value, then get that value; this PR does not fully address that situation. One could use a conditional in the generate method or the newly introduced `$this->getMakerParamValue()` could return `null` if that argument doesn't exist. The latter does not feel right IMO. As such, we are declaring all of the arguments first in `configure()` so we don't run into a `NotFoundException` or the like in `generate()`. Open to ideas on this part...

The `MakeRegistrationForm` maker was used to demonstrate the new feature. Other makers would be able to be converted in the future.


src/Maker/MakeRegistrationForm.php
```diff
@@ -66,32 +69,36 @@ public static function getCommandName(): string
...
+            'from-email-address' => false,
+            'from-email-name' => false,
+            'auto-login-authenticator' => false,
+            'firewall-name' => false,
+            'redirect-route-name' => false,
+            'add-unique-entity-constraint' => false,
+        ];
+
+        foreach ($arguments as $name => $required) {
+            $argument = new MakerArgument($name, null, $required);
```
jrushlow, line 95: We could also call `$this->arguments->createArgument($name, null, $required);` instead of creating a new `MakerArgument` then using `$this->arguments->addArgument(MakerArgument)` Either way provides the same result...

jrushlow:
Shelving this PR for now - experimenting with maker properties in upcoming docker feature instead...

PR #629, opened 2020-06-05
Working around a change in symfony/console where we must set a max attempt

weaverryan:
This was caused by https://github.com/symfony/symfony/pull/36696

The fix probably shouldn't be here. I'll create a PR later to propose a default, high, max attempts. More technical info about this will follow on that PR.


nicolas-grekas:
Should be fixed by https://github.com/symfony/symfony/pull/37122 (see my last suggestion there)
Maybe not worth it here.

weaverryan:
Closing in favor of #590, which just removes the test, which is not that important

PR #632, opened 2020-06-14
[ResetPassword & VerifyEmail] fix templated email "lifetime" being mutated by timezones

jrushlow:
Currently in the twig email templates, the `This link will expire in 1 hour(s).` is malformed if a timezone is set either in twig or in PHP.

In the controllers of both ResetPassword & VerifyEmail(RegistrationController) the token lifetime is expressed as `int X`. The lifetime is then passed to the twig templated email where it is ultimately passed to the `date` filter.  This PR enforces that the lifetime value should always be formatted in the UTC timezone.

This is only a temporary solution. Ideally, either the `ResetPasswordBundle` & `VerifyEmailBundle` or `MakerBundle` should use a math based method to transform the lifetime value to a human readable format. e.g
```
$lifetime = 3600;   // seconds

public function convertToHours(int $lifetime): string

$result = ::convertToHours($lifetime);
// 1 hour
```

Refs https://github.com/SymfonyCasts/reset-password-bundle/issues/119

### DIFF
--- a/src/Resources/skeleton/registration/twig_email.tpl.php
+++ b/src/Resources/skeleton/registration/twig_email.tpl.php
@@ -3,7 +3,7 @@
 <p>
     Please confirm your email address by clicking the following link: <br><br>
     <a href="{{ signedUrl|raw }}">Confirm my Email</a>.
-    This link will expire in {{ expiresAt|date('g') }} hour(s).
+    This link will expire in {{ expiresAt|date('g', 'UTC') }} hour(s).
 </p>
 
 <p>
--- a/src/Resources/skeleton/resetPassword/twig_email.tpl.php
+++ b/src/Resources/skeleton/resetPassword/twig_email.tpl.php
@@ -3,7 +3,7 @@
 <p>
     To reset your password, please visit
     <a href="{{ url('app_reset_password', {token: resetToken.token}) }}">here</a>
-    This link will expire in {{ tokenLifetime|date('g') }} hour(s)..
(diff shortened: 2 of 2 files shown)


src/Resources/skeleton/registration/twig_email.tpl.php
```diff
@@ -3,7 +3,7 @@
 <p>
     Please confirm your email address by clicking the following link: <br><br>
     <a href="{{ signedUrl|raw }}">Confirm my Email</a>.
-    This link will expire in {{ expiresAt|date('g') }} hour(s).
+    This link will expire in {{ expiresAt|date('g', 'UTC') }} hour(s).
```
javiereguiluz, line 6: I'm probably missing something, but the `g` format outputs "12-hour format of an hour without leading zeros". So, if it's 8:35 AM it outputs `8` and if it's  17:21 PM it outputs `5`.

that's why I think the text requires some rewording like this:

```
will expire in ... hour(s) --> will expire at ...
```

jrushlow:
Closing this PR in favor of a better solution. Both ResetPassword and VerifyEmail will soon have built in translation support for the token / signature expiration time. Once this is available, we'll update the templates in a new PR here in maker.. This should also facilitate an easier path forward to adding those bundles to the demo app.. 

See:
https://github.com/SymfonyCasts/reset-password-bundle/pull/135
https://github.com/SymfonyCasts/verify-email-bundle/pull/43

PR #635, opened 2020-06-19
Added type-hints to onAuthenticationSuccess

jenyak:
Added type-hints to `onAuthenticationSuccess` to support [changes](https://github.com/symfony/security-guard/commit/3105836b14bd984e9839a697ca9e65c2e6c7db8e) introduced in `symfony/security-guard 5.0`

### DIFF
--- a/src/Resources/skeleton/authenticator/EmptyAuthenticator.tpl.php
+++ b/src/Resources/skeleton/authenticator/EmptyAuthenticator.tpl.php
@@ -36,7 +36,7 @@ class <?= $class_name ?> extends AbstractGuardAuthenticator
         // todo
     }
 
-    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, <?= $provider_key_type_hint ?>$providerKey)
     {
         // todo
     }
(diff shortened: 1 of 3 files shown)


weaverryan:
Hey @jenyak!

Hmm. So yes... but because this bundle is still used in Symfony 4, this needs to be conditional: we can only add the `string` type-hint if the user is using Symfony 5 (or, more accurately, if `AuthenticatorInterface::onAuthenticationSuccess` has a `string` type for its 3rd arg - we could use Reflection to be sure). We would need to detect that and pass this into the template so we could do this conditionally.

Could you make that change?

Thanks!

jenyak:
Hi Ryan. Please look at the changes. I did as you suggested.

src/Resources/skeleton/authenticator/EmptyAuthenticator.tpl.php
```diff
@@ -36,7 +36,7 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
         // todo
     }
 
-    public function onAuthenticationSuccess(Request $request, TokenInterface $token, $providerKey)
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, <?= $provider_key_type_hint ?> $providerKey)
```
javiereguiluz, line 39: It's minor but ... will this generate an extra white space when `provider_key_type_hint` is empty? If that's the case, we may do this:

```diff
-TokenInterface $token, <?= $provider_key_type_hint ?> $providerKey)
+TokenInterface $token, <?= $provider_key_type_hint.' ' ?>$providerKey)
```

jenyak: Your right, it will generate an extra white space. Your suggestion also generate an extra white space.

jenyak: @javiereguiluz I made a changes to address this

weaverryan:
Thank you @jenyak!

PR #637, opened 2020-06-24
Add text email for password reset request

pierredup:
When no text template is specified, Symfony mailer converts the HTML template to text by stripping out the tags. This causes the text template to be malformed and includes references to links that don't exist. This PR adds a text template when requesting a password reset.

**Before:**

Screenshot 2020-06-24 at 22 32 17

**After:**

Screenshot 2020-06-24 at 22 35 31

### DIFF
--- a/src/Resources/skeleton/resetPassword/twig_email.tpl.php
+++ b/src/Resources/skeleton/resetPassword/twig_email.tpl.php
@@ -1,11 +1,9 @@
 <h1>Hi!</h1>
 
-<p>
-    To reset your password, please visit
-    <a href="{{ url('app_reset_password', {token: resetToken.token}) }}">here</a>
-    This link will expire in {{ tokenLifetime|date('g') }} hour(s)..
-</p>
+<p>To reset your password, please visit the following link</p>
 
-<p>
-    Cheers!
-</p>
+<a href="{{ url('app_reset_password', {token: resetToken.token}) }}">{{ url('app_reset_password', {token: resetToken.token}) }}</a>
+
+<p>This link will expire in {{ tokenLifetime|date('g') }} hour(s).</p>
+
+<p>Cheers!</p>


weaverryan:
Excellent catch!

To avoid needing 2 templates, could we improve the HTML one to make it friendlier? It seems like the biggest problem is the fact that the link just says "here"... which is probably not really a great thing for even an HTML email. What about updating the HTML email to say:

```
<h1>Hi!</h1>

<p>
    To reset your password, please visit the following link
</p>

<a href="{{ url('app_reset_password', {token: resetToken.token}) }}">{{ url('app_reset_password', {token: resetToken.token}) }}</a>

<p>
    This link will expire in {{ tokenLifetime|date('g') }} hour(s).
</p>

<p>
    Cheers!
</p>
```

weaverryan:
Btw, part of my motivation to avoid 2 templates is that it is very likely that the user will want to customize the email. If we ship with 2 templates, they will need to update both of them (and realize that there are 2).

src/Resources/skeleton/resetPassword/twig_email_txt.tpl.php
```diff
@@ -0,0 +1,9 @@
+Hi!
+
+To reset your password, please copy the below link into your browser.
+
+{{ url('app_reset_password', {token: resetToken.token}) }}
+
+This link will expire in {{ tokenLifetime|date('g') }} hour(s)..
```
javiereguiluz, line 7: Super minor comment: there's an extra dot at the end of this phrase.

javiereguiluz:
If possible, Ryan's solution would be better. Otherwise, let's merge this!

src/Resources/skeleton/resetPassword/twig_email_txt.tpl.php
pierredup: This comes directly from the current template https://github.com/symfony/maker-bundle/blob/bb1438ba9d121f58d40e1d3d5a02d2fea45c8ad4/src/Resources/skeleton/resetPassword/twig_email.tpl.php#L6 I thought it might be intentionally there

pierredup:
I'll have a look at implementing the suggestion, but the formatting would still be off.
Although I think this is something that should be fixed in the Symfony mailer side, as just stripping the HTML tags to create a text template causes some issues with the whitespace

pierredup:
I updated the template to the suggestion but inlined some of the text with the HTML tags. This keeps the formatting clean when generating the text template and doesn't have any additional extra spaces before the text.

Screenshot 2020-07-13 at 10 29 55

weaverryan:
This is awesome! Thank you Pierre!

PR #640, opened 2020-06-29
hello docker!

jrushlow:
Spin up a MySQL / MariaDB / Postgres database via docker compose! Whether you already use docker, or want to give it a try - `make:docker:database` will create or update a `docker-compose.yaml` in your projects root directory...

### DIFF
--- a/src/Maker/MakeDockerDatabase.php
+++ b/src/Maker/MakeDockerDatabase.php
@@ -1,0 +1,183 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Docker\DockerDatabaseServices;
+use Symfony\Bundle\MakerBundle\FileManager;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Util\ComposeFileManipulator;
+use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+final class MakeDockerDatabase extends AbstractMaker
+{
+    private $fileManager;
+    private $composeFilePath;
+
+    /**
+     * @var ComposeFileManipulator
+     */
+    private $composeFileManipulator;
+
+    /**
+     * @var string type of database selected by the user
+     */
+    private $databaseChoice;
+
+    /**
+     * @var string Service identifier to be set in docker-compose.yaml
+     */
+    private $serviceName = 'database';
+
+    /**
+     * @var string Version set in docker-compose.yaml for the service. e.g. latest
+     */
+    private $serviceVersion = 'latest';
+
+    public function __construct(FileManager $fileManager)
+    {
+        $this->fileManager = $fileManager;
+        $this->composeFilePath = sprintf('%s/docker-compose.yaml', $this->fileManager->getRootDirectory());
+    }
+
+    public static function getCommandName(): string
+    {
+        return 'make:docker:database';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig): void
+    {
+        $command
+            ->setDescription('Adds a database container to your docker-compose.yaml file.')
+        ;
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
+    {
+        $io->section('- Docker Compose Setup-');
+
+        $composeFileContents = '';
+        $statusMessage = 'Existing docker-compose.yaml not found: a new one will be generated!';
+
+        if ($this->fileManager->fileExists($this->composeFilePath)) {
+            $composeFileContents = $this->fileManager->getFileContents($this->composeFilePath);
+
+            $statusMessage = 'We found your existing docker-compose.yaml: Let\'s update it!';
+        }
+
+        $io->text($statusMessage);
+
+        $this->composeFileManipulator = new ComposeFileManipulator($composeFileContents);
+
+        $io->newLine();
+
+        $this->databaseChoice = strtolower($io->choice(
+            'Which database service will you be creating?',
+            ['MySQL', 'MariaDB', 'Postgres']
+        ));
+
+        $io->text([sprintf(
+            'For a list of supported versions, check out https://hub.docker.com/_/%s',
+            $this->databaseChoice
+        )]);
+
+        $this->serviceVersion = $io->ask('What version would you like to use?', DockerDatabaseServices::getSuggestedServiceVersion($this->databaseChoice));
+
+        if ($this->composeFileManipulator->serviceExists($this->serviceName)) {
+            $io->comment(sprintf('A <fg=yellow>"%s"</> service is already defined.', $this->serviceName));
+            $io->newLine();
+
+            $serviceNameMsg[] = 'If you are using the Symfony Binary, it will expose the connection config for';
+            $serviceNameMsg[] = 'this service as environment variables. The name of the service determines the';
+            $serviceNameMsg[] = 'name of those environment variables.';
+            $serviceNameMsg[] = '';
+            $serviceNameMsg[] = 'For example, if you name the service <fg=yellow>database_alt</>, the binary will expose a';
+            $serviceNameMsg[] = '<fg=yellow>DATABASE_ALT_URL</> environment variable.';
+
+            $io->text($serviceNameMsg);
+
+            $this->serviceName = $io->ask(sprintf('What name should we call the new %s service? e.g. database', $this->serviceName), null, [Validator::class, 'notBlank']);
+        }
+
+        $this->checkForPDOSupport($this->databaseChoice, $io);
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator): void
+    {
+        $io->newLine();
+
+        $service = DockerDatabaseServices::getDatabaseSkeleton($this->databaseChoice, $this->serviceVersion);
+
+        $this->composeFileManipulator->addDockerService($this->serviceName, $service);
+        $this->composeFileManipulator->exposePorts($this->serviceName, DockerDatabaseServices::getDefaultPorts($this->databaseChoice));
+
+        $generator->dumpFile($this->composeFilePath, $this->composeFileManipulator->getDataString());
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text(sprintf('The new <fg=yellow>"%s"</> service is now ready!', $this->serviceName));
+        $io->newLine();
+
+        $ports = DockerDatabaseServices::getDefaultPorts($this->databaseChoice);
+        $closing[] = 'Next:';
+        $closing[] = sprintf(' A) Run <fg=yellow>docker-compose up -d %s</> to start your database container', $this->serviceName);
+        $closing[] = sprintf('    or <fg=yellow>docker-compose up -d</> to start all of them.');
+        $closing[] = '';
+        $closing[] = ' B) If you are using the Symfony Binary, it will detect the new service automatically.';
+        $closing[] = '    Run <fg=yellow>symfony var:export --multiline</> to see the environment variables the binary is exposing.';
+        $closing[] = '    These will override any values you have in your .env files.';
+        $closing[] = '';
+        $closing[] = ' C) Run <fg=yellow>docker-compose stop</> will stop all the containers in docker-compose.yaml.';
+        $closing[] = '    <fg=yellow>docker-compose down</> will stop and destroy the containers.';
+        $closing[] = '';
+        $closing[] = sprintf(
+            'Port%s %s will be exposed to %s random port%s on your host machine.',
+            1 === \count($ports) ? '' : 's',
+            implode(' ', $ports),
+            1 === \count($ports) ? 'a' : '',
+            1 === \count($ports) ? '' : 's'
+        );
+
+        $io->text($closing);
+        $io->newLine();
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies): void
+    {
+        $dependencies->addClassDependency(
+            Yaml::class,
+            'yaml'
+        );
+    }
+
+    private function checkForPDOSupport(string $databaseType, ConsoleStyle $io): void
+    {
+        $extension = DockerDatabaseServices::getMissingExtensionName($databaseType);
+
+        if (null !== $extension) {
+            $io->note(
+                sprintf('Cannot find PHP\'s pdo_%s extension. Be sure it\'s installed & enabled to talk to the database.', $extension)
+            );
+        }
+    }
+}
--- a/src/Util/ComposeFileManipulator.php
+++ b/src/Util/ComposeFileManipulator.php
@@ -1,0 +1,132 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Component\Yaml\Dumper;
+
+/**
+ * Manipulate Docker Compose Files.
+ *
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ * @final
+ */
+class ComposeFileManipulator
+{
+    public const COMPOSE_FILE_VERSION = '3.7';
+
+    /** @var YamlSourceManipulator */
+    private $manipulator;
+
+    public function __construct(string $contents)
+    {
+        if ('' === $contents) {
+            $this->manipulator = new YamlSourceManipulator(
+                (new Dumper())->dump($this->getBasicStructure(), 2)
+            );
+        } else {
+            $this->manipulator = new YamlSourceManipulator($contents);
+        }
+
+        $this->checkComposeFileVersion();
+    }
+
+    public function getComposeData(): array
+    {
+        return $this->manipulator->getData();
+    }
+
+    public function getDataString(): string
+    {
+        return $this->manipulator->getContents();
+    }
+
+    public function serviceExists(string $name): bool
+    {
+        $data = $this->manipulator->getData();
+
+        if (\array_key_exists('services', $data)) {
+            return \array_key_exists($name, $data['services']);
+        }
+
+        return false;
+    }
+
+    public function addDockerService(string $name, array $details): void
+    {
+        $data = $this->manipulator->getData();
+
+        $data['services'][$name] = $details;
+
+        $this->manipulator->setData($data);
+    }
+
+    public function removeDockerService(string $name): void
+    {
+        $data = $this->manipulator->getData();
+
+        unset($data['services'][$name]);
+
+        $this->manipulator->setData($data);
+    }
+
+    public function exposePorts(string $service, array $ports): void
+    {
+        $portData = [];
+        $portData[] = sprintf('%s To allow the host machine to access the ports below, modify the lines below.', YamlSourceManipulator::COMMENT_PLACEHOLDER_VALUE);
+        $portData[] = sprintf('%s For example, to allow the host to connect to port 3306 on the container, you would change', YamlSourceManipulator::COMMENT_PLACEHOLDER_VALUE);
+        $portData[] = sprintf('%s "3306" to "3306:3306". Where the first port is exposed to the host and the second is the container port.', YamlSourceManipulator::COMMENT_PLACEHOLDER_VALUE);
+        $portData[] = sprintf('%s See https://docs.docker.com/compose/compose-file/#ports for more information.', YamlSourceManipulator::COMMENT_PLACEHOLDER_VALUE);
+
+        foreach ($ports as $port) {
+            $portData[] = $port;
+        }
+
+        $data = $this->manipulator->getData();
+
+        $data['services'][$service]['ports'] = $portData;
+
+        $this->manipulator->setData($data);
+    }
+
+    public function addVolume(string $service, string $hostPath, string $containerPath): void
+    {
+        $data = $this->manipulator->getData();
+
+        $data['services'][$service]['volumes'][] = sprintf('%s:%s', $hostPath, $containerPath);
+
+        $this->manipulator->setData($data);
+    }
+
+    private function getBasicStructure(string $version = self::COMPOSE_FILE_VERSION): array
+    {
+        return [
+            'version' => $version,
+            'services' => [],
+        ];
+    }
+
+    private function checkComposeFileVersion(): void
+    {
+        $data = $this->manipulator->getData();
+
+        if (empty($data['version'])) {
+            throw new RuntimeCommandException('docker-compose.yaml file version is not set.');
+        }
+
+        if (2.0 > (float) $data['version']) {
+            throw new RuntimeCommandException(sprintf('docker-compose.yaml version %s is not supported. Please update your docker-compose.yaml file to the latest version.', $data['version']));
+        }
+    }
+}
--- a/tests/Maker/FunctionalTest.php
+++ b/tests/Maker/FunctionalTest.php
@@ -12,9 +12,9 @@
 namespace Symfony\Bundle\MakerBundle\Tests\Maker;
 
 use PHPUnit\Framework\TestCase;
+use ReflectionClass;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Bundle\MakerBundle\Command\MakerCommand;
-use Symfony\Bundle\MakerBundle\Maker\AbstractMaker;
 use Symfony\Bundle\MakerBundle\Test\MakerTestKernel;
 use Symfony\Component\Finder\Finder;
 
@@ -33,17 +33,19 @@ class FunctionalTest extends TestCase
 
         $application = new Application($kernel);
         foreach ($finder as $file) {
-            $class = 'Symfony\Bundle\MakerBundle\Maker\\'.$file->getBasename('.php');
+            $maker = new ReflectionClass(sprintf('Symfony\Bundle\MakerBundle\Maker\%s', $file->getBasename('.php')));
 
-            if (AbstractMaker::class === $class) {
+            if ($maker->isAbstract()) {
                 continue;
             }
 
-            $commandName = $class::getCommandName();
             // if the command does not exist, this will explode
-            $command = $application->find($commandName);
+            $command = $application->find(
+                $maker->getMethod('getCommandName')->invoke(null)
+            );
+
             // just a smoke test assert
-            $this->assertInstanceOf(MakerCommand::class, $command);
+            self::assertInstanceOf(MakerCommand::class, $command);
         }
     }
 }
--- a/src/Docker/DockerDatabaseServices.php
+++ b/src/Docker/DockerDatabaseServices.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Docker;
+
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+
+/**
+ * @author  Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+final class DockerDatabaseServices
+{
+    /**
+     * @throws RuntimeCommandException
+     */
+    public static function getDatabaseSkeleton(string $name, string $version): array
+    {
+        switch ($name) {
+            case 'mariadb':
+                return [
+                    'image' => sprintf('mariadb:%s', $version),
+                    'environment' => [
+                        'MYSQL_ROOT_PASSWORD' => 'password',
+                    ],
+                ];
+            case 'mysql':
+                return [
+                    'image' => sprintf('mysql:%s', $version),
+                    'environment' => [
+                        'MYSQL_ROOT_PASSWORD' => 'password',
+                    ],
+                ];
+            case 'postgres':
+                return [
+                    'image' => sprintf('postgres:%s', $version),
+                    'environment' => [
+                        'POSTGRES_PASSWORD' => 'main',
+                        'POSTGRES_USER' => 'main',
+                        'POSTGRES_DB' => 'main',
+                    ],
+                ];
+        }
+
+        self::throwInvalidDatabase($name);
+    }
+
+    /**
+     * @throws RuntimeCommandException
+     */
+    public static function getDefaultPorts(string $name): array
+    {
+        switch ($name) {
+            case 'mariadb':
+            case 'mysql':
+                return ['3306'];
+            case 'postgres':
+                return ['5432'];
+        }
+
+        self::throwInvalidDatabase($name);
+    }
+
+    public static function getSuggestedServiceVersion(string $name): string
+    {
+        if ('postgres' === $name) {
+            return 'alpine';
+        }
+
+        return 'latest';
+    }
+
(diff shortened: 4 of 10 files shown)


src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->serviceVersion = $io->ask('What version would you like to use?', 'latest');
+
+        if ($this->composeFileManipulator->serviceExists($this->serviceName)) {
+            $this->serviceAlreadyDefinedQuestion($io);
+
+            $this->serviceName = $io->ask(sprintf(
+                'What name should we call the new %s service? e.g. %s',
+                $this->serviceName,
+                str_replace(' ', '-', Str::getRandomTerm())
+            ));
```
weaverryan, line 74: Actually, we need to be opinionated here. If the service is called `database`, then, iirc, the Symfony binary will expose `DATABASE_URL`. So, we should suggest `database` as the default. We should maybe even print a note about how the name will map to the exposed env var with the Symfony binary... but I'm not positive on that

src/Docker/DatabaseServices.php
```diff
@@ -0,0 +1,64 @@
...
+                    'image' => sprintf('mysql:%s', $version),
+                    'environment' => [
+                        'MYSQL_ROOT_PASSWORD' => 'password',
+                    ],
+                    'ports' => ['3306:3306'],
+                ];
+            case 'postgres':
+                return [
+                    'image' => sprintf('postgres:%s', $version),
+                    'ports' => ['5432:5432'],
```
weaverryan, line 43: We can (as Fabien does) simplify these to just `5432` - e.g. `ports: [5432]`

src/Docker/DatabaseServices.php
```diff
@@ -0,0 +1,64 @@
...
+                return [
+                    'image' => sprintf('mysql:%s', $version),
+                    'environment' => [
+                        'MYSQL_ROOT_PASSWORD' => 'password',
+                    ],
+                    'ports' => ['3306:3306'],
+                ];
+            case 'postgres':
+                return [
+                    'image' => sprintf('postgres:%s', $version),
```
weaverryan, line 42: Are no env vars needed for postgres! I know that Fabien includes these in his book:

```yml
POSTGRES_USER: main
POSTGRES_PASSWORD: main
POSTGRES_DB: main
```

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $this->databaseChoice = $io->choice(
+            'Which database service will you be creating?',
+            ['MySQL', 'MariaDB', 'Postgres']
+        );
+
+        $this->serviceName = strtolower($this->databaseChoice);
+
+        $io->text([sprintf(
+            'For a list of supported versions, check out https://hub.docker.com/_/%s',
```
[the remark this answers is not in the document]
jrushlow: what would happen if the api call fails? would that lockup maker?

src/Docker/DatabaseServices.php
jrushlow: good catch, I think postgres needs a password.. i'll double check the docs and adjust accordingly..

src/Maker/MakeDockerDatabase.php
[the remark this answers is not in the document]
LeJeanbono: Manualy ask, as you did here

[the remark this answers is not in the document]
jrushlow: only other concern would be a degraded UX on a slow connection.. don't get me wrong, i love the idea.. just trying to think of any pitfalls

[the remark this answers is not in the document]
weaverryan: We're thinking about showing a few common options, like `alpine`, `latest` and `12` for pgsql. If you care to get a specific version, then you'll know what you want :)

src/Docker/DatabaseServices.php
jrushlow: i'll have to hunt down exactly why, but i ran into a problem on a debian host - if the ports were not mapped out like `[$hostPort:$containerPort]` docker wouldn't expose the port to the host.. I'll reconfirm that edge..

src/Docker/DatabaseServices.php
jrushlow: done

src/Docker/DatabaseServices.php
JHeimbach: If you only specify the container port (5432) docker will choose an ephemeral host port, you run into problems if you choose your database host as localhost.

weaverryan: So this is tricky then. The Symfony binary (I believe) automatically detects the random port and uses it to expose the environment variable to the server. In that case, we *do* want the random behavior because it allows you to have multiple database servers (from multiple projects) running on your machine at once. 

This means that this is a decision point that largely depends on whether or not you'll be using the Symfony binary to interface with Docker.

Before we figure out what to do, is my understanding accurate?

Thanks :)

src/Docker/DatabaseServices.php
```diff
@@ -0,0 +1,80 @@
...
+                    'environment' => [
+                        'POSTGRES_PASSWORD' => 'main',
+                        'POSTGRES_USER' => 'main',
+                        'POSTGRES_DB' => 'main',
+                    ],
+                    'ports' => ['5432'],
+                ];
+        }
```
weaverryan, line 50: Throw an exception down here if none was matched (that will also enforce the `array` return type)

src/Docker/DatabaseServices.php
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @author  Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+final class DatabaseServices
+{
+    public static function getDatabaseSkeleton(string $name, string $version): array
+    {
+        switch (strtolower($name)) {
```
weaverryan, line 23: Could we remove the `strtolower`?

src/Docker/DatabaseServices.php
```diff
@@ -0,0 +1,80 @@
...
+                $ports = ['5432'];
+                break;
+        }
+
+        return $ports;
+    }
+
+    public static function getSuggestedServiceVersion(string $name): string
+    {
+        $name = strtolower($name);
```
weaverryan, line 72: Same here - are we guarding against some user input that might us capitals? Or is that case not really possible?

src/Maker/AbstractDockerMaker.php
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @var string Service identifier to be set in docker-compose.yaml.
+     */
+    protected $serviceName = '';
+
+    /**
+     * @var string Version set in docker-compose.yaml for the service. e.g. latest
+     */
+    protected $serviceVersion = '';
```
weaverryan, line 57: I think these might be better as protected methods. Then, if we need a property, we can default them to null. I know that might be tricky to get the flow right, but it smells weird how it is now.

src/Maker/AbstractDockerMaker.php
```diff
@@ -0,0 +1,100 @@
...
+
+            $io->text('Existing Docker Compose file found.');
+        }
+
+        $this->composeFileManipulator = new ComposeFileManipulator($composeFileContents);
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator): void
+    {
+    }
```
weaverryan, line 85: This shouldn't be needed, right?

src/Maker/AbstractDockerMaker.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->composeFilePath = sprintf('%s/docker-compose.yaml', $this->fileManager->getRootDirectory());
+
+        $composeFileContents = '';
+
+        if ($this->fileManager->fileExists($this->composeFilePath)) {
+            $composeFileContents = $this->fileManager->getFileContents($this->composeFilePath);
+
+            $this->existingComposeFile = true;
+
+            $io->text('Existing Docker Compose file found.');
```
weaverryan, line 77: Maybe:

> Existing docker-compose.yaml not found: a new one will be generated!

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+
+    public static function getCommandName(): string
+    {
+        return 'make:docker:database';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig): void
+    {
+        $command
+            ->setDescription('Create a database Docker container.')
```
weaverryan, line 43: > Adds a database container to your docker-compose.yaml file

I was thinking this might be more descriptive?

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+final class MakeDockerDatabase extends AbstractDockerMaker
+{
+    /**
+     * @var string
+     */
+    private $databaseChoice = '';
```
weaverryan, line 33: I usually default props to null if they're truly not set yet...

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+        );
+
+        $io->text([sprintf(
+            'For a list of supported versions, check out https://hub.docker.com/_/%s',
+            strtolower($this->databaseChoice)
+        )]);
+
+        $suggestedVersion = DatabaseServices::getSuggestedServiceVersion($this->databaseChoice);
+
+        $this->serviceVersion = $io->ask('What version would you like to use?', $suggestedVersion);
```
weaverryan, line 65: I realize that the serviceVersion or serviceName could be re-used across Docker makers. But having the prop in the parent class but asking it here looks weird. I'd love if they were fully in one place or the other.

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+
+        $this->serviceName = 'database';
+
+        if ($this->composeFileManipulator->serviceExists($this->serviceName)) {
+            $this->showServiceAlreadyDefinedComment($io);
+
+            $randomName = str_replace(' ', '-', Str::getRandomTerm());
```
weaverryan, line 72: Instead of a random term, use `database` here. It's just more realistic (even if `database` won't be the name they actually use here... since they already have it)

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+            ));
+        }
+
+        $io->section(sprintf('- %s -', $this->databaseChoice));
+
+        $defaults = [
+            sprintf('Port(s) %s are exposed to the host.', ...(DatabaseServices::getDefaultPorts($this->databaseChoice))),
+        ];
+
+        $io->text($defaults);
```
weaverryan, line 92: It seems like an odd thing to mention the ports before we do the generating - the ports feel information that might be at the end.

src/Maker/AbstractDockerMaker.php
```diff
@@ -0,0 +1,100 @@
...
+        $io->section('- Docker Compose Setup-');
+
+        $this->composeFilePath = sprintf('%s/docker-compose.yaml', $this->fileManager->getRootDirectory());
+
+        $composeFileContents = '';
+
+        if ($this->fileManager->fileExists($this->composeFilePath)) {
+            $composeFileContents = $this->fileManager->getFileContents($this->composeFilePath);
+
+            $this->existingComposeFile = true;
```
weaverryan, line 75: I don't think this property is used

src/Maker/AbstractDockerMaker.php
```diff
@@ -0,0 +1,100 @@
...
+     * @var string Version set in docker-compose.yaml for the service. e.g. latest
+     */
+    protected $serviceVersion = '';
+
+    public function __construct(FileManager $fileManager)
+    {
+        $this->fileManager = $fileManager;
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
```
weaverryan, line 64: We could refactor this into a `getComposeFileManipulator(ConsoleStyle $io): ComposeFileManipulator` type of function. You would then call it from your `interact` method. We (in this class) would store it on a `private` property and avoid creating it twice.

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+                'What name should we call the new %s service? e.g. %s',
+                $this->serviceName,
+                $randomName
+            ));
+        }
+
+        $io->section(sprintf('- %s -', $this->databaseChoice));
+
+        $defaults = [
+            sprintf('Port(s) %s are exposed to the host.', ...(DatabaseServices::getDefaultPorts($this->databaseChoice))),
```
weaverryan, line 89: Does `...` work like this!!?!?!

src/Maker/AbstractDockerMaker.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        $io->section('- Docker Compose Setup-');
+
+        $this->composeFilePath = sprintf('%s/docker-compose.yaml', $this->fileManager->getRootDirectory());
```
weaverryan, line 68: We can avoid this property (well, at least avoid making it protected - we could make it private) by adding a new protected method in this class to save the Composer file:

```
protected function dumpComposeFile(Generator $generator)
{
    // we do the dumping here
}
```

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+        $generator->dumpFile($this->composeFilePath, Yaml::dump($this->composeFileManipulator->getComposeData(), 20));
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text(sprintf('The new <fg=yellow>"%s"</> service is now ready.', $this->serviceName));
+        $io->newLine();
+
+        $closing[] = 'Next:';
+        $closing[] = 'Run <fg=yellow>docker-compose up -d</> to start your database container.';
```
weaverryan, line 112: WDYT about:

> ... to start all of your containers.

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+        $closing[] = 'If you are using the Symfony Binary, it will detect the new service automatically.';
```
weaverryan, line 113: And one more line:

> Run <fg=yellow>symfony var:export --multiline</> to see the environment variables exposed for Docker.

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,121 @@
...
+        $closing[] = '';
+        $closing[] = sprintf('<fg=yellow>docker-compose stop %s</> will stop the container without removing it.', $this->serviceName);
+        $closing[] = '<fg=yellow>docker-compose down</> will stop and remove all containers in the compose file.';
```
weaverryan, line 116: We should probably make these 2 consistent with either mentioning the service name or not. For me, the key thing is to say that "stop" stops and "down" fully destroys them. My preference would be to show both of them without the serviceName.

Also:

> will stop and destroy all of your containers.

src/Util/ComposeFileManipulator.php
```diff
@@ -0,0 +1,78 @@
...
+ * @internal
+ * @final
+ */
+class ComposeFileManipulator
+{
+    public const COMPOSE_FILE_VERSION = '3.7';
+
+    private $composeData;
+
+    public function __construct(string $contents = '')
```
weaverryan, line 28: No need to have a default value I think?

src/Util/ComposeFileManipulator.php
```diff
@@ -0,0 +1,78 @@
...
+        if (\array_key_exists('services', $this->composeData)) {
+            return \array_key_exists($name, $this->composeData['services']);
+        }
+
+        return false;
+    }
+
+    public function addDockerService(string $name, array $details): void
+    {
+        $this->composeData['services'][$name] = $details;
```
weaverryan, line 53: Edge case: but in some insane universe, the `services` might not exist (like, if they created their own... and I guess had zero services to start)

src/Util/ComposeFileManipulator.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public function removeDockerService(string $name): void
+    {
+        unset($this->composeData['services'][$name]);
+    }
+
+    public function exposePorts(string $service, array $ports): void
+    {
+        $this->composeData['services'][$service]['ports'] = $ports;
```
weaverryan, line 63: Same here and below - we should probably include some `isset()` checks and throw an exception if something is wrong. It would be totally unexpected - but better than having an offset does not exist error.

src/Docker/DatabaseServices.php
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+namespace Symfony\Bundle\MakerBundle\Docker;
+
+/**
+ * @author  Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+final class DatabaseServices
```
weaverryan, line 19: DockerDatabaseServices

... just trying to make this more unique and boring... to avoid any conflicts in a real app or anyone thinking it is more interesting than it really is

tests/Maker/FunctionalTest.php
```diff
@@ -35,7 +36,7 @@ public function testWiring()
         foreach ($finder as $file) {
             $class = 'Symfony\Bundle\MakerBundle\Maker\\'.$file->getBasename('.php');
 
-            if (AbstractMaker::class === $class) {
+            if (AbstractMaker::class === $class || AbstractDockerMaker::class === $class) {
```
weaverryan, line 39: I guess we could use reflection to check if the class abstract - I don't care too much either way

src/Docker/DatabaseServices.php
chapterjason: I agree, AFAIK `^[a-zA-Z0-9._-]+$` this is a validation pattern for this name. I didn't found anything in the docker documentation but I found this specs, I think it comes from YAML as a valid key. https://github.com/compose-spec/compose-spec/blob/master/schema/config_schema_v3.9.json

src/Docker/DatabaseServices.php
jrushlow: done

src/Docker/DatabaseServices.php
jrushlow: exactly what we're doing here but i just realized `case` is doing ` ==` not `===` so I can get rid of those here and above..

src/Docker/DatabaseServices.php
jrushlow: i just realized case is doing == not === so I can get rid of those here and below..

jrushlow: done

src/Docker/DatabaseServices.php
jrushlow: done

jrushlow: to clarify, this is done because we present via `$io` as `MariaDB` or `MySQL` in `interact()`, user selects `0`, `1`, `2` and we ultimately pass the string representation to `DatabaseServices::getDatabaseSkeleton()` or any of the other methods..

src/Docker/DatabaseServices.php
jrushlow: done

src/Maker/AbstractDockerMaker.php
jrushlow: removed

src/Maker/MakeDockerDatabase.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
jrushlow: why not both..
done

src/Util/ComposeFileManipulator.php
jrushlow: Thinking back, we can use the manipulator to create a new file or manipulate an existing one. For the latter we need the contents.. IIRC we went w/ default value to avoid setter injection...

src/Maker/AbstractDockerMaker.php
jrushlow: it was at one time.. #livingInThePast

src/Maker/MakeDockerDatabase.php
jrushlow: living in the past again - there was more too that in a previous life.. 
done

src/Maker/MakeDockerDatabase.php
jrushlow: `docker-compose stop` will stop all of the containers listed in the compose file.
`docker-compose stop mysql` will only stop the container named "mysql" in the compose file.
`docker-compose down` does not accept any args (service names) and stops + destroys all of the containers in the compose file.. 

For `down` I think leaving off the `in the compose file.` might lead to confusion. "Hey i ran down in the project over there, why is this container still running over here?"

src/Util/ComposeFileManipulator.php
jrushlow: i always thought if `services` didn't exist, the above would create it w/o issue.. have i been living my life wrong this whole time?

Then again for arrays that works, but i can't remember what the yaml manipulator does in that situation.

src/Util/ComposeFileManipulator.php
jrushlow: i agree - see above comments

tests/Maker/FunctionalTest.php
jrushlow: I concur, but would reflection add more overhead?

src/Docker/DatabaseServices.php
weaverryan: Ah, ok - that's the explanation I was missing - *why* we needed this in the first place. I'd vote for normalizing it immediately after we ask the question. Then we can keep things "tight" here.

src/Maker/MakeDockerDatabase.php
weaverryan: Ok, then I vote for

> <>docker-compose stop</> will stop all the containers in docker-compose.yaml
> <>docker-compose down</> will stop *and* destroy the containers

WDYT? I mention the `docker-compose.yaml` in the first (you're right that it's clearer with it). Then, I think it's clear enough that we don't need to repeat in the second, but we could.

src/Util/ComposeFileManipulator.php
weaverryan: We never (except in 2 tests) use `new ComposeFileManipulator()` with no args. If we removed the `= ''` part (so, just `string $contents`), then it just forces us to be explicit - e.g. `new ComposeFileManipulator('')` if you want to create from an empty file.

src/Util/ComposeFileManipulator.php
weaverryan: > i always thought if services didn't exist, the above would create it w/o issue.. have i been living my life wrong this whole time?

I just tried it, apparently *I* have been living my life wrong this whole time! 🙃

tests/Maker/FunctionalTest.php
weaverryan: I think the overhead of Reflection has largely (or completely) been disproven. The real overhead, I believe, would be that using Reflection would cause you to autoload these classes. So, it would be super minor - and not important in a test anyways.

src/Maker/MakeDockerDatabase.php
jrushlow: done

src/Util/ComposeFileManipulator.php
jrushlow: that makes sense.. i was thinking along the lines of avoiding confusion like "this thing needs a string, but i dont have one. hmm i wonder what happens if i pass an empty string.." where as an explicit `= ''` gets our point across better that and empty string is ok..

weaverryan: I hear that. From the other side, I think, if an argument is optional, it makes it easier to, sort of, "not notice". I'm cool either way, really

src/Maker/AbstractDockerMaker.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
jrushlow: yup yup

jrushlow: fixed

src/Maker/AbstractDockerMaker.php
jrushlow: done

src/Maker/AbstractDockerMaker.php
jrushlow: refactored

src/Maker/AbstractDockerMaker.php
jrushlow: Hmm i like it. But we also call `$this->composeFileManipulator` in the generate method.. it seem that having to `$manipulator = $this->getComposeFileManipulator($io)` twice in different methods of `MakeDockerDatabase` isn't as dry as providing access to `AbstractDockerMaker::composeFileManipulator`.

Perhaps hold off and see how things look with additional makers. Then refactor this to be a private w/ getter vs protected?. I can see a few different ways to tackle this, just not sure which would be the best for the next round of make dockers..

src/Maker/MakeDockerDatabase.php
jrushlow: what did we decided to do with this? keep as is to "future proof" for typed props or refactor?

src/Util/ComposeFileManipulator.php
jrushlow: open to keeping as is for now, then revisiting this in the next round of make dockers when we have a better idea how this is implemented outside of the single use case we have now?

src/Maker/AbstractDockerMaker.php
jrushlow: what did we decided to do with this? keep as is to "future proof" for typed props or refactor?

tests/Maker/FunctionalTest.php
jrushlow: done

src/Docker/DatabaseServices.php
jrushlow: done

src/Maker/AbstractDockerMaker.php
jrushlow: what abstract maker?

src/Maker/MakeDockerDatabase.php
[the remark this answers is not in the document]
jrushlow: we grab the "suggested version" from `DockerDatabaseServices::getSuggestedServiceVersion` depending on which database is selected.. right now thats either "alpine" or "latest"...

src/Util/ComposeFileManipulator.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,154 @@
...
+
+        if ($this->fileManager->fileExists($this->composeFilePath)) {
+            $composeFileContents = $this->fileManager->getFileContents($this->composeFilePath);
+
+            $statusMessage = 'We found your existing docker-compose.yaml: Let\'s update it!';
+        }
+
+        $io->text($statusMessage);
+
+        $this->composeFileManipulator = new ComposeFileManipulator($composeFileContents);
```
weaverryan, line 88: very nice flow

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,154 @@
...
+            $this->databaseChoice
+        )]);
+
+        $this->serviceVersion = $io->ask('What version would you like to use?', DockerDatabaseServices::getSuggestedServiceVersion($this->databaseChoice));
+
+        if ($this->composeFileManipulator->serviceExists($this->serviceName)) {
+            $io->comment(sprintf('A <fg=yellow>"%s"</> service is already defined.', $this->serviceName));
+            $io->newLine();
+
+            $envTest[] = 'If you are using the Symfony Binary, it will expose the env_var associated with this service.';
```
weaverryan, line 108: `env_var` => `env var` or `environment variable` - not sure `env_var` specifically is a thing :)

And maybe:

> ...it will expose the connection config of this service as an environment variable.
> The *name* of the service determines the name of the environment variable.
> For example, if you name the service 'database', the binary will expose a DATABASE_URL environment variable.

Not sure this is perfect, but it feels like it needs to get "filled out" a bit more

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,154 @@
...
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator): void
+    {
+        $io->newLine();
+
+        $service = DockerDatabaseServices::getDatabaseSkeleton($this->databaseChoice, $this->serviceVersion);
+
+        $this->composeFileManipulator->addDockerService($this->serviceName, $service);
+
+        $generator->dumpFile($this->composeFilePath, Yaml::dump($this->composeFileManipulator->getComposeData(), 20));
```
weaverryan, line 125: Wait a sec... this isn't quite right. We're never using the `YamlSourceManipulator`. I mean, we use it for a hot second in ComposeFileManipulator to load the data, but then we do *not* actually use that class to update the contents. We need to do that - instead of using `Yaml::dump()` - else we're going to blow away any existing comments and formatting.

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,154 @@
...
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text(sprintf('The new <fg=yellow>"%s"</> service is now ready.', $this->serviceName));
```
weaverryan, line 130:
```suggestion
        $io->text(sprintf('The new <fg=yellow>"%s"</> service is now ready!', $this->serviceName));
```

I couldn't resist :p

src/Maker/MakeDockerDatabase.php
```diff
@@ -0,0 +1,154 @@
...
+        $io->newLine();
+
+        $closing[] = 'Next:';
+        $closing[] = sprintf('Run <fg=yellow>docker-compose up -d %s</> to start your database container or <fg=yellow>docker-compose up -d</> to start all of them.', $this->serviceName);
+        $closing[] = 'If you are using the Symfony Binary, it will detect the new service automatically.';
+        $closing[] = 'Run <fg=yellow>symfony var:export --multiline</> to see the environment variables exposed for Docker.';
```
weaverryan, line 136:
```suggestion
        $closing[] = 'Run <fg=yellow>symfony var:export --multiline</> to see the environment variables the binary is exposing.';
```

... edge case... and I'm not even sure *what* the edge case would be... but this prints all env variables, not just the ones from docker.

src/Util/ComposeFileManipulator.php
```diff
@@ -0,0 +1,78 @@
...
+    public const COMPOSE_FILE_VERSION = '3.7';
+
+    private $composeData;
+
+    public function __construct(string $contents)
+    {
+        if ('' === $contents) {
+            $this->composeData = $this->getBasicStructure();
+        } else {
+            $this->composeData = (new YamlSourceManipulator($contents))->getData();
```
weaverryan, line 33: Per my comment about using `YamlSourceManipulator` to actually make changes to the YAML, we should probably maintain an instance of `YamlSourceManipulator` on a private property, then use `->getData()` on it whenever we need to look at data. And, all mutations would go through the YSM instead of on a local `composeData` array (which would now not exist).

src/Util/ComposeFileManipulator.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public function removeDockerService(string $name): void
+    {
+        unset($this->composeData['services'][$name]);
+    }
+
+    public function exposePorts(string $service, array $ports): void
+    {
+        $this->composeData['services'][$service]['ports'] = $ports;
```
weaverryan, line 63: This is where we should add the comments we talked about:

```
$finalPortData = [];
foreach ($ports as $port) {
    $finalPortData[] = sprintf('__COMMENT__ exposes port %s to a random port on your host machine', $port),
    $finalPortData[] => sprintf('__COMMENT__ use %s:%s to expose to port %s on your host machine', $port, $port, $port),
    $finalPortData[] = $port;
}

// but, using the YSM in reality
$this->composeData['services'][$service]['ports'] = $finalPortData;
```

That code probably contains syntax errors - but hopefully you get the idea :). It's not optimized for multiple ports... because we don't have that use-case yet (but it would work fine).

jrushlow: done

src/Util/ComposeFileManipulator.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
jrushlow: done

src/Maker/MakeDockerDatabase.php
jrushlow: refactored

weaverryan:
Thank you @jrushlow! This works great and I'm *super* excited about it :)

PR #645, opened 2020-07-03
Docs: improve make:form helper

### LINKED ISSUE #477: make:form only allows for entities to be bound
When trying to create a new form via `bin/console make:form` whilst supplying the bound class one can not specify a class which is not inside the `App\Entity` namespace even when one specifying the fully qualified class name. 

Entering:
```bash
bin/console make:form RegisterType App\Mode\Form\Register
```
Results in the following response
```
In ClassDetails.php line 45:

  Class App\Entity\AppModeFormRegister does not exist


make:form [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<name> [<bound-class>]]
```

I would expect to be able to bind any class to the form and not being bound to a specific namespace.

ckrack:
Add information about the second argument (bound-class).
Give hints on using the bound-class with classes other than entities.

Improves on #301 and fixes #477

### DIFF
--- a/src/Resources/help/MakeForm.txt
+++ b/src/Resources/help/MakeForm.txt
@@ -3,3 +3,14 @@ The <info>%command.name%</info> command generates a new form class.
 <info>php %command.full_name% UserType</info>
 
 If the argument is missing, the command will ask for the form class interactively.
+
+You can optionally specify the bound class in a second argument.
+This can be the name of an entity like <info>User</info>
+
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks @ckrack!

PR #646, opened 2020-07-03
redirect to check email route on failure

jrushlow:
default to failing silently on a reset password request in the `ResetPasswordController`. This will prevent the actual cause of the failure (user throttle limit, unknown user, etc..) from being displayed. We also redirect to the check email route by default.

The user has the option to reveal the cause of the failure via a flash or, preferably, implementing a logging solution to store the error message.


weaverryan:
Thanks Jesse!

PR #652, opened 2020-07-30
Fix logout setup on updater

matthieumota:
Hi, catch a little miss arg for updater in ```maker:auth```.


weaverryan:
Nice catch! Thanks Matthieu!

PR #658, opened 2020-08-09
Better guess entity name in Voter template

yahyaerturan:
`$subject instanceof \App\Entity\BlogPost;` part replaced with `$subject instanceof \App\Entity\<?= str_replace('Voter', null, $class_name) ?>;` to better guess entity name based on Voter naming convention.


weaverryan:
This has bothered me for a LONG time. Thank you @yahyaerturan!

PR #659, opened 2020-08-10
Remove "internal" from FileManager

wouterj:
I'm rewriting the makers of my eloquent bundle to the maker bundle. It works great, except for one thing: For the best consistent experience, I use Laravel's stub files and generator. This means I cannot use the `Generator` argument of `generate()`. However, I would like to use MakerBundle's `FileManager` (for nice output and FQCN to path transformations).

This requires `FileManager` to no longer be internal (at least, I don't want to depend on internal classes in a low-maintenance bundle). Would it be possible to make `FileManager` final only?

### DIFF
--- a/src/FileManager.php
+++ b/src/FileManager.php
@@ -21,7 +21,7 @@ use Symfony\Component\Finder\Finder;
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  * @author Ryan Weaver <weaverryan@gmail.com>
  *
- * @internal
+ * @final
  */
 class FileManager
 {


weaverryan:
Sorry for the slow reply @wouterj :). Can you tell me which methods you are specifically using? This class does a lot... in part because it's marked `@internal` so I'm not overly worried about its design. It could be that we isolate a few of the methods that are most useful to a new class that's not `@internal` (and make sure the design of that class(es) makes sense).

Cheers!

wouterj:
I'm using `getRelativePathForFutureClass()` (to get class paths) and `dumpFile()`. Basically, Laravel's stub file format is a bit weird mix of blade and raw text replacements, so I cannot use the `Generator` class. I still want the consistent file changed/created reports generated by other makers and rely on the smart file path guessing in this bundle.

Now I'm looking at the `Generator` class again and if `parseTemplate()` would live in the `Generator` class, I could also just extend the Generator class and override that method. That would allow keeping `FileManager` internal I think, but it does require moving a method to another class.

VincentLanglet:
Hi @wouterj @weaverryan 

What's the status of this ?

Moving methods does not seems enough to me.
The Generator is not internal, so I should be able to instantiate it in my tests.
```
$fileManager = new FileManager(...);
new Generator($fileManager, $namespace);
```
Since it's internal, I can't be sure that the FileManager::__construct() won't change.

Moreover since, the version 1.25, Generator need another argument `PhpCompatUtil`.
The same problem exists since `PhpCompatUtil` is internal.

jrushlow:
Howdy @VincentLanglet, for the time being we're going to keep `FileManager` internal. Anytime we add to the public API that of course increases the burden on the maintainers to prevent BC in future updates. In regards to `Generator::class` + `PhpCompatUtil`, you still can use the generator in your tests. Creating a mock of `PhpCompatUtil` should do the trick if you need to test against its functionality. 

Going to close this for now as it sounds like @wouterj has found an alternative approach. Thanks!

PR #663, opened 2020-08-13
swapping jellybean for popsicle

Leannapelham:
the capitalization in the wizard was confusing because Jelly Bean is two words and 'bean' was showing up lowercase.


weaverryan:
Woo! Go team! Thanks Leanna!

PR #664, opened 2020-08-22
Use Command:SUCCESS constant for command exit code when possible

Chi-teck:
This is follow-up for https://github.com/symfony/symfony/issues/35431

It would also align generated code with an example in the documentation.
https://symfony.com/doc/current/console.html#creating-a-command


lyrixx:
Thanks for your first PR 🎉 

For what is worth, I prefer to read `0` (It's the same for HTTP Status Code).
More over I would keep the code simpler.

But let's wait for feedback from other contributor :)

weaverryan:
Ah, cool! It's subjective, but I like this. And I noticed in the original PR that Javier git a lot of +1's when he said he liked the idea. So let's do it! Thanks @Chi-teck!

PR #671, opened 2020-09-11
[make:controller] Set index method return type

robmeijer:
This sets the return type of the default `index()` method in a new controller generated by `make:controller`.

This is consistent with the various other controllers generated by the bundle.


BatsaxIV:
Looks good to me!

kasteckis (approved):

weaverryan:
Thank you Rob!

PR #672, opened 2020-09-14
#395 - guess entity name in normalizer template

BatsaxIV:
Hi,

To fix https://github.com/symfony/maker-bundle/issues/395, I propose to do like in https://github.com/symfony/maker-bundle/pull/658. 
_This is only for Normalizer, because Voter was already fixed in https://github.com/symfony/maker-bundle/pull/658_

This intends to replace : https://github.com/symfony/maker-bundle/pull/416


weaverryan:
Thank you Jérémy! And we were missing a test! Nice catch

PR #676, opened 2020-09-19
Added alternative suggestion after creating an entity

kasteckis:
When I started learning symfony after creating my very first entity I started using migrations, I didn't even knew that I can run `bin/console d:s:u --force`.

I think new users should be suggested this option


DuboisS:
I was in the same situation as you.
But I'm affraid that someone will do it on a project deployed in production...

weaverryan:
> When I started learning symfony after creating my very first entity I started using migrations, I didn't even knew that I can run bin/console d:s:u --force.

You're correct that this is a fine thing to do, but for beginners, it's not actually something I *want* them to know. I want them to learn migrations and learn how to use migrations. Even I don't use `doctrine:schema:update` anymore. The problem is that if you use this instead of running a migration... then you can't run `make:migration` anymore, because the diff will be gone. So I'd prefer to push people towards the "proper" way of doing it :).

Thanks!

PR #679, opened 2020-09-21
Fix adding a key to a second-level hash after a key with empty value in YAML

### LINKED ISSUE #678: Cannot add a key to a second-level hash after a key with empty value in YAML
Given:
```yaml
volumes:
  abc-data:

networks:
```

Manipulating it with:
```php
$data['volumes']['foo-data'] = 'anything';
```

The expected result is:
```yaml
volumes:
  abc-data:
  foo-data: anything
```

Actual result: `YamlManipulationFailedException` with message `Failed updating YAML contents: the process was successful, but something was not updated. Expected new data: ...`.

The would-be output from the exception message:
```yaml
volumes:
  abc-data:

foo-data: anything
networks:
```

~Failing test case~ Fix PR: https://github.com/symfony/maker-bundle/pull/679

Adding the same key after a key with a non-empty value works as expected.

andrewmy:
Fixes #678

### DIFF
--- a/src/Util/YamlSourceManipulator.php
+++ b/src/Util/YamlSourceManipulator.php
@@ -909,7 +909,7 @@ class YamlSourceManipulator
         $previousIndentation = $this->indentationForDepths[$this->depth];
         $newIndentation = $previousIndentation;
 
-        if ($advancedContent === "\n") {
+        if ("\n" === $advancedContent) {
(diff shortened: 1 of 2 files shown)


weaverryan:
Amazing. Any time someone can dig through my mess of code in this class and fix a tough bug, I'm amazed... and it's very rare. Sorry for the slow merge!

PR #689, opened 2020-10-22
 [WIP] Feat: Type properties with PHP >= 7.4

Superkooka:
Edit: now this flag is global and you can defined if you want to use typing in the configuraiton file 🥳 Just one test dosn't pass.
When it's fix, this PR is ready to be merge.

----

Hi 👋 
With PHP 7.4 and the arrival of PHP 8.0 typing is starting to gain importance. So it's why I have adding the ``--typed`` flag. This flag type the properties of your entity.

For exemple, with PHP 7.4 or more, if I create this entity, I obtain the PHP below:
```shell
PS D:\symfony\sf-maker> php .\bin\console make:entity --typed

 Class name of the entity to create or update (e.g. AgreeableKangaroo):
 > TypedPizza
TypedPizza

 created: src/Entity/TypedPizza.php
 created: src/Repository/TypedPizzaRepository.php

 Entity generated! Now let's add some fields!
 You can always add more fields later manually or by re-running this command.

 > isCooked

 Field type (enter ? to see all types) [boolean]:
 >


 Can this field be null in the database (nullable) (yes/no) [no]:
 >

 updated: src/Entity/TypedPizza.php

 Add another property? Enter the property name (or press <return> to stop adding fields):
 >



  Success!


 Next: When you're ready, create a migration with php bin/console make:migration
```

```php
<?php

namespace App\Entity;

use App\Repository\TypedPizzaRepository;

/**
 * @ORM\Entity(repositoryClass=TypedPizzaRepository::class)
 */
class TypedPizza
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\Column(type="boolean")
     */
    private ?bool $isCooked;

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getIsCooked(): ?bool
    {
        return $this->isCooked;
    }

    public function setIsCooked(bool $isCooked): self
    {
        $this->isCooked = $isCooked;

        return $this;
    }
}
```

If I execute this command with a lower PHP version I have a confirmation warning to ask if I really want typed properties.
To run the test, this pr need PHP 7.4 inot the CI, see #690

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -147,7 +147,10 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         if (!$classExists) {
             $entityPath = $this->entityClassGenerator->generateEntityClass(
                 $entityClassDetails,
-                $input->getOption('api-resource')
+                $input->getOption('api-resource'),
+                false,
+                true,
+                $input->getOption('typed')
             );
 
             $generator->writeChanges();
@@ -188,7 +191,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
             if (\is_array($newField)) {
                 $annotationOptions = $newField;
                 unset($annotationOptions['fieldName']);
-                $manipulator->addEntityField($newField['fieldName'], $annotationOptions);
+                $manipulator->addEntityField($newField['fieldName'], $annotationOptions, [], $input->getOption('typed'));
 
                 $currentFields[] = $newField['fieldName'];
             } elseif ($newField instanceof EntityRelation) {
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -83,7 +83,7 @@ final class ClassSourceManipulator
         return $this->sourceCode;
     }
 
-    public function addEntityField(string $propertyName, array $columnOptions, array $comments = [])
+    public function addEntityField(string $propertyName, array $columnOptions, array $comments = [], bool $isTyped = false)
     {
         $typeHint = $this->getEntityTypeHint($columnOptions['type']);
         $nullable = $columnOptions['nullable'] ?? false;
@@ -94,7 +94,7 @@ final class ClassSourceManipulator
         if ('array' === $typeHint) {
             $defaultValue = new Node\Expr\Array_([], ['kind' => Node\Expr\Array_::KIND_SHORT]);
         }
-        $this->addProperty($propertyName, $comments, $defaultValue);
+        $this->addProperty($propertyName, $comments, $defaultValue, $isTyped ? $typeHint : null);
 
         $this->addGetter(
             $propertyName,
@@ -305,7 +305,7 @@ final class ClassSourceManipulator
         return $this->createBlankLineNode(self::CONTEXT_CLASS_METHOD);
     }
 
-    public function addProperty(string $name, array $annotationLines = [], $defaultValue = null)
+    public function addProperty(string $name, array $annotationLines = [], $defaultValue = null, $typeHint = null)
     {
         if ($this->propertyExists($name)) {
             // we never overwrite properties
@@ -315,6 +315,11 @@ final class ClassSourceManipulator
         $newPropertyBuilder = (new Builder\Property($name))->makePrivate();
         if ($annotationLines && $this->useAnnotations) {
             $newPropertyBuilder->setDocComment($this->createDocBlock($annotationLines));
+        }
+
+        if (null !== $typeHint) {
+            $newPropertyBuilder->setType(new Node\NullableType($typeHint));
+            $newPropertyBuilder->setDefault($defaultValue);
         }
 
         if (null !== $defaultValue) {
--- a/src/Command/MakerCommand.php
+++ b/src/Command/MakerCommand.php
@@ -56,7 +56,7 @@ final class MakerCommand extends Command
 
     protected function configure()
     {
-        $this->addOption('typed', null, null, 'Type the generated properties', $this->params->get('isTyped'));
+        $this->addOption('typed', null, InputOption::VALUE_OPTIONAL, 'Type the generated properties', $this->params->get('isTyped'));
         $this->maker->configureCommand($this, $this->inputConfig);
     }
 
@@ -88,8 +88,12 @@ final class MakerCommand extends Command
             ]);
         }
 
-        if (70400 < \PHP_VERSION_ID && $input->getOption('typed')) {
-            $response = $this->io->ask('Be careful, you seem to be using a version of PHP that is too old to support property typing. Are you sure you know what you\'re doing');
+        if (null === $input->getOption('typed')) {
+            $input->setOption('typed', true);
+        }
+
+        if (\PHP_VERSION_ID < 70400 && $input->getOption('typed')) {
+            $response = $this->io->confirm('Be careful, you seem to be using a version of PHP that is too old to support property typing. Are you sure you know what you\'re doing', false);
             $input->setOption('typed', $response);
         }
 
--- a/src/Command/MakerCommand.php
+++ b/src/Command/MakerCommand.php
@@ -56,7 +56,7 @@ final class MakerCommand extends Command
 
     protected function configure()
     {
-        $this->addOption('typed', null, InputOption::VALUE_OPTIONAL, 'Type the generated properties', $this->params->get('isTyped'));
+        $this->addOption('typed', null, InputOption::VALUE_OPTIONAL, 'Type the generated properties', $this->params->get('is_typed'));
         $this->maker->configureCommand($this, $this->inputConfig);
     }
 
--- a/src/Command/MakerCommand.php
+++ b/src/Command/MakerCommand.php
@@ -56,7 +56,7 @@ final class MakerCommand extends Command
 
     protected function configure()
     {
-        $this->addOption('typed', null, InputOption::VALUE_NONE, 'Type the generated properties', $this->params->get('isTyped'));
+        $this->addOption('typed', null, null, 'Type the generated properties', $this->params->get('isTyped'));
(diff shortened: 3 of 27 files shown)


weaverryan:
Hi @Superkooka!

Ah, thank you for this! It's been on my mind lately!!! ❤️ 

2 quick things come to mind:

A) If someone is using PHP 7.4, when would they *not* want to use typed properties? I'm wondering about the option: do we need it? Or, do you "envision" the question as a "sanity" check to make sure they aren't accidentally using php 7.4 on a code base that will *actually* use 7.3 (for example) on production?

B) There are various other spots where we should also use typed properties. A simple - but probably not exhaustive - thing to check for is to search this bundle for `addEntityField()`. That method is also called in `EntityRegenerator` and `UserClassBuilder`. This is part of the reason why having an option on a command feels odd to me. We would need to add the option on other commands too. But really, you probably either want (or don't want) typed properties *globally*. So a global option (or just detecting PHP 7.4) might be better. 

Thanks!

DuboisS:
Hey @weaverryan 

Regarding behavior, what do you think of this? For all the commands which can be used to generate class properties.

1) By checking the PHP required version in composer.json, if >= 7.4, we type automatically.

2) If this PHP required version is lower than 7.4 but the  PHP interpreter version (in the path) is >= 7.4, asking for type.
If user says yes, warn him to update his composer.json with a minimum of 7.4.

3) Else, is that both versions are lower than 7.4, we don't type.


Next, yes it should concern the whole bundle and not just make:entity. What would you recommend for this at the files level ? A new "Util" file ?

Thanks ;)

src/Maker/MakeEntity.php
```diff
@@ -83,6 +83,7 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
             ->addOption('api-resource', 'a', InputOption::VALUE_NONE, 'Mark this class as an API Platform resource (expose a CRUD API for it)')
             ->addOption('regenerate', null, InputOption::VALUE_NONE, 'Instead of adding new fields, simply generate the methods (e.g. getter/setter) for existing fields')
             ->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite any existing getter/setter methods')
+            ->addOption('typed', null, InputOption::VALUE_NONE, 'Type each filed of doctrine entity class')
```
innocenzi, line 86:
```suggestion
            ->addOption('typed', null, InputOption::VALUE_NONE, 'Type the generated properties')
```

src/Maker/MakeEntity.php
```diff
@@ -107,6 +108,12 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
             return;
         }
 
+        if ($input->getOption('typed') && !version_compare(PHP_VERSION, '7.4', '>=')) {
+            $io->warning(sprintf('To use the typed properties you need a PHP version >= 7.4 but you actually are in %s', PHP_VERSION));
```
innocenzi, line 112:
```suggestion
            $io->warning(sprintf('Typed properties are available starting from PHP 7.4, but you are running %s.', PHP_VERSION));
```

src/Maker/MakeEntity.php
```diff
@@ -107,6 +108,12 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+            $question = $io->confirm('Are you sure to do that ?', false);
```
innocenzi, line 113:
```suggestion
            $question = $io->confirm('Do you still want to type the properties?', false);
```

src/Resources/help/MakeEntity.txt
```diff
@@ -17,3 +17,7 @@ for the properties of existing entities:
 You can also *overwrite* any existing methods:
 
 <info>php %command.full_name% --regenerate --overwrite</info>
+
+You can also use PHP 7.4 typing for properties
```
innocenzi, line 21:
```suggestion
You may type properties if you are running PHP 7.4 or newer:
```

src/Util/ClassSourceManipulator.php
```diff
@@ -94,7 +94,12 @@ public function addEntityField(string $propertyName, array $columnOptions, array
         if ('array' === $typeHint) {
             $defaultValue = new Node\Expr\Array_([], ['kind' => Node\Expr\Array_::KIND_SHORT]);
         }
-        $this->addProperty($propertyName, $comments, $defaultValue);
+
+        if ($isTyped) {
+            $this->addProperty($propertyName, $comments, $defaultValue, $typeHint);
+        } else {
+            $this->addProperty($propertyName, $comments, $defaultValue);
+        }
```
innocenzi, line 102: You don't need to repeat `addProperty` twice there:
```suggestion
        $this->addProperty($propertyName, $comments, $defaultValue, $isTyped ? $typeHint : null);
```

src/Util/ClassSourceManipulator.php
```diff
@@ -305,14 +310,16 @@ public function createMethodLevelBlankLine()
...
-    public function addProperty(string $name, array $annotationLines = [], $defaultValue = null)
+    public function addProperty(string $name, array $annotationLines = [], $defaultValue = null, $type = null)
     {
         if ($this->propertyExists($name)) {
             // we never overwrite properties
             return;
         }
 
         $newPropertyBuilder = (new Builder\Property($name))->makePrivate();
+        null == $type ?: $newPropertyBuilder->setType(new Node\NullableType($type));
```
innocenzi, line 321: I think it'd be less confusing to have an actual condition there, instead of that inline ternary shorthand. 

```suggestion
        if (null !== $type) {
            $newPropertyBuilder->setType(new Node\NullableType($type));
        }
```

Either way, use the strict equality operator instead of `==`:
```suggestion
        null === $type ?: $newPropertyBuilder->setType(new Node\NullableType($type));
```

tests/Maker/MakeEntityTest.php
```diff
@@ -487,6 +487,37 @@ public function getTestDetails()
...
+                ->setArgumentsString('--typed')
+                ->assert(function (string $output, string $directory) {
+                    $this->assertFileExists($directory.'/src/Entity/User.php');
+
+                    $content = file_get_contents($directory.'/src/Entity/User.php');
+                    $this->assertStringContainsString('private ?string $name;', $content);
+                    $this->assertStringContainsString('private ?\DateTimeInterface $createdAt;', $content);
+                }),
+            ];
+        }
```
innocenzi, line 519: I think you should remove the comments here, or at least put them in the same line as what they are referring to (because currently, it's confusing which line, below or above, they are referring to).

innocenzi (changes requested):
Here are a few suggestions about the code style, and a few English fixes as well :+1:

tests/Maker/MakeEntityTest.php
jrushlow:
```
[
  'value',  // comment
  'value', // comment
],
```
would be preferred as long as it doesn't trip `php_cs_fixer`

src/Util/ClassSourceManipulator.php
```diff
@@ -94,7 +94,8 @@ public function addEntityField(string $propertyName, array $columnOptions, array
         if ('array' === $typeHint) {
             $defaultValue = new Node\Expr\Array_([], ['kind' => Node\Expr\Array_::KIND_SHORT]);
         }
-        $this->addProperty($propertyName, $comments, $defaultValue);
+
+        $this->addProperty($propertyName, $comments, $defaultValue, $isTyped ? $typeHint : null);
```
jrushlow, line 98: If a property is type hinted, we should probably automatically default the value to `null` or `[]` to avoid initialization errors.
e.g.
```
private ?string $someString = null;
```

See https://rushlow.dev/blog/php-74-typed-properties-bc-problems for more information on this.

tests/Maker/MakeEntityTest.php
```diff
@@ -487,6 +487,37 @@ public function getTestDetails()
             ->configureDatabase(false),
         ];
 
+        if (version_compare(PHP_VERSION, '7.4', '>=')) {
```
OskarStark, line 490:
```suggestion
        if (\PHP_VERSION_ID >= 70400) {
```

like used in other tests

Superkooka:
To avoid all the problems with the different versions of php (<7.4) I decided to use a global flag (--typed) as well as the possibility of activating this flag by default in a configuration file as long as the version minimum PHP supported by the maker bundle will not be >= 7.4. In the following day(s) I will rebase and force push on my branch to keep a clean history of commits

Superkooka:
Hi 👋 ! I don't have the time to fix the CI, so if someone can fix it, I just have to squash the commit and this PR is ready to merge.

aleksblendwerk:
Seems the tests pass so far but the whole job was just cancelled after 10 minutes before they all finished.

Could the tests be triggered again so we can see if there's still anything actually wrong?

Superkooka:
Just look to the "CI / PHP 7.1.3 + symfony/skeleton@stable" job, the other are automatically cancel after.

aleksblendwerk:
> Just look to the "CI / PHP 7.1.3 + symfony/skeleton@stable" job, the other are automatically cancel after.

That 7.1.3 run has tests failing that are OK on 7.4 though. I doubt 7.1 is the top priority here...

Superkooka:
I think if the CI crash in 7.1, it's also crash on 7.4, the 7.1 is always supported and I don't know how to run only one test 😄

jrushlow:
Hi @Superkooka great work so far! Just a heads up in PR #736 I've created a `canUseTypedProperties(): bool` method, https://github.com/symfony/maker-bundle/blob/44278a695902c491086ecc94e5d9a2bf1f1ccd80/src/Util/PhpCompatUtil.php

This is not merged yet but should be shortly. Once it's merged, I think rather than setting a flag - you'll be able to check the newly created `$use_typed_properties` global variable that will be available when creating templates. e.g. 
image

PR #690, opened 2020-10-22
fix: adding PHP 7.4 to the CI

Superkooka:


weaverryan:
Thank you @Superkooka!

PR #692, opened 2020-10-23
Enhancement: Add PHP 7.4 stable build which is not allowed to fail

OskarStark:


weaverryan:
The problem is the MAKER_TEST_VERSION=dev should be allowed to fail.

So, best would be to make a new matrix item for just PHP 7.4, but then we take more CI resources from the rest of the symfony org.

.travis.yml
```diff
@@ -21,11 +21,13 @@ matrix:
...
         - php: 7.3
           env: MAKER_TEST_VERSION=stable
+        - php: 7.4
+          env: MAKER_TEST_VERSION=stable          
         - php: 7.4
           env: MAKER_TEST_VERSION=dev
     allow_failures:
         - php: 7.4
-          env: MAKER_TEST_VERSION=dev
+          env: MAKER_TEST_VERSION=dev
```
OskarStark, line 30:
```suggestion
          env: MAKER_TEST_VERSION=dev
```

OskarStark:
I updated the code and the PR title accordingly 👍

OskarStark:
> So, best would be to make a new matrix item for just PHP 7.4, but then we take more CI resources from the rest of the symfony org.

I will tackle this problem in #693

jrushlow (approved):

OskarStark:
Superseded by #701

PR #696, opened 2020-10-23
Fix: asHumanWords()

OskarStark:
and added test cases.

Or would you expect sth else from this method? 🤔


weaverryan:
Thank you Oskar!

PR #699, opened 2020-10-26
Allow PHP 8

fabpot:
That will help test Symfony with PHP 8. That opens up support for PHP 8 annotations in the generated code as well.

I've not updated the Travix matrix as @OskarStark is doing some work there.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "minimum-stability": "dev",
     "require": {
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

weaverryan:
Thank you Fabien! I'll tag as soon as we have the tests straightened out + the new test matrix

PR #701, opened 2020-10-26
Enhancement: Use GithubActions for tests instead of Travis + PHP 8.0 tests

OskarStark:


.github/workflows/ci.yml
```diff
@@ -8,6 +8,9 @@ on:
 
 env:
     fail-fast: true
+    PHPUNIT_FLAGS: "-v"
+    SYMFONY_PHPUNIT_DIR: "$HOME/symfony-bridge/.phpunit"
+    SYMFONY_REQUIRE: ">=3.4"
```
jderusse, line 13: Does it work without flex installed globally?

OskarStark: There is a (commented step down below)

This is based on `travis.yml`, but it fails there too, check Travis build: https://github.com/symfony/maker-bundle/pull/699

.github/workflows/ci.yml
```diff
@@ -58,3 +59,62 @@ jobs:
...
+
+            - name: Cache Composer
+              uses: actions/cache@v2.1.2
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-${{ matrix.variant }}
+                  restore-keys: ${{ runner.os }}-${{ matrix.php-version }}-composer-
+
+#            - name: Require symfony/flex
+#              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-master
```
wouterj, line 109:
```suggestion
            - name: Require symfony/flex
              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-main
```

OskarStark: Thanks, I will fix Travis in a separate PR 👍

.github/workflows/ci.yml
```diff
@@ -56,3 +69,63 @@ jobs:
...
+
+            - name: Cache Composer
+              uses: actions/cache@v2.1.2
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-${{ matrix.variant }}
+                  restore-keys: ${{ runner.os }}-${{ matrix.php-version }}-composer-
+
+            - name: Require symfony/flex
+              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-main
```
nicolas-grekas, line 120: ~dev-main~

OskarStark: You mean we should remove it entirely?

.github/workflows/ci.yml
```diff
@@ -56,3 +59,79 @@ jobs:
...
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: "Cache composer"
+              uses: actions/cache@v2.1.2
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-${{ matrix.variant }}
+                  restore-keys: ${{ runner.os }}-${{ matrix.php-version }}-composer-
+
+            - name: "Require symfony/flex:dev-main"
```
OskarStark, line 125:
```suggestion
            - name: "Require symfony/flex"
```

.github/workflows/ci.yml
Nyholm: This is currently trying to install a package called "dev-main". 

Let's remove that entirely. 

If you really really want it there, then it should be `symfony/flex:dev-main`

Nyholm (commented):
Looking at the build failure: 

```
"Creating a "symfony/skeleton:dev-main" project at "./flex_projectdev-main"

                                                                  
  [InvalidArgumentException]                                      
  Could not find package symfony/skeleton with version dev-main.  
```

It seams that the `MakerTestEnvironment::$targetFlexVersion` is used for the skeleton too.. Is that right?

src/Test/MakerTestEnvironment.php
```diff
@@ -465,7 +465,7 @@ private function getTargetFlexVersion(): string
 
                     break;
                 case 'dev':
-                    $this->targetFlexVersion = 'dev-master';
+                    $this->targetFlexVersion = 'dev-main';
```
OskarStark, line 468: This one looks wrong

.github/workflows/ci.yml
OskarStark: This part is ported from `.travis.yaml` 🤔  and this steps does not fail, I think .this one is failing:
https://github.com/symfony/maker-bundle/pull/701/files#r512590650

OskarStark:
CleanShot 2020-10-27 at 13 32 58@2x

Ok `stable` builds with composer `1` & `2` are working 🎉

.github/workflows/ci.yml
weaverryan: What is missing? The link (unfortunately) doesn't go anywhere for me.

But in general, @Nyholm is correct - what you have here is not the correct syntax for a version. I have no idea why it's not failing with a `dev-main package not found` type of error.

.github/workflows/ci.yml
```diff
@@ -56,3 +59,103 @@ jobs:
...
+                    - '7.1.33'
+                    - '7.2.5'
+                    - '7.3'
+                    - '7.4'
+                maker-test-version:
+                    - 'stable'
+                allow-failures: [false]
+                composer-version:
+                    - '1'
+                    - '2'
```
weaverryan, line 95: This really "blows up" the matrix to a lot of cases. I'm trying to think of it's worth it... it seems like it IS a good thing to have tested once, but it seems pretty unlikely that something we do in the future would break on composer 1. I'd vote for just using Composer 2. WDYT?

.github/workflows/ci.yml
```diff
@@ -56,3 +59,103 @@ jobs:
...
+#                include:
+#                    - php-version: '7.4'
+#                      maker-test-version: 'dev'
+#                      allow-failures: true
+#                    - php-version: '8.0'
+#                      maker-test-version: 'stable'
+#                      allow-failures: true
+#                    - php-version: '8.0'
+#                      maker-test-version: 'dev'
+#                      allow-failures: true
```
weaverryan, line 105: What is the plan with php 8 and `maker-test-version: dev`?

.github/workflows/ci.yml
OskarStark: Fine 👍

.github/workflows/ci.yml
```diff
@@ -56,3 +59,100 @@ jobs:
...
+    test:
+        name: "PHP ${{ matrix.php-version }} + Composer v2 + MAKER_TEST_VERSION: ${{ matrix.maker-test-version }}"
+
+        runs-on: ubuntu-latest
+
+        services:
+            mysql:
+                image: mysql:5.7
+                env:
+                    MYSQL_ROOT_PASSWORD: root
```
weaverryan, line 72: could we use `MYSQL_ALLOW_EMPTY_PASSWORD: true` and use an empty password? That would allow phpunit.xml.dist to remain the same and (selfishly) make my life simpler locally ;)

OskarStark: Should work, will try tomorrow 👍🏻

.github/workflows/ci.yml
```diff
@@ -8,6 +8,9 @@ on:
 
 env:
     fail-fast: true
+    PHPUNIT_FLAGS: "-vvv"
```
OskarStark, line 11:
```suggestion
    PHPUNIT_FLAGS: "-v"
```

.github/workflows/ci.yml
OskarStark: Looks like everything is fine now

weaverryan:
Thank you Oskar for your great work on this! php8 is still failing, but it will be for awhile until related packages add php 8 support. Not much we can do there

PR #714, opened 2020-10-28
Enhancement: Use GithubActions for tests instead of AppVeyor

OskarStark:


OskarStark:
Unfortunately this is not possible by using a MYSQL docker container as a test database, as containers are only supported on linux runners 😞

PR #719, opened 2020-10-28
Fix: Caching of composer deps

OskarStark:


.github/workflows/ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
                 uses: "actions/cache@v2"
                 with:
                     path: "${{ steps.determine-composer-cache-directory.outputs.directory }}"
-                    key: "php-${{ matrix.php-version }}-composer-${{ hashFiles('composer.lock') }}"
+                    key: "php-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}"
```
jrushlow, line 52: Wouldn't this only allow us to bust cache if the composer.json changed? I went through this a while back over in the SymfonyCasts bundles and honestly cannot remember where I landed. But do remember there was an issue using the `.json` and getting the latest version of packages for tests.

OskarStark: No, AFAIK it just restores the cached/downlaoded libraries. In the next step we call `composer install` which will get the latest and newest libs, which are not in the cache right now, because there is no `composer.lock`.

If you would have (in a preojct) a `composer.lock` file, it is clear that a following `composer install` would result in the same state as of the cache.

localheinz: Since `composer.lock` isn't checked in, this change makes sense to me!

👍

jrushlow (approved):

OskarStark:
@weaverryan can you please merge this PR? Thanks

weaverryan:
Thanks Oskar!

PR #720, opened 2020-10-28
Enhancement: Based on composer.json we should test 7.1.3

OskarStark:
instead of 7.1.33

For reference see: https://github.com/symfony/maker-bundle/blob/ffaee624a20a462ec2da82942d3c9d8fc8dd1421/composer.json#L16


jrushlow (approved):

weaverryan:
Thanks!

PR #723, opened 2020-10-30
Enhancement: Do not install mysql-client, use the already installed one

OskarStark:
THis PR

* [x] fixes the builds where `mysql-client` (sometimes) could not be installed via `apt-get`. It looks like we don't need to install it, as there is a `mysql-client` already available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,6 @@ jobs:
 
             - name: "Verify MySQL connection from host"
               run: |
-                  sudo apt-get install -y mysql-client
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -111,9 +111,12 @@ jobs:
               run: |
                   sudo /etc/init.d/mysql start
 
-            - name: "Verify MySQL connection from host"
+            - name: "Install MySQL client"
               run: |
                   sudo apt-get install -y mysql-client
```
OskarStark, line 116:
```suggestion
                  echo "foo"
```

weaverryan:
Thank Oskar!

PR #724, opened 2020-10-30
skip cs related tests in PHP 8

jrushlow:
skips styling tests in PHP 8 until cs tooling is available

### DIFF
--- a/src/Test/MakerTestCase.php
+++ b/src/Test/MakerTestCase.php
@@ -52,6 +52,10 @@ abstract class MakerTestCase extends TestCase
         foreach ($files as $file) {
             $this->assertTrue($testEnv->fileExists($file), sprintf('The file "%s" does not exist after generation', $file));
 
+            if (\PHP_VERSION_ID >= 80000) {
+                continue;
+            }
+
             if ('.php' === substr($file, -4)) {
                 $csProcess = $testEnv->runPhpCSFixer($file);
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,6 @@ jobs:
 
         env:
             SYMFONY_SKELETON_STABILITY: ${{ matrix.symfony-skeleton-stability }}
-            PHP_CS_FIXER_IGNORE_ENV: ${{ matrix.php-cs-fixer-ignore }}
 
         strategy:
             matrix:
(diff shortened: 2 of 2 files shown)


OskarStark:
I would rather let them fail until PHP 8 is stable

src/Test/MakerTestCase.php
```diff
@@ -52,6 +52,10 @@ protected function executeMakerCommand(MakerTestDetails $testDetails)
         foreach ($files as $file) {
             $this->assertTrue($testEnv->fileExists($file), sprintf('The file "%s" does not exist after generation', $file));
 
+            if (version_compare(PHP_VERSION, '8') >= 0) {
```
Superkooka, line 55:
```suggestion
            if (\PHP_VERSION_ID >= 800000)) {
```

jrushlow:
Agreed but, cs tests block actual tests. Lesser of two evils is to skip cs based tests for now so we can test the actual code base, not the styling of the code. This of course will get reverted when cs tooling is 8x compatible.

OskarStark:
I thought we did in #715 🧐

cc @weaverryan

weaverryan:
> I thought we did in #715 🧐

The difference is that we're going to skip them completely now, instead of running them and telling php-cs-fixer to stop complaining that we're using it on an unsupported version. The reason for that change is that @jrushlow in #725 adds support for php 8 attributes. That forces us to skip php-cs-fixer entirely, because I believe it will explode on the syntax or at least think it's bad cs (but I'm just assuming, I could be wrong).

@jrushlow it is true that we should "undo" the parts in #715 in the CI that set the `PHP_CS_FIXER_IGNORE_ENV` env var for php 8 - otherwise we have 2 things that basically do the same thing.

OskarStark:
@jrushlow can you try to remove the env from CI ?

jrushlow:
@weaverryan @OskarStark yes and yes.

We could leave the CI code in place for that, but ultimately we are not going to need to it with the conditional in the test suite it self to skip cs. 

For reference to self in the future - when running the test suite locally w/o this PR:
1 of many on PHP8 RC3
```
1) Symfony\Bundle\MakerBundle\Tests\Maker\MakeControllerTest::testExecute with data set "controller_basic" (Symfony\Bundle\MakerBundle\Test\MakerTestDetails Object (...))
File 'src/Controller/FooBarController.php' has a php-cs problem: PHP needs to be a minimum version of PHP 5.6.0 and maximum version of PHP 7.4.*.
```

weaverryan:
Thanks Jesse!

PR #728, opened 2020-11-08
improve php8 attribute support in makers

jrushlow:
Before we just checked the value of `\PHP_VERSION_ID` and passed a `bool` to templates that could use attributes. Now we've expended attribute support by checking if `composer.lock` has a platform override set. If not then we fall back to `\PHP_VERSION`

We've also introduced a new `PhpCompatUtil::class` & `use_attributes` is globally available to all templates.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -14,6 +14,7 @@ namespace Symfony\Bundle\MakerBundle;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\Util\ClassNameDetails;
+use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
 
 /**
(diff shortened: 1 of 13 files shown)


src/Generator.php
```diff
@@ -14,6 +14,9 @@
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\Util\ClassNameDetails;
+use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
+
+trigger_deprecation('symfony/maker-bundle', '1.25', 'Initializing Generator without providing an instance of PhpCompatUtil is deprecated.');
```
weaverryan, line 19: This should be down in the constructor and done IF `$phpCompatUtil` is null

jrushlow: done

weaverryan (approved):

weaverryan:
Thanks Jesse!

PR #730, opened 2020-11-09
[CI] disable fail fast functionality

jrushlow:
allows tests to run across all of the matrix's even if one matrix fails.


weaverryan:
Thanks Jesse!

Issue #731, opened 2020-11-09
php-cs-fixer complains of no break statement


jrushlow:
php-cs-fixer squawks that there is not a break statement when an exception is thrown in a switch statement. As a result CI fails due to:

https://github.com/symfony/maker-bundle/blob/cda232bc04d320244050acbf2368ae7c9ed96e42/src/Util/ClassSourceManipulator.php#L986

An issue has been created in the fixer repo https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5247

This issue will be resolved accordingly pending the outcome of the issues referenced above.

jrushlow:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5248 closes this issue

jrushlow:
Issues resolved in `php-cs-fixer`

PR #735, opened 2020-11-12
stop php-cs-fixer from complaining in CI

jrushlow:
Worked around for the bug in `PHP-CS-Fixer` documented in issue #731


weaverryan:
Thanks!

PR #736, opened 2020-11-13
keeping aliens at bay with maker + new security 5.2 features

jrushlow:
Hello Security51! Allows `make:auth` to take advantage of the new security features that were introduced. Starting in Symfony `5.2` when you run `make:auth` MakerBundle will automatically check if you have set:
```
security:
    enable_authenticator_manager: true
```
If so, MakerBundle will generate the required classes to authenticate users leveraging the new Authenticators.

A positive side effect of this PR, all templates can check `$use_typed_properties` to determine if the host is capable of utilizing typed properties that were introduced in PHP 7.4.

e.g. - 

```
private <?= $use_typed_properties ? 'UrlGeneratorInterface ' : null ?>$urlGenerator;
```

Internally, we've added the `PhpCompatUtil::canUseTypedProperties()` method that is called anytime a Maker needs to generate a twig template. We then inject `$use_typed_properties` into all templates so the developer can add typed properties to a generated template without having to worry about a bunch of behind the scenes logic.

### DIFF
--- a/src/Maker/MakeAuthenticator.php
+++ b/src/Maker/MakeAuthenticator.php
@@ -36,10 +36,12 @@ use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Form\Form;
 use Symfony\Component\HttpKernel\Kernel;
 use Symfony\Component\Security\Guard\Authenticator\AbstractFormLoginAuthenticator;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
 use Symfony\Component\Yaml\Yaml;
 
 /**
- * @author Ryan Weaver <ryan@knpuniversity.com>
+ * @author Ryan Weaver   <ryan@symfonycasts.com>
+ * @author Jesse Rushlow <jr@rushlow.dev>
  *
  * @internal
  */
@@ -55,6 +57,8 @@ final class MakeAuthenticator extends AbstractMaker
     private $generator;
 
     private $doctrineHelper;
+
+    private $useSecurity52 = false;
 
     public function __construct(FileManager $fileManager, SecurityConfigUpdater $configUpdater, Generator $generator, DoctrineHelper $doctrineHelper)
     {
@@ -83,6 +87,15 @@ final class MakeAuthenticator extends AbstractMaker
         }
         $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
         $securityData = $manipulator->getData();
+
+        // Determine if we should use new security features introduced in Symfony 5.2
+        if ($securityData['security']['enable_authenticator_manager'] ?? false) {
+            $this->useSecurity52 = true;
+        }
+
+        if ($this->useSecurity52 && !class_exists(UserBadge::class)) {
+            throw new RuntimeCommandException('MakerBundle does not support generating authenticators using the new authenticator system before symfony/security-bundle 5.2. Please upgrade to 5.2 and try again.');
+        }
 
         // authenticator type
         $authenticatorTypeValues = [
@@ -138,10 +151,13 @@ final class MakeAuthenticator extends AbstractMaker
         $input->setOption('firewall-name', $firewallName = $interactiveSecurityHelper->guessFirewallName($io, $securityData));
 
         $command->addOption('entry-point', null, InputOption::VALUE_OPTIONAL);
-        $input->setOption(
-            'entry-point',
-            $interactiveSecurityHelper->guessEntryPoint($io, $securityData, $input->getArgument('authenticator-class'), $firewallName)
-        );
+
+        if (!$this->useSecurity52) {
+            $input->setOption(
+                'entry-point',
+                $interactiveSecurityHelper->guessEntryPoint($io, $securityData, $input->getArgument('authenticator-class'), $firewallName)
+            );
+        }
 
         if (self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type')) {
             $command->addArgument('controller-class', InputArgument::REQUIRED);
@@ -192,13 +208,21 @@ final class MakeAuthenticator extends AbstractMaker
 
         // update security.yaml with guard config
         $securityYamlUpdated = false;
+
+        $entryPoint = $input->getOption('entry-point');
+
+        if ($this->useSecurity52 && self::AUTH_TYPE_FORM_LOGIN !== $input->getArgument('authenticator-type')) {
+            $entryPoint = false;
+        }
+
         try {
             $newYaml = $this->configUpdater->updateForAuthenticator(
                 $this->fileManager->getFileContents($path = 'config/packages/security.yaml'),
                 $input->getOption('firewall-name'),
-                $input->getOption('entry-point'),
+                $entryPoint,
                 $input->getArgument('authenticator-class'),
-                $input->hasArgument('logout-setup') ? $input->getArgument('logout-setup') : false
+                $input->hasArgument('logout-setup') ? $input->getArgument('logout-setup') : false,
+                $this->useSecurity52
             );
             $generator->dumpFile($path, $newYaml);
             $securityYamlUpdated = true;
@@ -235,10 +259,8 @@ final class MakeAuthenticator extends AbstractMaker
         if (self::AUTH_TYPE_EMPTY_AUTHENTICATOR === $authenticatorType) {
             $this->generator->generateClass(
                 $authenticatorClass,
-                'authenticator/EmptyAuthenticator.tpl.php',
-                [
-                    'provider_key_type_hint' => $this->providerKeyTypeHint(),
-                ]
+                sprintf('authenticator/%sEmptyAuthenticator.tpl.php', $this->useSecurity52 ? 'Security52' : ''),
+                ['provider_key_type_hint' => $this->providerKeyTypeHint()]
             );
 
             return;
@@ -251,12 +273,13 @@ final class MakeAuthenticator extends AbstractMaker
 
         $this->generator->generateClass(
             $authenticatorClass,
-            'authenticator/LoginFormAuthenticator.tpl.php',
+            sprintf('authenticator/%sLoginFormAuthenticator.tpl.php', $this->useSecurity52 ? 'Security52' : ''),
             [
                 'user_fully_qualified_class_name' => trim($userClassNameDetails->getFullName(), '\\'),
                 'user_class_name' => $userClassNameDetails->getShortName(),
                 'username_field' => $userNameField,
                 'username_field_label' => Str::asHumanWords($userNameField),
+                'username_field_var' => Str::asCamelCase($userNameField),
                 'user_needs_encoder' => $this->userClassHasEncoder($securityData, $userClass),
                 'user_is_entity' => $this->doctrineHelper->isClassAMappedEntity($userClass),
                 'provider_key_type_hint' => $this->providerKeyTypeHint(),
@@ -322,7 +345,8 @@ final class MakeAuthenticator extends AbstractMaker
                 'main',
                 null,
                 $authenticatorClass,
-                $logoutSetup
+                $logoutSetup,
+                $this->useSecurity52
             );
             $nextTexts[] = '- Your <info>security.yaml</info> could not be updated automatically. You\'ll need to add the following config manually:\n\n'.$yamlExample;
         }
--- a/src/Security/SecurityConfigUpdater.php
+++ b/src/Security/SecurityConfigUpdater.php
@@ -12,9 +12,13 @@
 namespace Symfony\Bundle\MakerBundle\Security;
 
 use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
+use Symfony\Component\HttpKernel\Log\Logger;
 use Symfony\Component\Security\Core\Encoder\NativePasswordEncoder;
 
 /**
+ * @author Ryan Weaver   <ryan@symfonycasts.com>
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
  * @internal
  */
 final class SecurityConfigUpdater
@@ -22,12 +26,24 @@ final class SecurityConfigUpdater
     /** @var YamlSourceManipulator */
     private $manipulator;
 
+    /** @var Logger|null */
+    private $ysmLogger;
+
+    public function __construct(Logger $ysmLogger = null)
+    {
+        $this->ysmLogger = $ysmLogger;
+    }
+
     /**
      * Updates security.yaml contents based on a new User class.
      */
     public function updateForUserClass(string $yamlSource, UserClassConfiguration $userConfig, string $userClass): string
     {
         $this->manipulator = new YamlSourceManipulator($yamlSource);
+
+        if (null !== $this->ysmLogger) {
+            $this->manipulator->setLogger($this->ysmLogger);
+        }
 
         $this->normalizeSecurityYamlFile();
 
@@ -43,9 +59,13 @@ final class SecurityConfigUpdater
         return $contents;
     }
 
-    public function updateForAuthenticator(string $yamlSource, string $firewallName, $chosenEntryPoint, string $authenticatorClass, bool $logoutSetup): string
+    public function updateForAuthenticator(string $yamlSource, string $firewallName, $chosenEntryPoint, string $authenticatorClass, bool $logoutSetup, bool $useSecurity52): string
     {
         $this->manipulator = new YamlSourceManipulator($yamlSource);
+
+        if (null !== $this->ysmLogger) {
+            $this->manipulator->setLogger($this->ysmLogger);
+        }
 
         $this->normalizeSecurityYamlFile();
 
@@ -56,23 +76,50 @@ final class SecurityConfigUpdater
         }
 
         if (!isset($newData['security']['firewalls'][$firewallName])) {
-            $newData['security']['firewalls'][$firewallName] = ['anonymous' => true];
+            if ($useSecurity52) {
+                $newData['security']['firewalls'][$firewallName] = ['lazy' => true];
+            } else {
+                $newData['security']['firewalls'][$firewallName] = ['anonymous' => 'lazy'];
+            }
         }
 
         $firewall = $newData['security']['firewalls'][$firewallName];
 
-        if (!isset($firewall['guard'])) {
-            $firewall['guard'] = [];
-        }
+        if ($useSecurity52) {
+            if (isset($firewall['custom_authenticator'])) {
+                if (\is_array($firewall['custom_authenticator'])) {
+                    $firewall['custom_authenticator'][] = $authenticatorClass;
+                } else {
+                    $stringValue = $firewall['custom_authenticator'];
+                    $firewall['custom_authenticator'] = [];
+                    $firewall['custom_authenticator'][] = $stringValue;
+                    $firewall['custom_authenticator'][] = $authenticatorClass;
+                }
+            } else {
+                $firewall['custom_authenticator'] = $authenticatorClass;
+            }
 
-        if (!isset($firewall['guard']['authenticators'])) {
-            $firewall['guard']['authenticators'] = [];
-        }
+            if (!isset($firewall['entry_point']) && $chosenEntryPoint) {
+                $firewall['entry_point_empty_line'] = $this->manipulator->createEmptyLine();
+                $firewall['entry_point_comment'] = $this->manipulator->createCommentLine(
+                    ' the entry_point start() method determines what happens when an anonymous user accesses a protected page'
+                );
+                $firewall['entry_point'] = $authenticatorClass;
+            }
+        } else {
+            if (!isset($firewall['guard'])) {
+                $firewall['guard'] = [];
+            }
 
-        $firewall['guard']['authenticators'][] = $authenticatorClass;
+            if (!isset($firewall['guard']['authenticators'])) {
+                $firewall['guard']['authenticators'] = [];
+            }
 
-        if (\count($firewall['guard']['authenticators']) > 1) {
-            $firewall['guard']['entry_point'] = $chosenEntryPoint ?? current($firewall['guard']['authenticators']);
+            $firewall['guard']['authenticators'][] = $authenticatorClass;
+
+            if (\count($firewall['guard']['authenticators']) > 1) {
+                $firewall['guard']['entry_point'] = $chosenEntryPoint ?? current($firewall['guard']['authenticators']);
+            }
         }
 
         if (!isset($firewall['logout']) && $logoutSetup) {
@@ -86,10 +133,10 @@ final class SecurityConfigUpdater
         }
 
         $newData['security']['firewalls'][$firewallName] = $firewall;
-        $this->manipulator->setData($newData);
-        $contents = $this->manipulator->getContents();
 
-        return $contents;
+        $this->manipulator->setData($newData);
+
+        return $this->manipulator->getContents();
     }
 
     private function normalizeSecurityYamlFile()
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -168,6 +168,7 @@ class Generator
 
         $variables['relative_path'] = $this->fileManager->relativizePath($targetPath);
         $variables['use_attributes'] = $this->phpCompatUtil->canUseAttributes();
+        $variables['use_typed_properties'] = $this->phpCompatUtil->canUseTypedProperties();
 
         $templatePath = $templateName;
         if (!file_exists($templatePath)) {
--- a/src/Resources/skeleton/authenticator/Security52EmptyAuthenticator.tpl.php
+++ b/src/Resources/skeleton/authenticator/Security52EmptyAuthenticator.tpl.php
@@ -1,0 +1,44 @@
+<?php echo "<?php\n" ?>
+
+namespace <?php echo $namespace ?>;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator;
+use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
+
+class <?php echo $class_name ?> extends AbstractAuthenticator
+{
+    public function supports(Request $request): ?bool
+    {
+        // TODO: Implement supports() method.
+    }
+
+    public function authenticate(Request $request): PassportInterface
+    {
+        // TODO: Implement authenticate() method.
+    }
+
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
+    {
+        // TODO: Implement onAuthenticationSuccess() method.
+    }
+
+    public function onAuthenticationFailure(Request $request, AuthenticationException $exception): ?Response
+    {
+        // TODO: Implement onAuthenticationFailure() method.
+    }
+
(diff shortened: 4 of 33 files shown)


wouterj (commented):
Looking good! Just a very minor comment for now.

You probably know already, but the `EmptyAuthenticator.tpl.php` should also get a 5.1 version.

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,80 @@
...
+        /** @var string */
+        $password = $request->get('password', '');
+
+        /** @var string|null */
+        $csrf = $request->get('_csrf_token');
+
+        $credentials = new PasswordCredentials($password);
+
+        return new Passport(
+            $user, $credentials, [
```
maxhelias, line 65: Since 5.2, we need to pass a UserBadge instead of User. Something like this
```
new UserBadge($credentials['username'], function ($username) {
    $user = $this->userProvider->loadUserByUsername($username);
    if (!$user instanceof UserInterface) {
        throw new AuthenticationServiceException('The user provider must return a UserInterface object.');
    }

    return $user;
})
```

jrushlow:
@jrushlow TODO:

1) `make:user` _INPUT_ (`User`, `yes`, `email`, `yes`)
2) `make:auth` ===

image

Gracefully handle security features not permitted w/ security 51

src/Maker/MakeAuthenticator.php
```diff
@@ -249,19 +264,35 @@ private function generateAuthenticatorClass(array $securityData, string $authent
...
+
+            if (class_exists($userRepoClassNameDetails->getFullName())) {
+                $userRepoFullyQualifiedName = trim($userRepoClassNameDetails->getFullName(), '\\');
+                $userRepoName = $userRepoClassNameDetails->getShortName();
+            }
+        }
+
+        $this->generator->generateClass($authenticatorClass, $template, [
                 'user_fully_qualified_class_name' => trim($userClassNameDetails->getFullName(), '\\'),
+                'user_repository_fully_qualified_class_name' => $userRepoFullyQualifiedName,
```
jrushlow, line 287: The idea is to use the `UserRepository::class` created by `make:user` directly instead of leaning on an `EntityManagerInterface` instance to get the user as we currently do with the "old" login form authenticator. This of course introduces its own set of challenges if a `UserRepository` (or whatever the name of the repository is) does not exist. Thoughts on moving in this direction?

Granted, I have not taken into account the new features in 5.2 that @wouterj mentioned w/ the lazy user loading at the time of this comment.

weaverryan: Yea, I would prefer to inject the repository class if there is a custom repository class. But, if we do that, we will need to:

A) Read the Doctrine metadata to see if there *is* a custom repository (and get its class name if there is one).
B) If there is *not* one, fallback to using EntityManagerInterface. 

Needing to do (B) is kind of annoying... but it also seems aggressive to throw an error if the user doesn't have a custom `UserRepository`. So, probably we should do it / try it.

wouterj: I'm not very sure about the policy of MakerBundle, but wouldn't it be possible to only support the experimental authenticator maker in Symfony 5.2? That would make this logic a lot easier, as this logic is done by the Security system. All that's needed is to pass the username to the password and configure the user provider (as is done in the "legacy" part of the maker).

jrushlow: @wouterj ya you're correct, thats what im planning on doing for 5.2. But where it gets tricky is in 5.1 - i started working on implementing this way back when and time came hard to come by. So now we're in a situation where `<= 5.0` we have the original `make:auth`, `=== 5.1` we'll have this implementation as it written now, and in `>= 5.2` we'll use those new features. Hopefully `>5.3` there wont be any significant changes.. But in the meantime I think this is where we're at with this.

when we merge this the goal is to have `make:auth` work in all 3 scenerios

wouterj: Yeah, I understand (especially the lack of time bit).

What I'm wondering is whether the `5.1` scenario should be focused upon. Or if we can skip it and support the original maker in `<=5.1` and the new experimental authenticator in `5.2`. Given that 5.2 is going to released in just 2 weeks from now, and 5.1 reaching end of life in just 2.5 months from now, is anyone going to use the maker in a Symfony 5.1 project? (open question, this is my first MakerBundle PR review, so I honestly don't know)

jrushlow: I'm with ya - I am chewing on the idea myself.. I would almost guarantee if we did skip support in 5.1, someone would open an issue in the coming months about it. That aside - 

Pro's of skipping 5.1 support:
- easier to maintain in the future
- quicker to get out the door now

Cons:
 - Loosing out on the ability to coin `Security51` (Pop culture reference to Area 51 here in the states).
 - Potentially cutting off / alienating a segment of "less experienced" users who are stuck on 5.1 for whatever reason in the future.
 - walking back the functionality that I've written to support 5.1 then saying we didn't want to support this functionality because it makes life easier.

That last 2 cons are the biggest things holding me back from coming out and saying yup, lets skip it. But that is not to say I'm not completely against the idea considering it would make life easier going forward. 

Curious as to what others think as well.

weaverryan: Supporting only 5.2 would be very cool for me

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,74 @@
...
+
+    public function authenticate(Request $request): PassportInterface
+    {
+        /** @var string $password */
+        $password = $request->request->get('password', '');
+
+        /** @var string $userIdentifier */
+        $userIdentifier = $request->request->get('<?= $username_field ?>', '');
+
+        /** @var string|null */
```
jrushlow, line 47: I'm not a fan of having to use `/** @var.... */` for these vars. But I also hate, even more, having to correct these / silence Psalm errors after running a maker. Open to ideas if anyone has a better way to clean this up & keep psalm from complaining in userland.

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,80 @@
...
+        $csrf = $request->get('_csrf_token');
+
+        $credentials = new PasswordCredentials($password);
+
+        return new Passport(
+            $user, $credentials, [
+                new PasswordUpgradeBadge($password, $this->userRepository),
+                new CsrfTokenBadge('authenticate', $csrf),
+            ]
+        );
```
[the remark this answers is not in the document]
jrushlow: done

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,74 @@
...
+        $userIdentifier = $request->request->get('<?= $username_field ?>', '');
+
+        /** @var string|null */
+        $csrf = $request->get('_csrf_token');
+
+        return new Passport(
+            new UserBadge($userIdentifier),
+            new PasswordCredentials($password),
+            [
+                new PasswordUpgradeBadge($password, $this->userRepository),
```
jrushlow, line 54:
```suggestion
                new PasswordUpgradeBadge($password),
```
- Placeholder - 
@wouterj has a PR in the works to eliminate the need to pass a `PasswordUpgraderInterface` as a require 2nd arg.

jrushlow: https://github.com/symfony/symfony/pull/39111

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
wouterj: Installing https://github.com/psalm/psalm-plugin-symfony in the project should do the trick (and if it doesn't, please open an issue or add the relevant stub).

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
wouterj: For reference, the upstream PR was just merged.

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,73 @@
...
+    public function __construct(UrlGeneratorInterface $urlGenerator)
+    {
+        $this->urlGenerator = $urlGenerator;
+    }
+
+    public function supports(Request $request): ?bool
+    {
+        return self::LOGIN_ROUTE === $request->attributes->get('_route')
+            && $request->isMethod('POST');
+    }
```
wouterj, line 36: :bulb: No review comment, but DX idea: we can make this the default implementation in `AbstractFormLoginAuthenticator`. If someone needs a different support call, they can always override but this implementation would cover like 80% of all login forms, wouldn't it?

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,73 @@
...
+        $userIdentifier = $request->request->get('<?= $username_field ?>', '');
+
+        /** @var string|null */
+        $csrf = $request->get('_csrf_token');
+
+        return new Passport(
+            new UserBadge($userIdentifier),
+            new PasswordCredentials($password),
+            [
+                new PasswordUpgradeBadge($password),
```
wouterj, line 53: :bulb:  Also no review comment: didn't we decide to automatically upgrade passwords if `UserBadge` has a password upgrading user provider and `PasswordCredentials` is given, @weaverryan?

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,73 @@
...
+        );
+    }
+
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
+    {
+        if ($targetPath = $this->getTargetPath($request->getSession(), $firewallName)) {
+            return new RedirectResponse($targetPath);
+        }
+
+        // For example : return new RedirectResponse($this->urlGenerator->generate('some_route'));
```
wouterj, line 65:
```suggestion
        // For example:
        //return new RedirectResponse($this->urlGenerator->generate('some_route'));
```
This allows for quicker uncommenting and implementation of the example code.

src/Security/SecurityConfigUpdater.php
```diff
@@ -43,7 +46,7 @@ public function updateForUserClass(string $yamlSource, UserClassConfiguration $u
         return $contents;
     }
 
-    public function updateForAuthenticator(string $yamlSource, string $firewallName, $chosenEntryPoint, string $authenticatorClass, bool $logoutSetup): string
+    public function updateForAuthenticator(string $yamlSource, string $firewallName, $chosenEntryPoint, string $authenticatorClass, bool $logoutSetup, bool $useSecurity51 = false): string
```
wouterj, line 49: Is there a need to have the default `false` value? (given that this class is internal, we don't need to care about BC here)

src/Security/SecurityConfigUpdater.php
```diff
@@ -61,18 +64,28 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
 
         $firewall = $newData['security']['firewalls'][$firewallName];
 
-        if (!isset($firewall['guard'])) {
-            $firewall['guard'] = [];
-        }
+        if ($useSecurity51) {
+            if (!isset($firewall['custom_authenticator'])) {
+                $firewall['custom_authenticator'] = $authenticatorClass;
+            }
```
wouterj, line 70: Shouldn't this append the custom authenticator to the list of custom authenticators?

E.g. if there already is one of these 2:
```yaml
security:
  firewalls:
    main:
      # ...
      custom_authenticator: 'App\Security\FirstAuthenticator'
```
```yaml
security:
  firewalls:
    main:
      # ...
      custom_authenticator:
        - 'App\Security\FirstAuthenticator'
```

I think the result should be:
```yaml
security:
  firewalls:
    main:
      # ...
      custom_authenticator:
        - 'App\Security\FirstAuthenticator'
        - 'App\Security\NewMakerAuthenticator'
```

Or is this not how other make commands behave as well? (I'm not an experienced maker bundle user)

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
jrushlow: I like it, to go one step further on that - what if we had `security.firewalls.Xmain.login.path = some_route`? this would allow us to do away with the constant and make it easier to change the login route for the majority of use cases. Thoughts?

src/Maker/MakeAuthenticator.php
```diff
@@ -84,6 +87,15 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
         $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
         $securityData = $manipulator->getData();
 
+        // Determine if we should use new security features introduced in Symfony 5.1
+        if (!empty($securityData['security']['enable_authenticator_manager']) && $securityData['security']['enable_authenticator_manager']) {
+            $this->useSecurity51 = true;
+        }
+
+        if ($this->useSecurity51 && Kernel::VERSION_ID < 50200) {
```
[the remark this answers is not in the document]
jrushlow: done

src/Security/SecurityConfigUpdater.php
jrushlow: good catch! done..

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
weaverryan: I just pinged you on this on Slack. I think it IS still required, though maybe it shouldn't be in a perfect world.

src/Security/SecurityConfigUpdater.php
weaverryan: I think Wouter is right. Should we (can't we) use `custom_authenticators`? So... we would need to check to see if either keys are configured?

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
jrushlow: no longer needed https://github.com/symfony/symfony/pull/39213 thanks for that @wouterj 

- removed badge in the template

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
jrushlow: abstract support added in https://github.com/symfony/symfony/pull/39213

- removed from template

src/Security/SecurityConfigUpdater.php
wouterj: Yes, indeed. `custom_authenticators` and `custom_authenticator` are both allowed.

jrushlow:
coding standards are failing due to unrelated `sprintf()` usage. fix is available in #757

src/Security/SecurityConfigUpdater.php
jrushlow: done

src/Maker/MakeAuthenticator.php
```diff
@@ -84,6 +88,15 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
         $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
         $securityData = $manipulator->getData();
 
+        // Determine if we should use new security features introduced in Symfony 5.1
+        if (!empty($securityData['security']['enable_authenticator_manager']) && $securityData['security']['enable_authenticator_manager']) {
+            $this->useSecurity51 = true;
```
weaverryan, line 93: We're only supporting 5.2, right? If so, let's make sure all the 51 stuff -> 52. And also, we would need some way to detect that the user is using 5.1 and yell at them that we only support 5.2.

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->urlGenerator = $urlGenerator;
+    }
+
+    public function authenticate(Request $request): PassportInterface
+    {
+        /** @var string $password */
+        $password = $request->request->get('password', '');
+
+        /** @var string $userIdentifier */
+        $userIdentifier = $request->request->get('<?= $username_field ?>', '');
```
weaverryan, line 37: if possible/reasonable, I would love to have this be called `$email = ` or `$username = ` based on whatever their field is really called.

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
```diff
@@ -0,0 +1,66 @@
...
+
+        /** @var string|null */
+        $csrf = $request->get('_csrf_token');
```
weaverryan, line 40: One thing the security system doesn't do for you currently is set the "last username" onto the session so it pre-fills after an invalid attempt. So we need to do it here. Here is the code we need - https://github.com/symfony/symfony/blob/9d40bd80f609fba8a329d22df20226fbf908ba14/src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php#L147

weaverryan: Oh, and about this line, apart from the psalm issues... which I'm not sure if we should care / if it's worth it to worry about at the expense of cluttering up the code in here with documentation - we could avoid the `$csrf` variable entirely and just pass `$request->get->...` into the badge below.

src/Security/SecurityConfigUpdater.php
```diff
@@ -55,24 +75,47 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
             $newData['security']['firewalls'] = [];
         }
 
-        if (!isset($newData['security']['firewalls'][$firewallName])) {
+        if (!isset($newData['security']['firewalls'][$firewallName]) && !$useSecurity51) {
             $newData['security']['firewalls'][$firewallName] = ['anonymous' => true];
+        } elseif (!isset($newData['security']['firewalls'][$firewallName]) && $useSecurity51) {
+            $newData['security']['firewalls'][$firewallName] = ['lazy' => true];
         }
```
weaverryan, line 82: Logically it might read better as:

```php
if (!isset($newData['security']['firewalls'][$firewallName])) {
    if ($useSecurity51) {

    } else {

    }
}
```

Also, it's minor, but I think we were already missing that `lazy` should be included in ALL firewalls, at least back to Symfony 4.4. And... I guess we should check the version of Symfony to see if `lazy` exists? Or, we (in a separate PR) could drop support for 3.4 - bump it to 4.4, as 3.4 is no longer supported.

src/Security/SecurityConfigUpdater.php
```diff
@@ -55,24 +75,47 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
...
-        if (!isset($firewall['guard']['authenticators'])) {
-            $firewall['guard']['authenticators'] = [];
-        }
+            if (!isset($firewall['custom_authenticator'])) {
+                $firewall['custom_authenticator'] = $authenticatorClass;
+            }
+
+            if (!isset($firewall['entry_point'])) {
+                $firewall['entry_point'] = $authenticatorClass;
+            }
```
weaverryan, line 104: Oh, hmm. This is not something I had thought of. I think it's not quite this simple. Hmm. Some info:

1) Not all custom authenticators provide an entrypoint. You need to implement an extra interface. Right now, our "login form" DOES, but our "blank" does not. Should that possibly be a question we ask for blank authenticators? Do you want to provide an entry point (but of course, we would need to explain what an entry point is)?

2) If the authenticator that we're generating DOES provide an entrypoint, then setting `entry_point` if it is not already set is a nice, safe thing to do. If the user did not previously have any "authentication mechanisms" (which includes things like http_basic and form_login) that provided an entry point, then setting this is technically unnecessary. But again, it seems like a safe thing to do.

3) However, if `entry_point` was NOT set... and you did already have exactly 1 previous "authentication mechanism" that provided an entry point, then this is a unique situation. By setting this, we are *changing* their entry point from the previous one to the new one. Probably... we should do this... because (for starters), we can't even know if they had previously a single auth mechanism that provided an entry point reliably... without doing some serious craziness. Anyways... this is a long way of saying... in this situation, we should STILL do exactly what you're doing here. But having a comment above `entry_point` in the config to explain things might help:

```yml
# the entry_point start() method determines what happens when an anonymous user access a protected page
```

weaverryan (commented):
Very nice work - down to some last details!

src/Maker/MakeAuthenticator.php
wouterj: `class_exists(UserBadge::class)` can be used to detect 5.2 security

src/Security/SecurityConfigUpdater.php
wouterj: `lazy: true` is introduced in 5.1, it used to be `anonymous: lazy`.

src/Security/SecurityConfigUpdater.php
```diff
@@ -55,24 +75,47 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
...
+                    $firewall['custom_authenticator'] = [];
+                    $firewall['custom_authenticator'][] = $stringValue;
+                    $firewall['custom_authenticator'][] = $authenticatorClass;
+                }
+            }
 
-        if (!isset($firewall['guard']['authenticators'])) {
-            $firewall['guard']['authenticators'] = [];
-        }
+            if (!isset($firewall['custom_authenticator'])) {
```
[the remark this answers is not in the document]
jrushlow: done

src/Security/SecurityConfigUpdater.php
jrushlow: 2 & 3 done

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
jrushlow: done

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
jrushlow: done

src/Resources/skeleton/authenticator/Security51LoginFormAuthenticator.tpl.php
jrushlow: done

src/Security/SecurityConfigUpdater.php
jrushlow: done

src/Maker/MakeAuthenticator.php
jrushlow: done

src/Security/SecurityConfigUpdater.php
jrushlow: done

src/Maker/MakeAuthenticator.php
```diff
@@ -84,6 +88,15 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
         $manipulator = new YamlSourceManipulator($this->fileManager->getFileContents($path));
         $securityData = $manipulator->getData();
 
+        // Determine if we should use new security features introduced in Symfony 5.2
+        if (!empty($securityData['security']['enable_authenticator_manager']) && $securityData['security']['enable_authenticator_manager']) {
```
weaverryan, line 92: I think we can do:

```suggestion
        if ($securityData['security']['enable_authenticator_manager']) ?? false) {
```

src/Maker/MakeAuthenticator.php
```diff
@@ -84,6 +88,15 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+            $this->useSecurity52 = true;
+        }
+
+        if ($this->useSecurity52 && !class_exists(UserBadge::class)) {
+            throw new RuntimeCommandException('Maker does not support Authenticators before symfony/security-bundle 5.2.');
```
weaverryan, line 97:
```suggestion
            throw new RuntimeCommandException('MakerBundle does not support generating authenticators using the new authenticator system before symfony/security-bundle 5.2. Please upgrade to 5.2 and try again.');
```

src/Maker/MakeAuthenticator.php
```diff
@@ -192,13 +205,21 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
 
         // update security.yaml with guard config
         $securityYamlUpdated = false;
+
+        $entryPoint = $input->getOption('entry-point');
+
+        if ($this->useSecurity52 && self::AUTH_TYPE_FORM_LOGIN !== $input->getArgument('authenticator-type')) {
+            $entryPoint = false;
+        }
+
```
weaverryan, line 214: Up in `interact()`, we still call `$interactiveSecurityHelper->guessEntryPoint()`, which still uses guard logic and may ask the user about the entry point interactively. I think we don't want that in the new system.

src/Maker/MakeAuthenticator.php
```diff
@@ -233,13 +254,15 @@ private function generateAuthenticatorClass(array $securityData, string $authent
     {
         // generate authenticator class
         if (self::AUTH_TYPE_EMPTY_AUTHENTICATOR === $authenticatorType) {
-            $this->generator->generateClass(
-                $authenticatorClass,
-                'authenticator/EmptyAuthenticator.tpl.php',
-                [
+            $template = 'authenticator/EmptyAuthenticator.tpl.php';
```
weaverryan, line 257:
```suggestion
            $template = sprintf('authenticator/%sEmptyAuthenticator.tpl.php', $this-> useSecurity52 ? 'Security52' : '');
```

then no if below

weaverryan: Hmm, in fact, I'd inline this into the `generateClass` 2nd argument.

src/Security/SecurityConfigUpdater.php
```diff
@@ -56,23 +76,50 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
...
+                }
+            } else {
+                $firewall['custom_authenticator'] = $authenticatorClass;
+            }
 
-        if (!isset($firewall['guard']['authenticators'])) {
-            $firewall['guard']['authenticators'] = [];
-        }
+            if (!isset($firewall['entry_point']) && $chosenEntryPoint) {
+                $firewall['entry_point_mt_line'] = $this->manipulator->createEmptyLine();
```
weaverryan, line 103:
```suggestion
                $firewall['entry_point_empty_line'] = $this->manipulator->createEmptyLine();
```

I understood it, but that'll be less confusing :)

weaverryan (commented):
We're very close now! Nice work on this complex feature!

src/Maker/MakeAuthenticator.php
jrushlow: done

src/Maker/MakeAuthenticator.php
jrushlow: done

weaverryan:
Thanks for the hard work Jesse!

wouterj:
Great!!! Thanks @jrushlow and @weaverryan!

Issue #625, opened 2020-05-28
MakerBundle does not work on SF 3.4 (  Attempted to call an undefined method named "setHref" of class "Symfony\Component\Console\Formatter\OutputFormatterStyle".)


lyrixx:
On a fresh sf 3.4 install:

```
>/home/gregoire/dev/tmp/symfony-36852(master %) bin/console make:form -v 

 The name of the form class (e.g. GentleJellybeanType):
 > Test 

 The name of Entity or fully qualified model class name that the new form will be bound to (empty for none):
 > 


In MakerFileLinkFormatter.php line 50:
                                                                                                                              
  [Symfony\Component\Debug\Exception\UndefinedMethodException]                                                                
  Attempted to call an undefined method named "setHref" of class "Symfony\Component\Console\Formatter\OutputFormatterStyle".  
                                                                                                                              

Exception trace:
  at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/maker-bundle/src/Util/MakerFileLinkFormatter.php:50
 Symfony\Bundle\MakerBundle\Util\MakerFileLinkFormatter->makeLinkedPath() at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/maker-bundle/src/FileManager.php:84
 Symfony\Bundle\MakerBundle\FileManager->dumpFile() at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/maker-bundle/src/Generator.php:195
 Symfony\Bundle\MakerBundle\Generator->writeChanges() at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/maker-bundle/src/Maker/MakeForm.php:114
 Symfony\Bundle\MakerBundle\Maker\MakeForm->generate() at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/maker-bundle/src/Command/MakerCommand.php:104
 Symfony\Bundle\MakerBundle\Command\MakerCommand->execute() at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/console/Application.php:1005
 Symfony\Component\Console\Application->doRunCommand() at /home/gregoire/dev/tmp/symfony-36852/vendor/symfony/framework-bundle/Console/Application.php:86
[...]

PR #740, opened 2020-11-17
Update MakerFileLinkFormatter.php

ehrlich-media:
I checked https://github.com/symfony/console/tree/3.4 
The Method does not exist in any Version of "symfony/console". Why is it calling here?
$outputFormatterStyle->setHref($formatted);

Maybe not nesseary?

### DIFF
--- a/src/Util/MakerFileLinkFormatter.php
+++ b/src/Util/MakerFileLinkFormatter.php
@@ -47,7 +47,6 @@ final class MakerFileLinkFormatter
         }
 
         $outputFormatterStyle = new OutputFormatterStyle();
-        $outputFormatterStyle->setHref($formatted);
 
         return $outputFormatterStyle->apply($relativePath);
     }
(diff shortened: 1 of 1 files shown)


ehrlich-media:
See here:
https://github.com/symfony/maker-bundle/issues/625

weaverryan:
It looks like we need to check if this method exists and *only* call it if it does. So, you're right - but we shouldn't remove the call entirely, but only call it if it exists - it's used to make clickable links. I'm not sure why our tests didn't catch this problem on 3.4...

src/Util/MakerFileLinkFormatter.php
```diff
@@ -47,7 +47,6 @@ public function makeLinkedPath(string $absolutePath, string $relativePath): stri
         }
 
         $outputFormatterStyle = new OutputFormatterStyle();
```
jrushlow, line 49:
```suggestion
        $outputFormatterStyle = new OutputFormatterStyle();
        
        if (method_exists(OutputFormatterStyle::class, 'setHref')) {
            $outputFormatterStyle->setHref($formatted);
        }
```

jrushlow (changes requested):

jrushlow:
closes #625

Superkooka (approved):

lyrixx (approved):

lyrixx:
Thank you @ehrlich-media.

PR #746, opened 2020-11-25
Allow released versions of the deprecation contract

stof:
This ensures that the package can be installed with `stable` as minimum stability.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "symfony/config": "^3.4|^4.0|^5.0",
         "symfony/console": "^3.4|^4.0|^5.0",
         "symfony/dependency-injection": "^3.4|^4.0|^5.0",
-        "symfony/deprecation-contracts": "^2.3",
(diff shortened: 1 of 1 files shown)


stof:
The cause of introducing a faulty constraint is probably that the contributor adding it in https://github.com/symfony/maker-bundle/pull/728 ran `composer require symfony/deprecation-contract`, letting Composer guess the constraint. But this guessing respect the minimum-stability of the project, and this project defines `dev` as the min stability.

jrushlow:
> contributor adding it in #728 ran composer require symfony/deprecation-contracts
That's exactly what happened. Good catch!

stof:
> Good catch!

Well, that was actually catched because someone complain on Slack that `composer require symfony/maker-bundle` was broken.

weaverryan:
Thank you for the quick PR Stof!

weaverryan:
Tagged! I think we could bump even lower to `^2.1`, right?

PR #747, opened 2020-11-25
lower symfony/contracts dependency version a tad bit more

jrushlow:
PR #728 introduced `symfony/deprecations-contracts` at version `^2.3` which prevented some users from using maker bundle.

PR #746 fixed the reported issue by lowering the version constraint to `^2.2`. But in reality we can lower it even further to `2.1` as
this is when `symfony/deprecation-contracts` was introduced.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "symfony/config": "^3.4|^4.0|^5.0",
         "symfony/console": "^3.4|^4.0|^5.0",
         "symfony/dependency-injection": "^3.4|^4.0|^5.0",
-        "symfony/deprecation-contracts": "^2.2",
+        "symfony/deprecation-contracts": "^2.1",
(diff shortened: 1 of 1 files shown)


jrushlow:
@stof think we can go even lower w/ this constraint. Only thought I have - will going lower than `2.2` have any negative affects for PHP8 compatibility? I cant think of anything off the top of my head as composer should take care of the versions needed respectively.

stof:
Well, support for PHP 8 exists only in 2.2+ of the contracts. I think it is fine to depend on 2.2+. There is nothing forbidding to upgrade from 2.1 to 2.2 in a project (the only blockage to upgrade to the previously required 2.3 is that it does not exist yet...)

jrushlow:
good point, thanks for the feedback @stof

PR #755, opened 2020-12-02
fix(make:entity): handle new value if nullable in OneToOne association, close #195

### LINKED ISSUE #195: Make:entity --regenerate : OneToOne relationship bug in the setter
Hello,

I have one entity Organization with a team property
```
class RefOrganization

    /**
     * One Organization has One team
     * @ORM\OneToOne(targetEntity="RefTeam", mappedBy="organization")
     */
    private $team;
```

I have a second entity Team with a Organization property
```
class RefTeam

    /**
     * One Team has one Organization
     * @ORM\OneToOne(targetEntity="RefOrganization", inversedBy="team")
     * @ORM\JoinColumn(name="ORGANIZATION_ID", referencedColumnName="ID", nullable=true)
     */
    private $organization;
```
With the following code, I got a OneToOne relationship between both entities.
However, when lauching the Make:Entity --regenerate to get back all setters / getters, 

In the RefOrganization entity, it generates the following setter for the team :

```
    public function setTeam(?RefTeam $team): self
    {
        $this->team = $team;

        // set (or unset) the owning side of the relation if necessary
        $newOrganization = $team === null ? null : $this;
        if ($newOrganization !== $team->getOrganization()) {
            $team->setOrganization($newOrganization);
        }

        return $this;
    }
```

When creating the RefOrganization without team (as it's nullable), we have an error because of the : 

```
if ($newOrganization !== $team->getOrganization()) 
```
call method getOrganization of null object 

Regards

Kocal:
Hi :wave: 

This PR is a proposal for #195 for handling the new value (if nullable) in a OneToOne association.

The following code was not valid, it fails at `$user->getUserProfile()` if we call `UserProfile#setUser(null)`:
```php
class UserProfile
{
    // ...
    public function setUser(?User $user): self
    {
        $this->user = $user;

        // set (or unset) the owning side of the relation if necessary
        $newUserProfile = null === $user ? null : $this;
        if ($user->getUserProfile() !== $newUserProfile) {
            $user->setUserProfile($newUserProfile);
        }

        return $this;
    }
}
```

The following code takes care of:
- unset the user profile of the previous User (`$this->user`) if we assign a nullable User
- set the profile to the new user, and prevent an infinite loop with `user->getUserProfile() !== $this`
```php
class UserProfile 
{
    public function setUser(?User $user): self
    {
        // unset the owning side of the relation if necessary
        if ($user === null && $this->user !== null) {
            $this->user->setUserProfile(null);
        }

        // set the owning side of the relation if necessary
        if ($user !== null && $user->getUserProfile() !== $this) {
            $user->setUserProfile($this);
        }

        $this->user = $user;

        return $this;
    }
}
```

WDYT? Thanks!

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -232,6 +232,12 @@ final class ClassSourceManipulator
     public function addSetter(string $propertyName, $type, bool $isNullable, array $commentLines = [])
     {
         $builder = $this->createSetterNodeBuilder($propertyName, $type, $isNullable, $commentLines);
+        $builder->addStmt(
+            new Node\Stmt\Expression(new Node\Expr\Assign(
+                new Node\Expr\PropertyFetch(new Node\Expr\Variable('this'), $propertyName),
+                new Node\Expr\Variable($propertyName)
+            ))
+        );
         $this->makeMethodFluent($builder);
         $this->addMethod($builder->getNode());
     }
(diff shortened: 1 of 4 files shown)


weaverryan:
Wonderful work! Thanks for this great fix Hugo!

PR #757, opened 2020-12-04
fix incorrect usage of sprintf

jrushlow:
php-cs-fixer released an update which catches useless `sprintf()` calls. A couple were found and fixed.

refs https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5280

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -298,7 +298,7 @@ final class MakeRegistrationForm extends AbstractMaker
                 UniqueEntity::class,
                 [
                     'fields' => [$usernameField],
-                    'message' => sprintf('There is already an account with this '.$usernameField),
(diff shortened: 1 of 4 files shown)


src/Maker/MakeRegistrationForm.php
```diff
@@ -298,7 +298,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
                 UniqueEntity::class,
                 [
                     'fields' => [$usernameField],
-                    'message' => sprintf('There is already an account with this '.$usernameField),
+                    'message' => sprintf('There is already an account with this %s', $usernameField),
```
OskarStark, line 301: The message sounds weird

jrushlow: you not wrong. it made sense at the time...

weaverryan:
Thanks Jesse!

PR #760, opened 2020-12-04
only use attributes if PHP >= 8 && Symfony >= 5.2

### LINKED ISSUE #750: make controller generates controller with attributes but symfony does not support it yet
Symfony 5.2 will support attributes for routes, but it's not the case for Symfony 5.1. I just made a fresh install and used the maker, the generated route was the following:

```php
#[Route('/main', name: 'main')]
public function index(): Response
{
       
}
```

This doesn't work. It seems critical but I didn't see any issue in that way so here it is.

- PHP version: 8.0
- Bundle version: 1.24.2

jrushlow:
fixes #750

### DIFF
--- a/src/Util/PhpCompatUtil.php
+++ b/src/Util/PhpCompatUtil.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Util;
 
 use Symfony\Bundle\MakerBundle\FileManager;
+use Symfony\Component\HttpKernel\Kernel;
 
 /**
  * @author Jesse Rushlow <jr@rushlow.dev>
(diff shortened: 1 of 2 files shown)


src/Util/PhpCompatUtil.php
```diff
@@ -32,7 +33,7 @@ public function canUseAttributes(): bool
     {
         $version = $this->getPhpVersion();
 
-        return version_compare($version, '8alpha', '>=');
+        return version_compare($version, '8alpha', '>=') && Kernel::VERSION_ID >= 50200;
```
weaverryan, line 36: We should probably make this method more specific. This is really about *routing* attributes. But soon we will probably start using Doctrine attributes and then other things. Possibly we could rename this to `canUseRouteAttributes()` and add other methods later.

weaverryan:
I did leave one comment I think we should address later, but as this is internal and fixes a bug, merging now. Thanks!

PR #766, opened 2020-12-13
Update LoginFormAuthenticator.tpl.php

sebheitzmann:
I think that we should always load the user with the UserProvider even if we could load it with the doctrine repo.


weaverryan:
Hi @sebheitzmann!

I think your PR is one that many people would support and is sensible. But, I actually don't think this is a good idea. The reason is that the user provider works great, until someone needs to make a more complex query. If we generate the user provider code, 90% of users are happy and never need to think about it. But the other 10% of users - that need to do some extra query to fetch the user (e.g. `WHERE email = '...' AND company_id='value-from-a-drop-down-on-the-login-form'`) - will have a harder time knowing how to do that. For users that know what they're doing, the user provider is great. Generating it all the time, however, makes life harder for beginners :).

Cheers!

PR #770, opened 2020-12-15
fix invalid expiresAt values for ResetPassword & VerifyEmail templates

### LINKED ISSUE #727: ResetPassword : error tokenLifetime display
Hi,

Display of tokenLifetime value is incorrect.
Example: when lifetime = 7200, tokenLifetime display 3 hours in the email instead of 2 hours.

https://github.com/symfony/maker-bundle/blob/f1093e98de2ac9816ec336fba68367da377b2677/src/Resources/skeleton/resetPassword/twig_check_email.tpl.php#L6

https://github.com/symfony/maker-bundle/blob/f1093e98de2ac9816ec336fba68367da377b2677/src/Resources/skeleton/resetPassword/twig_email.tpl.php#L7

jrushlow:
Depends on:
- [x] https://github.com/SymfonyCasts/verify-email-bundle/pull/43
- [x] https://github.com/SymfonyCasts/reset-password-bundle/pull/135

closes #727

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -38,6 +38,7 @@ use Symfony\Component\Form\Extension\Core\Type\PasswordType;
 use Symfony\Component\Mailer\MailerInterface;
 use Symfony\Component\Routing\RouterInterface;
 use Symfony\Component\Validator\Validation;
+use SymfonyCasts\Bundle\VerifyEmail\Model\VerifyEmailSignatureComponents;
 use SymfonyCasts\Bundle\VerifyEmail\SymfonyCastsVerifyEmailBundle;
 
 /**
@@ -368,7 +369,15 @@ final class MakeRegistrationForm extends AbstractMaker
         $missing = false;
         $composerMessage = 'composer require';
 
-        if (!class_exists(SymfonyCastsVerifyEmailBundle::class)) {
+        // verify-email-bundle 1.1.1 includes support for translations and a fix for the bad expiration time bug.
+        // we need to check that if the bundle is installed, it is version 1.1.1 or greater
+        if (class_exists(SymfonyCastsVerifyEmailBundle::class)) {
+            $reflectedComponents = new \ReflectionClass(VerifyEmailSignatureComponents::class);
+
+            if (!$reflectedComponents->hasMethod('getExpirationMessageKey')) {
+                throw new RuntimeCommandException('Please upgrade symfonycasts/verify-email-bundle to version 1.1.1 or greater.');
+            }
+        } else {
             $missing = true;
             $composerMessage = sprintf('%s symfonycasts/verify-email-bundle', $composerMessage);
         }
--- a/src/Resources/skeleton/registration/twig_email.tpl.php
+++ b/src/Resources/skeleton/registration/twig_email.tpl.php
@@ -3,7 +3,7 @@
 <p>
     Please confirm your email address by clicking the following link: <br><br>
     <a href="{{ signedUrl|raw }}">Confirm my Email</a>.
-    This link will expire in {{ expiresAt|date('g') }} hour(s).
+    This link will expire in {{ expiresAtMessageKey|trans(expiresAtMessageData, 'VerifyEmailBundle') }}.
 </p>
 
 <p>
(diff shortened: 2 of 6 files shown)


src/Resources/skeleton/resetPassword/twig_email.tpl.php
```diff
@@ -4,6 +4,6 @@
 
 <a href="{{ url('app_reset_password', {token: resetToken.token}) }}">{{ url('app_reset_password', {token: resetToken.token}) }}</a>
 
-<p>This link will expire in {{ tokenLifetime|date('g') }} hour(s).</p>
+<p>This link will expire in {{ resetToken.getExpirationMessageKey()|trans(resetToken.getExpirationMessageData(), 'ResetPasswordBundle') }}.</p>
```
weaverryan, line 7: Fun fact: we need to check the version of the bundle in question (e.g. reset-password here) and either:

A) Toggle 2 "modes" of generated code based on the version
B) Throw an exception and tell people to upgrade to version 1.WHATEVER

Option (B) seems harsh, but I think it's probably a good idea. It's an opportunity for us to notify them that a newer version is available that fixes a bug we know about. And, it will probably be rare (as users will typically install the bundle right before running this command.

src/Resources/skeleton/registration/twig_email.tpl.php
```diff
@@ -3,7 +3,7 @@
 <p>
     Please confirm your email address by clicking the following link: <br><br>
     <a href="{{ signedUrl|raw }}">Confirm my Email</a>.
-    This link will expire in {{ expiresAt|date('g') }} hour(s).
+    This link will expire in {{ expiresAtMessageKey|trans(expiresAtMessageData, 'VerifyEmailBundle') }}.
```
weaverryan, line 6: iirc, we don't require the bundle to be installed before running this command. And so, the best we could do is detect if verifier is installed AND if it's out-of-date, and throw an exception to tell the user to upgrade. It's probably an edge-case though...

src/Resources/skeleton/resetPassword/twig_email.tpl.php
weaverryan:
```suggestion
<p>This link will expire in {{ resetToken.expirationMessageKey|trans(resetToken.expirationMessageData, 'ResetPasswordBundle') }}.</p>
```

src/Resources/skeleton/registration/twig_email.tpl.php
jrushlow: done

src/Resources/skeleton/resetPassword/twig_email.tpl.php
jrushlow: done

src/Maker/MakeRegistrationForm.php
```diff
@@ -368,7 +369,15 @@ private function getMissingComponentsComposerMessage(): ?string
...
 
-        if (!class_exists(SymfonyCastsVerifyEmailBundle::class)) {
+        // verify-email-bundle 1.2 includes support for translations and a fix for the bad expiration time bug.
+        // we need to check that if the bundle is installed, it is version 1.2 or greater
+        if (class_exists(SymfonyCastsVerifyEmailBundle::class)) {
+            $reflectedComponents = new \ReflectionClass(VerifyEmailSignatureComponents::class);
+
+            if (!$reflectedComponents->hasMethod('getExpirationMessageKey')) {
+                throw new RuntimeCommandException('Please upgrade symfonycasts/verify-email-bundle to version 1.2.0 or greater.');
+            }
```
weaverryan, line 379: Hmm, I think this would cause the entire commend to explode *right* at the end of it working. I didn't really think about that. Probably we should just return a strongly-worded message here to show to the user (instead of `composer require symfonycasts/..`) or we need to do this check *right* when the command is starting, before we ask any questions (the idea would be that, if they have verifier already installed, we will assume they probably ARE going to generate with email verification, so we'll stop them immediately and tell them to upgrade).

jrushlow: Currently the very first question that gets asked is "Do you want to verify emails?..." (paraphrasing) - then internally we:

- check if the bundle is installed. If it is not we: (Original Logic)
image

- If the bundle is installed but does not have the new trans + expiresAt fix we :
image

The latter is how we are handling it in `make:reset`. The same type of exception + "halt" is done if `reset-password-bundle` is not installed and you try to run the `make:reset-password` command.

For verify, I think the "exception + halt" strategy would be best if someone runs `make:registration-form` -> `y` to the verify email question while already having the old bundle installed.

jrushlow: For clarity in the future - the above screen shot shows `...version 1.2.0...` the actual release is `1.1.1`

weaverryan:
Thanks Jesse!

PR #771, opened 2020-12-17
[make:auth] Fixed newline characters

wouterj:
The wrong quotes were used, causing `\n\n` to be printed to the terminal (instead of real newlines)


jrushlow (approved):
good catch!

weaverryan:
Thanks Wouter!

PR #773, opened 2020-12-31
feat: adding typing PHP 7.4

Superkooka:
Now the maker-bundle type the properties of class. By default is desactivate, of course I can just check the version of PHP but the person can use PHP 7.4 on his "php path" but work on PHP 7.3 and 7.3 dosen't support typing. It's why you can use the flag "--typed" globaly on all command or use a configuration key to activate it by default on the projet. 

Exemple of entity generated with typing: 
```php
<?php

namespace App\Entity;

use App\Repository\EntityEtGrominetRepository;

/**
 * @ORM\Entity(repositoryClass=EntityEtGrominetRepository::class)
 */
class EntityEtGrominet
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private ?int $id;

    /**
     * @ORM\Column(type="string", length=255)
     */
    private ?string $name = null;

    /**
     * @ORM\Column(type="datetime")
     */
    private ?\DateTimeInterface $createdAt = null;

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getName(): ?string
    {
        return $this->name;
    }

    public function setName(string $name): self
    {
        $this->name = $name;

        return $this;
    }

    public function getCreatedAt(): ?\DateTimeInterface
    {
        return $this->createdAt;
    }

    public function setCreatedAt(\DateTimeInterface $createdAt): self
    {
        $this->createdAt = $createdAt;

        return $this;
    }
}

```

In the configuration file, to activate typing by default

```yml
maker:
  is_typed: true
```

~~At the moment, one test (and only one :/) dosn't pass, I don't know why. If someone can looking It's the MakeRegistrationFormTest with the data set "verify_email_functional_test". For more information run the test localy or look at gh action :grin:~~

I've close the #689 PR to make a more clean PR

### DIFF
--- a/src/Maker/MakeResetPassword.php
+++ b/src/Maker/MakeResetPassword.php
@@ -26,6 +26,7 @@ use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Security\InteractiveSecurityHelper;
 use Symfony\Bundle\MakerBundle\Util\ClassNameDetails;
 use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
+use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
 use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
 use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Command\Command;
@@ -54,6 +55,7 @@ class MakeResetPassword extends AbstractMaker
     private $fileManager;
     private $doctrineHelper;
     private $entityClassGenerator;
+    private $phpCompatUtil;
 
     private $fromEmailAddress;
     private $fromEmailName;
@@ -68,6 +70,7 @@ class MakeResetPassword extends AbstractMaker
         $this->fileManager = $fileManager;
         $this->doctrineHelper = $doctrineHelper;
         $this->entityClassGenerator = $entityClassGenerator;
+        $this->phpCompatUtil = new PhpCompatUtil($fileManager);
     }
 
     public static function getCommandName(): string
@@ -326,7 +329,8 @@ class MakeResetPassword extends AbstractMaker
         $generator->writeChanges();
 
         $manipulator = new ClassSourceManipulator(
-            $this->fileManager->getFileContents($requestEntityPath)
+            $this->fileManager->getFileContents($requestEntityPath),
+            $this->phpCompatUtil
         );
 
         $manipulator->addInterface(ResetPasswordRequestInterface::class);
@@ -369,7 +373,8 @@ CODE
         );
 
         $manipulator = new ClassSourceManipulator(
-            $this->fileManager->getFileContents($pathRequestRepository)
+            $this->fileManager->getFileContents($pathRequestRepository),
+            $this->phpCompatUtil
         );
 
         $manipulator->addInterface(ResetPasswordRequestRepositoryInterface::class);
--- a/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
+++ b/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
@@ -29,7 +29,7 @@ class <?= $class_name ?> extends AbstractController
 {
     use ResetPasswordControllerTrait;
 
-    private $resetPasswordHelper;
+    private <?= $use_typed_properties ? "ResetPasswordHelperInterface " : null ?>$resetPasswordHelper;
 
     public function __construct(ResetPasswordHelperInterface $resetPasswordHelper)
     {
--- a/src/Doctrine/EntityRegenerator.php
+++ b/src/Doctrine/EntityRegenerator.php
@@ -19,6 +19,7 @@ use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
+use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
 
 /**
  * @internal
@@ -30,6 +31,7 @@ final class EntityRegenerator
     private $generator;
     private $entityClassGenerator;
     private $overwrite;
+    private $phpCompatUtil;
 
     public function __construct(DoctrineHelper $doctrineHelper, FileManager $fileManager, Generator $generator, EntityClassGenerator $entityClassGenerator, bool $overwrite)
     {
@@ -38,6 +40,7 @@ final class EntityRegenerator
         $this->generator = $generator;
         $this->entityClassGenerator = $entityClassGenerator;
         $this->overwrite = $overwrite;
+        $this->phpCompatUtil = new PhpCompatUtil($fileManager);
     }
 
     public function regenerateEntities(string $classOrNamespace)
@@ -201,6 +204,7 @@ final class EntityRegenerator
     {
         return new ClassSourceManipulator(
             $this->fileManager->getFileContents($classPath),
+            $this->phpCompatUtil,
             $this->overwrite,
             // use annotations
             // if properties need to be generated then, by definition,
--- a/src/Maker/MakeAuthenticator.php
+++ b/src/Maker/MakeAuthenticator.php
@@ -23,6 +23,7 @@ use Symfony\Bundle\MakerBundle\Security\SecurityConfigUpdater;
 use Symfony\Bundle\MakerBundle\Security\SecurityControllerBuilder;
 use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
+use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
 use Symfony\Bundle\MakerBundle\Util\YamlManipulationFailedException;
 use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
 use Symfony\Bundle\MakerBundle\Validator;
@@ -60,12 +61,15 @@ final class MakeAuthenticator extends AbstractMaker
 
     private $useSecurity52 = false;
 
+    private $phpCompatUtil;
+
     public function __construct(FileManager $fileManager, SecurityConfigUpdater $configUpdater, Generator $generator, DoctrineHelper $doctrineHelper)
     {
         $this->fileManager = $fileManager;
         $this->configUpdater = $configUpdater;
         $this->generator = $generator;
         $this->doctrineHelper = $doctrineHelper;
+        $this->phpCompatUtil = new PhpCompatUtil($fileManager);
     }
 
     public static function getCommandName(): string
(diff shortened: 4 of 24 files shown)


weaverryan:
Thanks for this @Superkooka - I'm super excited for it!

It looks like the only big decision point is about the `--typed` option and the `is_typed:` config flag. Ideally, I'd like to avoid both, but I am not stubbornly stuck in that position ;). It seems to me there are 2 reasons to make it configurable:

> 1) To avoid someone using PHP 7.4/8... but on a project that actually uses PHP 7.3 on production.

We already read the `config.platform.php` from `composer.json`. If a user does not have this set, and they are using PHP 7.4/8 on their project, then they will likely already have problems (e.g. Composer may download dependencies that are too new). So I think this is a bug in the user's app - and MakerBundle would be behaving identically to Composer.

> 2) The project is on PHP 7.4 or higher, but the user simply does not prefer to used typed properties.

This might be valid? Is there any reason that someone might *not* want typed properties?

Thanks!

src/DependencyInjection/Configuration.php
```diff
@@ -32,6 +32,7 @@ public function getConfigTreeBuilder()
         $rootNode
             ->children()
                 ->scalarNode('root_namespace')->defaultValue('App')->end()
+                ->booleanNode('is_typed')->defaultFalse()->end()
```
OskarStark, line 35: I would prefer something more meaningful:
```suggestion
                ->booleanNode('typed_properties')->defaultFalse()->end()
```

nicolas-grekas:
Can't we read the composer.json file and figure out this way? Removing the option would be nice I agree.

src/Doctrine/EntityClassGenerator.php
```diff
@@ -31,7 +31,7 @@ public function __construct(Generator $generator, DoctrineHelper $doctrineHelper
         $this->doctrineHelper = $doctrineHelper;
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $isTyped = false): string
```
OskarStark, line 34: Wouldn't it be worth to create some `Config` object which holds such values instead of injecting them all over the place? It increases complexity a lot

jrushlow:
im think because we already check if we can use typed properties and have a global template var https://github.com/symfony/maker-bundle/blob/0f1d3ed2584349dc8700d7908e8a92b3742b1c99/src/Generator.php#L171 that tells us - we shouldnt need a flag or config param for this.. i can however see an argument that "what if i dont want typed properties..." then setting an optional config `uses_typed_properties: false` or something along those lines would make sense. 

but i still cant think of a valid use case where one would not want typed properties if PHP supports it.

OskarStark:
> but i still cant think of a valid use case where one would not want typed properties if PHP supports it.

Only for consistency across the code base

nicolas-grekas:
> but i still cant think of a valid use case where one would not want typed properties if PHP supports it.

actually, there are security reasons to use typed props. By default we should encourage ppl to use them. If they really don't want to, they should just edit the generated code.

src/Doctrine/EntityClassGenerator.php
jrushlow: we should only need to check `$uses_typed_properties` in the templates with a conditional for this. see https://github.com/symfony/maker-bundle/blob/0f1d3ed2584349dc8700d7908e8a92b3742b1c99/src/Resources/skeleton/authenticator/Security52LoginFormAuthenticator.tpl.php#L25

src/Resources/skeleton/authenticator/Security52LoginFormAuthenticator.tpl.php
```diff
@@ -22,7 +22,7 @@ class <?= $class_name; ?> extends AbstractLoginFormAuthenticator
 
     public const LOGIN_ROUTE = 'app_login';
 
-    private <?= $use_typed_properties ? 'UrlGeneratorInterface ' : null ?>$urlGenerator;
+    private <?= $is_typed ? 'UrlGeneratorInterface ' : null ?>$urlGenerator;
```
jrushlow, line 25: because `$use_typed_properties` is a global var that is available in all templates, we should use that to check if properties can be typed.

jrushlow (changes requested):

src/Doctrine/EntityClassGenerator.php
Superkooka: Thanks @OskarStark, it's not a bad idea but in my opinion, a simple ParameterBag is sufficient for 2 parameters

Superkooka:
> Thanks for this @Superkooka - I'm super excited for it!
> 
> It looks like the only big decision point is about the `--typed` option and the `is_typed:` config flag. Ideally, I'd like to avoid both, but I am not stubbornly stuck in that position ;). It seems to me there are 2 reasons to make it configurable:
> 
> > 1. To avoid someone using PHP 7.4/8... but on a project that actually uses PHP 7.3 on production.
> 
> We already read the `config.platform.php` from `composer.json`. If a user does not have this set, and they are using PHP 7.4/8 on their project, then they will likely already have problems (e.g. Composer may download dependencies that are too new). So I think this is a bug in the user's app - and MakerBundle would be behaving identically to Composer.
> 
> > 1. The project is on PHP 7.4 or higher, but the user simply does not prefer to used typed properties.
> 
> This might be valid? Is there any reason that someone might _not_ want typed properties?
> 
> Thanks!

Yes I'm totaly agree with you. I think the user need to define if the projet is typed one time so the configuration key, in my opinion is the most appropriate. With ``05190b2`` only configuration key is supported 😄

src/DependencyInjection/Configuration.php
```diff
@@ -32,6 +32,7 @@ public function getConfigTreeBuilder()
         $rootNode
             ->children()
                 ->scalarNode('root_namespace')->defaultValue('App')->end()
+                ->booleanNode('typed_properties')->defaultFalse()->end()
```
weaverryan, line 35: I think this should default to `null`, meaning that the user has no "preference". Then, we should inject this into the `PhpCompatUtil` class and use it in `canUseTypedProperties`. Basically, if this option is `null`, then that method will do what it does now: return true if php 7.4 or higher is used. If this option IS passed (either as true or false), then that value would be returned directly from that method.

Then we can continue using the `canUseTypedProperties()` method everywhere to figure out if we should use typed properties.

src/Doctrine/EntityClassGenerator.php
```diff
@@ -50,6 +50,7 @@ public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $
                 'api_resource' => $apiResource,
                 'should_escape_table_name' => $this->doctrineHelper->isKeyword($tableName),
                 'table_name' => $tableName,
+                'is_typed' => $isTyped,
```
weaverryan, line 53: I don't think we need this - there should already be a `use_typed_properties` variable that's added automatically https://github.com/symfony/maker-bundle/blob/a47408fa6e39034a5abe324a46f4e6733266267a/src/Generator.php#L171

We can use that already in the template

weaverryan (commented):
Great progress! I just dropped 2 comments which should cause some changes.

Thanks!

src/DependencyInjection/Configuration.php
Superkooka: Humm for ``canUseTypedProperties`` I'm not agree, this function check on the composer.lock if the keys ['platform-overrides']['php'] exist but on the most php project this key is not defined, and if the key is not defind the function compare the version of the PHP Interpretor and this is a problem in my opinion :/ If you don't see any problem with this I can still use ``canUseTypedProperties``

Superkooka:
There are more than one month you have send that: 
> I think this should default to `null`, meaning that the user has no "preference". Then, we should inject this into the `PhpCompatUtil` class and use it in `canUseTypedProperties`. Basically, if this option is `null`, then that method will do what it does now: return true if php 7.4 or higher is used. If this option IS passed (either as true or false), then that value would be returned directly from that method.
> 
> Then we can continue using the `canUseTypedProperties()` method everywhere to figure out if we should use typed properties.

And I have answerd:
> Humm for `canUseTypedProperties` I'm not agree, this function check on the composer.lock if the keys ['platform-overrides']['php'] exist but on the most php project this key is not defined, and if the key is not defind the function compare the version of the PHP Interpretor and this is a problem in my opinion :/ If you don't see any problem with this I can still use `canUseTypedProperties`

So there are news about it ? I still use `` $variables['use_typed_properties'] = $this->phpCompatUtil->canUseTypedProperties();`` ? If it's the case I solve conflict, just edit that and you can merge. If you prefer my "initial version checking" I just have to solve conflict and you can merge 😀

src/Maker/MakeResetPassword.php
```diff
@@ -54,12 +55,14 @@ class MakeResetPassword extends AbstractMaker
     private $fileManager;
     private $doctrineHelper;
     private $entityClassGenerator;
+    private $params;
 
-    public function __construct(FileManager $fileManager, DoctrineHelper $doctrineHelper, EntityClassGenerator $entityClassGenerator)
+    public function __construct(FileManager $fileManager, DoctrineHelper $doctrineHelper, EntityClassGenerator $entityClassGenerator, ParameterBagInterface $params)
```
jrushlow, line 60: The changes to this maker are not needed. All of the templates already have a `$use_typed_properties` global variable available which is ultimately set by the https://github.com/symfony/maker-bundle/blob/313b5669a5370bf36cd34fa8f5b5bbbfa5fb8aa8/src/Util/PhpCompatUtil.php#L39 method.

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -29,7 +29,7 @@ class <?= $class_name ?> extends AbstractController
 {
     use ResetPasswordControllerTrait;
 
-    private $resetPasswordHelper;
+    private<?= $is_typed ? ' ResetPasswordHelperInterface' : null ?> $resetPasswordHelper;
```
jrushlow, line 32: Example usage of the global `$use_typed_properties` variable that is available in all templates and should be used anytime we are determining if typed properties should be used in all templates.

```suggestion
    private <?= $use_typed_properties ? 'ResetPasswordHelperInterface ' : null ?>$resetPasswordHelper;
```

src/Command/MakerCommand.php
```diff
@@ -39,13 +40,15 @@ final class MakerCommand extends Command
     private $io;
     private $checkDependencies = true;
     private $generator;
+    private $params;
 
-    public function __construct(MakerInterface $maker, FileManager $fileManager, Generator $generator)
+    public function __construct(MakerInterface $maker, FileManager $fileManager, Generator $generator, ParameterBagInterface $params)
```
jrushlow, line 45: The `ParameterBagInterface` is not needed. If we are going to allow a new `typed_properties` config param, it should be injected into the `PhpCompatUtil::class` instead.

jrushlow (changes requested):
Howdy @Superkooka - A few changes and this should be in pretty good shape.

elzozo13:
I this abandoned? Was looking forward for the feature since first time I saw it in december.

Superkooka:
I don't have the time to edit/create tests 😦

SirMishaa:
What about this PR? Can we restart tests to have logs?

jrushlow:
Closing - users upstream no longer exists..

PR #774, opened 2021-01-02
Add make:makefile

DuboisS:
With @Superkooka

Handle:
- `make help`
- Twig linter command
- Doctrine / Doctrine fixtures bundle
- Webpack encore (yarn/npm)
- Save existing Makefile if user wants to generate new one

Also, most of the commands in the Makefile use parallel jobs.
For the sensitives commands with `rm -rf`, we ask the user for a confirmation.

Example of Makefile with doctrine (sqlite) / fixtures / webpack encore : https://pastebin.com/Scbysu03

image
image
image
image

### DIFF
--- a/src/Maker/MakeMakefile.php
+++ b/src/Maker/MakeMakefile.php
@@ -1,0 +1,192 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\FileManager;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Question\Question;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * @author Steven Dubois <contact@duboiss.fr>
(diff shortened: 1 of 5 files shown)


jrushlow:
Hi and thanks for taking the time to submit this PR @DuboisS. I love the concept - personally I do this locally for things like `make up` -> `docker-compose up --no-start`. However, I'm not sure this is something we should maintain in MakerBundle. The biggest issue that I've run into when going using MakeFiles is keeping track of flags.. e.g. `make serve` by default runs in the background.. What If I don't want to pass the `-d` flag to run this in the background?

On the flip side, keeping track of argument changes for each command could be a very daunting task in my opinion. E.g. The Symfony binary adds/removes a flag for say `new`. For a couple of different commands that wouldn't be that bad, but for this - I'm not sure its a path we should go down.

Having said that, please don't feel discouraged in contributing to Symfony! I love the idea, but just not sure it is a good fit for `MakerBundle`. Curious as to what others think.

weaverryan:
Yea, this is a really, really cool command... but it feels out of scope for MakerBundle, as Makefiles aren't really something that's officially part of Symfony.

DuboisS:
> What If I don't want to pass the -d flag to run this in the background?

It is true, but it is the only command of this kind.
Some can also develop with their own web server (apache, nginx...). So we could just remove make start and it's fixed.

Regarding all of the remaining command, they seem like they can be used by everyone (exceptions for docker users), no matter the situation.

The idea is to have a simple Makefile that embeds some basic features that can benefit a maximum number of people.

> but it feels out of scope for MakerBundle, as Makefiles aren't really something that's officially part of Symfony.
It's like make:docker:database, a shortcut for developers in symfony.

Concerning file maintenance, I understand that it can be scary but it's very rare that a change takes place in the symfony command/folder architecture.

jrushlow:
Howdy @DuboisS 

> It's like make:docker:database, a shortcut for developers in symfony.

The `make:docker:database` command came to life because the Symfony Binary / Cloud have support for docker. Originally we were going to create a whole suite of `make:docker:`'s but I believe we are going to shift our focus on encouraging the use of  https://github.com/dunglas/symfony-docker which has the ability to leverage Symfony's Flex system. 

Aside from that, great work on this! I do feel it will get better traction as its own bundle. Perhaps `SymfonyMakeFileBundle`? Again, please do not get discouraged from contributing and let us know if you decide to turn this into a bundle. 

Thanks @DuboisS

PR #776, opened 2021-01-06
[make:registration] allow email verification without authentication

jrushlow:
By passing the user id as an extra query param to `VerifyEmailHelper::generateSignature()` - users are able to verify their email address without being authenticated.

As a precautionary note, answering `no` to `Do you want to require the user to be authenticated to verify their email?` will allow anyone with the link generated by `VerifyEmailHelper` to validated that users email address. It should also be advised that answering `no` could possibly leak personally identifiable information in log files if the user `id` is changed to say, a users email address.

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -16,6 +16,7 @@ use Doctrine\Common\Annotations\Annotation;
 use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Doctrine\DoctrineHelper;
 use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
@@ -55,11 +56,14 @@ final class MakeRegistrationForm extends AbstractMaker
 
     private $router;
 
-    public function __construct(FileManager $fileManager, FormTypeRenderer $formTypeRenderer, RouterInterface $router)
+    private $doctrineHelper;
+
+    public function __construct(FileManager $fileManager, FormTypeRenderer $formTypeRenderer, RouterInterface $router, DoctrineHelper $doctrineHelper)
     {
         $this->fileManager = $fileManager;
         $this->formTypeRenderer = $formTypeRenderer;
         $this->router = $router;
+        $this->doctrineHelper = $doctrineHelper;
     }
 
     public static function getCommandName(): string
@@ -83,6 +87,7 @@ final class MakeRegistrationForm extends AbstractMaker
             ->addArgument('username-field')
             ->addArgument('password-field')
             ->addArgument('will-verify-email')
+            ->addArgument('verify-email-anonymously')
             ->addArgument('id-getter')
             ->addArgument('email-getter')
             ->addArgument('from-email-address')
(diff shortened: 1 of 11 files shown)


src/Maker/MakeRegistrationForm.php
```diff
@@ -138,9 +139,16 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
 
+        // This must be preset to true to avoid code being generated if $willVerify === false
+        $input->setArgument('verify-email-with-auth', true);
+
         if ($willVerify) {
             $this->checkComponentsExist($io);
 
+            $verifyWithAuth = $io->confirm('Do you want to require the user to be authenticated to verify their email?', true);
+
+            $input->setArgument('verify-email-with-auth', $verifyWithAuth);
```
weaverryan, line 150: we can combine these into 1 line and avoid the `$verifyWithAuth` argument.

src/Resources/skeleton/registration/RegistrationController.tpl.php
```diff
@@ -32,10 +35,16 @@ class <?= $class_name; ?> extends <?= $parent_class_name; ?><?= "\n" ?>
 {
 <?php if ($will_verify_email): ?>
     private $emailVerifier;
+<?php if (!$verify_email_with_auth): ?>
+    private <?= $use_typed_properties ? 'EntityManagerInterface ' : null ?>$manager;
+<?php endif; ?>
```
weaverryan, line 40: I think we should autowire this into the method, not the constructor. That's my default mode - `EmailVerifier` is kind of an exception because we know it will definitely be needed in all methods.

src/Resources/skeleton/registration/RegistrationController.tpl.php
```diff
@@ -104,11 +113,26 @@ public function register(Request $request, UserPasswordEncoderInterface $passwor
...
+<?php if ($verify_email_with_auth): ?>
         $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
+<?php else: ?>
+        $id = $request->get('id');
+
+        if (null === $id) {
+            return $this->redirectToRoute('app_register');
+        }
+
+        $repository = $this->manager->getRepository(<?= $user_class_name ?>::class);
```
weaverryan, line 125: I can't remember: is there a reason we don't autowire the repository directly instead of the EM?

weaverryan (commented):
Very nice! Only minor comments

src/Maker/MakeRegistrationForm.php
jrushlow: done

src/Resources/skeleton/registration/RegistrationController.tpl.php
jrushlow: done

src/Resources/skeleton/registration/RegistrationController.tpl.php
jrushlow: done

src/Maker/MakeRegistrationForm.php
```diff
@@ -138,9 +139,17 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
 
+        // This must be preset to true to avoid code being generated if $willVerify === false
+        $input->setArgument('verify-email-with-auth', true);
+
         if ($willVerify) {
             $this->checkComponentsExist($io);
 
+            $input->setArgument(
+                'verify-email-with-auth',
+                $io->confirm('Do you want to require the user to be authenticated to verify their email?', true)
```
weaverryan, line 150: Hmm, I'm trying to think about how a person would understand this... and the wording on it. Maybe we give some extra info before?

> To verify their email address, you can either require that users are authenticated when they click
> the verification link *or* allow them to be anonymous (e.g. maybe they click the link on a different
> device).

> Allow anonymous users to verify their email?

(and default to no)

jrushlow: think we have this nailed down now

weaverryan:
Thank you Jesse!

PetarBelberov:
After [make:registration] where I need to change the code to allow email verification without authentication?

weaverryan:
@PetarBelberov if you're using the newest version of Maker, which includes this PR, then `make:registration` will ask you interactively whether or not you want to allow verification without authentication. And so, you should not need to make any changes. 

If you've previously executed this command, adapt the changes from these 2 files into your app:

* https://github.com/symfony/maker-bundle/pull/776/files#diff-cd522c6e6b467964c7f6841acfa6d7d77bfcafba5019fde91e59e8ca11ebc92f
* https://github.com/symfony/maker-bundle/pull/776/files#diff-3d33dffe2b9b57eee72392dc4336b80bb41045c8ca266eed9eccae3f04425834

We also have it documented here: https://github.com/symfonycasts/verify-email-bundle#anonymous-validation

Cheers!

PetarBelberov:
@weaverryan, yes, the issue was from the current version I used. I am a lucky guy with this brand new update. Thank you for the help. All the best!

PR #778, opened 2021-01-09
Move symfony/process dependency from require-dev to require

franmomu:
Hi!

I was trying to use [`MakerTestCase`](https://github.com/symfony/maker-bundle/blob/a47408fa6e39034a5abe324a46f4e6733266267a/src/Test/MakerTestCase.php) and I'm finding some little things. 

In this case, `symfony/process` is being used under the src directory in two files:
- [src/Test/MakerTestProcess.php](https://github.com/symfony/maker-bundle/blob/a47408fa6e39034a5abe324a46f4e6733266267a/src/Test/MakerTestProcess.php#L14)
- [src/Test/MakerTestEnvironment.php](https://github.com/symfony/maker-bundle/blob/a47408fa6e39034a5abe324a46f4e6733266267a/src/Test/MakerTestEnvironment.php#L18-L19)

So trying to execute tests I'm getting: `Error: Class 'Symfony\Component\Process\Process' not found`

### DIFF
--- a/src/Test/MakerTestCase.php
+++ b/src/Test/MakerTestCase.php
@@ -15,6 +15,7 @@ use Composer\Semver\Semver;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bundle\MakerBundle\MakerInterface;
 use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Component\Process\Process;
 
 abstract class MakerTestCase extends TestCase
 {
@@ -32,6 +33,10 @@ abstract class MakerTestCase extends TestCase
 
     protected function executeMakerCommand(MakerTestDetails $testDetails)
     {
+        if (!class_exists(Process::class)) {
+            throw new \LogicException('The MakerTestCase cannot be run as the Process component is not installed. Try running "compose require --dev symfony/process".');
+        }
+
         if (!$testDetails->isSupportedByCurrentPhpVersion()) {
             $this->markTestSkipped();
         }


OskarStark (approved):

wouterj:
Hi! Thank you for taking the time to submit PRs!

However, I believe the current state is correct. `symfony/process` is an optional dependency of MakerBundle (you only need it when you want to write tests for custom makers). I don't think we want to install `symfony/process` in all projects that use the maker commands.

OskarStark:
Isn't maker bundle itself a dev dependency? So this should not affect the end user or do I miss sth?

jrushlow:
I'm not sure if this will fly, `MakerTestProcess::class` is marked as `internal` and as such the dependency on `symfony/process` is only needed when developing `MakerBundle` it self. @OskarStark is correct that technically `MakerBundle` _should_ be a dev dependency for the consumer. However, there is still an argument to be made that once we create a dependency - in this case move one, it puts a BC burden on the us for in the future.

wouterj:
> So this should not affect the end user or do I miss sth?

Dev dependencies of projects are installed (unless you're using ``--no-dev``, which is highly unlikely if you run the command locally). So it does affect end users (end users of MakerBundle are developers on local machines).

---

In Symfony, such optional dependencies are never required directly - to minimize the number of dependent packages. However, we often include a check + exception with nice error message if the optional dependency was not used. As @jrushlow mentions, these classes are internal - but they are used in `MakerTestCase` (which is the official entry point to test your own makers).

@franmomu what do you think of updating this PR to add something like a check for `class_exists(Process::class)` and a message like `Install the "symfony/process" package to use the MakerTestCase` in to the `MakerTestCase::executeMakerCommand()` >method?

franmomu:
> @franmomu what do you think of updating this PR to add something like a check for `class_exists(Process::class)` and a message like `Install the "symfony/process" package to use the MakerTestCase` in to the `MakerTestCase::executeMakerCommand()` >method?

Done!

src/Test/MakerTestCase.php
```diff
@@ -32,6 +33,10 @@ public function testExecute(MakerTestDetails $makerTestDetails)
 
     protected function executeMakerCommand(MakerTestDetails $testDetails)
     {
+        if (!class_exists(Process::class)) {
+            throw new \LogicException('The MakerTestCase cannot be run as the Process component is not installed. Try running "compose require symfony/process".');
```
wouterj, line 37:
```suggestion
            throw new \LogicException('The MakerTestCase cannot be run as the Process component is not installed. Try running "compose require --dev symfony/process".');
```

As it's required for tests, I think it's safe to mention the `--dev` flag in the error message.

wouterj (approved):
:+1: thank you for the quick updates!

src/Test/MakerTestCase.php
franmomu: you're right, changed.

weaverryan:
Thanks @franmomu!

PR #780, opened 2021-01-09
Fixed PHP 7.1.3 tests by using PHPUnit 8.5

wouterj:
This fixes the broken PHP 8 tests, as they were using PHPUnit 7.5 (which doesn't support PHP 8).

This version is also proposed to the official recipes and will probably be merged shortly: https://github.com/symfony/recipes/pull/868

### DIFF
--- a/src/Test/MakerTestEnvironment.php
+++ b/src/Test/MakerTestEnvironment.php
@@ -373,12 +373,18 @@ final class MakerTestEnvironment
             $this->fs->symlink($rootPath.'/vendor/symfony/phpunit-bridge', $this->flexPath.'/vendor/symfony/phpunit-bridge');
         }
 
-        // temporarily ignoring indirect deprecations - see #237
         $replacements = [
+            // temporarily ignoring indirect deprecations - see #237
             [
                 'filename' => '.env.test',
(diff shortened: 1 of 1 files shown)


Superkooka:
This bundle support at least PHP 7.1.3 but PHPUnit support PHP ^7.2, so it's incopatible

jrushlow:
we need to have a fallback / conditionally use PHPUnit 7x for PHP 7.1

OskarStark:
Recipe is now merged @wouterj wou

wouterj:
Alright after the merge, we now would need to do exactly the opposite for PHP 7.1 :smile: 

I've updated the PR to remove the PHPUNIT version line completely. If not explicitly set, the `simple-phpunit` binary already chooses the correct version based on the available PHP version ([ref](https://github.com/symfony/symfony/blob/5.x/src/Symfony/Bridge/PhpUnit/bin/simple-phpunit.php#L96-L104)).

jrushlow (approved):

jrushlow:
thanks @wouterj - nice to finally see a green check next to PHP8.0!

weaverryan:
Thanks Wouter!

PR #785, opened 2021-01-12
[Reset-Password] use expiration translation in check email route

jrushlow:
I missed this template in #770

- [x] Depends on https://github.com/SymfonyCasts/reset-password-bundle/pull/143

### DIFF
--- a/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
+++ b/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
@@ -76,12 +76,12 @@ class <?= $class_name ?> extends AbstractController
     public function checkEmail(): Response
     {
         // We prevent users from directly accessing this page
-        if (!$this->canCheckEmail()) {
+        if (null === ($resetToken = $this->getTokenObjectFromSession())) {
             return $this->redirectToRoute('app_forgot_password_request');
         }
 
         return $this->render('reset_password/check_email.html.twig', [
-            'tokenLifetime' => $this->resetPasswordHelper->getTokenLifetime(),
+            'resetToken' => $resetToken,
         ]);
     }
 
@@ -155,9 +155,6 @@ class <?= $class_name ?> extends AbstractController
             '<?= $email_field ?>' => $emailFormData,
         ]);
 
-        // Marks that you are allowed to see the app_check_email page.
-        $this->setCanCheckEmailInSession();
-
         // Do not reveal whether a user account was found or not.
         if (!$user) {
             return $this->redirectToRoute('app_check_email');
@@ -189,6 +186,9 @@ class <?= $class_name ?> extends AbstractController
         ;
 
         $mailer->send($email);
+
+        // Store the token object in session for retrieval in check-email route.
+        $this->setTokenObjectInSession($resetToken);
 
         return $this->redirectToRoute('app_check_email');
     }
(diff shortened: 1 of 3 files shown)


jrushlow:
Fixes https://github.com/SymfonyCasts/reset-password-bundle/issues/142

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -75,13 +76,18 @@ public function request(Request $request, MailerInterface $mailer): Response
...
+        $generatedAt = $this->getTokenGeneratedAtFromSession();
+
         // We prevent users from directly accessing this page
-        if (!$this->canCheckEmail()) {
+        if (null === $generatedAt || !$this->canCheckEmail()) {
             return $this->redirectToRoute('app_forgot_password_request');
         }
 
+        // We need to generate a dummy token to retrieve the translation data for the template.
+        $dummyToken = new ResetPasswordToken('fake', new \DateTimeImmutable(), $generatedAt);
```
weaverryan, line 87: This is kind of a bummer (which I'm sure you already know!).

Could we just store the entire `ResetPasswordToken` in the session? That would replace the current boolean we use in the `canCheckEmail()` check and the new date. 

If we're worried about the token string inside `ResetPasswordToken` being stored in the session, we could call a method to clear the token string from the `ResetPasswordToken` object before storing it in the session (and then, throw an exception if the user tries to access it). A new controller trait could handle cloning the object, clearing the token, and setting it on the session.

WDYT? It's a bit odd... but then we're basically keeping the `ResetPasswordToken` object in the session instead of dates or booleans.

jrushlow: I like it.. I was thinking about that earlier only being concerned by the token (clearing it as suggested would solve that problem) and performance / tangible costs (e.g. AWS Redis for larger applications). I doubt the latter would ever become a factor in the majority of most applications, but if it did, I'm sure the dev team could implement a more cost effective solution. 

I'll refactor to go this route - v2 of reset bundle may provide us with a better solution in the future. None of which are feasible now without BC.

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -75,13 +76,18 @@ public function request(Request $request, MailerInterface $mailer): Response
 <?php } ?>
     public function checkEmail(): Response
     {
+        $generatedAt = $this->getTokenGeneratedAtFromSession();
+
         // We prevent users from directly accessing this page
-        if (!$this->canCheckEmail()) {
+        if (null === $generatedAt || !$this->canCheckEmail()) {
```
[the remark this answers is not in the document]
jrushlow: done - we not check if the reset token object is present

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
jrushlow: done

weaverryan:
Thanks Jesse!

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -155,9 +155,6 @@ private function processSendingPasswordResetEmail(string $emailFormData, MailerI
             '<?= $email_field ?>' => $emailFormData,
         ]);
 
-        // Marks that you are allowed to see the app_check_email page.
-        $this->setCanCheckEmailInSession();
```
venyii, line 159: One small change that this introduces: Now an unknown email won't get to the check-email page since the next `if` clause will exit the function before `setTokenObjectInSession` can be called (can't be called anyway without a token).

This means that you'll be redirected in the `checkEmail` function (without user feedback) since no token is present. Is this an intentional change? Right now you'd know if the email existed by checking if a redirect occurs or not 🙂

PR #790, opened 2021-01-19
[MakeVoter] add missing type hints inherited from abstract Voter

jrushlow:
`Symfony\Component\Security\Core\Authorization\Voter\Voter::class` References:

- [Supports Method](https://github.com/symfony/symfony/blob/294195157c3690b869ff6295713a69ff38b3039c/src/Symfony/Component/Security/Core/Authorization/Voter/Voter.php#L70)

- [VoteOnAttribute Method](https://github.com/symfony/symfony/blob/294195157c3690b869ff6295713a69ff38b3039c/src/Symfony/Component/Security/Core/Authorization/Voter/Voter.php#L80)

### DIFF
--- a/src/Resources/skeleton/security/Voter.tpl.php
+++ b/src/Resources/skeleton/security/Voter.tpl.php
@@ -8,7 +8,7 @@ use Symfony\Component\Security\Core\User\UserInterface;
 
 class <?= $class_name ?> extends Voter
 {
-    protected function supports($attribute, $subject)
+    protected function supports(string $attribute, $subject)
     {
         // replace with your own logic
         // https://symfony.com/doc/current/security/voters.html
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/security/Voter.tpl.php
```diff
@@ -8,15 +8,15 @@
 
 class <?= $class_name ?> extends Voter
 {
-    protected function supports($attribute, $subject)
+    protected function supports(string $attribute, $subject)
```
weaverryan, line 11: The type-hint isn't added to Symfony until 5.0. I can never remember the rules on this, but it seems like a user cannot *add* the type on their class if the parent doesn't have it: https://3v4l.org/uUIYl - but I can't remember if this is the whole story...

weaverryan: This also makes me wonder if we should at least test the LTS Symfony version in our test matrix...

jrushlow:
duplicate of #853

PR #797, opened 2021-01-28
[Tests] use plaintext password encoder in tests

jrushlow:


jrushlow:
Whoooops.. not needed for these tests

PR #799, opened 2021-01-30
[DX] [ResetPassword] lets stop abusing input arguments

jrushlow:
We're currently abusing input arguments by using them as data stores internally. This PR is the first of many to refactor makers to replace input argument with proper class properties, add method return types where possible, and general housekeeping to improve code readability. 

The first maker to use this approach was `make:docker:database` found [here](https://github.com/symfony/maker-bundle/blob/main/src/Maker/MakeDockerDatabase.php).

One decision we should make now before proceeding with refactoring additional makers is in regards to the docBlock format we should use for properties. Symfony uses the expanded format - but this can lead to reduced readability with a class that has multiple properties. As a solution to this we could break tradition and use single line docBlocks strictly for internal maker properties. 

```php
// Traditional expanded properties

    /**
     * @var string type of database selected by the user
     */
    private $databaseChoice;

    /**
     * @var string Service identifier to be set in docker-compose.yaml
     */
    private $serviceName = 'database';

    /**
     * @var string Version set in docker-compose.yaml for the service. e.g. latest
     */
    private $serviceVersion = 'latest';
``` 

```php
// Single line

    /** @var string The "FROM" email address to be used in email template. */
    private $fromEmailAddress;

    /** @var string The "FROM" name used in the email template. */
    private $fromEmailName;

    /** @var string Redirect route to be used after a successful password reset. Route does not have to exist. */
    private $controllerResetSuccessRedirect;
```

Another consideration to be made is `types` and `descriptions`. When the minimum supported PHP version for Symfony reaches `7.4`, we will be able to use typed properties. Until then I think we could omit types declared in property docBlocks unless we actually need a docBlock to begin with. Which leads me to, should we include descriptions for all properties as shown above (not included in the actual PR) or only when describing the property can not be done clearly with the property name itself. 

*Disclaimer - All of the above is scoped strictly to Makers and input argument replacements. It should not be taken out of context for the public API or generated code.

### DIFF
--- a/src/Maker/MakeResetPassword.php
+++ b/src/Maker/MakeResetPassword.php
@@ -29,7 +29,6 @@ use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
 use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
 use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Mailer\MailerInterface;
 use Symfony\Component\Yaml\Yaml;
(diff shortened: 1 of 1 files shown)


weaverryan:
Hi!

Thanks for this :).

1) I'm fine with the single-line format - `/** ... *`
2) Perhaps we should - as a convention - prefix all these properties? e.g. `$inputFromEmailAddress` and `$inputFromEmailName`?
3) In general, if a doc block doesn't add anything, then let's not include it - I'm 👍 for that

Cheers!

weaverryan (approved):

weaverryan:
Very nice - thanks Jesse!

PR #800, opened 2021-01-30
[Internal] [User] removed unused service from make:user

jrushlow:
PR #590 added `DoctrineHelper` to `MakeResetPassword` but the service is never used.


weaverryan:
Thanks!

PR #801, opened 2021-01-30
[make:auth] handle session deprecations in 5.3

jrushlow:
- fixed deprecated ReflectionType::__toString() (MakeAuthenticatorTest) Deprecated in PHP 7.1
- removes unneeded optional param - triggers required param order deprecation in PHP8 (GeneratorTest)
- handles Session deprecation's introduced in Symfony 5.3 in test suites

### DIFF
--- a/src/Maker/MakeAuthenticator.php
+++ b/src/Maker/MakeAuthenticator.php
@@ -403,11 +403,12 @@ final class MakeAuthenticator extends AbstractMaker
     private function providerKeyTypeHint(): string
     {
         $reflectionMethod = new \ReflectionMethod(AbstractFormLoginAuthenticator::class, 'onAuthenticationSuccess');
-        $typeHint = (string) $reflectionMethod->getParameters()[2]->getType();
-        if ($typeHint) {
-            $typeHint .= ' ';
+        $type = $reflectionMethod->getParameters()[2]->getType();
+
+        if (!$type instanceof \ReflectionNamedType) {
+            return '';
         }
 
-        return $typeHint;
+        return sprintf('%s ', $type->getName());
     }
 }
--- a/tests/fixtures/MakeAuthenticatorLoginFormUserEntity/tests/SecurityControllerTest.php
+++ b/tests/fixtures/MakeAuthenticatorLoginFormUserEntity/tests/SecurityControllerTest.php
@@ -6,6 +6,7 @@ use App\Entity\User;
 use App\Security\AppCustomAuthenticator;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Symfony\Component\HttpKernel\Kernel;
 
 class SecurityControllerTest extends WebTestCase
 {
@@ -60,7 +61,17 @@ class SecurityControllerTest extends WebTestCase
         $client->submit($form);
 
         $this->assertStringContainsString('TODO: provide a valid redirect', $client->getResponse()->getContent());
-        $this->assertNotNull($token = $client->getContainer()->get('security.token_storage')->getToken());
+
+        $tokenStorage = static::$container->get('security.token_storage');
+
+        // Handle Session deprecations in Symfony 5.3+
+        if (Kernel::VERSION_ID >= 50300) {
+            $tokenStorage->disableUsageTracking();
+        }
+
+        $token = $tokenStorage->getToken();
+
+        self::assertNotNull($token);
         $this->assertInstanceOf(User::class, $token->getUser());
     }
 }
--- a/tests/fixtures/MakeAuthenticatorLoginFormUserEntityLogout/tests/SecurityControllerTest.php
+++ b/tests/fixtures/MakeAuthenticatorLoginFormUserEntityLogout/tests/SecurityControllerTest.php
@@ -6,6 +6,8 @@ use App\Entity\User;
 use App\Security\AppCustomAuthenticator;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 
 class SecurityControllerTest extends WebTestCase
 {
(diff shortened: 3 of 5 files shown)


weaverryan:
I suspect that the failing tests are caused by - https://github.com/symfony/symfony/pull/38616 - the timing is pretty much perfect. What we'll need to do is run that test locally (using the flags so we get the `dev` dependencies) and see if we can repeat. After we run the test (and it fails), you should be able to move into the tmp directory of the project and execute the test inside that project to see the failure again, because the actual failure is coming from when we run the test. It's coming from this line: https://github.com/symfony/maker-bundle/blob/main/tests/fixtures/MakeAuthenticatorLoginFormUserEntity/tests/SecurityControllerTest.php#L63

If we *are* able to reproduce all of this, then we could try to create a tiny, reproducer project that shows the (potential) bug and open an issue in symfony. I don't want to ping the author `jderusse` until we have something more easy for him to take a look at.

Thanks!

jrushlow:
Most of the session related test issues on the `dev` branches were fixed by https://github.com/symfony/recipes/pull/905 referenced in https://github.com/symfony/symfony/issues/40282. The remaining session based issues will be fixed by either 

```php
// login a user in WebTestCase

$usageTrackingTokenStorage = static::$container->get('security.token_storage');
$usageTrackingTokenStorage->disableUsageTracking();  // <- bypases trying to get the session from the request stack
$token = $usageTrackingTokenStorage->getToken();
$this->assertNotNull($token); // passes
```
or by adding a event listener on the request with a callable to the test as done in https://github.com/symfony/symfony/pull/38616/files#diff-c1a4581fd182f7bd35bbb4f0297fd62d1df49fb0b2a4b4210707f3a8131766d1R53-R55

jrushlow:
`dev`  CRUD failures may be related to https://github.com/symfony/recipes/pull/892 which disables `http_method_override` by default

jrushlow:
splitting up the test fixes into individual PR's - there are too many issues across different parts of maker to fix everything under one roof.

src/Maker/MakeAuthenticator.php
```diff
@@ -403,11 +403,11 @@ public function configureDependencies(DependencyBuilder $dependencies, InputInte
     private function providerKeyTypeHint(): string
     {
         $reflectionMethod = new \ReflectionMethod(AbstractFormLoginAuthenticator::class, 'onAuthenticationSuccess');
-        $typeHint = (string) $reflectionMethod->getParameters()[2]->getType();
-        if ($typeHint) {
-            $typeHint .= ' ';
+
+        if (null !== $type = $reflectionMethod->getParameters()[2]) {
```
weaverryan, line 407: This looks backwards to me? Shouldn't it be `null === $type`, then `return ''`, else (at the bottom) return `sprintf('%s ', $type->getName())`?

tests/fixtures/MakeAuthenticatorLoginFormUserEntity/tests/SecurityControllerTest.php
```diff
@@ -60,7 +61,17 @@ public function testCommand()
         $client->submit($form);
 
         $this->assertStringContainsString('TODO: provide a valid redirect', $client->getResponse()->getContent());
-        $this->assertNotNull($token = $client->getContainer()->get('security.token_storage')->getToken());
+
+        // Handle Session deprecations in Symfony 5.3+
+        if (Kernel::VERSION_ID >= 50300) {
+            $tokenStorage = static::$container->get('security.token_storage');
```
weaverryan, line 67: Why is the way that we fetch the service different in the two cases? This new way seems more correct.

And, would it be possible to use a `method_exists()` instead of the Symfony version?

```php
$tokenStorage = static::$container->get('security.token_storage');
if (method_exists($tokenStorage, 'disableUsageTracking')) {
    $tokenStorage->disableUsageTracking();
}
$token = $tokenStorage->getToken();
```

Checking the kernel is sort of a "last resort". If we can rely on something hard in the code, that's even better.

src/Maker/MakeAuthenticator.php
jrushlow: good catch! refactored to be a bit more clear && fix my mistake.

src/Maker/MakeAuthenticator.php
```diff
@@ -403,11 +403,12 @@ public function configureDependencies(DependencyBuilder $dependencies, InputInte
     private function providerKeyTypeHint(): string
     {
         $reflectionMethod = new \ReflectionMethod(AbstractFormLoginAuthenticator::class, 'onAuthenticationSuccess');
-        $typeHint = (string) $reflectionMethod->getParameters()[2]->getType();
-        if ($typeHint) {
-            $typeHint .= ' ';
+        $type = $reflectionMethod->getParameters()[2]->getType();
+
+        if (!$type instanceof \ReflectionNamedType) {
```
jrushlow, line 408: `getType()` returns `\ReflectionNamedType` | `null` since 7.1. As maker is not used in production, I opted to check `instanceOf` vs the more performant `null ===` for better DX.

tests/fixtures/MakeAuthenticatorLoginFormUserEntityLogout/tests/SecurityControllerTest.php
```diff
@@ -60,10 +62,23 @@ public function testCommand()
...
+    }
+
+    /**
+     * Handle Session deprecations in Symfony 5.3+
+     */
+    private function getToken(): ?TokenInterface
+    {
+        $tokenStorage = static::$container->get('security.token_storage');
+
+        if (class_exists(SessionNotFoundException::class)) {
```
jrushlow, line 78: > And, would it be possible to use a method_exists() instead of the Symfony version?

Yes, kinda. The `disableUsageTracking()` method has always existed in the [UsageTrackingTokenStorage::class](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Security/Core/Authentication/Token/Storage/UsageTrackingTokenStorage.php) since it was introduced in Symfony 4.4

For context, PR 38616 [changed](https://github.com/symfony/symfony/pull/38616/files#diff-2aeaa6c8af14b7abd53c7398c9c7e1bf105541e8470e2f9adc74c2019f81d8b1) how the token is retrieved when calling `UsageTrackingTokenStorage->getToken()` if usage tracking is enabled (which the kernel enables). 

Although I'm not sure we _need_ to, but to support Symfony `4.0` - `4.3` we can't rely on `disableUsageTracking()` to exist. And for Symfony `4.4` - `5.2` I don't think we _should_ disable usage tracking as thats not realistic from the users perspective although we could. E.g. `class_exists(UsageTrackingTokenStorage)`

The `SessionNotFoundException::class` was introduced in `38616` (Symfony `5.3`) and was the only class introduced that didn't look like it would be removed anytime soon.

weaverryan: let's keep the kernel check then - thanks for the explantion

weaverryan:
Thanks Jesse!

PR #805, opened 2021-02-01
Enhancement: Fullfill interface and move comments to the method description

OskarStark:
Before this fix, static analyzers are complaining:
```
Return type (void) of method App\Entity\User::getSalt() should be compatible with return type (string|null) of method Symfony\Component\Security\Core\User\UserInterface::getSalt()
```

cc @chalasr

### DIFF
--- a/src/Security/UserClassBuilder.php
+++ b/src/Security/UserClassBuilder.php
@@ -165,13 +165,20 @@ final class UserClassBuilder
             // add an empty method only
             $builder = $manipulator->createMethodBuilder(
                 'getPassword',
-                null,
-                false,
-                ['@see UserInterface']
+                'string',
+                true,
+                [
+                    'This method is not needed for apps that do not check user passwords.',
+                    '',
+                    '@see UserInterface',
+                ]
             );
+
             $builder->addStmt(
-                $manipulator->createMethodLevelCommentNode(
-                    'not needed for apps that do not check user passwords'
+                new Node\Stmt\Return_(
+                    new Node\Expr\ConstFetch(
+                        new Node\Name('null')
+                    )
                 )
             );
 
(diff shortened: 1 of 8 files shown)


chalasr:
Thanks Oskar. 👍 once tests are fixed

src/Security/UserClassBuilder.php
```diff
@@ -165,14 +165,17 @@ private function addGetPassword(ClassSourceManipulator $manipulator, UserClassCo
...
+                    '',
+                    '@see UserInterface',
+                ]
             );
+
             $builder->addStmt(
-                $manipulator->createMethodLevelCommentNode(
-                    'not needed for apps that do not check user passwords'
-                )
+                new Node\Stmt\Return_(null)
```
OskarStark, line 178: I think this is wrong 😑 

I want to have

    return null;

Any idea?

OskarStark: This returns `return;`, but I would like to have `return null;` 🤔

OskarStark: I used this test script:
```php
<?php

include ('vendor/autoload.php');

use PhpParser\Error;
use PhpParser\NodeDumper;
use PhpParser\ParserFactory;

$code = <<<'CODE'
<?php

function getPassword(): ?string
{
    return null;
}
CODE;

$parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
try {
    $ast = $parser->parse($code);
} catch (Error $error) {
    echo "Parse error: {$error->getMessage()}\n";
    return;
}

$dumper = new NodeDumper;
echo $dumper->dump($ast) . "\n";
```

which returns:
```
array(
    0: Stmt_Function(
        attrGroups: array(
        )
        byRef: false
        name: Identifier(
            name: getPassword
        )
        params: array(
        )
        returnType: NullableType(
            type: Identifier(
                name: string
            )
        )
        stmts: array(
            0: Stmt_Return(
                expr: Expr_ConstFetch(
                    name: Name(
                        parts: array(
                            0: null
                        )
                    )
                )
            )
        )
    )
)
```

so 
```php
        $builder->addStmt(
            new Node\Stmt\Return_(
                new Node\Expr\ConstFetch(
                    new Node\Name('null')
                )
            )
        );
```

should be correct

OskarStark:
Test failures looks unrelated

jrushlow (approved):
thanks @OskarStark for this PR!

chalasr (approved):

weaverryan:
Awesome work - thank you Oskar!

PR #806, opened 2021-02-03
feat: hint how to use Panther in functional tests

dunglas:
This will help to discover how to use Panther.

### DIFF
--- a/src/Resources/skeleton/test/Functional.tpl.php
+++ b/src/Resources/skeleton/test/Functional.tpl.php
@@ -12,14 +12,21 @@ class <?= $class_name ?> extends <?= $panther_is_available ? 'PantherTestCase' :
 {
     public function testSomething()
     {
-        $client = static::createClient();
+        $client = static::createClient(); // Use BrowserKit to simulate a web browser
+<?php if ($panther_is_available): ?>
+        //$client = static::createPantherClient(); // Or use the real Google Chrome and have your scripts executed
+<?php endif ?>
         $crawler = $client->request('GET', '/');
 
 <?php if ($web_assertions_are_available): ?>
+<?php if (!$panther_is_available): ?>
         $this->assertResponseIsSuccessful();
+<?php endif ?>
         $this->assertSelectorTextContains('h1', 'Hello World');
 <?php else: ?>
+<?php if (!$panther_is_available): ?>
         $this->assertSame(200, $client->getResponse()->getStatusCode());
+<?php endif ?>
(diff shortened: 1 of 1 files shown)


dunglas:
I also removed assertions that not available with WebDriver (only when Panther is enabled). A second step, could be to ask users if they want to generate a `WebTestCase` or a `PantherTestCase` (in another PR).

jrushlow:
Thanks @dunglas for this! 

> I also removed assertions that not available with WebDriver (only when Panther is enabled).

I haven't used Panther before - are there alternatives to assert if a response is successful or if a response contains a specific status code?

jrushlow:
I believe we have a dependency issue?
```
 There was 1 error:

1) App\Tests\FooBarTest::testSomething
Error: Call to undefined method Symfony\Bundle\FrameworkBundle\KernelBrowser::createWebDriverByFromLocator()

/home/runner/work/maker-bundle/maker-bundle/tests/tmp/cache/maker_app_0678c151a0327b3d957ca476c9c30360/vendor/symfony/panther/src/WebTestAssertionsTrait.php:224
/home/runner/work/maker-bundle/maker-bundle/tests/tmp/cache/maker_app_0678c151a0327b3d957ca476c9c30360/vendor/symfony/panther/src/WebTestAssertionsTrait.php:56
/home/runner/work/maker-bundle/maker-bundle/tests/tmp/cache/maker_app_0678c151a0327b3d957ca476c9c30360/tests/FooBarTest.php:15
```

dunglas:
> I haven't used Panther before - are there alternatives to assert if a response is successful or if a response contains a specific status code?

No, it's impossible by design. The WebDriver spec explicitly say that you can only access to elements visible by the end user through the WebDriver API, and it's not the case of the HTTP status code (and headers). We could maybe find some browser-specific hacks, but it would be ugly and against the spirit of the spec.

dunglas:
> I believe we have a dependency issue?

Weird, it may be a bug introduced in the version I just tagged. How can I reproduce this error?

weaverryan:
You could run:

```
./vendor/bin/simple-phpunit --filter=functional_with_panther
```

That'll take a minute to run. You can then `cd tests/tmp/cache/maker_app_0678c151a0327b3d957ca476c9c30360`(the exact path may be difference - check the error, or there should only be one similarly-named directory in `cache/`) and you'll be looking at a full, standalone Symfony app with Panther installed and a file in `tests/`.

dunglas:
Closing in favor of #806.

PR #807, opened 2021-02-03
Add a new make:test command

dunglas:
Replaces #806.

As a result of discussions with @weaverryan, this PR introduces a new generic test maker. It allows creating unit (PHPUnit), integration/service (WebTestCase) and E2E/UI (PantherTestCase) tests.

`make:unit-test` and `make:functional-test` have been deprecated, but BC is preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
+1.29
+====
+
+* [make:test] Added the maker
+* [make:unit-test] Deprecated the maker
+* [make:functional-test] Deprecated the maker
+
 1.27
+====
 
 * [make:registration-form] Added a new question to generate code that will allow
   users to click on the "verify email" link in their email without needing to be
--- a/src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
+++ b/src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
@@ -29,6 +29,10 @@ class MakeCommandRegistrationPass implements CompilerPassInterface
     {
         foreach ($container->findTaggedServiceIds(self::MAKER_TAG) as $id => $tags) {
             $def = $container->getDefinition($id);
+            if ($def->isDeprecated()) {
+                continue;
+            }
+
             $class = $container->getParameterBag()->resolveValue($def->getClass());
             if (!is_subclass_of($class, MakerInterface::class)) {
                 throw new InvalidArgumentException(sprintf('Service "%s" must implement interface "%s".', $id, MakerInterface::class));
@@ -49,6 +53,15 @@ class MakeCommandRegistrationPass implements CompilerPassInterface
             }
 
             $commandDefinition->addTag('console.command', $tagAttributes);
+
+            /*
+             * @deprecated remove this block when removing make:unit-test and make:functional-test
+             */
+            if (method_exists($class, 'getCommandAliases')) {
+                foreach ($class::getCommandAliases() as $alias) {
+                    $commandDefinition->addTag('console.command', ['command' => $alias, 'description' => 'Deprecated alias of "make:test"']);
+                }
+            }
 
             $container->setDefinition(sprintf('maker.auto_command.%s', Str::asTwigVariable($class::getCommandName())), $commandDefinition);
         }
--- a/src/Maker/MakeTest.php
+++ b/src/Maker/MakeTest.php
@@ -1,0 +1,189 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestAssertionsTrait;
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputAwareMakerInterface;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Component\BrowserKit\History;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\CssSelector\CssSelectorConverter;
+use Symfony\Component\Panther\PantherTestCaseTrait;
+
+/**
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Ryan Weaver <weaverryan@gmail.com>
+ */
+final class MakeTest extends AbstractMaker implements InputAwareMakerInterface
+{
+    private const DESCRIPTIONS = [
+        'TestCase' => 'basic PHPUnit tests',
+        'KernelTestCase' => 'basic tests that have access to Symfony services',
+        'WebTestCase' => 'to run browser-like scenarios, but that don\'t execute JavaScript code',
+        'ApiTestCase' => 'to run API-oriented scenarios',
+        'PantherTestCase' => 'to run e2e scenarios, using a real-browser or HTTP client and a real web server',
+    ];
+    private const DOCS = [
+        'TestCase' => 'https://symfony.com/doc/current/testing.html#unit-tests',
+        'KernelTestCase' => 'https://symfony.com/doc/current/testing/database.html#functional-testing-of-a-doctrine-repository',
+        'WebTestCase' => 'https://symfony.com/doc/current/testing.html#functional-tests',
+        'ApiTestCase' => 'https://api-platform.com/docs/distribution/testing/',
+        'PantherTestCase' => 'https://github.com/symfony/panther#testing-usage',
+    ];
+
+    public static function getCommandName(): string
+    {
+        return 'make:test';
+    }
+
+    /**
+     * @deprecated remove this method when removing make:unit-test and make:functional-test
+     */
+    public static function getCommandAliases(): iterable
+    {
+        yield 'make:unit-test';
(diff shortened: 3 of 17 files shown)


src/Maker/MakeTest.php
```diff
@@ -0,0 +1,209 @@
...
+            case 'web':
+            case 'functional':
+                $input->setArgument(type, 'integration');
+
+                return;
+
+            case 'e2e':
+            case 'ui':
+            case 'panther':
+                $input->setArgument(type, 'e2');
```
nicolas-grekas, line 201: `e2e`?

src/Resources/skeleton/test/Functional.tpl.php
```diff
@@ -10,16 +10,24 @@
 
 class <?= $class_name ?> extends <?= $panther_is_available ? 'PantherTestCase' : 'WebTestCase' ?><?= "\n" ?>
 {
-    public function testSomething()
+    public function testSomething(): void
```
nicolas-grekas, line 13: these `void` on test methods are pure visual noise...

src/Maker/MakeTest.php
```diff
@@ -0,0 +1,209 @@
...
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Component\BrowserKit\History;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\CssSelector\CssSelectorConverter;
+use Symfony\Component\Panther\PantherTestCaseTrait;
+
+/**
+ * MakeTest.
```
nicolas-grekas, line 29: canbe removed

src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
```diff
@@ -50,6 +54,13 @@ public function process(ContainerBuilder $container)
 
             $commandDefinition->addTag('console.command', $tagAttributes);
 
+            // TODO: remove this block when removing make:unit-test and make:functional-test
```
nicolas-grekas, line 57: `@deprecated` instead of TODO everywhere, to help when grepping?

src/Resources/skeleton/test/Functional.tpl.php
dunglas: I know you don't like them but my goal here is pragmatic: without that PHPStorm displays shows a ugly notice and I don't like that freshly generated code triggers warnings in the most popular IDE.

src/Maker/MakeTest.php
```diff
@@ -0,0 +1,209 @@
...
+    public static function getCommandDescription(): string
+    {
+        return 'Creates a new test class';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConf)
+    {
+        $command
+            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the test class (e.g. <fg=yellow>BlogPostTest</>)')
+            ->addArgument('type', InputArgument::OPTIONAL, 'The type of test: <fg=yellow>unit</> (PHPUnit), <fg=yellow>integration</> (WebTestCase) or <fg=yellow>e2e</> (PantherTestCase)')
```
wouterj, line 58: I think we need to have a clear terminology throughout the official Symfony resources. In https://github.com/symfony/symfony-docs/pull/14731 and during discussions with @kbond, @Nyholm and @weaverryan, we came up with these terms:

* *Unit tests*, the PHPUnit tests we all know
* *Integration tests*, either PHPUnit tests without too much mocking or `KernelTestCase`. These tests test a combination of multiple classes
* *Application tests* (or "functional tests"), uses `WebTestCase` or `PantherTestCase`. These tests test a whole application, from booting Symfony to sending the response. Whether they use javascript or not is mostly an "implementation detail" of that test.

This slightly differs from the terms used here (integration is the non-javascript "application" variant, and e2e and the javascript "application" variant). I would love to hear your views on this subject (we need to have good terminology here, so it's open for all discussion imho).

dunglas: I commented about that in the Google Docs this morning. I do agree that we need to use the same terminology everywhere. However, I don't like that proposal very much.

1. we'll introduce new confusing terms while they are terms already broadly used and well-defined. I used unit/integration/e2e as in https://martinfowler.com/articles/practical-test-pyramid.html. Using unit/service/UI as in Mike Cohn's original pyramid would work too. But reinventing new terms add more confusion than it solves.
2. we need to make a clear difference between PHP-based integrations tests (`KernelTestCase` / `WebTestCase` / `ApiTestCase`) and UI and e2e tests (Panther). They don't serve the same purpose, and even if they can share the same API, with E2E test you also need specific helpers to interact with the JavaScript application: you need to be able to test that clicking on a button triggers an AJAX request that results the sending of a message...

Maybe this PR misses some test cases, as far as I know the following test cases exist:

* `TestCase`: basic PHPUnit test
* `KernelTestCase`:  to use Symfony services
* `WebTestCase`: to run browser-like scenarios, but don't execute client-side code
* `ApiTestCase`: to run API-oriented scenarios (needs API Platform)
* `PantherTestCase`:  to run e2e scenarios, using real-browsers or HTTP clients, and a real web server

Instead of unit/integration/e2e or unit/service/ui, maybe this generator could use allow to select the `TestCase` to use as base class? WDYT?

src/Resources/skeleton/test/Functional.tpl.php
jrushlow: psalm will also complain without declaring a return type.. Seeing as we declare return types for the majority of the generated methods, tests should be no different to keep consistency...

nicolas-grekas: the rise of the machines :)

src/Maker/MakeTest.php
jrushlow: I think a lot of users will question what `e2e` is. They may get its end to end, but not know exactly what that is or how it differs from a functional, integration, and/or acceptance test. I don't think we need to reinvent the wheel - but defiantly provide a brief statement of what that is + the usual "how to" in symfony docs.

weaverryan: I am also not very familiar with e2e - i'm familiar with "functional".

But, my main point is that this is a small detail: most people will not pass an argument. They will, instead, allow the maker to ask interactively what type they want. And when we do THAT, we can allow all the names (though, we will need to choose a "main" name.

src/Maker/MakeTest.php
```diff
@@ -0,0 +1,209 @@
...
+
+        if (null !== $input->getArgument('type')) {
+            $this->normalizeType($input);
+
+            return;
+        }
+
+        $input->setArgument(
+            'type',
+            $io->choice($command->getDefinition()->getArgument('type')->getDescription(), ['unit', 'integration', 'e2e'])
```
weaverryan, line 90: What about not re-using the `type` argument definition? I'd rather say something like this (which probably includes printing some thins before the question) - I'm actually re-using your suggestion from another comment, which I really like:

> Which test type would you like?
> 
> 1) TestCase: basic PHPUnit test
> 2) KernelTestCase: basic tests that have access to Symfony services
> 3) WebTestCase: to run browser-like scenarios, but that don't execute JavaScript code
> 4) ApiTestCase: to run API-oriented scenarios (needs API Platform)
> 5) PantherTestCase: to run e2e scenarios, using aa real-browser or HTTP client and a real web server

CHANGELOG.md
```diff
@@ -1,4 +1,12 @@
+1.29
```
OskarStark, line 1: Not 1.28?

weaverryan: 1.28 is out already - I forgot to update the CHANGELOG because there weren't any really significant changes. But I should indeed fix that. But this is correct

src/Maker/MakeTest.php
```diff
@@ -0,0 +1,196 @@
...
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Ryan Weaver <weaverryan@gmail.com>
+ */
+final class MakeTest extends AbstractMaker implements InputAwareMakerInterface
+{
+    /**
+     * @internal
+     */
+    const DESCRIPTIONS = [
```
weaverryan, line 39: `private const`? We require 7.1, so I think that' safe

src/Maker/MakeTest.php
```diff
@@ -0,0 +1,196 @@
...
+        'TestCase' => 'basic PHPUnit tests',
+        'KernelTestCase' => 'basic tests that have access to Symfony services',
+        'WebTestCase' => 'to run browser-like scenarios, but that don\'t execute JavaScript code',
+        'ApiTestCase' => 'to run API-oriented scenarios',
+        'PantherTestCase' => 'to run e2e scenarios, using aa real-browser or HTTP client and a real web server',
```
weaverryan, line 41: `aa` -> `a`

src/Maker/MakeTest.php
dunglas: I thought `private const` was newer than that! I update.

src/Maker/MakeTest.php
```diff
@@ -0,0 +1,189 @@
...
+        $input->setArgument(
+            'type',
+            $io->choice('Which test type would you like?', self::DESCRIPTIONS)
+        );
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $testClassNameDetails = $generator->createClassNameDetails(
+            $input->getArgument('name'),
```
weaverryan, line 122: Do we ever ask this interactively?

dunglas: Yes, it's delegated to the method in the abstract class.

tests/Maker/MakeFunctionalTestTest.php
```diff
@@ -15,19 +15,13 @@
...
 use Symfony\Bundle\MakerBundle\Test\MakerTestDetails;
 
+/**
+ * @group legacy
+ */
 class MakeFunctionalTestTest extends MakerTestCase
 {
     public function getTestDetails()
     {
-        yield 'functional_maker' => [MakerTestDetails::createTest(
```
dunglas, line 22: This test is failing because of a bug in `MakerTestCase`. If a previous test (such as our new `MakeTestTest`) installs Panther, then the autoloader is confused for some reason and detect the dependency has available, so the generated test extends `PantherTestCase`  instead of `WebTestCase`. But when the test is executed in isolation, this time `PantherTestCase` isn't available, and it fails.

As this class is now deprecated, I don't think that it is worth it to fix `MakerTestCase`.

weaverryan:
Thanks Kévin!

PR #810, opened 2021-02-06
Tweaks to the new make:test

weaverryan:
This follows #807 - I played with it locally and had some minor tweaks :).

Cheers!

### DIFF
--- a/src/Maker/MakeTest.php
+++ b/src/Maker/MakeTest.php
@@ -19,6 +19,7 @@ use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputAwareMakerInterface;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\BrowserKit\History;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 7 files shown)


src/Maker/MakeTest.php
```diff
@@ -77,10 +78,11 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
         }
 
         $command
-            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the test class (e.g. <fg=yellow>BlogPostTest</>)')
             ->addArgument('type', InputArgument::OPTIONAL, 'The type of test: '.implode(', ', $typesDesc))
+            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the test class (e.g. <fg=yellow>BlogPostTest</>)')
```
weaverryan, line 82: Once I played with this, I decided to reverse the arguments. It occurred to me that knowing the "type" as early as possible could help us. For example, I don't do it in this PR, but we could suggest different example filenames on the next question based on the type.

src/Maker/MakeTest.php
```diff
@@ -106,14 +108,40 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+                $io->choice('Which test type would you like?', self::DESCRIPTIONS)
+            );
+        }
 
-            return;
+        if ('ApiTestCase' === $input->getArgument('type') && !class_exists(ApiTestCase::class)) {
+            $io->warning([
+                'API Platform is required for this test type. Install it with',
+                'composer require api',
+            ]);
```
weaverryan, line 122: Unfortunately, the dependency system is called before the interactive step. So previously, no error or warning was shown. That's actually ok - I like showing a warning better than exploding with an error that they must stop and run the `composer require` command before starting the `make:test` command again.

src/Maker/MakeTest.php
```diff
@@ -106,14 +108,40 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+        }
+
+        if (null === $input->getArgument('name')) {
+            $io->writeln([
+                '',
+                'Choose a class name for your test, like:',
+                ' * <fg=yellow>UtilTest</> (to create tests/UtilTest.php)',
+                ' * <fg=yellow>Service\\UtilTest</> (to create tests/Service/UtilTest.php)',
+                ' * <fg=yellow>\\App\Tests\\Service\\UtilTest</> (to create tests/Service/UtilTest.php)',
```
jrushlow, line 138: Could we (should we) suggest the class based on the test? e.g. if the type is `UnitTest` - we suggest `'App\\Tests\\UnitTests\\{NAME}` or something along those lines?

src/Maker/MakeTest.php
jrushlow: Agreed, seeing as we're not actually collecting a bunch of input that is dependent on `Api Platform` at this time - the warning makes sense..

src/Maker/MakeTest.php
weaverryan: I thought about that too. Let's save it for a future PR... I don't think there is clear agreement on the directory conventions of these things. I personally don't sub-divide things into `UnitTest` or `IntegrationTest` directories, but I know that some people do this.

src/Maker/MakeTest.php
```diff
@@ -91,13 +93,13 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
         switch ($currentCommand) {
             case 'make:unit-test':
                 $input->setArgument('type', 'TestCase');
-                $io->caution('The "make:unit-test" command is deprecated, use "make:test" instead.');
+                $io->warning('The "make:unit-test" command is deprecated, use "make:test" instead.');
 
                 return;
 
             case 'make:functional-test':
                 $input->setArgument('type', trait_exists(PantherTestCaseTrait::class) ? 'WebTestCase' : 'PantherTestCase');
```
jrushlow, line 101: Just caught this - shouldn't this be asked? What if I want a `WebTestCase` to check the status codes of a couple routes but want `PantherTestCase` for browser based tests.

weaverryan: I guess even before this, if you used `make:functional-test`, we already looked to see if you had Panther available and then used it automatically. It's a minor thing, since this is deprecated. The new system of actually asking is much better. But I'll keep this the way it is now... which is the way it was before mostly.

jrushlow (approved):

Issue #777, opened 2021-01-08
[Docker][database] the maker does not work because of the YamlSourceManipulator


lyrixx:
I used https://github.com/dunglas/symfony-docker but I edit the files to use .yaml instead of .yml ([see the PR](https://github.com/dunglas/symfony-docker/pull/103)).

But I get an error when running the maker: 

```
>…m/lyrixx/SFLive-Paris2016-Workflow(docker +) bin/console make:docker:data -v

- Docker Compose Setup-
-----------------------

 We found your existing docker-compose.yaml: Let's update it!

 Which database service will you be creating?:
  [0] MySQL
  [1] MariaDB
  [2] Postgres
 > 2

 For a list of supported versions, check out https://hub.docker.com/_/postgres

 What version would you like to use? [alpine]:
 > 



In YamlSourceManipulator.php line 868:
                                                                     
  [Symfony\Bundle\MakerBundle\Util\YamlManipulationFailedException]  
  Cannot find the original value "80:80"                             
                                                                     

Exception trace:
  at /home/gregoire/dev/github.com/lyrixx/SFLive-Paris2016-Workflow/vendor/symfony/maker-bundle/src/Util/YamlSourceManipulator.php:868
 Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator->findEndPositionOfValue() at /home/gregoire/dev/github.com/lyrixx/SFLive-Paris2016-Workflow/vendor/symfony/maker-bundle/src/Util/YamlSourceManipulator.php:568
 Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator->advanceBeyondValue() at /home/gregoire/dev/github.com/lyrixx/SFLive-Paris2016-Workflow/vendor/symfony/maker-bundle/src/Util/YamlSourceManipulator.php:215
 Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator->updateData() at /home/gregoire/dev/github.com/lyrixx/SFLive-Paris2016-Workflow/vendor/symfony/maker-bundle/src/Util/YamlSourceManipulator.php:207
 Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator->updateData() at /home/gregoire/dev/github.com/lyrixx/SFLive-Paris2016-Workflow/vendor/symfony/maker-bundle/src/Util/YamlSourceManipulator.php:207
[...]

PR #782, opened 2021-01-11
add failing test for issue 777

jrushlow:
WIP
fixes #777

PR #816, opened 2021-02-16
[YSM] fix array key regex

jrushlow:
This fixes an issue if an existing array has a `:` within the value, `YamlSourceManipulator` could mistake part of the value for the array key when determining position. 

E.g.
```yaml
config:
    - '80:80'
```

YSM would pick the `0:` within `80:80` as the array key.


OskarStark (approved):

jrushlow:
This should fix #777. But does not handle passing a literal `"80:80"` as a new array value and getting `'80:80` in return. PR #782 is a WIP for this "issue" with an unknown future.

weaverryan:
Thanks for hunting this down Jesse!

PR #822, opened 2021-02-22
[make:command] lets use attributes if possible

### LINKED ISSUE #820: Generate ConsoleCommand attribute on PHP8
See https://github.com/symfony/symfony/pull/40234

jrushlow:
Console command attribute class renamed from `ConsoleCommand` => `AsCommand` https://github.com/symfony/symfony/pull/40556

closes #820

### DIFF
--- a/src/Maker/MakeCommand.php
+++ b/src/Maker/MakeCommand.php
@@ -16,6 +16,7 @@ use Symfony\Bundle\MakerBundle\DependencyBuilder;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Command\LazyCommand;
 use Symfony\Component\Console\Input\InputArgument;
@@ -63,6 +64,7 @@ final class MakeCommand extends AbstractMaker
             [
                 'command_name' => $commandName,
                 'set_description' => !class_exists(LazyCommand::class),
+                'use_command_attribute' => 80000 <= \PHP_VERSION_ID && class_exists(AsCommand::class),
             ]
         );
 
--- a/src/Resources/skeleton/command/Command.tpl.php
+++ b/src/Resources/skeleton/command/Command.tpl.php
@@ -2,6 +2,9 @@
 
 namespace <?= $namespace; ?>;
 
+<?php if ($use_attributes && $use_command_attribute): ?>
+use Symfony\Component\Console\Attribute\AsCommand;
+<?php endif; ?>
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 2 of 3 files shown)


src/Resources/skeleton/command/Command.tpl.php
```diff
@@ -9,11 +9,19 @@
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
+<?php if ($use_attributes): ?>
```
weaverryan, line 12: It's more complex than this. `$use_attributes` tells us if the PHP version supports attributes. But in this case, we *also* need to see if the version of Symfony has the `ConsoleCommand` class. Also, this will need a `use` statement on top.

I know we're in the process of fixing some test failures to due upstream changes (so it's hard to look at CI), but let's make sure that if we misconfigure this somehow (e.g. forget the `use` statement) that it *does* cause a test to fail.

src/Maker/MakeCommand.php
```diff
@@ -57,12 +58,23 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             sprintf('The "%s" command name is not valid because it would be implemented by "%s" class, which is not valid as a PHP class name (it must start with a letter or underscore, followed by any number of letters, numbers, or underscores).', $commandName, Str::asClassName($commandName, 'Command'))
         );
 
+        $attributeClassExists = false;
```
nicolas-grekas, line 61: `$use_command_attribute = 80000 <= PHP_VERSION_ID && class_exists(AsCommand::class)`

jrushlow: done

src/Maker/MakeCommand.php
```diff
@@ -63,6 +64,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             [
                 'command_name' => $commandName,
                 'set_description' => !class_exists(LazyCommand::class),
+                'use_command_attribute' => 80000 <= \PHP_VERSION_ID && class_exists(AsCommand::class),
```
jrushlow, line 67: Under a different PR, I think we can add a helper to the `PhpCompatUtil` to eliminate having to check the PHP version before calling `class_exists()`. But the complexity required to make that happen is beyond the scope of this PR.

https://bugs.php.net/bug.php?id=80938 explains why this must be done like this.

src/Resources/skeleton/command/Command.tpl.php
```diff
@@ -2,18 +2,29 @@
 
 namespace <?= $namespace; ?>;
 
+<?php if ($use_attributes && $use_command_attribute): ?>
```
weaverryan, line 5: Do we need to check both? `$use_command_attribute` already checks for PHP 8.0 or higher. So it seems like it fully describes which situation we are in, right?

jrushlow: I think we _could_ drop `$use_attributes` here, but then we would loose the ability to determine if attributes should be used based on the `platform-overrides`. Then again as I'm saying this, regardless of platform overrides, they wouldn't be able to use attributes anyway if PHP was not 8+ at runtime.. 

Thinking we drop `$use_attributes` for now and when I'm able to accomplish the helper in `PhpCompatUtil`, we can re introduce / refactor our logic here.

weaverryan: Ah, you're right about the `config.platform.php` thing. So then... shouldn't we keep the `$use_attributes`? Otherwise, it's possible that I'm using PHP 8... but my `config.platform.php` is set to 7.4... and suddenly we generate code that uses attributes?

weaverryan:
Thanks Jesse!

PR #825, opened 2021-02-24
[make:crud] use POST method instead of DELETE

jrushlow:
Starting in Symfony 5.3, `framework.http_method_override` defaults to `false`. This prevents CRUD form submission for `PUT`, `PATCH` &/or `DELETE` methods.  See symfony/recipes#892

~This PR conditionally sets  `http_method_override` to true when using `make:crud` if using Symfony 5.3+~

We stop using the `DELETE` method override in favor of using `POST` on the delete operation.

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -114,10 +114,10 @@ class <?= $class_name ?> extends <?= $parent_class_name; ?><?= "\n" ?>
     }
 
 <?php if ($use_attributes) { ?>
-    #[Route('/{<?= $entity_identifier ?>}', name: '<?= $route_name ?>_delete', methods: ['DELETE'])]
+    #[Route('/{<?= $entity_identifier ?>}', name: '<?= $route_name ?>_delete', methods: ['POST'])]
 <?php } else { ?>
     /**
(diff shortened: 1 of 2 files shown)


src/Maker/MakeCrud.php
```diff
@@ -95,6 +101,15 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
     {
+        if (Kernel::VERSION_ID >= 50300) {
+            $this->enableHttpMethodOverride($generator);
+
+            $io->newLine();
+            $overrideMessage = 'Enabled framework.http_method_override - see <href=https://symfony.com/doc/5.3/reference/configuration/framework.html#http-method-override>https://symfony.com/doc/5.3/reference/configuration/framework.html#http-method-override</>';
+            $io->text($overrideMessage);
+            $io->newLine();
+        }
```
weaverryan, line 111: You did a nice job with this. But I think the better solution is to stop relying on this setting. In other words, update `make:crud` to just send simple POST requests. I don't see any issue with this - I'm not sure it's very useful to imitate HTTP methods in a browser setting.

jrushlow: done

weaverryan:
Thanks Jesse!

PR #828, opened 2021-03-02
TEST - testing

jrushlow:
DO NOT MERGE

PR #829, opened 2021-03-02
[CI] run tests on ubuntu 18.04

jrushlow:
Use ubuntu 1804 for tests that rely on PHP `<7.4`. Github released base image `ubuntu-latest` which uses `ubuntu-20.04` - the 20.04 does not have PHP 7.0 - 7.3. Previously, `ubuntu-latest` was based on `ubuntu-18.04`

In maker this caused our stable tests against 7.1, .2 & .3 to fail with a PDO Driver Error. Which is ultimately because the base 18.04 image does not have the required PHP version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ jobs:
     coding-standards:
         name: "Coding Standards (${{ matrix.php-version }})"
 
-        runs-on: "ubuntu-latest"
+        runs-on: "ubuntu-18.04"
 
         strategy:
             fail-fast: false
@@ -52,7 +52,7 @@ jobs:
     test:
         name: "PHP ${{ matrix.php-version }} + symfony/skeleton@${{ matrix.symfony-skeleton-stability }}"
 
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-18.04
 
         services:
             mysql:


stof:
AFAIK, `shivammathur/setup-php` can install older PHP versions on ubunut-20-04. What changes is which pre-installed version is available.

The difference is that when a version is not pre-installed, properly configuring the list of extensions you need is more important (as there is no preinstalled extensions either). tests in https://github.com/symfony/maker-bundle/pull/828/checks?check_run_id=2016163291 are failing because they try to use PDO mysql without installing that extension.

jrushlow:
> AFAIK, shivammathur/setup-php can install older PHP versions on ubunut-20-04. What changes is which pre-installed version is available.

That is my understanding as well. We could continue to use the `ubuntu-latest` (20.04) then install php 7.1 - 7.3 and require the PDO extension. However using 18.04 negates the need to install the previous versions of PHP & the PDO extension.

weaverryan:
Thanks Jesse! I'm also interested in `shivammathur/setup-php` - it seems to be quite the standard these days. But I'm even happier to have a healthier test suite now ;)

jrushlow:
> I'm also interested in shivammathur/setup-php - it seems to be quite the standard these days.

We are using this now, but because of the upstream change to the GH default base image for `ubuntu-latest` and how `setup-php` is built on top of / configures those images (think of composer + autoscripts) - the "problem" appeared because we did not have the need to declare the PDO extension as it is included in the default base image.

PR #830, opened 2021-03-03
[make:validator] Fix @var typehint comments

mmarton:
Hi!

I found some malformed `@var` typehint comments  in the code annd fixed them.

(This is my first PR)

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -161,7 +161,7 @@ final class DoctrineHelper
         if ($this->isDoctrineInstalled()) {
             $allMetadata = $this->getMetadata();
 
-            /* @var ClassMetadata $metadata */
+            /** @var ClassMetadata $classname */
             foreach (array_keys($allMetadata) as $classname) {
                 $entityClassDetails = new ClassNameDetails($classname, $this->entityNamespace);
                 $entities[] = $entityClassDetails->getRelativeName();
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -161,7 +161,6 @@ final class DoctrineHelper
         if ($this->isDoctrineInstalled()) {
             $allMetadata = $this->getMetadata();
 
-            /** @var ClassMetadata $classname */
             foreach (array_keys($allMetadata) as $classname) {
                 $entityClassDetails = new ClassNameDetails($classname, $this->entityNamespace);
                 $entities[] = $entityClassDetails->getRelativeName();
--- a/src/Resources/skeleton/validator/Validator.tpl.php
+++ b/src/Resources/skeleton/validator/Validator.tpl.php
@@ -9,7 +9,7 @@ class <?= $class_name ?> extends ConstraintValidator
 {
     public function validate($value, Constraint $constraint)
     {
-        /** @var <?= $constraint_class_name ?> $constraint */
+        /* @var <?= $constraint_class_name ?> $constraint */
 
         if (null === $value || '' === $value) {
             return;
--- a/src/Resources/skeleton/validator/Validator.tpl.php
+++ b/src/Resources/skeleton/validator/Validator.tpl.php
@@ -9,7 +9,7 @@ class <?= $class_name ?> extends ConstraintValidator
 {
     public function validate($value, Constraint $constraint)
     {
-        /* @var $constraint \<?= $constraint_class_name ?> */
+        /** @var <?= $constraint_class_name ?> $constraint */
 
         if (null === $value || '' === $value) {
             return;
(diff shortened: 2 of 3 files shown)


src/Resources/skeleton/validator/Validator.tpl.php
```diff
@@ -9,7 +9,7 @@ class <?= $class_name ?> extends ConstraintValidator
 {
     public function validate($value, Constraint $constraint)
     {
-        /* @var $constraint \<?= $constraint_class_name ?> */
+        /** @var \<?= $constraint_class_name ?> $constraint */
```
mmarton, line 12: This was the most problematic, because phpstan kept failing on the generated validator classes and had to manually fix the typehints

src/Doctrine/DoctrineHelper.php
```diff
@@ -117,7 +117,7 @@ public function getEntitiesForAutocomplete(): array
         if ($this->isDoctrineInstalled()) {
             $allMetadata = $this->getMetadata();
 
-            /* @var ClassMetadata $metadata */
+            /** @var ClassMetadata $metadata */
```
mmarton, line 120: This one looks unnecessary, it hits a variable that doesn't exist

jrushlow: This should be `$allMetadata` and adding the second `*` breaks CI https://github.com/symfony/maker-bundle/pull/830/checks?check_run_id=2022138726

src/Resources/skeleton/validator/Validator.tpl.php
jrushlow, line 12:
```suggestion
        /** @var <?= $constraint_class_name ?> $constraint */
```

jrushlow: Even though this does give a small perf boost - its not consistent with the rest of the generated code in maker bundle

jrushlow (changes requested):
Thanks @mmarton!

src/Resources/skeleton/validator/Validator.tpl.php
mmarton: It was already there I just changed the order here

jrushlow: Correct, but CI would flag this as an issue if it could "see" the actual class being declared. We only use the root name space for a select list of PHP functions.

src/Doctrine/DoctrineHelper.php
mmarton: It applies to the $classname variable not the $allMetadata

mmarton:
@jrushlow I updated the commented parts, but it's still failing.

src/Doctrine/DoctrineHelper.php
```diff
@@ -117,7 +117,7 @@ public function getEntitiesForAutocomplete(): array
         if ($this->isDoctrineInstalled()) {
             $allMetadata = $this->getMetadata();
 
-            /* @var ClassMetadata $metadata */
+            /** @var ClassMetadata $classname */
```
jrushlow, line 120:
```suggestion
            /* @var ClassMetadata $classname */
```

src/Resources/skeleton/validator/Validator.tpl.php
```diff
@@ -9,7 +9,7 @@ class <?= $class_name ?> extends ConstraintValidator
 {
     public function validate($value, Constraint $constraint)
     {
-        /* @var $constraint \<?= $constraint_class_name ?> */
+        /** @var <?= $constraint_class_name ?> $constraint */
```
jrushlow, line 12:
```suggestion
        /* @var <?= $constraint_class_name ?> $constraint */
```

jrushlow (changes requested):
@mmarton This looks good! I believe there is a bug in cs fixer that is causing the tests to fail. Using the inline comment `/*` instead of the inline docBlock `/**` will take care of our CI and everything will be all set.

src/Doctrine/DoctrineHelper.php
jrushlow, line 120: Wait, I just realized, this is wrong. This docBlock is not needed. `$allMetadata` is an array like
```php
$allClassMetadata = ['App\Entity\User' => Doctrine\ORM\Mapping\ClassMetadata];
```
If anything we would do something like
```diff
  if ($this->isDoctrineInstalled()) {
+ /** @var array<string, ClassMetadata> $allMetadata */
  $allMetadata = $this->getMetadata();
```

But as this class is `internal` && `$getMetaData()` specifies a return type - we should just remove this docBlock all together.
```suggestion
```

jrushlow (changes requested):

src/Resources/skeleton/validator/Validator.tpl.php
mmarton: But that would make my fix completely pointless. If I'm not using docblock phpstorm or phpstan won't recognise this as a typehint, just as a regualr comment...
image
vs
image

mmarton: @jrushlow could you help me with this one pls?

jrushlow (approved):
Thank you @mmarton!

PR #835, opened 2021-03-09
require twig as a direct dependency

jrushlow:


nicolas-grekas:
Doesn't this mean that controller will now always generate the "twig version", and never the json one anymore, which would then become dead-code for this bundle?

wouterj:
This same argument was also given by @weaverryan [in the past](https://github.com/symfony/maker-bundle/issues/33#issuecomment-346075403):

> I disagree with adding Twig as a dependency. I don't want API-only projects to end up with a templates/ directory just so they can generate some guard authenticators or something else.

Is there anything that changed regarding this topic?

jrushlow:
@weaverryan and I were talking about this a few days ago with the main driver being that in tests - if we `->addExtraDependencies('twig')` we may potentially be covering up missing dependencies that should be required in maker. 

>Doesn't this mean that controller will now always generate the "twig" version, and never the "json" one anymore, which would then become dead-code for this bundle?

We may have to explicitly ask something along the lines of "Do you want json, twig, or both to be generated...." But with `make:controller` we have the `--no-template` option that skips generating the templates.. i think that this could be better suited as a question vs an arg for better visibility.

>I don't want API-only projects to end up with a templates/ directory just so they can generate some guard authenticators or something else.

Thats a pretty good point. I'm not sure what the better tradeoff here would be, generating an empty directory in API only apps (which may end up using templates in the future anyways) or having "better tests"? Definitely open to ideas / solutions on how to achieve the best of both worlds.

For additional context, `make:auth`, `make:controller`, && `make:reset-password` are the only 3 makers affected by this. But `reset-password` will soon have the ability to generate templates for api and/or twig based apps

jrushlow:
An alternative would be to keep things as they're now, and anytime we have the need to generate twig templates is implement an exit early / warn early strategy for these makers if twig doesn't exist (like we do in `make:reset-password`).

e.g. 
1) ask if they want to gen a twig template
2) check if twig xor twig-bundle exists
3) exit as soon as possible "You must have twig installed.." || "Warning - Twig is required but you can finish using maker..."

We will also face the same dilemma for other makers/tests that heavily rely on extra dependencies.. e.g. `security-bundle`, `reset-bundle`, etc.. `reset-password` is not a biggie as theres only 1 maker that requires that, but something like `security-bundle` is used by several makers and is another contender to be a MakerBundle dependecy

wouterj:
> Ryan and I were talking about this a few days ago with the main driver being that in tests - if we ->addExtraDependencies('twig') we may potentially be covering up missing dependencies that should be required in maker.

I'm not sure if I can follow this:
* Some makers require twig for some code paths (e.g. if you choose the login form in `make:auth`).
* This means that, for this maker to succeed the application does require `symfony/twig-bundle` already.
* This means that requiring `twig` in the tests only replicates the state of a user's application when running the maker

In which situation would this lead to possible missing dependencies?

I do think it's good to run makers with optional dependencies against a clean application and see if it shows nice error messages. I looked at `make:auth`, and it seems to already do so in case Twig is missing.

PR #837, opened 2021-03-11
Extend Yaml manipulator add_item_and_sort test

7ochem:
This case is the subject of issue #618 

Apparently this does work now, because it passes the test. Maybe it still is a good idea to extend this test so no bugs will creep in in the future


jrushlow (changes requested):
I would prefer that we create a new test case for this change. In the original test, we are sorting the array using the `SORT_REGULAR` algorithm. This PR proposes we sort the array using a different algorithm.  

Something along the lines of `add_item_and_sort_natural.test` would probably work.

jrushlow (approved):
Thanks @7ochem

PR #843, opened 2021-03-21
[BC BREAK] fix (DoctrineHelper): Doctrine bundle 2.3.0 compatibility

### LINKED ISSUE #841: [ERROR] Only annotation mapping is supported by make:entity after upgrading bundles
After upgrading bundles like this:
```
Upgrading doctrine/doctrine-bundle (2.2.3 => 2.3.0)
Upgrading doctrine/migrations (3.1.0 => 3.1.1)
Upgrading symfony/doctrine-bridge (v5.2.4 => v5.2.5)
Upgrading symfony/maker-bundle (v1.29.1 => v1.30.0)  
```

when I try to change or add a new entity like:

`php bin/console make:entity MyEntity` I get error

```
Only annotation mapping is supported by make:entity, but the <info>App\Entity\MyEntity</info>       
class uses a different format. If you would like this command to generate the properties & getter/setter       
methods, add your mapping configuration, and then re-run this command with the <info>--regenerate</info>       
flag. 
```

When I have downgraded to previous version it works OK.

Geekimo:
Fixes #841 

Doctrine Bundle 2.3.0 introduced a new Mapping\MappingDriver class, which doesn't give access to the MappingDriverChain anymore.
A PR has been made for it [1304](https://github.com/doctrine/DoctrineBundle/pull/1304) and is required for this to work.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "composer/semver": "^3.0@dev",
-        "doctrine/doctrine-bundle": "^1.8|^2.0",
+        "doctrine/doctrine-bundle": "^2.0",
         "doctrine/orm": "^2.3",
         "friendsofphp/php-cs-fixer": "^2.8",
         "friendsoftwig/twigcs": "^4.1.0|^5.0.0",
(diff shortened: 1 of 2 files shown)


Geekimo:
Tests failing actually because fix is missing on Doctrine bundle side.

nicolas-grekas:
Closing in favor of https://github.com/symfony/maker-bundle/pull/844, thanks for submitting!

PR #847, opened 2021-03-23
Fix dealing with Definition objects in SetDoctrineAnnotatedPrefixesPass

### LINKED ISSUE #845: Fatal error on cache:clear with 1.30.1
Hello

The added `SetDoctrineAnnotatedPrefixesPass` has an issue on line 41
```php
$isAnnotated = 'doctrine.orm.'.$managerName.'_annotation_metadata_driver' === (string) $arguments[0];
```

Object of class Symfony\Component\DependencyInjection\Definition could not be converted to string

This happens on symfony 4.4.20, I'm assuming it applies to all versions of symfony 4.4
The __toString method doesn't exist in class Definition in 4.4 so that's the issue.

nicolas-grekas:
Fix #845

Can happen when [`DoctrineOrmMappingsPass::createAnnotationMappingDriver()`](https://github.com/doctrine/DoctrineBundle/blob/155a870ecbd035963e6b450e0f235e69c48cab30/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php#L124) is used in a third-party bundle.

### DIFF
--- a/src/DependencyInjection/CompilerPass/SetDoctrineAnnotatedPrefixesPass.php
+++ b/src/DependencyInjection/CompilerPass/SetDoctrineAnnotatedPrefixesPass.php
@@ -11,8 +11,11 @@
 
 namespace Symfony\Bundle\MakerBundle\DependencyInjection\CompilerPass;
 
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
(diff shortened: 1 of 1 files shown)


sadikoff (approved):

weaverryan:
Thanks Nicolas!

PR #824, opened 2021-02-23
[make:auth] support new password hasher interface

jrushlow:
- adds support for the new `UserPasswordHasherInterface` in `make:auth`. If the new interface exists, it is automatically used.

- introduces the concept of `TemplateClassDetails` that has getter methods that return preformatted strings for our most common template variable types. 

Depends on: 
- [x] #849 
- [x] #923 

fixes #821

PR #849, opened 2021-03-24
[make:user] user entities implement PasswordAuthenticatedUserInterface

jrushlow:
Starting in Symfony 5.3 - PR https://github.com/symfony/symfony/pull/40267 introduces the new `PasswordAuthenticatedUserInterface` ~& `LegacyPasswordAuthenticatedUserInterface`~ aimed at decoupling passwords from the `UserInterface`.

- Generated `User` entity implements `PasswordAuthenticatedUserInterface` if it exists.
- Adds correct type hint to the generated `UserRepository::upgradePassword()` method.

Authentication related changes are handled in #824

// cc @chalasr

### DIFF
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -15,6 +15,8 @@ use Doctrine\Common\Persistence\ManagerRegistry as LegacyManagerRegistry;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Util\ClassNameDetails;
+use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
 
 /**
  * @internal
@@ -69,6 +71,15 @@ final class EntityClassGenerator
     {
         $shortEntityClass = Str::getShortClassName($entityClass);
         $entityAlias = strtolower($shortEntityClass[0]);
+
+        $passwordUserInterfaceName = UserInterface::class;
+
+        if (interface_exists(PasswordAuthenticatedUserInterface::class)) {
+            $passwordUserInterfaceName = PasswordAuthenticatedUserInterface::class;
+        }
+
+        $interfaceClassNameDetails = new ClassNameDetails($passwordUserInterfaceName, 'Symfony\Component\Security\Core\User');
+
         $this->generator->generateClass(
             $repositoryClass,
             'doctrine/Repository.tpl.php',
@@ -77,6 +88,7 @@ final class EntityClassGenerator
                 'entity_class_name' => $shortEntityClass,
                 'entity_alias' => $entityAlias,
                 'with_password_upgrade' => $withPasswordUpgrade,
+                'password_upgrade_user_interface' => $interfaceClassNameDetails,
                 'doctrine_registry_class' => $this->managerRegistryClassName,
                 'include_example_comments' => $includeExampleComments,
             ]
--- a/src/Security/UserClassBuilder.php
+++ b/src/Security/UserClassBuilder.php
@@ -13,6 +13,8 @@ namespace Symfony\Bundle\MakerBundle\Security;
 
 use PhpParser\Node;
 use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
 
 /**
@@ -22,7 +24,7 @@ use Symfony\Component\Security\Core\User\UserInterface;
  */
 final class UserClassBuilder
 {
-    public function addUserInterfaceImplementation(ClassSourceManipulator $manipulator, UserClassConfiguration $userClassConfig)
+    public function addUserInterfaceImplementation(ClassSourceManipulator $manipulator, UserClassConfiguration $userClassConfig): void
     {
         $manipulator->addInterface(UserInterface::class);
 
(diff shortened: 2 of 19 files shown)


chalasr (approved):
Looks good to me. Thanks a lot @jrushlow!

src/Doctrine/EntityClassGenerator.php
```diff
@@ -69,6 +71,15 @@ public function generateRepositoryClass(string $repositoryClass, string $entityC
...
         $shortEntityClass = Str::getShortClassName($entityClass);
         $entityAlias = strtolower($shortEntityClass[0]);
+
+        $passwordUserInterfaceName = UserInterface::class;
+
+        if (interface_exists(PasswordAuthenticatedUserInterface::class)) {
+            $passwordUserInterfaceName = PasswordAuthenticatedUserInterface::class;
+        }
+
+        $interfaceClassNameDetails = new ClassNameDetails($passwordUserInterfaceName, 'Symfony\Component\Security\Core\User');
```
weaverryan, line 81: Why do we need `ClassNameDetails` here? Can't we just use the normal class strings?

You might be doing this due to the logic in the template to figure out *which* location the `use` statement should be. I'd prefer passing in a second flag to know that - e.g. `is_using_password_authenticated_interface`. That should simplify the template

src/Security/UserClassBuilder.php
```diff
@@ -22,21 +25,39 @@
...
 
         $this->addGetRoles($manipulator, $userClassConfig);
 
-        $this->addGetPassword($manipulator, $userClassConfig);
-
-        $this->addGetSalt($manipulator, $userClassConfig);
+        if (06000 < Kernel::VERSION_ID) {
+            $this->addGetPassword($manipulator, $userClassConfig);
+            $this->addGetSalt($manipulator, $userClassConfig);
+        }
```
weaverryan, line 39: I understand why it's written this way, but I'm having a hard time following the flow.

Could we make `addUserPasswordInterface()` private and call it from `addUserInterfaceImplementation()` (we could rename both methods if we want to for clarity). Then `addUserPasswordInterface()` might be called `addPasswordImplementation()` and it would have all the logic for adding the interface (if needed) and generating the fields. And we wouldn't need to check if the `getPassword()` method already exists.

src/Security/UserClassBuilder.php
```diff
@@ -22,21 +25,39 @@
...
         $this->addGetUsername($manipulator, $userClassConfig);
 
         $this->addGetRoles($manipulator, $userClassConfig);
 
-        $this->addGetPassword($manipulator, $userClassConfig);
-
-        $this->addGetSalt($manipulator, $userClassConfig);
+        if (06000 < Kernel::VERSION_ID) {
+            $this->addGetPassword($manipulator, $userClassConfig);
+            $this->addGetSalt($manipulator, $userClassConfig);
```
weaverryan, line 38: In both `getPassword()` and `getSalt()`, when using 5.3 or 5.4, the messaging should change. For `getSalt()` it should have some comment about how the method is deprecated - and you can remove it once using 6.0. For `getPassword()`, the same (unless they are implementing the password interface because they actually need it).

tests/Maker/MakeUserTest.php
```diff
@@ -37,6 +37,21 @@ public function getTestDetails()
...
 
+        yield 'user_security_entity_with_password_authenticated_user_interface' => [MakerTestDetails::createTest(
+            $this->getMakerInstance(MakeUser::class),
+            [
+                // user class name
+                'User',
+                'y', // entity
+                'email', // identity property
+                'y', // with password
+                'n', // legacy password interface in >5.3
```
weaverryan, line 48: What question is the answers?

tests/Security/fixtures/expected/PasswordAuthenticatedUser.php
```diff
@@ -0,0 +1,29 @@
...
+class PasswordAuthenticatedUserFixture implements PasswordAuthenticatedUserInterface
+{
+    /**
+     * @var string The hashed password
+     * @ORM\Column(type="string")
+     */
+    private $password;
+
+    /**
+     * @see UserInterface
```
weaverryan, line 16:
```suggestion
     * @see PasswordAuthenticatedUserInterface
```

jrushlow: done

src/Security/UserClassBuilder.php
jrushlow: done

src/Doctrine/EntityClassGenerator.php
jrushlow: We need the full class name for the `use` statement and the shortName for the type hint in `upgradePassword()`.

 We currently already have a `$with_password_upgrade` var available in the templates - we could add a `$is_using_password_authenticated_interface` var as well, but i think that would make things even messier in the template when we need to determine the sort order of the use statements.  Ultimately what we really need is a use statement sorter to solve the messy part - but I'm open to ideas.

Also to consider, #824  introduces the concept of a `TemplateClassDetails` object the returns formatted strings for the particular use case. e.g.

```php
$userDetails = new TemplateClassDetails(User::class, true);    // second arg === $isTyped

$userDetails->getUseStatement();
// use App\Entity\User;

$userDetails->getConstructorArg();
// User $user

$userDetails->getMethodArg();  // similar to the above but spacing is slightly different
// User $user

$userDetails->getVariable();
// $user

$userDetails->getProperty();
// $this->user

// All of the above add appropriate spacing, casing, trailing commas, new lines, etc..
```
While 824 is not ready yet, ideally _if_ we merge that in with, a lot of stuff in this PR will be refactored to leverage the new helper and cleanup the templates.

tests/Maker/MakeUserTest.php
jrushlow: umm good question.. think that was for an earlier WIP that i reverted.. 

removed

src/Security/UserClassBuilder.php
jrushlow: Updated the doc blocks for both methods to `This method can be removed in Symfony 6.0 - is not needed for apps that do not check user passwords.` unless the a user is checking passwords.

For `getSalt()`, once https://github.com/symfony/symfony-docs/issues/15065 is completed, I figure we can refactor the docBlocks to point to that doc vs pointing to an interface that may not exist / have a lengthy explanation.

src/Doctrine/EntityClassGenerator.php
weaverryan: Ok then!

> Ultimately what we really need is a use statement sorter to solve the messy part

👍

weaverryan:
Thanks Jesse!

PR #851, opened 2021-03-24
[tests][make:migration] handle shebang line in output

jrushlow:
PHP 7.4 dev tests failing due to 

```
1) Symfony\Bundle\MakerBundle\Tests\Maker\MakeMigrationTest::testExecute with data set "migration_with_previous_migration_question" (Symfony\Bundle\MakerBundle\Test\MakerTestDetails Object (...))
Asserting that very specific output is shown - some should be hidden
Failed asserting that actual size 15 matches expected size 14.
```

image

This PR accounts for the shebang line in the make:migration test assertion.

### DIFF
--- a/tests/Maker/MakeMigrationTest.php
+++ b/tests/Maker/MakeMigrationTest.php
@@ -76,7 +76,15 @@ class MakeMigrationTest extends MakerTestCase
             ->assert(function (string $output, string $directory) {
                 $this->assertStringContainsString('You have 1 available migrations to execute', $output);
                 $this->assertStringContainsString('Success', $output);
-                $this->assertCount(14, explode("\n", $output), 'Asserting that very specific output is shown - some should be hidden');
+
+                $output = explode("\n", $output);
+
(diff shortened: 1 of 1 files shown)


jrushlow:
I'm not sure what component introduces the shebang line - so this may be a patch we should not merge if this is a bug.

tests/Maker/MakeMigrationTest.php
```diff
@@ -76,7 +76,15 @@ public function getTestDetails()
             ->assert(function (string $output, string $directory) {
                 $this->assertStringContainsString('You have 1 available migrations to execute', $output);
                 $this->assertStringContainsString('Success', $output);
-                $this->assertCount(14, explode("\n", $output), 'Asserting that very specific output is shown - some should be hidden');
+
+                $output = explode("\n", $output);
+
+                $expectedCount = str_contains($output[0], 'php') ? 15 : 14;
```
weaverryan, line 82: Would it be better to filter out the new line? Especially if the shebang is printing to the console, we don't really want that.

jrushlow:
fixed upstream

PR #853, opened 2021-03-29
[make:voter] generate type hints

jrushlow:
starting in Symfony 5.0 the abstract voter methods introduced type hints (https://github.com/symfony/symfony/commit/8c46b95ec2958cec3e3f7e6dab858c4c29b20770). We now conditionally generate the `string` type hint for the `$attribute` argument.

Minor cleanup of the maker itself.

### DIFF
--- a/src/Resources/skeleton/security/Voter.tpl.php
+++ b/src/Resources/skeleton/security/Voter.tpl.php
@@ -8,7 +8,7 @@ use Symfony\Component\Security\Core\User\UserInterface;
 
 class <?= $class_name ?> extends Voter
 {
-    protected function supports($attribute, $subject)
+    protected function supports(<?= $use_type_hints ? 'string ' : null ?>$attribute, $subject): bool
     {
         // replace with your own logic
         // https://symfony.com/doc/current/security/voters.html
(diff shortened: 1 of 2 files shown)


src/Resources/skeleton/security/Voter.tpl.php
```diff
@@ -8,15 +8,15 @@
 
 class <?= $class_name ?> extends Voter
 {
-    protected function supports($attribute, $subject)
+    protected function supports(<?= $use_type_hints ? 'string ' : null ?>$attribute, $subject): bool
```
weaverryan, line 11: I think we have this conversation on every PR :p. Would generating the types break if I were using Symfony 4.4, where the parent class does NOT have those hints? If so, then we should do this - but only if we're on Symfony 5.

jrushlow: No - i keep forgetting how this works myself, but if the abstract does not declare a return type, the child is able to declare one. For arguments it would break iirc

>Note:
>
>When overriding a parent method, the child's method must match any return type declaration on the parent. If the parent doesn't define a return type, then the child method may do so.

jrushlow: >If so, then we should do this - but only if we're on Symfony 5.

The new `$use_type_hints` var checks if we're using symfony 5 or greater (double check my comparison operator :D) for the arg. But this check isnt needed for the return type.

weaverryan: > No - i keep forgetting how this works myself, but if the abstract does not declare a return type, the child is able to declare one. For arguments it would break iirc...

Ah yes. It's legal to make a return type "more restrictive" than your parent, but it is NOT legal to make an argument type stricter (which would now mean you accept LESS input)

weaverryan:
Thanks Jesse!

PR #855, opened 2021-03-29
[reset-password] allow anyone to access check email

### LINKED ISSUE #808: ResetPassword: broken flow for non-existing account
When trying to reset the password for an account that does not exist (wrong email address), the user is redirected to the same form without an error message, whereas, from what I read in the code, the intent was to show them the confirmation screen anyway (so we don't reveal if the account exists or not).

The current flow for an non-existing account is:
1. user submits the password reset request form with an unknown email address (form is valid, though) ;
2. user is redirected to the confirmation page (app_check_email)
  `// Do not reveal whether a user account was found or not.`
3. there is no reset password token, so user is redirected to the password reset request form back again:
  `// We prevent users from directly accessing this page`

Not sure how to fix this because:
- in the current state, the bad redirection reveals that there is no account, so the non-disclosure promise is broken;
- for UX, it's even worse than assuming the disclosure because there is no explicit error message displayed to the user;
- the confirmation page needs a token, and we don't want to generate a token for a non-existing account, I guess - or do we?

jrushlow:
fixes #808 

- [x] Requires https://github.com/SymfonyCasts/reset-password-bundle/pull/156


src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -75,9 +75,9 @@ public function request(Request $request, MailerInterface $mailer): Response
 <?php } ?>
     public function checkEmail(): Response
     {
-        // We prevent users from directly accessing this page
+        // Generate a fake token - the user does not exist or someone hit this page directly.
         if (null === ($resetToken = $this->getTokenObjectFromSession())) {
-            return $this->redirectToRoute('app_forgot_password_request');
+            $resetToken = $this-resetPasswordHelper->generateFakeResetToken();
```
weaverryan, line 80: We can only do this if the method exists. Honestly, we should probably just ensure that the user is using a new enough version of reset bundle. I'd rather say "Hey! You need to upgrade!" vs generating the old code... and then they don't have the better code even after they upgrade reset later.

jrushlow: agreed and done

src/Maker/MakeResetPassword.php
```diff
@@ -96,12 +95,10 @@ public function configureDependencies(DependencyBuilder $dependencies): void
...
-        // reset-password-bundle 1.3 includes helpers to get/set a ResetPasswordToken object from the session.
-        // we need to check that version 1.3 is installed
-        if (class_exists(ResetPasswordToken::class)) {
-            if (!method_exists(ResetPasswordControllerTrait::class, 'getTokenObjectFromSession')) {
-                throw new RuntimeCommandException('Please upgrade symfonycasts/reset-password-bundle to version 1.3 or greater.');
-            }
+        // reset-password-bundle 1.6 includes the ability to generate a fake token.
+        // we need to check that version 1.6 is installed
+        if (class_exists(ResetPasswordHelper::class) && !method_exists(ResetPasswordHelper::class, 'generateFakeResetToken')) {
+            throw new RuntimeCommandException('Please use "composer upgrade symfonycasts/reset-password-bundle" - version 1.6 or greater is required.');
```
weaverryan, line 101:
```suggestion
            throw new RuntimeCommandException('Please run "composer upgrade symfonycasts/reset-password-bundle". Version 1.6 or greater of this bundle is required.');
```

src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
```diff
@@ -75,9 +75,9 @@ public function request(Request $request, MailerInterface $mailer): Response
 <?php } ?>
     public function checkEmail(): Response
     {
-        // We prevent users from directly accessing this page
+        // Generate a fake token - the user does not exist or someone hit this page directly.
```
weaverryan, line 78:
```suggestion
        // Generate a fake token if the user does not exist or someone hit this page directly.
        // This prevents exposing whether or not a user was found with the given email address or not
```

weaverryan (commented):
very minor stuff!

weaverryan:
Thanks Jesse!

PR #858, opened 2021-04-01
[make:controller] avoid require doctrine/annotation when can use attributes

Jibbarth:
Hi makers :wave: !

This PR allows to make controller when we can use PHP8 attributes without forcing to install `doctrine/annotation`

But we should be aware that currently, without `doctrine/annotation`, we don't have a nice flex recipe that enable routing annotation read on Controller folder.

See https://github.com/symfony/recipes/pull/916 for reference.

So maybe we should create the config/routes/annotation.yaml in case we can use attributes AND file does not exist ? WDYT ?

### DIFF
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -15,7 +15,6 @@ use Doctrine\Common\Annotations\Annotation;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
-use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Str;
@@ -38,10 +37,10 @@ final class MakeController extends AbstractMaker
 {
     private $phpCompatUtil;
 
-    public function __construct(FileManager $fileManager, PhpCompatUtil $phpCompatUtil = null)
+    public function __construct(PhpCompatUtil $phpCompatUtil = null)
     {
         if (null === $phpCompatUtil) {
-            $phpCompatUtil = new PhpCompatUtil($fileManager);
+            @trigger_error(sprintf('Passing a "%s" instance is mandatory since version 1.42.0', PhpCompatUtil::class), \E_USER_DEPRECATED);
         }
 
         $this->phpCompatUtil = $phpCompatUtil;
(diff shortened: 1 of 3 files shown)


src/Maker/MakeController.php
```diff
@@ -29,6 +31,19 @@
...
 final class MakeController extends AbstractMaker
 {
+    private $phpCompatUtil;
+
+    public function __construct(FileManager $fileManager, PhpCompatUtil $phpCompatUtil = null)
+    {
+        if (null === $phpCompatUtil) {
+            $phpCompatUtil = new PhpCompatUtil($fileManager);
+
+            trigger_deprecation('symfony/maker-bundle', '1.31', 'Initializing MakeController without providing an instance of PhpCompatUtil is deprecated.');
```
Jibbarth, line 41: This is heavily inspired on https://github.com/symfony/maker-bundle/blob/main/src/Generator.php#L40, tell me if it's not needed in this case

dunglas (approved):

src/Maker/MakeController.php
dunglas: I'm not sure if it is worth it. When `maker-bundle` will require PHP 8 as a minimum version, we'll most likely drop the dependency to `PhpCompatUtil`, so in my opinion it's not useful to make this dependency mandatory.

jrushlow: No need for the deprecation here. `PhpCompatUtil` is internal & `MakeController` is final - I don't see a use case where someone would use a version >=1.31 and not be able to initiate `MakeController` w/o the utility.

jrushlow (changes requested):
Thanks @Jibbarth This PR will have to wait for the resolution of symfony/recipes#916 before it can be merged.

> So maybe we should create the config/routes/annotation.yaml in case we can use attributes AND file does not exist ? WDYT ?

I'm not a fan of this idea. That responsibility falls on recipes / flex in my opinion.

src/Maker/MakeController.php
Jibbarth: Deprecation removed :ok_hand:

src/Maker/MakeController.php
```diff
@@ -29,6 +31,17 @@
...
 {
+    private $phpCompatUtil;
+
+    public function __construct(FileManager $fileManager, PhpCompatUtil $phpCompatUtil = null)
+    {
+        if (null === $phpCompatUtil) {
+            $phpCompatUtil = new PhpCompatUtil($fileManager);
+        }
+
+        $this->phpCompatUtil = $phpCompatUtil;
```
ismail1432, line 42:
```suggestion
        $this->phpCompatUtil = $phpCompatUtil ?? new PhpCompatUtil($fileManager);
```

jrushlow (approved):
Thank you @Jibbarth! we can finally get this merged in!

jrushlow:
Note: To preserve BC, `doctrine/annotations` is still required for Symfony 5.4 _IF_ the `MakeController::class` is used w/o passing an instance of `PhpCompatUtil::class` via the constructor.

PR #860, opened 2021-04-03
Add support for Symfony UX Turbo

dunglas:
This PR adds a new `--broadcast` to the `make:entity` command. This generates the required template to use the broadcasting feature of UX Turbo, and also add the `#[Broadcast]` attribute to the generated entity.

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -35,6 +35,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Question\Question;
+use Symfony\UX\Turbo\Attribute\Broadcast;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
@@ -84,6 +85,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         $command
             ->addArgument('name', InputArgument::OPTIONAL, sprintf('Class name of the entity to create or update (e.g. <fg=yellow>%s</>)', Str::asClassName(Str::getRandomTerm())))
             ->addOption('api-resource', 'a', InputOption::VALUE_NONE, 'Mark this class as an API Platform resource (expose a CRUD API for it)')
+            ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Add the ability to broadcast entity updates using Symfony UX Turbo?')
             ->addOption('regenerate', null, InputOption::VALUE_NONE, 'Instead of adding new fields, simply generate the methods (e.g. getter/setter) for existing fields')
             ->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite any existing getter/setter methods')
             ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeEntity.txt'))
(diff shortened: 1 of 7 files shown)


src/Resources/skeleton/doctrine/broadcast_twig_template.tpl.php
```diff
@@ -0,0 +1,19 @@
...
+{% block update %}
+<turbo-stream action="update" target="<?= $class_name ?>_{{ entity.id }}">
+    <template>
+        #{{ entity.id }} updated
+    </template>
+</turbo-stream>
+{% endblock %}
+
+{% block remove %}
+<turbo-stream action="remove" target="<?= $class_name ?>_{{ entity.id }}"></turbo-stream>
```
nicolas-grekas, line 18: should be `id` eveywhere instead of `entity.id`

dunglas: Why not but on the other hand here we're sure that `entity.id` works and this hints the existence of the `entity` variable.

nicolas-grekas: For remove, it doesn't work, but for others yes

nicolas-grekas: Oh, and id works even if the property is not named id

dunglas: Good catch for remove. (When using maker-bundle the property is always named id).

jrushlow (approved):

src/Maker/MakeEntity.php
```diff
@@ -84,6 +85,7 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
         $command
             ->addArgument('name', InputArgument::OPTIONAL, sprintf('Class name of the entity to create or update (e.g. <fg=yellow>%s</>)', Str::asClassName(Str::getRandomTerm())))
             ->addOption('api-resource', 'a', InputOption::VALUE_NONE, 'Mark this class as an API Platform resource (expose a CRUD API for it)')
+            ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Broadcast entities updates using Symfony UX Turbo')
```
weaverryan, line 88:
```suggestion
            ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Add the ability to broadcast entity updates using Symfony UX Turbo?')
```

src/Resources/help/MakeEntity.txt
```diff
@@ -9,6 +9,11 @@ automatically be available for this entity class:
 
 <info>php %command.full_name% --api-resource</info>
 
+Symfony can also broadcast all changes made to the entities to the client using Symfony
```
weaverryan, line 12:
```suggestion
Symfony can also broadcast all changes made to the entity to the client using Symfony
```

src/Resources/skeleton/doctrine/broadcast_twig_template.tpl.php
```diff
@@ -0,0 +1,28 @@
+{#
+Available variables:
+* id: the normalized id of the entity
+* entity: the entity object
+* action: the current action
+* options: the options passed to the Broadcast attribute
```
weaverryan, line 6: Maybe include a link to the docs? https://github.com/symfony/ux/tree/main/src/Turbo#broadcast-doctrine-entities-update

Because the user will still need to add a `{{ turbo_stream_listen(book) }}` somewhere in a template, if I understand things correctly.

weaverryan (commented):
A few minor comments - but this looks great!

weaverryan (approved):
Awesome!

weaverryan:
Thanks Kévin!

PR #862, opened 2021-04-19
[make:user] implement getUserIdentifier if required

jrushlow:
handle `getUsername()` -> `getUserIdentifier()` in Symfony 5.3 https://github.com/symfony/symfony/pull/40403


src/Security/UserClassBuilder.php
```diff
@@ -97,10 +97,18 @@ private function addGetUsername(ClassSourceManipulator $manipulator, UserClassCo
             );
         }
 
+        $getterIdentifierName = 'getUsername';
+
+        $reflectedInterface = new \ReflectionClass(UserInterface::class);
+
+        if (false !== stripos($reflectedInterface->getDocComment(), '@method string getUserIdentifier()')) {
+            $getterIdentifierName = 'getUserIdentifier';
+        }
```
weaverryan, line 106: What if we just look for the presence of this concrete method: https://github.com/symfony/symfony/blob/527b885fc68da965ada884e484d043f79f703f45/src/Symfony/Component/Security/Core/User/User.php#L96-L99 - looking for `@method` is a no go because it would move to be a real method in 6.0

weaverryan (commented):
Minus the tests and one comment, this is looking very nice 👍

src/Security/UserClassBuilder.php
jrushlow: ahh yes i like that a lot better.. saves us from having to worry about adding another conditional for 6+

jrushlow: done

src/Resources/skeleton/security/UserProvider.tpl.php
```diff
@@ -3,7 +3,7 @@
 namespace <?= $namespace; ?>;
 
 use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
-use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
+use Symfony\Component\Security\Core\Exception\<?= $uses_user_identifier ? 'UserNotFoundException' : 'UsernameNotFoundException' ?>;
 <?= ($password_upgrader = interface_exists('Symfony\Component\Security\Core\User\PasswordUpgraderInterface')) ? "use Symfony\Component\Security\Core\User\PasswordUpgraderInterface;\n" : '' ?>
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Security\Core\User\UserProviderInterface;
```
weaverryan, line 9: We really need to, at some point, write a util class where we plug all the `use` statements into... then we just dump `echo $useStatements->render()` in the template :p

src/Security/UserClassBuilder.php
```diff
@@ -97,10 +98,17 @@ private function addGetUsername(ClassSourceManipulator $manipulator, UserClassCo
             );
         }
 
+        $getterIdentifierName = 'getUsername';
+
+        // Check if we're using Symfony 5.3+ - UserInterface::getUsername() was replaced with UserInterface::getUserIdentifier()
+        if (class_exists(InMemoryUser::class)) {
```
weaverryan, line 104: This class DOES only exist on 5.3. But why did you choose this class specifically? Could we use something more direct like `method_exists(Symfony\Component\Security\Core\User\User::class, 'getUserIdentifier')`?

weaverryan (commented):
Just one tiny comment!

src/Security/UserClassBuilder.php
jrushlow: `User::class` has been deprecated as of 5.3 && Renamed / Will be renamed to `InMemoryUser::class`

https://github.com/symfony/symfony/blob/abeb8e4ca9adfcc1494df22af9fd9a1c0181b532/src/Symfony/Component/Security/Core/User/User.php#L21

https://github.com/symfony/symfony/pull/40443

src/Resources/skeleton/security/UserProvider.tpl.php
jrushlow: No kidding.. use statements are getting harder and harder to manage lately.. have a solution on the todo list ;)

weaverryan:
Thanks Jesse!

PR #863, opened 2021-04-19
Fix regenerate xml

jhonaikerf:
#207 
#579


jhonaikerf:
@weaverryan can you check this?

src/Doctrine/EntityRegenerator.php
```diff
@@ -102,7 +102,17 @@ public function regenerateEntities(string $classOrNamespace)
...
+
+                    foreach ($classMetadata->reflClass->getTraits() as $trait) {
+                        foreach ($trait->getProperties() as $property) {
+                            $traitProperties[] = $property->getName();
+                        }
+                    }
+
+                    if (\in_array($fieldName, $traitProperties) || isset($mapping['declared'])) {
+                        continue;
+                    }
```
weaverryan, line 115: Can you explain this a bit? And we'll need a test case that shows what this is fixing :)

src/Doctrine/DoctrineHelper.php
```diff
@@ -195,7 +196,13 @@ public function getMetadata(string $classOrNamespace = null, bool $disconnected
...
             }
 
-            foreach ($cmf->getAllMetadata() as $m) {
+            try {
+                $loaded = $cmf->getAllMetadata();
+            } catch (ReflectionException $e) {
+                $loaded = $this->isInstanceOf($cmf, AbstractClassMetadataFactory::class) ? $cmf->getLoadedMetadata() : [];
+            }
+
+            foreach ($loaded as $m) {
```
weaverryan, line 205: Can you include the test case from #579 so we can see that it's passing?

weaverryan (commented):
Thanks for the PR @jhonaikerf and sorry for the slow review. Mostly, we need some tests that show what you're fixing with these changes.

Thanks!

PR #864, opened 2021-04-19
[make:command] template: add void return type to configure method

duboiss:


jrushlow (approved):
thanks @duboiss

weaverryan:
Thanks Steven!

PR #869, opened 2021-05-04
[make:serializer:encoder] set public constant visibility modifier

jrushlow:
Since PHP 7.1 - class constants can have a visibility modifier. The Symfony coding standards require the visibility modifier to be set. 

`make:serializer:encoder` now defaults to the `public` modifier for the `FORMAT` const.

### DIFF
--- a/src/Resources/skeleton/serializer/Encoder.tpl.php
+++ b/src/Resources/skeleton/serializer/Encoder.tpl.php
@@ -7,7 +7,7 @@ use Symfony\Component\Serializer\Encoder\EncoderInterface;
 
 class <?= $class_name ?> implements EncoderInterface, DecoderInterface
 {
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/serializer/Encoder.tpl.php
```diff
@@ -7,7 +7,7 @@
 
 class <?= $class_name ?> implements EncoderInterface, DecoderInterface
 {
-    const FORMAT = '<?= $format ?>';
+    public const FORMAT = '<?= $format ?>';
```
weaverryan, line 10: Hmm, but what's the use-case (why make it public) for it being public?

weaverryan:
I'm merging - but I'm still not sure about the tiny detail of whether the visiblity should be public or private :)

PR #872, opened 2021-05-08
Use object typehint when generating entities

HypeMC:
Use object typehint when adding object entity fields.

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -1199,6 +1199,8 @@ final class ClassSourceManipulator
                 return '\\'.\DateInterval::class;
 
             case 'object':
+                return 'object';
+
             case 'binary':
             case 'blob':
             default:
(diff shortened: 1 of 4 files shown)


src/Util/ClassSourceManipulator.php
```diff
@@ -1113,6 +1113,8 @@ private function getEntityTypeHint($doctrineType)
                 return '\\'.\DateInterval::class;
 
             case 'object':
+                return \PHP_VERSION_ID < 70200 ? null : 'object';
```
jrushlow, line 1116: This check should be done in the `PhpCompatUtil` -> https://github.com/symfony/maker-bundle/blob/main/src/Util/PhpCompatUtil.php

Something along the lines of `PhpCompatUtil::canUseObjectReturnType()`

jrushlow (changes requested):
I like the idea! Anytime we can set explicit types, the better. We do however need to move the PHP version conditional logic into the `PhpCompatUtil::class`

src/Util/ClassSourceManipulator.php
HypeMC: @jrushlow Done

HypeMC: @jrushlow Since the maker bundle now requires PHP 7.2.5+ I've removed all conditional logic.

jrushlow (approved):
Awesome! Thank you @HypeMC

PR #875, opened 2021-05-10
[ci][make-controller] set framework.test in test env

jrushlow:
fixes 
```
1) Symfony\Bundle\MakerBundle\Tests\Maker\MakeControllerTest::testExecute with data set "controller_basic" (Symfony\Bundle\MakerBundle\Test\MakerTestDetails Object (...))
Error while running the PHPUnit tests *in* the project: 

 
PHPUnit 8.5.15 by Sebastian Bergmann and contributors.

Testing Project Test Suite
E                                                                   1 / 1 (100%)

Time: 162 ms, Memory: 20.00 MB

There was 1 error:

1) App\Tests\GeneratedControllerTest::testController
LogicException: You cannot create the client used in functional tests if the "framework.test" config is not set to true.

/var/htdocs/tests/tmp/cache/maker_app_f06e46ebe75d42ea9e9b63872d47e20b5.x-dev/vendor/symfony/framework-bundle/Test/WebTestCase.php:52
/var/htdocs/tests/tmp/cache/maker_app_f06e46ebe75d42ea9e9b63872d47e20b5.x-dev/tests/GeneratedControllerTest.php:11
```

### DIFF
--- a/tests/fixtures/MakeController/config/packages/framework.yaml
+++ b/tests/fixtures/MakeController/config/packages/framework.yaml
@@ -1,2 +1,0 @@
-framework:
-  ide: 'phpstorm://open?file=%%f&line=%%l'


tests/fixtures/MakeController/config/packages/test/framework.yaml
```diff
@@ -0,0 +1,2 @@
+framework:
+    test: true
```
weaverryan, line 2: This shouldn't be needed, so we need to dig deeper to see what's going on. 

In 5.3, this file IS gone in the recipe - but it's because it's replaced by the per-env setting: https://github.com/symfony/recipes/blob/9a3759a60b0a5ff48b30f949d2ac89a7d2816e47/symfony/framework-bundle/5.3/config/packages/framework.yaml#L20-L22

So if that isn't working, we need to figure out why. Good job pushing that forward!

jrushlow: Alrighty, I've figured out the problem. We had a `framework.yaml` fixture for this particular test that I believe is not needed. This in turn overrides the flex recipe for `framework.yaml` in the generated project. End of the day, `when@test` didn't exist nor did the old `test/framework.yaml` 

All should be fine and well now

weaverryan:
Thanks Jesse!

PR #876, opened 2021-05-11
[make:entity] Add @method tags to Doctrine repositories

HypeMC:
Doctrine supports querying by a certain field through the use of [the `__call` magic method](https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/working-with-objects.html#by-simple-conditions).

The idea here is to add the appropriate `@method` tags to custom repositories for each field that supports this functionality.
The following three methods are added:
 - `@method <Class>|null findOneBy<Field>($<Field>, array $orderBy = null)`
 - `@method <Class>[]      findBy<Field>($<Field>, array $orderBy = null, $limit = null, $offset = null)`
 - `@method int          countBy<Field>($<Field>)`

I still need to write some tests, but would like to hear your thoughts on this idea.

### TODO
 - [ ] Tests

### DIFF
--- a/tests/Doctrine/fixtures/expected_xml/src/Repository/UserRepository.php
+++ b/tests/Doctrine/fixtures/expected_xml/src/Repository/UserRepository.php
@@ -11,6 +11,9 @@ use Doctrine\Persistence\ManagerRegistry;
  * @method UserXml|null findOneBy(array $criteria, array $orderBy = null)
  * @method UserXml[]    findAll()
  * @method UserXml[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
+ * @method UserXml|null findOneByName($name, array $orderBy = null)
+ * @method UserXml[]    findByName($name, array $orderBy = null, $limit = null, $offset = null)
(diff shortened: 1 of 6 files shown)


tests/Doctrine/fixtures/expected_xml/src/Repository/UserRepository.php
```diff
@@ -11,6 +11,9 @@
  * @method UserXml|null findOneBy(array $criteria, array $orderBy = null)
  * @method UserXml[]    findAll()
  * @method UserXml[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
+ * @method UserXml|null findOneByName($name, array $orderBy = null)
+ * @method UserXml[]    findByName($name, array $orderBy = null, $limit = null, $offset = null)
+ * @method int          countByName($name)
```
weaverryan, line 16: Would this add 3 new `@method` lines for every property in the entity?

This is a really nice-looking PR, but that's a lot of noise (and I don't love this magic feature of Doctrine).

PR #877, opened 2021-05-12
[make:entity] Default to "datetime_immutable" when creating entities

nicolas-grekas:
datetime should die, it's a footgun.


jrushlow (approved):
thanks @nicolas-grekas! we should've done this decades ago :)

weaverryan:
Thanks Nicolas!

PR #882, opened 2021-05-18
[security] fix cve-2021-21424 legacy Symfony 3.x + MakerBundle

### LINKED ISSUE #880: Fix CVE (v1.31.1) also for usage in Symfony 3.4 (v1.29)
At the moment there is a CVE ([CVE-2021-21424][]: Prevent user enumeration via response content in authentication mechanisms) which is only fixed in v1.31.1 for Symfony version >= 4.

There is still a version missing for Symfony 3.4 which is (sadly) still in use in many production environments.

Can we has CVE fix?

Thx :)

jrushlow:
Applies security fix for CVE-2021-21424 in newly generated authenticators created by MakerBundle in Symfony 3.x applications.

fixes #880


chalasr (approved):
👍 Note that we probably need a similar changelog for v1.31.1 on `main` (should have been done when releasing, sorry.)

src/Resources/skeleton/authenticator/LoginFormAuthenticator.tpl.php
```diff
@@ -74,7 +74,7 @@ public function getUser($credentials, UserProviderInterface $userProvider)
 
         if (!$user) {
             // fail authentication with a custom error
```
stof, line 76: the comment should probably be removed, as it is not a custom error

PR #887, opened 2021-05-21
Update repository and entity templates

drupol:
Hello,

This PR is a draft. I created to gather feedback from the maintainers (_and eventually other people_) about it.

- [x] Add a new option to the `make:entity` command: `--use-standalone-services`
- [x] Does not change the behavior of the regular `make:entity` command
- [x] Update the `make:entity` documentation
- [x] Implements a new custom interface for each Repository
- [ ] Has updated tests

Basically the goal is to have clean code using constructor injection, composition over inheritance and dependency inversion principles.

When using `bin/console make:entity --use-standalone-services`:

Repositories are:
- [x] Set as `final`
- [x] Extends their own interface which extends [ObjectRepository](https://github.com/doctrine/persistence/blob/3.0.x/lib/Doctrine/Persistence/ObjectRepository.php) 
- [x] Are injected with the `EntityManagerInterface`
- [x] Does not extend anything
- [x] Implements methods `find`, `findAll`, `findBy`, `findOneBy`, `findAll` with proper typing
- [x] Pure services, they do not need to be tagged with `doctrine.repository_service` anymore

Entities:
- [x] Does not know about the repository it is linked to.

Related blog post: https://tomasvotruba.com/blog/2017/10/16/how-to-use-repository-with-doctrine-as-service-in-symfony/ from @TomasVotruba


weaverryan:
Hi! I agree with this in general. Doctrine repos (the EntityRepository system where you fetch from the em) are a relic from before containers existed. However, I’m worried that generating this by default will cause surprises for less advanced users that may still be fetching repos from the entity manager (like by following blog posts) and the error they will get will be hard to debug. The benefit is cleaner code - but I don’t like that drawback: it means that MakerBundle completely opts users out of a well-known system.

drupol:
I understand and I share the point of view.

When do you think we should do this change then? Maybe we should propose this for the next major release?
At a certain point, we will have to change for the greater good, no ?

TomasVotruba:
> However, I’m worried that generating this by default will cause surprises for less advanced users that may still be fetching repos from the entity manager

:+1: 

Instead of all or nothing, this can be solved with opt-in command option. E.g. `make --type=services`

drupol:
@weaverryan 

I'm giving a second thought about this. That change will be completely transparent for the end-user.
I understand that they might find it "different" when they are going to edit a repository, but it's not a huge change either.

@TomasVotruba 
I like the idea of having an extra option. We could imagine to deprecate the old one (_through a console message?_) and propose to use the new one.

I'm ok to work further on this, but I'd like to have a clear idea on what we do. Please, advise.

weaverryan:
Adding an option like `--type=services` seems like a reasonable way to try this feature out and make users opt in :). I don't like adding more "questions" to a maker (because eventually we will ask users all kinds of edge-case things), but options (though harder to "discover") don't have this drawback.

Later, if we like it, and the ecosystem feels ready, we could indeed deprecate NOT passing this flag (or more likely, we would add a config option - e.g. to the user's yml/php file and force them to opt in there).

Cheers!

drupol:
Dear @weaverryan ,

I updated the draft PR, could you please let me know if I'm going in the right direction?

Thanks.

src/Doctrine/EntityClassGenerator.php
```diff
@@ -45,7 +45,7 @@ public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $
 
         $entityPath = $this->generator->generateClass(
             $entityClassDetails->getFullName(),
-            'doctrine/Entity.tpl.php',
+            $useComposition === true ? 'doctrine/Entity-use-composition.tpl.php': 'doctrine/Entity.tpl.php',
```
jrushlow, line 48:
```suggestion
            $useComposition === true ? 'doctrine/Entity-use-composition.tpl.php' : 'doctrine/Entity.tpl.php',
```

src/Doctrine/EntityClassGenerator.php
```diff
@@ -83,7 +84,7 @@ public function generateRepositoryClass(string $repositoryClass, string $entityC
 
         $this->generator->generateClass(
             $repositoryClass,
-            'doctrine/Repository.tpl.php',
+            $useComposition === true ? 'doctrine/Repository-use-composition.tpl.php': 'doctrine/Repository.tpl.php',
```
jrushlow, line 87:
```suggestion
            $useComposition === true ? 'doctrine/Repository-use-composition.tpl.php' : 'doctrine/Repository.tpl.php',
```

src/Resources/skeleton/doctrine/Entity-use-composition.tpl.php
```diff
@@ -0,0 +1,40 @@
...
+<?php if ($should_escape_table_name): ?> * @ORM\Table(name="`<?= $table_name ?>`")
+<?php endif ?>
+ */
+<?php if ($api_resource && $use_attributes): ?>
+#[ApiResource]
+<?php endif ?>
+<?php if ($broadcast && $use_attributes): ?>
+#[Broadcast]
+<?php endif ?>
+class <?= $class_name."\n" ?>
```
jrushlow, line 27: We should be able to use the existing `doctrine/Entity.tpl.php` template. The only difference that I see is this tpl doesnt include the `@ORM\Entity(repositoryClass=.....` docBlock. We can accomplish the same in the existing template by adding a conditional

src/Resources/skeleton/doctrine/Repository-use-composition.tpl.php
```diff
@@ -0,0 +1,96 @@
...
+final class <?= $class_name; ?><?= $with_password_upgrade ? " implements PasswordUpgraderInterface\n" : "\n" ?>
+{
+    private EntityRepository $repository;
+
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->repository = $entityManager->getRepository(<?= $entity_class_name; ?>::class);
+    }
+
+    public function find($id, $lockMode = null, $lockVersion = null): ?<?= $entity_class_name; ?><?= "\n"?>
```
jrushlow, line 22: we should use argument type hints here where possible. Especially if the repo does not implement any sort of interface for these methods.

src/Resources/skeleton/doctrine/Repository-use-composition.tpl.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        return $this->repository->find($id, $lockMode, $lockVersion);
+    }
+
+    public function findOneBy(array $criteria, array $orderBy = null): ?<?= $entity_class_name; ?><?= "\n"?>
```
jrushlow, line 27: I'm wondering if it would make more sense to have these methods in a `RepositoryTrait::class` to keep the end users code a dry.. Personally I'm not a fan of have lengthy generated PHP files that are duplicated in a code base.

jrushlow (changes requested):
Good work @drupol I like where this is going - a couple of comments below. 

Also in regards to `doctrine/Repository-use-composition.tpl.php`, I think we should rename this template to better describe its purpose without having to look at the PR that introduces it. Possibly `doctrine/StandaloneRepository.tpl.php` or something along those lines. Open to ideas on naming.

src/Resources/skeleton/doctrine/Repository-use-composition.tpl.php
drupol: One of the purpose of this change is to have a strict typing, in PHP and no more in PHPDoc.

If we use a trait, there will be no more typing because the trait will be common to all repositories. And thus pretty useless.

src/Resources/skeleton/doctrine/Repository-use-composition.tpl.php
drupol: I will do the change.

src/Resources/skeleton/doctrine/Entity-use-composition.tpl.php
drupol: I though about that... but having one single template for both use cases will be a bit messy don't you think ?

drupol: Ok I did it, I tried to be as clean as possible.

The hardest task was to comply with the empty lines and spacing.

In the end, only one line of the reference generated file (fixture in the test) is modified.
The modification is about adding the return type `:array`.

src/Resources/skeleton/doctrine/Repository-use-composition.tpl.php
drupol: The `ObjectRepository` interface doesn't have any type hint, so I will leave it like this.

src/Resources/skeleton/doctrine/Entity-use-composition.tpl.php
drupol: I fixed that, thanks for the suggestion.

Issue #861, opened 2021-04-19
[make:entity] Add a way to set $include_example_comments to false 


MrYamous:
Hi, 

When make:entity command is used, two example functions are added to the related repository
In the repository template, there is a $include_example_comments variable, this variable is set to true by default, but there seems to be no way to change its value (or I haven't found it, sorry about that)

What would you think of adding a way to set this value to false? With a `no-example` option maybe? I don't know if this could be a good name

bernard-ng:
Hello @MrYamous I thought about it too, it would be a good idea indeed

jrushlow:
I'm not opposed to the idea.. I often find myself removed all of the commented code samples from the repo classes after using `make:entity`.

I think the simplest approach would be to ask "Do you want to generate code examples..." that defaults to `true` or something along those lines. `$include_example_comments` could then be set accordingly.

PR's welcome!

jrushlow:
We have no immediate plans to implement this into MakerBundle. However, PR's are always welcome!

PR #888, opened 2021-05-21
[make:entity] Ask if maker should include example comments in repository

Kelows:
Hello, here is what this PR does:

 - Asks the user if he needs code samples in the generated repository with default at 'true'

here is the corresponding issue: #861


weaverryan:
Hi there!

I'm very wary of asking more questions. However, I would support this as a bundle option: users who care can change the bundle config to not generate these if they don't want.

Cheers!

Kelows:
Hello Ryan,

for me, you're right, it would be better to set it as a general option but i don't know how to do this, sorry

i think you can close it then

Cheers!

PR #891, opened 2021-06-01
[tests] housekeeping

jrushlow:
- removes `override_url` config logic (feature has was introduced && deprecated in doctrine-bundle 2.3)
- remove duplicate directory getter
- run generated projects `bin/phpunit` instead of "makers" `simple-phpunit`
- add return types to touched classes

### DIFF
--- a/src/Test/MakerTestEnvironment.php
+++ b/src/Test/MakerTestEnvironment.php
@@ -78,7 +78,7 @@ final class MakerTestEnvironment
         return file_get_contents($this->path.'/'.$path);
     }
 
-    private function changeRootNamespaceIfNeeded()
+    private function changeRootNamespaceIfNeeded(): void
     {
         if ('App' === ($rootNamespace = $this->testDetails->getRootNamespace())) {
             return;
@@ -128,7 +128,7 @@ final class MakerTestEnvironment
         $this->processReplacements($replacements, $this->path);
     }
 
-    public function prepare()
+    public function prepare(): void
     {
         if (!$this->fs->exists($this->flexPath)) {
             $this->buildFlexSkeleton();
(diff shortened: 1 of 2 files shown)


src/Test/MakerTestEnvironment.php
```diff
@@ -259,37 +259,37 @@ public function getGeneratedFilesFromOutputText()
...
 
-    public function runInternalTests()
+    public function runInternalTests(): ?MakerTestProcess
     {
         $finder = new Finder();
         $finder->in($this->path.'/tests')->files();
         if ($finder->count() > 0) {
             // execute the tests that were moved into the project!
-            return MakerTestProcess::create(sprintf('php %s', $this->rootPath.'/vendor/bin/simple-phpunit'), $this->path)
+            return MakerTestProcess::create(sprintf('php %s', $this->path.'/bin/phpunit'), $this->path)
```
jrushlow, line 285: This changes needs extra scrutiny - test time is cut in half (usually an indicator im doing something wrong) but I'm not seeing an actual problem. Before we were calling `/maker-bundle/vendor/bin/simple-phpunit /maker-bundle/cache/generated-project`. This PR calls the generated projects phpunit directly e.g. `/maker-bundle/cache/generated-project/bin/phpunit`

weaverryan: I think (?) I originally did this... to avoid PHPUnit being downloaded over and over again inside *every* generated project. I figured: PHPUnit is already there or Maker, so let's just use it! So, I don't see any problem with this.

As far as why the test time is cut in half, that is interesting. I believe you already told me on Slack that the number of tests before and after are identical - so we didn't somehow "lose" some tests. One theory is that my plan backfired: by using `simple-phpunit` from MakerBundle, PHPUnit *was* being downloaded fresh on every run. That's just a guess, but if I'm right, it would make a significant difference

weaverryan (commented):
This approach looks good to me - let's track down that last failing test

weaverryan:
Merging this now to get our tests closer to green. Thanks Jesse!

Issue #796, opened 2021-01-28
make:message do not find key when setting transport


jmeyo:
When creating new message with make:message and choosing a transport, I got a 

Symfony\Bundle\MakerBundle\Util\YamlManipulationFailedException

It seems the yaml parser doesn't like keys surrounded by quotes as used in the documentation.

```
$ bin/console m:message

 The name of the message class (e.g. SendEmailMessage):
 > SendEmailMessage

 Which transport do you want to route your message to? [[no transport]]:
  [0] [no transport]
  [1] async_priority_high
  [2] async
  [3] async_ad
  [4] failed
 > 2

14:12:41 ERROR     [console] Error thrown while running command "'m:message'". Message: "Cannot find the key "App\Message\OtherExistingMessage" ["exception" => Symfony\Bundle\MakerBundle\Util\YamlManipulationFailedException^ { …},"command" => "'m:message'","message" => "Cannot find the ksey "App\Message\OtherExistingMessage""]

In YamlSourceManipulator.php line 581:
                                                                 
  Cannot find the ksey "App\Message\OtherExistingMessage"  
                                                                 

make:message [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<name>]


```

jmeyo:
Easy workaround is to comment all your message routing, use the maker, and then uncomment.

jmeyo:
The getkeyregex could be improved
https://github.com/symfony/maker-bundle/blob/01a9058c7f10b4cda859fc0a99e28629b927604a/src/Util/YamlSourceManipulator.php#L685-L688

 to something like 
```php
   private function getKeyRegex($key)
    {
        return sprintf('#%s( )*\'?:#', preg_quote($key));
    }```

https://github.com/jmeyo/maker-bundle

But then generated keys should also be surrounded by quotes in end result when transport is chosen, and there might be some side effects

OskarStark:
Can you send a PR and add a testcase?

Yonninout:
I was on the same case today. 

In my messenger.yaml config files the rounting array was written with simple quoted class name like this : 

```yaml
        routing:
            'App\Message\NewEnrichmentFile': async
            'App\Message\UpdateEnrichmentFile': async
            'App\Message\ExportEnrichmentFile': async
            'App\Message\NewPackaging': async
[...]

PR #896, opened 2021-06-11
Fix keys not found when surrounded by quotes

valentinloiseau:
Fix bug reported in #796 : running ``make:message`` cannot find existing messages when configuring with quotes according to [documentation](https://symfony.com/doc/current/messenger.html#routing-messages-to-a-transport).


weaverryan:
Very nice - thank you Valentin!

PR #899, opened 2021-06-17
Use proper name for parameter of upgradePassword

Tobion:
encode is the wrong term symfony tries to remove. The paramter is also called $newHashedPassword in the PasswordUpgraderInterface


weaverryan (approved):
Good catch. Test failures are unrelated - we're still catching up from the new Symfony and Doctrine releases.

jrushlow (approved):

weaverryan:
Thanks Tobias!

PR #902, opened 2021-06-19
[make:entity] Add configuration whether to include repository examples

### LINKED ISSUE #861: [make:entity] Add a way to set $include_example_comments to false

simivar:
Closes #861 and is built upon #888 but with configuration instead of additional question

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -32,6 +32,12 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->children()
                 ->scalarNode('root_namespace')->defaultValue('App')->end()
+                ->arrayNode('entity')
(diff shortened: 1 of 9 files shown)


jrushlow (changes requested):
Thanks @simivar for the work on this! Would you be able to add a test case demonstrating that this works as intended?

simivar:
@jrushlow gladly, but I wasn't sure how to exactly do those. Do you have any tips?

PR #903, opened 2021-06-21
[make:auth] use userIdentifier instead of username on login_form

seb-jean:
Change `username` to `userIdentifier`

### DIFF
--- a/src/Resources/skeleton/authenticator/login_form.tpl.php
+++ b/src/Resources/skeleton/authenticator/login_form.tpl.php
@@ -11,7 +11,7 @@
 <?php if ($logout_setup): ?>
     {% if app.user %}
         <div class="mb-3">
-            You are logged in as {{ app.user.username }}, <a href="{{ path('app_logout') }}">Logout</a>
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -11,7 +11,7 @@
 <?php if ($logout_setup): ?>
     {% if app.user %}
         <div class="mb-3">
-            You are logged in as {{ app.user.username }}, <a href="{{ path('app_logout') }}">Logout</a>
+            You are logged in as {{ app.user.userIdentifier }}, <a href="{{ path('app_logout') }}">Logout</a>
```
weaverryan, line 14: I think this needs to be conditional, right? Based on what version of Symfony the code is being generated for? 

Looks like a legit bug though :)

jrushlow (approved):
Awesome Thanks @seb-jean.. This will work nicely with #1104 and we won't need to use a conditional anymore.

jrushlow:
fixes #1082

PR #870, opened 2021-05-05
[make:crud] Fix templates path use in include

leblanc-simon:
The maker CRUD allow to set a namespace in the controller name (example : `Admin\ArticleController`). In this case, maker create namespace and folders for the templates.

```
s.leblanc@cronos $ symfony console make:crud Article

 Choose a name for your controller class (e.g. ArticleController) [ArticleController]:
 > Admin\ArticleController

 created: src/Controller/Admin/ArticleController.php
 created: src/Form/ArticleType.php
 created: templates/admin/article/_delete_form.html.twig
 created: templates/admin/article/_form.html.twig
 created: templates/admin/article/edit.html.twig
 created: templates/admin/article/index.html.twig
 created: templates/admin/article/new.html.twig
 created: templates/admin/article/show.html.twig
```

The controller is OK, the render function is call with the good path : `admin/article/index.html.twig`

But in the template, when we include the form or the delete form, the path is build with route name (`admin_article`) instead of path (`admin/article`)

This path fix this issue

PR #906, opened 2021-06-23
[make:crud] Fix templates include paths when using namespace.

WedgeSama:
This problem got an older PR from 2019 and still not yet merge cause not finished (#479)

####################
In templates, the path to included templates was wrong when using namespace. It previously used the `route_name` to generate the template path. Using `templates_path` is better in this case and fix the problem.

Example:
- Old generated code
```twig
{{ include('admin_comment/_delete_form.html.twig') }}
```
- New generated code
```twig
{{ include('admin/comment/_delete_form.html.twig') }}
```


WedgeSama:
Sorry, i have missed the PR #870 that covering the problem.

PR #910, opened 2021-06-27
[YamlSourceManipulator] Tweak regex pattern for regex key

lubo13:
We need to tweak a regex pattern, because at the moment if you try to update yaml with following config:

```
services:
  _defaults:
    autowire: true
    autoconfigure: true
    bind:
      $tracer: 'something'
```

You will receive:
Screenshot from 2021-06-27 17-02-02

### DIFF
--- a/tests/Util/yaml_fixtures/add_item_on_config_with_bind.test
+++ b/tests/Util/yaml_fixtures/add_item_on_config_with_bind.test
@@ -1,0 +1,29 @@
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    bind:
+      $httpClient: '@http.client'
+
+  app.service:
+    class: App\Services\Service
+    arguments:
+      - argument_1: something
+        argument_2: something
+===
+$data['services']['app.service']['arguments'][0]['argument_3'] = 'something_new';
+===
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    bind:
+      $httpClient: '@http.client'
+
+  app.service:
+    class: App\Services\Service
+    arguments:
+      - argument_1: something
+        argument_2: something
+        argument_3: something_new
+        
--- a/tests/Util/yaml_fixtures/add_item_on_config_with_bind.test
+++ b/tests/Util/yaml_fixtures/add_item_on_config_with_bind.test
@@ -10,6 +10,7 @@ services:
     arguments:
       - argument_1: something
         argument_2: something
+
 ===
 $data['services']['app.service']['arguments'][0]['argument_3'] = 'something_new';
 ===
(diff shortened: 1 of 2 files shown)


weaverryan (commented):
Hi! This is awesome! Could you also add a test case? These are fairly simply - there is a directory of test cases - https://github.com/symfony/maker-bundle/tree/main/tests/Util/yaml_fixtures - if you created a new file here, it would run :)

Thanks!

tests/Util/yaml_fixtures/add_item_on_config_with_bind.test
```diff
@@ -0,0 +1,28 @@
...
+        argument_2: 'something'
+===
+$data['app.service']['arguments']['argument_3'] = "something";
+===
+services:
+ _defaults:
+   autowire: true
+   autoconfigure: true
+   bind:
+     $httpClient: '@http.client'
```
weaverryan, line 21: The indentation is slightly off here, which is making the test fail (it's making the expected not match)

lubo13: I miss this thank you

tests/Util/yaml_fixtures/add_item_on_config_with_bind.test
```diff
@@ -0,0 +1,28 @@
...
+    autoconfigure: true
+    bind:
+      $httpClient: '@http.client'
+
+  app.service:
+    class: App\Services\Service
+    arguments:
+      - argument_1: 'something'
+        argument_2: 'something'
+        argument_3: 'something'
```
weaverryan, line 28: One more thing. The bug you're fixing is related to the `$httpClient` above. Unfortunately, this array syntax down here (while totally legal) is, I think, not yet supported by `YamlSourceManipulator`. And so, the test is STILL failing. I'd recommend updating this test to not use this syntax - instead use:

```yaml
  app.service:
    class: App\Services\Service
    arguments:
        argument_1: 'something'
        argument_2: 'something'
        argument_3: 'something'
```

If you're interested, you could open a separate PR with a failing test case for this array syntax :)

lubo13: Yes, I see Thank you, I just tried somethings and change it how you propose

lubo13: @weaverryan Can you take a look again, please

Thank you

jrushlow:
@lubo13 Could you rebase this PR against the `main` branch. Thank you in advance!

lubo13:
> @lubo13 Could you rebase this PR against the `main` branch. Thank you in advance!

jrushlow (approved):
Thank you @lubo13 for the contribution!

PR #913, opened 2021-06-30
[make:registration] conditionally generate verify email flash in template

### LINKED ISSUE #905: Make verify_email_error flash conditional
Hi!

This line - https://github.com/symfony/maker-bundle/blob/702558c5fc5a27796eedad12a82bb27be76247dc/src/Resources/skeleton/registration/twig_template.tpl.php#L4 - is generated, even if you choose not to verify the user's email.

We should make that dynamic :)

jrushlow:
fixes #905 - previously the generated registration form included a loop to display verify-email flash message regardless if the `SymfonyCasts/verify-email-bundle` was used. This PR only generates the loop if the user chooses to utilize the verify email bundle when running `make:registration`

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -266,7 +266,7 @@ final class MakeRegistrationForm extends AbstractMaker
             'Security\\'
         );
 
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks Jesse!

PR #914, opened 2021-06-30
[make:registration] cleanup maker and template

jrushlow:
- Implements a new use statement sorter for templates
- use class properties instead of abusing command arguments

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle;
 
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\Util\ClassNameDetails;
 use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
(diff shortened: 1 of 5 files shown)


src/Generator.php
```diff
@@ -239,4 +239,9 @@ public function generateTemplate(string $targetPath, string $templateName, array
             $variables
         );
     }
+
+    public static function getParentClassName(): string
```
weaverryan, line 243:
```suggestion
    public static function getControllerBaseClass(): string
```

And really, this should return the FQCN. If we need to tweak `generateController()` to only grab the "last" piece, then let's do that.

src/Util/TemplateComponentGenerator.php
```diff
@@ -0,0 +1,39 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+/**
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
```
weaverryan, line 17: After we've used this a bunch and finalized it, we should consider removing the `@internal`.

jrushlow: agreed - I'm thinking that once we roll out the other pieces to the "new template generator", make all of those components available for use in userland code.

src/Generator.php
jrushlow: this I've come up with a more elegant solution than depending on a string.

jrushlow: done

weaverryan:
Thanks Jesse!

PR #915, opened 2021-06-30
[CI] Fixes test error about missing annotation_reader

weaverryan:
Hi!

Starting in Composer 2.1 & Symfony 5.3, whether the annotations system is auto-enabled depends on whether doctrine/annotations will be available *without* taking into account dev dependencies. In MakerBundle, when we boot a test container, doctrine/annotations is a dev dependency only (appropriately), which incorrectly hints to the container
that the system should not be initialized. So, we force it.

Cheers!


jrushlow (approved):
Awesome find!

PR #918, opened 2021-06-30
[internal] add route generator for controllers

jrushlow:
Concept to cleanup templates that depend on route annotations / attributes for better DX when contributing to MakerBundle.

The `ResetPasswordController` is an exception due to the existing docBlocks that are generated. Will revisit that controller in the future.

### DIFF
--- a/src/Util/TemplateComponentGenerator.php
+++ b/src/Util/TemplateComponentGenerator.php
@@ -18,6 +18,13 @@ namespace Symfony\Bundle\MakerBundle\Util;
  */
 final class TemplateComponentGenerator
 {
+    private $phpCompatUtil;
+
+    public function __construct(PhpCompatUtil $phpCompatUtil)
+    {
+        $this->phpCompatUtil = $phpCompatUtil;
+    }
+
     public static function generateUseStatements(array $classesToBeImported): string
     {
         $transformed = [];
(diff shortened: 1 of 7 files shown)


src/Util/TemplateComponentGenerator.php
```diff
@@ -36,4 +43,48 @@ public static function generateUseStatements(array $classesToBeImported): string
 
         return $statements;
     }
+
+    /** @legacy Annotation Support can be dropped w/ Symfony 6 LTS */
+    public function generateRouteForControllerMethod(string $routePath, string $routeName, array $methods = [], bool $indent = true, bool $trailingNewLine = true): string
```
weaverryan, line 48: Do we ever use the last 2 args?

jrushlow: They are intended to handle cases where we have a docBlock with more than just a route annotation. E.g. https://github.com/symfony/maker-bundle/blob/f093d906c667cba7e3f74487d9e5e55aaf25a031/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php#L39

But I'm still on the fence if implementing this in those situations would make the template less manageable than it is now. I was going to take a stab at refactoring the reset template above in a separate PR after I get some of the more important stuff knocked out.

weaverryan:
Thanks Jesse!

Issue #892, opened 2021-06-04
Add support for PHP 8 Attributes to make:entity


markwatney2016:
With https://github.com/doctrine/DoctrineBundle/releases/tag/2.4.0 the Doctrine Bundle added support for the ORM 2.9 AttributeDriver. It would now be great to also have support for those PHP 8 Attributes in the Maker Bundle.

Current situation:

```
 [ERROR] Only annotation mapping is supported by make:entity, but the App\Entity\User class
         uses a different format. If you would like this command to generate the properties &
         getter/setter methods, add your mapping configuration, and then re-run this command
         with the --regenerate flag.  
```

bastien70:
Hello, when this PR will be released ?

karaka200:
When will this problem be solved?

devnix:
I would stop making "When X will be solved/released", as it will probably result in a locked conversation.

PR #920, opened 2021-07-04
[make:entity] add php8 attribute support

SimonMarx:
Hey everyone,

i saw the open issue https://github.com/symfony/maker-bundle/issues/892 and i had the same problem.
Since my application is running on php8 and i do not want to create all my entities manually, i've added attribute support for the `make:entity` command.
The Annotation support is still available and (i think) untouched.

**Things I was / am unsure about:**

- Before i made changes, the process checked if the class is annotated. I was unsure if i should add a new check `isAttributed`, or decide which kind of manipulation strategy the ClassSourceManipulator should use (annotation or attribute) depending on the given MappingDriver. As you can see in the code, i selected option number 2.
- The method `DoctrineHelper::isClassAnnotated` is theoretically not needed anymore. But removing public methods is forbidden, do i have to flag/annotate them somehow in the doc blocks?
- Commands like `make:user` does not support Attributes now, should that logic be implemented in this Pull Request?

To be honest, this is my first public pull request and im a bit nervous  :cold_sweat: . I tried to follow all rules from the contributing guidelines, but maybe i missed some rules, so I am grateful for any feedback, tips, dos and don'ts.

Thanks in advance for your feedbacks and here is a screenshot of an generated entity with attributes :smiley: 

image


davelima:
Working fine to me! There is something i can do to help getting it merged? :raised_hands:

dunglas (approved):
lgtm

jrushlow:
@SimonMarx can you rebase this against the main branch so we can see what the tests look like? thanks in advance!

adlpz:
This change contains a bug in the Mapping Driver resolution that will return the wrong driver when multiple drivers are in place. I've submitted a pull request to @SimonMarx 's branch here: https://github.com/SimonMarx/maker-bundle/pull/2

I'd suggest getting that one through before rebasing and merging this one.

Geekimo:
Thanks, I'll update as soon as your branch is merged :)

SimonMarx:
@adlpz @Geekimo thanks for your PRs and sorry for making you wait, for some reason I didn't get notifications about the Pull Requests :( 

I've merged the pr's now and rebased the branch and fixed the review note from @dunglas

src/DependencyInjection/CompilerPass/SetDoctrineAnnotatedPrefixesPass.php
```diff
@@ -59,10 +59,9 @@ public function process(ContainerBuilder $container)
                     $methodCalls[$i] = $arguments;
                 }
 
-                $isAnnotated = false !== strpos($arguments[0], '_annotation_metadata_driver');
                 $annotatedPrefixes[$managerName][] = [
                     $arguments[1],
-                    $isAnnotated ? new Reference($arguments[0]) : null,
+                    new Reference($arguments[0]),
```
weaverryan, line 64: Can you explain this change? thx :)

src/Doctrine/DoctrineHelper.php
```diff
@@ -122,6 +124,43 @@ public function isClassAnnotated(string $className): bool
         return false;
     }
 
+    public function doesClassUsesAttributes(string $className): bool
```
weaverryan, line 127:
```suggestion
    public function doesClassUseAttributes(string $className): bool
```

src/Doctrine/DoctrineHelper.php
```diff
@@ -122,6 +124,43 @@ public function isClassAnnotated(string $className): bool
...
+        $managerName = array_search($em, $this->getRegistry()->getManagers(), true);
+
+        foreach ($this->annotatedPrefixes[$managerName] as [$prefix, $attributeDriver]) {
+            if (0 === strpos($className, $prefix)) {
+                return null !== $attributeDriver;
+            }
+        }
+
+        return false;
+    }
```
weaverryan, line 162: Could we share code from `isClassAnnotated()`? Perhaps we create a new:

```php
private function doesClassUseDriver(string $className, string $driverClass): bool
```

which contains all the logic, but where `$driverClass` is used instead of `AnnotationDriver` or `AttributeDriver`. Then, this method would be:

```php
public function doesClassUseAttributes(string $className): bool
{
    return $this->doesClassUseDriver(string $className, AttributeDriver::class): bool
}
```

src/Maker/MakeEntity.php
```diff
@@ -848,6 +860,26 @@ private function doesEntityUseAnnotationMapping(string $className): bool
         return $this->doctrineHelper->isClassAnnotated($className);
     }
 
+    private function doesEntityUseAttributeMapping(string $className): bool
```
weaverryan, line 863: Instead of a new method, let's rename `doesEntityUseAnnotationMapping()` to `doesEntityUseAnnotationOrAttributeMapping()` and update it to check for both.

src/Doctrine/DoctrineHelper.php
```diff
@@ -268,4 +307,27 @@ public function isKeyword(string $name): bool
 
         return $connection->getDatabasePlatform()->getReservedKeywordsList()->isKeyword($name);
     }
+
+    /**
+     * this method tries to find the correct MappingDriver for the given namespace/class
+     * To determine which MappingDriver belongs to the class we check the prefixes configured in Doctrine and use the
+     * prefix that has the closest match to the given $namespace.
+     */
+    public function getMappingDriverForNamespace(string $namespace): ?MappingDriver
```
weaverryan, line 316: Is this method needed? It's used in `MakeEntity`, but it's ultimately only used to determine if "attributes" are being used. Could we re-use `doesClassUseAttributes` for that?

src/Doctrine/EntityClassGenerator.php
```diff
@@ -33,7 +33,7 @@ public function __construct(Generator $generator, DoctrineHelper $doctrineHelper
         $this->doctrineHelper = $doctrineHelper;
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, bool $doctrineAttributes = false): string
```
weaverryan, line 36: Instead of passing this as an arg, this should use `$this->doctrineHelper` to determine if attributes should be used. This is important because `->generateEntityClass()` is called in 3 other parts of the code base

src/Util/ClassSourceManipulator.php
```diff
@@ -83,18 +85,23 @@ public function getSourceCode(): string
         return $this->sourceCode;
     }
 
-    public function addEntityField(string $propertyName, array $columnOptions, array $comments = [])
+    public function addEntityField(string $propertyName, array $columnOptions, array $comments = [], array $attributes = [])
```
weaverryan, line 88: There is no use-case yet (unless I'm missing something) for someone to need to pass extra attributes to this method. And so, do not add the new `$attributes` argument.

src/Util/ClassSourceManipulator.php
```diff
@@ -71,6 +72,7 @@ public function __construct(string $sourceCode, bool $overwrite = false, bool $u
         $this->printer = new PrettyPrinter();
 
         $this->setSourceCode($sourceCode);
+        $this->useAttributes = $useAttributes;
```
weaverryan, line 75: This is getting complex! A few things:

1) Rename this to `bool $preferPhp8Attributes` - it seems more clear to me and differentiates it from `$useAnnotations`. Really, `$useAnnotations` could be renamed to `$shouldAnnotationProperties`, but let's not do that now.

2) This class is instantiated in multiple places. Unfortunately, we're going to need to make sure we pass in this new flag in all of those places (try running `git grep 'new ClassSourceMan'`).

weaverryan (commented):
Thanks for your work on this - this is a complex PR, but I'd like to push it forward to the finish line as quickly as we can :).

Cheers!

src/Doctrine/DoctrineHelper.php
```diff
@@ -268,4 +307,27 @@ public function isKeyword(string $name): bool
...
+    {
+        $lowestCharacterDiff = null;
+        $foundDriver = null;
+
+        foreach ($this->annotatedPrefixes as $key => $mappings) {
+            foreach ($mappings as [$prefix, $driver]) {
+                $diff = substr_compare($namespace, $prefix, 0);
+
+                if ($diff >= 0 && (null === $lowestCharacterDiff || $diff < $lowestCharacterDiff)) {
```
php-programmist, line 325: Variable `$lowestCharacterDiff` always is `null`. You need to add `$lowestCharacterDiff = $diff;` in `if` body

src/DependencyInjection/CompilerPass/SetDoctrineAnnotatedPrefixesPass.php
SimonMarx: The `DoctrineHelper::annoatedPrefixes` property stores all configured Namespaces for Doctrine with its drivers.
In `DoctrineHelper::getMappingDriverForNamespace` we check which Namespace (Prefix) uses which driver. For that we need to configure the `DoctrineHelper` in the pass with all configured drivers from the doctrine configuration and not only when they are using the Annotation driver.

src/Maker/MakeEntity.php
SimonMarx: Whats about `doesEntityUseSupportedMapping` method instead of using the Driver names in the method?

src/Doctrine/DoctrineHelper.php
SimonMarx: good point, i don't know why i created a new method for that...

src/Maker/MakeEntity.php
weaverryan: That's a nice idea :)

src/Util/ClassSourceManipulator.php
weaverryan: After more thought, I think we should NOT pass this argument to this class at all. We should not try to pass a flag to the constructor that says that we "prefer attributes" or not. In theory, if attributes are supported, we should ALWAYS prefer them. Unfortunately, it's not that simple: even if you're using PHP 8, depending on your version of Doctrine & Symfony, your app might support the `Route` annotation but not the `Column` annotation. And so, I think it makes more sense to pass `$attributes` as an argument to each method where needed. So, similar to `addProperty()`, the method would have both a `$annotationLines` and `$attributes` arguments (we could remove/deprecate `$annotationLines` later). But unlike how you have it currently, we wouldn't try to be "smart" inside that method to choose one other the other: if `$annotationLines` is passed, we render those. Also, if `$attributes` is passed, we render those. The determination of which should be used would be done by whoever is calling that method.

src/Util/ClassSourceManipulator.php
weaverryan: Pass in a `bool $useAttributes` argument instead, which will tell this method if it should use annotations or attributes when building the `ORM\Column` metadata.

src/Util/ClassSourceManipulator.php
SimonMarx: Since the `ClassSourceManipulator` isnt a Singleton or Service and is new initialized in every make command, the command itself could check for requirements.  E.g. the  `make:entity` command have the check if Symfony is >= 5.2 and doctrine is >= 2.8.    In that case you can keep your settings global as constructor parameter.

Well there is maybe another problem when 2 different packages need to have annotations/attributes on the same class and one of the packages does not support attributes. Let's take as an example Entities which are handled by Doctrine and ApiPlatform. When ApiPlatform does not support Attributes, you have to be able to pass `$attributes` and `$annotationLines` to make sure a that the caller of the method can decide which properties/methods should be annotated and which are attributed.

What do you think?

weaverryan: That’s exactly the situation I was thinking of. So we should have no new constructor argument, but instead should allow both $annotationLines and $attributes to be passed to each method :)

weaverryan:
@SimonMarx you’ve done some really great work on this. Will you have some time to finish it? If not, either @jrushlow or I could take your work and keep going. Have you looked at the tests yet? My guess is that we’ll need a set of new “expected fixtures” classes for the ClassSourceManipulatorTest that is used for asserting against attributes instead of annotations (and then “toggle” which expected set of classes to use based on the PHP version in the test and dependency versions). Have you looked into this aspect? We have some experience doing this in other places - like the 5.3 security.

Thanks!

SimonMarx:
@weaverryan i will code the stuff we talked about tomorrow and push it to the repository. To be honest, i do not have that much experience in writing tests (shame on me) and it would be great if you could help with that part.

weaverryan:
No worries - we can help with that. Let me know when you’ve got the updates :)

src/Doctrine/DoctrineHelper.php
SimonMarx: Finally i found the reason i created this function.

In case we create an entity for the first time, the ManagerRegistry does not know this entity yet. Only after the Entity File is saved for the first time, the manager knows it.

To decide which MappingDriver should be used we need somehow the configuration from doctrine. Doctrine offers us to use the `prefix: 'App\Entity'` option to configure whole Namespaces as entities, and exactly for that the function is.

SimonMarx:
@weaverryan finally. I changed some stuff. Can you take another look?

Still one Problem with the Version check for Symfony / Doctrine and PHP. I thought i could implement a check for doctrine in the DoctrineHelper like that:

```
    public function isDoctrineSupportingAttributes(): bool
    {
        return $this->isDoctrineInstalled() && Comparator::greaterThanOrEqualTo(\constant('Doctrine\ORM\Version::VERSION'), '2.9.0') && Comparator::greaterThanOrEqualTo(\PHP_VERSION, '8.0');
    }
```

Error: ` Attempted to load class "Comparator" from namespace "Composer\Semver".`

The Composer\Semver\Comparator is only available in dev dependencies and somehow not usable does not matter if i installed in dev mode or production mode. Do you have best practices to compare Versions from composer packages which are installed? We need 3 checks.

- Is doctrine able to handle attributes (need to be greater or equal 2.9) | inside the DoctrineHelper
- Is symfony able to handle attributes (needs to be greater or equal 5.2?) | Inside the MakerCommand
- If both cases are true, do we have PHP > 8 | I think that should be inside the DoctrineHelper and MakerCommand

We also could create a new Helper Class (if that doesn't already exists somewhere) to compare Versions by its package names?

weaverryan:
> Still one Problem with the Version check for Symfony / Doctrine and PHP. I thought i could implement a check for doctrine in the DoctrineHelper like that:

What we usually do is look for a class that only exists in Doctrine 2.9 or higher.  But... I don't think there *is* such a class - https://github.com/doctrine/DoctrineBundle/pull/1322/files

What we *can* look for this the `doctrine.orm.metadata.attribute.class` parameter. Probably we would need to:

A) register a compiler pass that checks to see if this parameter exists
B) If it does/doesn't, inject a boolean `$areDoctrineAttributesSupported` into some helper service that will use this to determine, overall, if attributes should be used.

src/Util/ClassSourceManipulator.php
```diff
@@ -58,6 +58,8 @@ final class ClassSourceManipulator
     public function __construct(string $sourceCode, bool $overwrite = false, bool $useAnnotations = true, bool $fluentMutators = true)
     {
         $this->overwrite = $overwrite;
+        // is this needed? https://github.com/symfony/maker-bundle/pull/920 we discussed about moving "useAttributes" into the methods itselfs
+        // which would make "useAnnotations" obsolete
```
weaverryan, line 62: This property really means:

> Should Doctrine metadata (whether that is annotations or attributes) be added when `addProperty()` is called?

So, it's almost a "mode" that we want to activate on this entire manipulator. I think it's fine as a property, but it could be renamed to:

> shouldAddDoctrineMetadatatoNewProperties

A mouthful, but much more precise.

src/Util/ClassSourceManipulator.php
```diff
@@ -83,18 +85,24 @@ public function getSourceCode(): string
         return $this->sourceCode;
     }
 
-    public function addEntityField(string $propertyName, array $columnOptions, array $comments = [])
+    public function addEntityField(string $propertyName, array $columnOptions, array $comments = [], array $attributes = [], bool $useAttributes = false)
```
weaverryan, line 88:
```suggestion
    public function addEntityField(string $propertyName, array $columnOptions, array $comments = [], array $attributes = [], bool $useAttributesForMapping = false)
```

src/Util/ClassSourceManipulator.php
```diff
@@ -123,7 +131,16 @@ public function addEmbeddedEntity(string $propertyName, string $className)
...
+        $attributes = [
+            $this->buildAttributeNode(
+                'ORM\\Embedded',
+                [
+                    'class' => new ClassNameValue($className, $typeHint),
+                ]
+            ),
+        ];
+
+        $this->addProperty($propertyName, $annotations, null, $attributes);
```
weaverryan, line 143: We're passing both the annotations and attributes to `addProperty()`? Won't that cause both to be added? We might need another `$useAttributesForMapping` argument to the `addEmbeddedEntity()` that helps us determine which to create.

weaverryan (commented):
It's look much closer :). I apologize for the complexity of `ClassSourceManipulator` - I created it, and we're now pushing it beyond what it was designed to do, and it's becoming a bit of a mess. Honestly, I'm not re-thinking our original conversation about passing argument to each method about attributes/vs annotations. It might make more sense to activate it via the constructor. It would be a `bool $useAttributesForDoctrineMapping = false`. Then we would use that everywhere else in the class. We would still need to go find all places that it's instantiated and pass that in when it's relevant.

SimonMarx:
@weaverryan  so what should we do know, pass the `$useAttributesForDoctrineMapping` to the constructor for of `ClassSourceManipulator` or just leave it in the methods for now, im a bit confused :D 



I added a new Compiler pass to set the parameter `maker.compatible_check.doctrine.supports_attributes` which is available in the container. This parameter is used now in the `DoctrineHelper`. Its not the best solution but it works.. A helper class which provides methods to compare multiple packages, versions and requirements would be cleaner and more reusable in different places. But i think this will be to much overhead for now since many people waiting for the attribute support.

I also added 3 checks into the MakeEntity Command.

- Check if Symfony is supporting attributes
- Check if doctrine supporting attributes
- Check if php supporting attributes

These checks only triggers when the namespace/class given in the `make:entity` should use the `AttributeDriver`.
Not sure if we need the symfony version check, but i implemented it anyways.

weaverryan:
>  so what should we do know, pass the $useAttributesForDoctrineMapping to the constructor for of ClassSourceManipulator or just leave it in the methods for now, im a bit confused :D

sorry for the back and forth. Right now, I think we should do the constructor arg. But if you disagree, let me know - you’re deeper in the code than I am currently :)

SimonMarx:
> > so what should we do know, pass the $useAttributesForDoctrineMapping to the constructor for of ClassSourceManipulator or just leave it in the methods for now, im a bit confused :D
> 
> sorry for the back and forth. Right now, I think we should do the constructor arg. But if you disagree, let me know - you’re deeper in the code than I am currently :)

i think there are pros / cons for both sides, the constructor variant is more comfortable. The methods variant potentially covers more possbile cases.

Since the `ClassSourceManipulator` is a mix between a "Make whatever class you want with it" and "Its designed to create Entity classes" its hard to say whats the best way to go for the future (I already was thinking about to create an issue to discuss a re-design of the manipulator to move entity specific methods out of it).

In case the `ClassSourceManipulator` should be a manipulator for any class, i think for future cases it can be helpful when you can decide which attributes and annotations should be placed for a property, which is hard when you already defined `$useAttributes` in the constructor and exclude annotations when its set to true.

In my opinion the confusing part is just the design or naming of the `ClassSourceManipulator` and its methods.

When using the `public function addProperty(string $name, array $annotationLines = [], $defaultValue = null): void` method. we should not talk about `annotations`. Instead we should see it more as `$docBlocks`. And when we are using it as `$docBlocks` it will make sense to give the possiblity to pass both `$docBlocks` and `$attributes` when creating properties.

But I don't know the plans of the makers bundle or where clear boundaries are drawn as to what it should not do.
So whatever fits better in your plans or design, i will do it :D

weaverryan:
> But I don't know the plans of the makers bundle or where clear boundaries are drawn as to what it should not
> do.
> So whatever fits better in your plans or design, i will do it :D

Yup, I agree with everything you're saying 😆. The class should be split into 2 at least - perhaps a `ClassSourceManipulator` and then an `EntityClassSourceManipulator` that uses the `ClassSourceManipulator` internally.

Anyways, that's for later, but I'd be super happy if you wanted to take that on :). And, because `ClassSourceManipulator` is internal, we can do that refactoring freely without worrying about backwards-compatibility.

For this PR then, the goal is to "make the simplest changes to `ClassSourceManipulator` to get the job done". In other words, save any fixing of the class for later. From that perspective, adding a single new constructor argument called `$useAttributesForDoctrineMapping` I think will be the best.

Thanks for your work and thoughts on this :)

src/Resources/config/makers.xml
```diff
@@ -27,6 +27,7 @@
             <service id="maker.maker.make_crud" class="Symfony\Bundle\MakerBundle\Maker\MakeCrud">
                 <argument type="service" id="maker.doctrine_helper" />
                 <argument type="service" id="maker.renderer.form_type_renderer" />
+                <argument key="$attributeMappingSupport">%maker.compatible_check.symfony.supports_attributes%</argument>
```
fabpot, line 30: The name loosk wrong here:

```suggestion
                <argument key="$attributeMappingSupport">%maker.compatible_check.doctrine.supports_attributes%</argument>
```

Moreover, I think this argument should be added to the definition of `DoctrineHelper` and not `MakeCrud`?

fabpot (changes requested):

src/Resources/config/makers.xml
SimonMarx: Somehow its find its way in there... I removed it and the `DoctrineHelper` gets the parameter now.

SimonMarx:
@weaverryan so i moved the decision of using attributes or annotations into the constructor.

We can pass now attributes and annotations to the `addProperty` method, but all entity related methods should now generate an array of `$attributes` or `$annotations` depending on the constructor parameter `$useAttributesForDoctrineMapping`, so we prevent generating both for one property.

is there anything to do for me, except the tests (which would be great if someone else can make that part :D).

php-programmist:
> is there anything to do for me

Can you fix this, please?
https://github.com/symfony/maker-bundle/pull/920/files#r690295458

src/Resources/config/services.xml
```diff
@@ -38,6 +38,7 @@
             <service id="maker.doctrine_helper" class="Symfony\Bundle\MakerBundle\Doctrine\DoctrineHelper">
                 <argument /> <!-- entity namespace -->
                 <argument type="service" id="doctrine" on-invalid="ignore" />
+                <argument key="$attributeMappingSupport">%maker.compatible_check.doctrine.supports_attributes%</argument>
```
fabpot, line 41: I get "Invalid constructor argument 4 for service "maker.doctrine_helper": argument 3 must be defined before. Check your service definition.".

jrushlow:
Great work on this @SimonMarx! - I'm going to give you a hand with the tests by adding a few commits so we can get this merged in and out to the masses. 

If you add any additional commits, please ensure your local branch is up to date with your remotes. After everything is all green, I'll do a rebase with the main branch right before we get this merged.

src/Resources/skeleton/doctrine/Entity.tpl.php
```diff
@@ -9,29 +9,39 @@
 <?php if ($broadcast): ?>use Symfony\UX\Turbo\Attribute\Broadcast;
 <?php endif ?>
 
+<?php if(!$use_attributes || !$doctrine_use_attributes): ?>
```
fabpot, line 12: Small nitpick (other occurrences should be fixed as well)
```suggestion
<?php if (!$use_attributes || !$doctrine_use_attributes): ?>
```

src/DependencyInjection/CompilerPass/CompatibleCheckPass.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\DependencyInjection\CompilerPass;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class CompatibleCheckPass implements CompilerPassInterface
```
fabpot, line 17: I would rename this class to be more explicit about what it really does. Something like `DoctrineAttributesCheckPass`?

fabpot:
I've just tested this PR on the Symfony book, and it works quite well. @SimonMarx Can you also update the `make:user` command in this pull request so that we can use attributes for everything? That would help me test this PR on the whole book. Thank you for the amazing work so far.

src/DependencyInjection/CompilerPass/CompatibleCheckPass.php
SimonMarx: For now it is just a Doctrine check, but i was thinking about using the pass for multiple compatible checks later. e.g. is symfony supporting attributes, or ApiPlatform and so on. Thats the reason i named it without a package as prefix.

SimonMarx:
@fabpot the `make:user` is now supporting attributes when the User entity is in a configured namespace using type `attribute`. Same for the `make:reset-password` command.
Is there anything else using entities and attributes/annotations i did not see yet?

fabpot:
@SimonMarx Great! I've just tested with the book, regenerated everything with attributes, and it works really well. Well done!

Perf:
Great job @SimonMarx !
Would be great to have this PR merged asap! Or there are still some changes required @fabpot @weaverryan ?

src/Doctrine/DoctrineHelper.php
```diff
@@ -85,10 +93,19 @@ public function getEntityNamespace(): string
...
-    public function isClassAnnotated(string $className): bool
+    public function doesClassUseDriver(string $className, string $driverClass): bool
     {
-        /** @var EntityManagerInterface $em */
-        $em = $this->getRegistry()->getManagerForClass($className);
+        try {
+            /** @var EntityManagerInterface $em */
+            $em = $this->getRegistry()->getManagerForClass($className);
+        } catch (\ReflectionException $exception) {
+            // this exception will be thrown by the registry if the class isnt created yet.
```
OskarStark, line 102:
```suggestion
            // this exception will be thrown by the registry if the class isn't created yet.
```

src/Doctrine/DoctrineHelper.php
```diff
@@ -85,10 +93,19 @@ public function getEntityNamespace(): string
...
+            // an example case is the "make:entity" command, which needs to know which driver is used for the class to determine
+            // if the class should be generated with attributes or annotations. If this exception is thrown, we will check based on the
+            // namespaces for the given $className and compare it with the doctrine configuration to get the correct MappingDriver
```
OskarStark, line 105:
```suggestion
            // namespaces for the given $className and compare it with the doctrine configuration to get the correct MappingDriver.
```

src/DependencyInjection/CompilerPass/CompatibleCheckPass.php
SimonMarx: Pass name is changed to `DoctrineAttributesCheckPass` in the latest commit.

weaverryan:
> Would be great to have this PR merged asap! Or there are still some changes required @fabpot @weaverryan ?

I believe we just need the tests to get cleaned up - but I would also like to merge asap. @jrushlow is going to help with the tests, as I think we'll need to update the tests to "assert" that the expected HTML uses annotations OR attributes dynamically, based on the situation - and @jrushlow has handled that before.

fabpot:
From a user perspective, I think it would be great if more people would test this PR before merging this PR (but I also know that people are more comfortable testing things when they are released).

GwendolenLynch:
Test project created with `symfony new --full` on PHP 8.0.10

```yaml
# config/packages/doctrine.yaml
doctrine:
    orm:
        mappings:
            App:
                is_bundle: false
                type: attribute
                dir: '%kernel.project_dir%/src/Entity'
                prefix: 'App\Entity'
                alias: App
```

- `bin/console make:entity` :+1:
  - `text` :+1:
  - `string` :+1:
  - `int` :+1:
  - `bool` :+1:
  - `json` :+1:
  - `float` :+1:
  - `datetime_immutable` :+1:
- `bin/console make:user` :+1:
- `bin/console make:reset-password` :+1:

One thing that might be worth considering is that I (incorrectly) assumed that the project was installed with `type: attribute`, so a mention in the docs, and/or updating the installer. :smiley_cat:

src/Resources/skeleton/doctrine/Entity.tpl.php
```diff
@@ -26,15 +34,18 @@
...
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: 'integer')]
+    <?php endif ?>private $id;
 
     public function getId(): ?int
     {
         return $this->id;
     }
-}
+}
```
jderusse, line 51: revert

src/Util/ClassSourceManipulator.php
```diff
@@ -54,12 +54,17 @@ final class ClassSourceManipulator
     private $newStmts;
 
     private $pendingComments = [];
+    /**
+     * @var bool
```
jderusse, line 58: remove already inferred by constructor typehint

src/Doctrine/DoctrineHelper.php
```diff
@@ -51,13 +53,19 @@ final class DoctrineHelper
      */
     private $annotatedPrefixes;
 
+    /**
+     * @var bool
```
jderusse, line 57: remove. Already inferred by constructor typehint

src/Doctrine/EntityRegenerator.php
```diff
@@ -105,7 +105,7 @@ public function regenerateEntities(string $classOrNamespace): void
                     continue;
                 }
 
-                $manipulator->addEntityField($fieldName, $mapping);
+                $manipulator->addEntityField($fieldName, $mapping, [], []);
```
jderusse, line 108: revert, `[]` is already the default value for the 2 last args

src/Doctrine/EntityRegenerator.php
```diff
@@ -129,8 +129,7 @@ public function regenerateEntities(string $classOrNamespace): void
                             ->setIsNullable($getIsNullable($mapping))
                             ->setTargetClassName($mapping['targetEntity'])
                             ->setTargetPropertyName($mapping['inversedBy'])
-                            ->setMapInverseRelation(null !== $mapping['inversedBy'])
-                        ;
+                            ->setMapInverseRelation(null !== $mapping['inversedBy']);
```
jderusse, line 132: revert.

Please do not change things that is not related to the PR.

src/Maker/MakeEntity.php
```diff
@@ -88,8 +92,7 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
             ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Add the ability to broadcast entity updates using Symfony UX Turbo?')
             ->addOption('regenerate', null, InputOption::VALUE_NONE, 'Instead of adding new fields, simply generate the methods (e.g. getter/setter) for existing fields')
             ->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite any existing getter/setter methods')
-            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeEntity.txt'))
-        ;
```
jderusse, line 92: same

src/Maker/MakeEntity.php
```diff
@@ -160,6 +163,22 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+            if (\PHP_VERSION_ID < 80000) {
+                throw new RuntimeCommandException(sprintf('Your PHP version (%s) does not supports attributes, PHP 8 is needed.', \PHP_VERSION));
+            }
+
+            if (!$this->doctrineHelper->isDoctrineSupportingAttributes()) {
+                throw new RuntimeCommandException('Your doctrine version does not supports attributes. You need at least doctrine in version 2.9');
+            }
+
+            if (Kernel::VERSION_ID < 50200) {
+                throw new RuntimeCommandException('Your symfony version does not support attributes. Attribute supports begins with >= Symfony 5.2 ');
```
jderusse, line 178: What feature in `symfony/http-kernel: 5.2` do we need?

Beware, symfony is a set of component. People can install `http-kernel 5.1` alongside `dependency-injection 5.2`

src/Maker/MakeEntity.php
```diff
@@ -221,7 +243,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             if (\is_array($newField)) {
                 $annotationOptions = $newField;
                 unset($annotationOptions['fieldName']);
-                $manipulator->addEntityField($newField['fieldName'], $annotationOptions);
+                $manipulator->addEntityField($newField['fieldName'], $annotationOptions, [], []);
```
jderusse, line 246: same here and at many places in this PR

src/Maker/MakeResetPassword.php
```diff
@@ -371,8 +372,14 @@ private function generateRequestEntity(Generator $generator, ClassNameDetails $r
 
         $generator->writeChanges();
 
+        $useAttributesForDoctrineMapping = $this->doctrineHelper->isDoctrineSupportingAttributes() && $this->doctrineHelper->doesClassUsesAttributes($requestClassNameDetails->getFullName()) && Kernel::VERSION_ID >= 50200;
```
jderusse, line 375: Same comment. You should not rely on version of 1 component.
Instead you should check that the classes you need are present.

src/Maker/MakeUser.php
```diff
@@ -152,19 +156,25 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             $classPath = $this->entityClassGenerator->generateEntityClass(
                 $userClassNameDetails,
                 false, // api resource
-                $userClassConfiguration->hasPassword() && interface_exists(PasswordUpgraderInterface::class) // security user
+                $userClassConfiguration->hasPassword() && interface_exists(PasswordUpgraderInterface::class) // security user,
```
jderusse, line 159: revert, maker-bundle is compatible with php 7.1,  
trailing comma are allowed in +7.3

+ do not change things that are not related to your PR

src/Resources/skeleton/doctrine/Entity.tpl.php
```diff
@@ -2,13 +2,15 @@
 
 namespace <?= $namespace ?>;
 
-<?php if ($api_resource): ?>use ApiPlatform\Core\Annotation\ApiResource;
+<?php if ($api_resource && class_exists(\ApiPlatform\Metadata\ApiResource::class)): ?>use ApiPlatform\Metadata\ApiResource;
+<?php elseif ($api_resource): ?>use ApiPlatform\Core\Annotation\ApiResource;
```
jderusse, line 6: is it related to this PR?

tests/Doctrine/EntityRegeneratorTest.php
```diff
@@ -113,7 +113,7 @@ private function doTestRegeneration(string $sourceDir, Kernel $kernel, string $n
             });
 
         $fileManager = new FileManager($fs, $autoloaderUtil, new MakerFileLinkFormatter(null), $tmpDir);
-        $doctrineHelper = new DoctrineHelper('App\\Entity', $container->get('doctrine'));
+        $doctrineHelper = new DoctrineHelper('App\\Entity', $container->get('doctrine'), false, null);
```
jderusse, line 116: revert, `false` and `null` are the defaults values

jderusse (commented):
Tested on my pet project. it work like expected. However, I've few comments on the code.

jrushlow:
Again awesome work @SimonMarx - I'm going to push up a few changes to this PR to address some of the reviews and add in some tests. Please hold off on pushing any more commits. We're very close now to getting the mergable..

src/Doctrine/EntityRegenerator.php
jrushlow: done

src/Doctrine/EntityRegenerator.php
jrushlow: done

src/Doctrine/DoctrineHelper.php
jrushlow: done

src/Util/ClassSourceManipulator.php
jrushlow: done

src/Maker/MakeEntity.php
jrushlow: done

src/Maker/MakeEntity.php
jrushlow: done

src/Maker/MakeUser.php
jrushlow: done

tests/Doctrine/EntityRegeneratorTest.php
jrushlow: done

karaka200:
Working fine to me, But When I use `symfony console make:auth` create a `SecurityController` that use `annotations`

src/Doctrine/DoctrineHelper.php
```diff
@@ -85,43 +93,67 @@ public function getEntityNamespace(): string
...
+    {
+        return $this->doesClassUseDriver($className, AnnotationDriver::class);
+    }
+
+    public function doesClassUsesAttributes(string $className): bool
+    {
+        return $this->doesClassUseDriver($className, AttributeDriver::class);
+    }
+
+    public function canUseAttributes(): bool
```
weaverryan, line 152: This looks identical in purpose to `isDoctrineSupportingAttributes()`, but this method looks incorrect and `isDoctrineSupportingAttributes` looks correct. Unless I'm missing something, we should use `isDoctrineSupportingAttributes()` everywhere and remove this method.

src/Doctrine/DoctrineHelper.php
```diff
@@ -265,4 +297,36 @@ public function isKeyword(string $name): bool
...
+
+    /**
+     * this method tries to find the correct MappingDriver for the given namespace/class
+     * To determine which MappingDriver belongs to the class we check the prefixes configured in Doctrine and use the
+     * prefix that has the closest match to the given $namespace.
+     *
+     * this helper function is needed to create entities with the configuration of doctrine if they are not yet been registered
+     * in the ManagerRegistry
+     */
+    public function getMappingDriverForNamespace(string $namespace): ?MappingDriver
```
weaverryan, line 309: I think this method can be private. It IS used in `MakeEntity`. However, I think it could be replaced by `-> doesClassUsesAttributes()` and then this could be private.

Well, this method is also used to populate the `$mappingDriver` variable in `MakeEntity`, which is then passed to `MakeEntity::createClassManipulator()`. But, I think instead we should pass a `string $className` to `createClassManipulator` and then it can call methods on `DoctrineHelper` to determine if attributes should be used.

src/Maker/MakeEntity.php
```diff
@@ -160,6 +163,12 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             'Entity\\'
         );
 
+        $mappingDriver = $this->doctrineHelper->getMappingDriverForNamespace($entityClassDetails->getFullName());
+
+        if ($mappingDriver instanceof AttributeDriver && !$this->doctrineHelper->canUseAttributes()) {
+            throw new RuntimeCommandException('To use attributes you\'ll need atleast PHP 8, Doctrine 2.9, and Symfony 5.2.');
```
weaverryan, line 169:
```suggestion
            throw new RuntimeCommandException('To use attributes you\'ll need at least PHP 8, Doctrine 2.9, and Symfony 5.2.');
```

Also, you need `doctrine/orm` 2.9, right? And also `doctrine/doctrine-bundle` at some version?

src/Maker/MakeResetPassword.php
```diff
@@ -371,8 +371,14 @@ private function generateRequestEntity(Generator $generator, ClassNameDetails $r
 
         $generator->writeChanges();
 
+        $useAttributesForDoctrineMapping = $this->doctrineHelper->isDoctrineSupportingAttributes() && $this->doctrineHelper->doesClassUsesAttributes($requestClassNameDetails->getFullName()) && $this->doctrineHelper->canUseAttributes();
```
weaverryan, line 374: As mentioned earlier, I think the `canUseAttributes()` is totally redundant and cab be removed: `isDoctrineSupportingAttributes()` covers us.

src/Maker/MakeUser.php
```diff
@@ -160,11 +164,17 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+        $useAttributesForDoctrineMapping = $userClassConfiguration->isEntity() && ($this->doctrineHelper->isDoctrineSupportingAttributes()) && Kernel::VERSION_ID >= 50200 && $this->doctrineHelper->doesClassUsesAttributes($userClassNameDetails->getFullName());
+
         // B) Implement UserInterface
         $manipulator = new ClassSourceManipulator(
             $this->fileManager->getFileContents($classPath),
-            true
+            true,
+            !$useAttributesForDoctrineMapping,
+            true,
+            $useAttributesForDoctrineMapping
```
weaverryan, line 175: This constructor is complex, in large part because that nasty, confusing 3rd argument (which is my fault).

Anyways, I've seen this twice so far and it gave me an idea: add a `DoctrineHelper::createClassSourceManipulator(string $className)` method that handles all of this ugliness for us.

src/Resources/skeleton/doctrine/Entity.tpl.php
weaverryan: Yea, this is from #968 - what's it doing in here?

src/Test/MakerTestDetails.php
```diff
@@ -352,4 +352,13 @@ public function getSkipMessage(): ?string
...
     }
+
+    public function useDoctrineAttributeMapping(): self
+    {
+        return $this->addReplacement(
+        'config/packages/doctrine.yaml',
+                'type: annotation',
+                'type: attribute'
+        );
+    }
```
weaverryan, line 363: This will likely break not long after merging this PR, as it will allow for this recipe to be merged: https://github.com/symfony/recipes/pull/971/files

I can't immediately think of anything better to do. So, we might just need to wait for that recipe to break our tests and then fix them.

weaverryan (commented):
Wow, tests are green and this is close! I've left some comments for things we should fix before merge - I think @jrushlow is going to keep pushing this across the finish line, but @SimonMarx you've been wonderful on this PR, so if you have some time, just coordinate with jrushlow so you don't both do the work. And thanks again!

src/Doctrine/DoctrineHelper.php
jrushlow: done

src/DependencyInjection/CompilerPass/DoctrineAttributesCheckPass.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+namespace Symfony\Bundle\MakerBundle\DependencyInjection\CompilerPass;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class DoctrineAttributesCheckPass implements CompilerPassInterface
+{
+    public const DOCTRINE_SUPPORTS_ATTRIBUTE = 'maker.compatible_check.doctrine.supports_attributes';
```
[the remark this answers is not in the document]
jrushlow: done

src/Doctrine/DoctrineHelper.php
```diff
@@ -265,4 +297,36 @@ public function isKeyword(string $name): bool
...
+                    $lowestCharacterDiff = $diff;
+                    $foundDriver = $driver;
+                }
+            }
+        }
+
+        return $foundDriver;
+    }
+
+    public function isDoctrineSupportingAttributes(): bool
```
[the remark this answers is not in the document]
jrushlow: done - grouped w/ other related isser's

src/Maker/MakeResetPassword.php
jrushlow: done

src/Maker/MakeUser.php
```diff
@@ -160,11 +164,17 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
         // need to write changes early so we can modify the contents below
         $generator->writeChanges();
 
+        $useAttributesForDoctrineMapping = $userClassConfiguration->isEntity() && ($this->doctrineHelper->isDoctrineSupportingAttributes()) && Kernel::VERSION_ID >= 50200 && $this->doctrineHelper->doesClassUsesAttributes($userClassNameDetails->getFullName());
```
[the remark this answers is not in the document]
jrushlow: done - unneeded as 50200 is already checked in `isDoctrineSupportingAttributes` -> `PhpCompatUtil::canUseAttributes`

src/Resources/skeleton/doctrine/Entity.tpl.php
jrushlow: done - think this made it in with a merge maybe? reverted

src/Doctrine/DoctrineHelper.php
jrushlow: possible solution implemented. needs testing

src/Maker/MakeEntity.php
```diff
@@ -793,9 +805,13 @@ private function askRelationType(ConsoleStyle $io, string $entityClass, string $
         return $io->askQuestion($question);
     }
 
-    private function createClassManipulator(string $path, ConsoleStyle $io, bool $overwrite): ClassSourceManipulator
+    private function createClassManipulator(string $path, ConsoleStyle $io, bool $overwrite, ?MappingDriver $mappingDriver = null): ClassSourceManipulator
```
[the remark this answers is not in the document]
jrushlow: done

src/Maker/MakeEntity.php
```diff
@@ -163,9 +163,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             'Entity\\'
         );
 
-        $mappingDriver = $this->doctrineHelper->getMappingDriverForNamespace($entityClassDetails->getFullName());
-
-        if ($mappingDriver instanceof AttributeDriver && !$this->doctrineHelper->isDoctrineSupportingAttributes()) {
+        if (!$this->doctrineHelper->isDoctrineSupportingAttributes()) {
```
weaverryan, line 166: This would always throw an exception if PHP 8 attributes are supported, right?

We also need `&& $this->doctrineHelper-> doesClassUsesAttributes($entityClassDetails->getFullName())`

jrushlow: good catch

src/Test/MakerTestDetails.php
jrushlow: I'll keep an eye on the PR and our MB test results once we merge this so we can push a fix for the tests pretty quick.

src/Maker/MakeEntity.php
weaverryan: I'd say, let's be specific! Something like:

> To use Doctrine entity attributes, you'll need PHP 8, doctrine/orm 2.9... doctrine/doctrine-bundle X.? and symfony/framework-bundle 5.2

Notice the doctrine-bundle version if `X.?` - so don't just copy and paste it 😉

jrushlow: works for me!

weaverryan:
@SimonMarx I noticed there are 2 of other committers on this PR (besides @jrushlow and I). Can you explain who they are?

* @Geekimo 
* @adlpz 

Were you collaborating with these users?

weaverryan:
@SimonMarx  - Also, see #978 for a cleaned up version of this PR, I need your "ok" on that PR (as well as determining where the other contributors came from).

Thanks!

PR #922, opened 2021-07-05
[make:crud] fix conflict routes

Olyno:
Hi 👋🏻 

I come with a fix for the routes generated by the ``make:crud``. Routes with the "id" parameter conflict with other routes created. And since by default an id is a number, it's best to specify its content with the requirements in my opinion.

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -83,10 +83,10 @@ class <?= $class_name ?> extends <?= $parent_class_name; ?><?= "\n" ?>
     }
 
 <?php if ($use_attributes) { ?>
-    #[Route('/{<?= $entity_identifier ?>}', name: '<?= $route_name ?>_show', methods: ['GET'])]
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -83,10 +83,10 @@ public function new(Request $request): Response
     }
 
 <?php if ($use_attributes) { ?>
-    #[Route('/{<?= $entity_identifier ?>}', name: '<?= $route_name ?>_show', methods: ['GET'])]
+    #[Route('/{<?= $entity_identifier ?>}', name: '<?= $route_name ?>_show', methods: ['GET'], requirements: ['<?= $entity_identifier ?>' => '\d+'])]
```
jrushlow, line 86: Howdy, thanks for the work on this. My biggest concern with this is if my entity has a string based id, e.g. uuid's, these routes will no longer work. For us to properly implement this dynamically, we would need to know the scalar type of the  `id` for the entity.

jrushlow (changes requested):

src/Resources/skeleton/crud/controller/Controller.tpl.php
Olyno: Should I make a better regex to fix that?

Olyno: Any news? @jrushlow

Issue #821, opened 2021-02-21
Form-login authenticators should use the new UserPasswordHasherInterface


chalasr:
As it replaces the deprecated `UserPasswordEncoderInterface` and related service/aliases as of Symfony 5.3.

chalasr:
Fixed in #923, thanks @jrushlow!

PR #923, opened 2021-07-05
use password hasher for make:registration & make:reset-password, includes other improvements

jrushlow:
Use the new password hasher if it exists, or fallback to the old password encoder for `make:reset-password` & `make:registration`.

- introduces the concept of using `ClassNameDetails` objects in `make:registration` templates.
- adds a new `TemplateComponentsGenerator::getPropertyType($classNameDetails)` object that simplifies using typed properties in templates.

_contains `@legacy` tags for internal use:_
- Remove `Generator::getControllerBaseClass()` when `FrameworkBundle`  min supported version >= 4.1
- make:registration `$passwordHasher` conditional can be removed when min Symfony supported version >=5.3
- make:reset-password `$passwordHasher` conditional can be removed when min Symfony supported version >=5.3
- injected controller var's `password_variable_name` & `use_password_hasher` can be removed when min Symfony supported version >=5.3

refs #821

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -247,7 +247,7 @@ class Generator
 
     public static function getControllerBaseClass(): ClassNameDetails
     {
-        // Support for Controller::class can be dropped when FrameworkBundle minimum supported version is >=4.1
+        // @legacy Support for Controller::class can be dropped when FrameworkBundle minimum supported version is >=4.1
         $class = method_exists(AbstractController::class, 'getParameter') ? AbstractController::class : Controller::class;
 
         return new ClassNameDetails($class, '\\');
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -42,6 +42,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Mailer\MailerInterface;
 use Symfony\Component\Mime\Address;
+use Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Routing\RouterInterface;
 use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
@@ -60,11 +61,8 @@ use SymfonyCasts\Bundle\VerifyEmail\SymfonyCastsVerifyEmailBundle;
 final class MakeRegistrationForm extends AbstractMaker
 {
     private $fileManager;
-
     private $formTypeRenderer;
-
     private $router;
-
     private $doctrineHelper;
 
     private $userClass;
@@ -276,6 +274,16 @@ final class MakeRegistrationForm extends AbstractMaker
             'Controller\\'
         );
 
+        /*
+         * @legacy Conditional can be removed when MakerBundle no longer
+         *         supports Symfony < 5.2
+         */
+        $passwordHasher = UserPasswordEncoderInterface::class;
+
+        if (interface_exists(UserPasswordHasherInterface::class)) {
+            $passwordHasher = UserPasswordHasherInterface::class;
+        }
+
         $useStatements = [
             Generator::getControllerBaseClass()->getFullName(),
             $formClassDetails->getFullName(),
@@ -283,7 +291,7 @@ final class MakeRegistrationForm extends AbstractMaker
             Request::class,
             Response::class,
             Route::class,
-            UserPasswordEncoderInterface::class,
+            $passwordHasher,
         ];
 
         if ($this->willVerifyEmail) {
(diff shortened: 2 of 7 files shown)


src/Maker/MakeRegistrationForm.php
```diff
@@ -313,15 +321,17 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
                     'user_class_name' => $userClassNameDetails->getShortName(),
                     'password_field' => $this->passwordField,
                     'will_verify_email' => $this->willVerifyEmail,
+                    'email_verifier_details' => new TemplateClassDetails($verifyEmailServiceClassNameDetails->getFullName(), $this->phpCompatUtil->canUseTypedProperties()),
```
weaverryan, line 324: I don't like needing to pass that 2nd arg - it means me think that a factory might be better. What if we had a `$generator->createClassHelper(string $fqcn): TemplateClassHelper` method?

I also changed `Details` to `Helper`... that's subjective :)

weaverryan: Also, maybe we call the variable `email_verifier_class`? I know that makes it sound like a string, but it's more descriptive than "details". And, we could add a `__toString()` if we wanted to, which would return the fcqn.

src/Resources/skeleton/registration/RegistrationController.tpl.php
```diff
@@ -7,11 +7,11 @@
 class <?= $class_name; ?> extends <?= $parent_class_name; ?><?= "\n" ?>
 {
 <?php if ($will_verify_email): ?>
-    private $emailVerifier;
+<?= $email_verifier_details->getPropertyStatement() ?>
```
weaverryan, line 10: If there any advantage to having the `getPropertyStatement()` do the indentation and `;` for us? That seems like extra magic - and a little bit less readability here.

src/Resources/skeleton/registration/RegistrationController.tpl.php
```diff
@@ -7,11 +7,11 @@
...
 
-    public function __construct(EmailVerifier $emailVerifier)
+    public function __construct(<?= $email_verifier_details->getMethodArgument() ?>)
     {
-        $this->emailVerifier = $emailVerifier;
+        <?= $email_verifier_details->getConstructorArgument(true, false) ?>
```
weaverryan, line 14: I don't see the advantage in these methods of having the property name itself encoded in them. The method is useful for creating the property (which may or may not need a type), and it's even useful (though less-so) for the type-hint on an argument, but I don't see the value in this line (or other lines further down that previously said `$this->emailVerifier...`.

It makes me wonder if a generic helper function just for generating the property type would be simpler:

```php
    private <?= $generator->property($email_verifier_class, 'emailVerifier') ?>;
```

This is faux code - but the first arg would be the FCQN and the 2nd arg would be the property name. 

That still leaves:

> how to type-hint methods in a simpler way without needing to pass 2 variables into the template, one for the fcqn and one for the short name.

We could solve that by passing in the "class name details" objects into the template and using those directly. So, perhaps we pass a "class name details" to the above method. And later for an argument type-hint, we use:

```php
public function __construct(<?= $email_verifier_class->getShortName() ?> $emailVerifier)
```

This would only need really be necessary if a class is dynamic. In most cases, we would just hardcode the class - e.g. `RouterInterface $router`.

weaverryan (commented):
I like the aim for SURE - but I'm not convinced on the details yet :)

src/Generator.php
```diff
@@ -156,6 +157,14 @@ public function createClassNameDetails(string $name, string $namespacePrefix, st
         return new ClassNameDetails($className, $fullNamespacePrefix, $suffix);
     }
 
+    /**
+     * @experimental The method signature may change, moved, or be removed.
```
jrushlow, line 161: We should keep this experimental until we've implemented with a few different makers as the `Generator::class` is not `internal` or `final`

src/Maker/MakeRegistrationForm.php
jrushlow: I think I strayed away from this initially as the `Generator` is not internal. But adding an `@experimental` tag to the method takes care of that. Love it and done.

src/Generator.php
jrushlow: that didnt last long

src/Maker/MakeRegistrationForm.php
```diff
@@ -276,14 +274,24 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             'Controller\\'
         );
 
+        /*
+         * @legacy Conditional can be removed when MakerBundle no longer
+         *         supports Symfony < 6.0
```
weaverryan, line 279: Couldn't we drop this once MakerBundle no longer supports 5.2? The thinking is: `UserPasswordHasherInterface` is new in 5.3. If an app is using 5.3, then we will ALWAYS give them the new class. So once MakerBundle requires 5.3 or higher, we can drop this. That's much sooner than 6.0 :)

src/Maker/MakeResetPassword.php
```diff
@@ -186,22 +198,51 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             'Form\\'
         );
 
+        /*
+         * @legacy Conditional can be removed when MakerBundle no longer
+         *         supports Symfony < 6.0.
```
weaverryan, line 203: Same here - 5.3 is the min version when we can drop this.

weaverryan (commented):
Looks awesome! It's a much smaller/smoother update now. Just a small note about the legacy version

src/Maker/MakeRegistrationForm.php
jrushlow: agreed - i was thinking when the old interface is removed, not when the new one was added.. lets do it!

src/Maker/MakeResetPassword.php
jrushlow: done

weaverryan:
Thanks Jesse!

PR #926, opened 2021-07-13
feat: make:entity php 8 attributes support

Geekimo:
Based on #920
- Updated tests
- Re-added deleted annotated class detection, and added attribute mapping detection.

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -16,6 +16,7 @@ use Doctrine\Common\Persistence\Mapping\ClassMetadata as LegacyClassMetadata;
 use Doctrine\Common\Persistence\Mapping\MappingException as LegacyPersistenceMappingException;
 use Doctrine\DBAL\Connection;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\ORM\Mapping\MappingException as ORMMappingException;
 use Doctrine\ORM\Mapping\NamingStrategy;
 use Doctrine\ORM\Tools\DisconnectedClassMetadataFactory;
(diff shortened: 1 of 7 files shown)


src/Doctrine/DoctrineHelper.php
```diff
@@ -268,4 +307,27 @@ public function isKeyword(string $name): bool
...
+     * prefix that has the closest match to the given $namespace.
+     */
+    public function getMappingDriverForNamespace(string $namespace): ?MappingDriver
+    {
+        $lowestCharacterDiff = null;
+        $foundDriver = null;
+
+        foreach ($this->annotatedPrefixes as $key => $mappings) {
+            foreach ($mappings as [$prefix, $driver]) {
+                $diff = substr_compare($namespace, $prefix, 0);
```
php-programmist, line 323: This works incorrect. I have installed package gesdinet/jwt-refresh-token-bundle wich has annotated entity. And I have next:
substr_compare('App\Entity\Test', 'App\Entity', 0);//5
substr_compare('App\Entity\Test', 'Gesdinet\JWTRefreshTokenBundle\Entity', 0);//-6
And AnnotationDriver returns instead of AttributeDriver

php-programmist (changes requested):
Not working for me.
I have installed package gesdinet/jwt-refresh-token-bundle wich has annotated entity. And I have next situation:
substr_compare('App\Entity\Test', 'App\Entity', 0);//5
substr_compare('App\Entity\Test', 'Gesdinet\JWTRefreshTokenBundle\Entity', 0);//-6
And AnnotationDriver returns instead of AttributeDriver

src/Doctrine/DoctrineHelper.php
Geekimo: Hi !
This behavior sounds logical, as the purpose of this part of the code is to return the driver used for a mappping.
So for Gesdinet\JWTRefreshTokenBundle\Entity, which is annotated, the fact that you have a AnnotationDriver is correct.
Am I missing something ?

php-programmist: I have 2 prefixes (name spaces):
1) App\Entity - Attribute Driver
2) Gesdinet\JWTRefreshTokenBundle\Entity - Annotation Driver
I executed next command:
`php bin/console make:entity Test`
I expected that would be created App\Entity\Test entity with attributes, but I got - annotations!
I figured out that method `getMappingDriverForNamespace` returns incorrect Driver. Now this method returns the last driver of `$this->annotatedPrefixes` array, not the closest match.
I see 2 problems in mentioned block:
1) `$lowestCharacterDiff` always `null`. And if statement (row 325) always `true`
2) For entity `App\Entity\Test` closest match must be `App\Entity`. But
`substr_compare('App\Entity\Test', 'App\Entity', 0); ` results as `5`
`substr_compare('App\Entity\Test', 'Gesdinet\JWTRefreshTokenBundle\Entity', 0);` results as `-6`
-6 is less than 5, and this algorithm choosing incorrect driver
In my fork I using next variant of `if` block:
```php
if (null === $lowestCharacterDiff || ($diff < $lowestCharacterDiff && $diff > 0)) {
      $lowestCharacterDiff = $diff;
      $foundDriver = $driver;
}
```
I am not absolutely sure that it is correct, but it work for me.

Geekimo: Ah ok, your first message wasn't that explicit.
I'm gonna check differently this part.

Geekimo: This point has been fixed by @adlpz in #920

jrushlow:
Good catch @fabpot. 

Feature consolidated in PR #978 from this PR & PR #920. Thanks @Geekimo for your contribution to this awesome new feature!

PR #927, opened 2021-07-14
[No Merge] Debugging ci versioning

weaverryan:
We have a mystery with local vs CI where we get all 5.4 packages locally for the dev build but a mixture of 5.3 and 5.4 (exactly as if symfony/flex were not filtering only on the CI)


weaverryan:
Debugging complete :)

PR #931, opened 2021-07-26
Fix: usage of deprecated $container property

Nek-:
This is a deprecation that arrived with Symfony 5.3. 

This should be merged only before Symfony 6.0 IMO. Or another solution depending on the Symfony version should be implemented. Your call.


jrushlow:
Thanks for this @Nek- This has been fixed in a separate PR.

PR #933, opened 2021-08-05
[make:entity] Remove deprecated json_array type from available list types.

bdaler:
In a fix #482 json_array type was removed from using but it's available in a list of types. In section Other Types


jrushlow (approved):
Thank you @bdaler

weaverryan:
Thanks @bdaler!

Issue #911, opened 2021-06-30
[make:reset-password] Missing symfony/form requirement (From SymfonyCasts/reset-password-bundle)


jrushlow:
_Original Author @apphancer_

When running `make:reset-password` there should be a missing package error if the following are missing:

- symfony/form
- symfony/validator

At the moment, if symfony/mailer is not available, the `make:reset-password` command returns this as error suggesting to run ` composer require symfony/mailer`.
It would be helpful if the same happened for `symfony/form`.

Transfered issue to maker-bundle from https://github.com/SymfonyCasts/reset-password-bundle/issues/173

PR #930, opened 2021-07-20
Add all missed dependencies to make:reset-password

upyx:
Closes #911

PR #935, opened 2021-08-06
[make:reset-password] Add required packages to MakeResetPassword.

bdaler:
As mentioned in issue: https://github.com/symfony/maker-bundle/issues/911 add the packages `symfony/form` and `symfony/validator` as required.

### DIFF
--- a/src/Maker/MakeResetPassword.php
+++ b/src/Maker/MakeResetPassword.php
@@ -102,6 +102,8 @@ class MakeResetPassword extends AbstractMaker
     {
         $dependencies->addClassDependency(SymfonyCastsResetPasswordBundle::class, 'symfonycasts/reset-password-bundle');
         $dependencies->addClassDependency(MailerInterface::class, 'symfony/mailer');
+        $dependencies->addClassDependency(Forms::class, 'symfony/form');
+        $dependencies->addClassDependency(Validation::class, 'symfony/validator');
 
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks for this @bdaler!

Before I merge, one question is: why don't our tests catch this? 

My guess is due to these lines: https://github.com/symfony/maker-bundle/blob/main/tests/Maker/MakeResetPasswordTest.php#L141 - at least some of these should not be here.

These "extra dependencies" should be used sparingly - usually only if a maker command has some optional dependency, but we want to install it in the test to make sure the code generates correctly WITH that optional dependency. Looking at the list in the test, I'm wondering if we should remove ALL of these - even `symfony-bundle` and `twig` should probably be moved to REAL dependencies in the maker itself - the command relies on security being installed and Twig. The same is true (I think) for `doctrine`, `mailer`, & `doctrine` (`doctrine/annotations` *should* be an optional dependency.... I think... in case you're using PHP 8 attributes, which is a totally different complication - so probably that one should be left alone for now).

@jrushlow can you verify that i'm correct? That we should probably remove all of the "extra dependencies" in the test (with the exception, perhaps, of `doctrine/annotations`) and move them to real dependencies of the maker?

weaverryan:
Update: I had missed #930 - I believe that has the correct behavior.

PR #963, opened 2021-08-27
add return types for symfony 6

jrushlow:
- adds return types required to support Symfony 6

Also implements return types in other areas of the bundle as well.

Methodology:
- `internal` xor `final` classes return types added on all methods where possible
- All other classes, deprecation(s) to be added under a separate PR for methods that do not have a return type. return types to be added in the next major release.


weaverryan (approved):
Good careful work! 👍

Issue #964, opened 2021-09-01
make:entity fails with entity name "Repository"


jzawadzki:
I know that `Repository` isn't THE BEST name for an entity, but I found that `make:entity` have problem with generation of such.

Maker creates Repository called `Repository` instead of `RepositoryRepository` and because of that, their `use` sections conflicts (same name of Entity and Repository)

Steps to reproduce:
```
symfony new test;
cd test;
composer req orm;
composer req maker;
php bin/console make:entity Repository
```

Result:
```
PHP Fatal error:  Cannot declare class App\Entity\Repository because the name is already in use in /home/jerzy/test/src/Entity/Repository.php on line 11
```

I see two possible solutions:
1) Fix the bug 
2) Block "Repository" as reserved keyword

OskarStark:
I would propose to fix the bug, to me Repository is a valid word. For example when working with code which handles GitHub repositories or anything else.

jzawadzki:
In the meantime I found the cause:
https://github.com/symfony/maker-bundle/blob/main/src/Str.php#L36
 ```
/**
  * Ensures that the given string ends with the given suffix. If the string
  * already contains the suffix, it's not added twice. It's case-insensitive
  * (e.g. value: 'Foocommand' suffix: 'Command' -> result: 'FooCommand').
  */
public static function addSuffix(string $value, string $suffix): string
```

jrushlow:
I agree w/ @OskarStark - adding `Repository` as a "reserved word" is not ideal and will lead to problems. e.g. you can create a `User::class` entity, but using the word `user` for a property in a relation (`Members::getUsers()`) causes problems in some setups.. 

Let's fix the actual problem instead. @jzawadzki , are you up to submitting a PR?

PR #967, opened 2021-09-03
Changed behaviour of Str::addSuffix so it doesn't care about duplicate suffixes

jzawadzki:
This PR tries to resolve the issue: https://github.com/symfony/maker-bundle/issues/964

I'm still not sure why `Str::addSuffix` wanted to be sure that suffix is not added twice. So, let someone with bigger knowledge of the bundle and class name generation review the PR :)

Thanks!

### DIFF
--- a/src/Str.php
+++ b/src/Str.php
@@ -32,13 +32,11 @@ final class Str
     }
 
     /**
-     * Ensures that the given string ends with the given suffix. If the string
-     * already contains the suffix, it's not added twice. It's case-insensitive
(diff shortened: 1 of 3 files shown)


OskarStark:
We should maybe have "addPrefix()" and "ensurePrefix()", where add adds it no matter the string looks before and ensure ensures it is only added once?

EnsureEnd() is the method used in Symfony string component

jrushlow (approved):
Thank you @jzawadzki!

PR #968, opened 2021-09-06
[make:entity] APIP: use new attribute if exists

nikophil:
Hi,

here is an update to use new `ApiResource` attribute from api platofrm v3 when providing `--api-resource` flag in `make:entity` command.

This cannot be tested yet since v3 is not released

### DIFF
--- a/src/Doctrine/EntityRegenerator.php
+++ b/src/Doctrine/EntityRegenerator.php
@@ -44,7 +44,7 @@ final class EntityRegenerator
     {
         try {
             $metadata = $this->doctrineHelper->getMetadata($classOrNamespace);
-        } catch (MappingException | LegacyCommonMappingException | PersistenceMappingException $mappingException) {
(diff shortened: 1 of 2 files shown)


src/Doctrine/EntityRegenerator.php
```diff
@@ -44,7 +44,7 @@ public function regenerateEntities(string $classOrNamespace): void
     {
         try {
             $metadata = $this->doctrineHelper->getMetadata($classOrNamespace);
-        } catch (MappingException | LegacyCommonMappingException | PersistenceMappingException $mappingException) {
+        } catch (MappingException|LegacyCommonMappingException|PersistenceMappingException $mappingException) {
```
nikophil, line 47: just fixing CS problem

jrushlow (approved):
Awesome! Thanks @nikophil

weaverryan:
Thanks Nicolas!

PR #970, opened 2021-09-13
make PhpCompatUtil::getPhpVersion() public

nikophil:
Hello,

what do you think if we make the method `PhpCompatUtil::getPhpVersion()` public?

This could be needed for some libs which leverage maker bundle

This would be helpful to resolve this problem: https://github.com/api-platform/core/pull/4423#pullrequestreview-747675271

Another solution for my problem would be to expose a global `is_php8` variable to all templates. but maybe that's not a good solution, because way _may_ need to differentiate minor versions

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -173,6 +173,7 @@ class Generator
         $variables['relative_path'] = $this->fileManager->relativizePath($targetPath);
         $variables['use_attributes'] = $this->phpCompatUtil->canUseAttributes();
         $variables['use_typed_properties'] = $this->phpCompatUtil->canUseTypedProperties();
+        $variables['use_union_types'] = $this->phpCompatUtil->canUseUnionTypes();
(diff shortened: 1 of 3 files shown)


jrushlow:
We already have a global `use_attributes` & `use_typed_properties` variable available in the maker templates - https://github.com/symfony/maker-bundle/blob/203699de17bde367060f3ff12128cdea1fa783dd/src/Generator.php#L175

What additional features are needed in PHP8 that are not covered by these 2 methods? If union return types is needed, adding a `canUseUnionTypes()` method would be the way to go vs. opening up the `getPhpVersion()` scope as the class itself is `internal`.

nikophil:
Hello @jrushlow 

thanks for your feedback. I indeed felt it was a little bit redundant, and not flexible enough.

Here is a fix. Cheers!

Issue #975, opened 2021-09-21
make:entity ApiResource question not called when entity pass as an argument


weaverryan:
Described well here: https://symfonycasts.com/screencast/api-platform/api-resource#comment-5541407296

Basically, if you pass the entity class name to `make:entity` - e.g. `bin/console make:entity CheeseListing`, then it does NOT ask:

> Mark this class as an API Platform resource?

The problem is that, in `interact()`, if the `name` argument is passed, we immediately return: https://github.com/symfony/maker-bundle/blob/884f10da90939612f26989fa6a789e46c1194930/src/Maker/MakeEntity.php#L97-L101

That should not happen. If the `name` argument is passed, then:

A) We obviously should not ask for the entity name - https://github.com/symfony/maker-bundle/blob/884f10da90939612f26989fa6a789e46c1194930/src/Maker/MakeEntity.php#L115-L119
B) We should also not ask for it here: https://github.com/symfony/maker-bundle/blob/884f10da90939612f26989fa6a789e46c1194930/src/Maker/MakeEntity.php#L104-L110
C) But we should ask about the api-resource and broadcast: https://github.com/symfony/maker-bundle/blob/884f10da90939612f26989fa6a789e46c1194930/src/Maker/MakeEntity.php#L121-L143

So basically, this if (https://github.com/symfony/maker-bundle/blob/884f10da90939612f26989fa6a789e46c1194930/src/Maker/MakeEntity.php#L99-L101) statement should not have a `return`... the first two sections should probably just be moved inside of it: https://github.com/symfony/maker-bundle/blob/884f10da90939612f26989fa6a789e46c1194930/src/Maker/MakeEntity.php#L103-L119

Cheers!

PR #976, opened 2021-09-22
ApiResource question now called when entity is passed as an argument

shikharvashistha:
Fixes : #975

Greetings of the day @weaverryan,@weaverryan

I've refactored the code as desired. Please review the pr and let me know in case of any changes if required.

Thanks & Regards
@shikharvashistha

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -111,26 +111,35 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
     public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
     {
         if ($input->getArgument('name')) {
-            return;
+            if (
+                !$input->getOption('broadcast') &&
+                class_exists(Broadcast::class) &&
+                !class_exists($this->generator->createClassNameDetails($entityClassName, 'Entity\\')->getFullName())
+            ) {
+                $description = $command->getDefinition()->getOption('broadcast')->getDescription();
+                $question = new ConfirmationQuestion($description, false);
+                $isBroadcast = $io->askQuestion($question);
+
+                $input->setOption('broadcast', $isBroadcast);
+            }
+            if ($input->getOption('regenerate')) {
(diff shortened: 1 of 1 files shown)


src/Maker/MakeEntity.php
```diff
@@ -97,26 +97,35 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
 
-        $argument = $command->getDefinition()->getArgument('name');
-        $question = $this->createEntityClassQuestion($argument->getDescription());
-        $entityClassName = $io->askQuestion($question);
+            $argument = $command->getDefinition()->getArgument('name');
+            $question = $this->createEntityClassQuestion($argument->getDescription());
+            $entityClassName = $io->askQuestion($question);
 
-        $input->setArgument('name', $entityClassName);
+            $input->setArgument('name', $entityClassName);
```
weaverryan, line 127: I'm seeing a failure in the CI: https://github.com/symfony/maker-bundle/pull/976/checks?check_run_id=3670902563#step:14:28

We still nee the old if statement - https://github.com/symfony/maker-bundle/blob/3ca4dcc50af202114324cfad6bbe8775df6fa4b5/src/Maker/MakeEntity.php#L99 - the difference is that, instead of a `return` inside, the `regenerate` and "asking for a class name" - https://github.com/symfony/maker-bundle/blob/3ca4dcc50af202114324cfad6bbe8775df6fa4b5/src/Maker/MakeEntity.php#L103-L119 - should be moved inside if that if statement :)

shikharvashistha: Done ! 👍

jrushlow (changes requested):
@shikharvashistha Thanks for the work on this. Can you rebase this PR so we can see what the test suite looks like? Thank you in advance!

shikharvashistha:
> @shikharvashistha Thanks for the work on this. Can you rebase this PR so we can see what the test suite looks like? Thank you in advance!

:+1:

PR #986, opened 2021-10-03
[make:auth] Add `RememberMeBadge`

### LINKED ISSUE #848: make:auth Login form: we should add RememberMeBadge
In the new authenticator system, if you want to support "remember me" functionality, then your authenticator needs to return a `RememberMeBadge`. This advertises that it is "ok to activate remember me on this authenticator". It is safe to always add in our generated code because if the user doesn't have remember me activated (either in config or they don't have the remember me checkbox), nothing will happen. For example, it is ALWAYS returned from the core `FormLoginAuthenticator`.

Probably we will also need one comment line above the new badge to explain its purpose.

EDIT: And I wonder if we should even ask "Do you want to support remember me?" during this process? We could then ask "Do you want remember me to be activated via a checkbox or always activated"? We could use this to determine how the template is generated AND to automatically add the correct `remember_me` config to `security.yaml`.

bechir:
Fix #848 

@weaverryan in #848:
> EDIT: And I wonder if we should even ask "Do you want to support remember me?" during this process? We could then ask "Do you want remember me to be activated via a checkbox or always activated"? We could use this to determine how the template is generated AND to automatically add the correct remember_me config to security.yaml.

- [x] Ask "Do you want to support remember me?" and "Do you want remember me to be activated via a checkbox or always activated"?
- [x] Add RememberMeBadge in `LoginFormAuthenticator` when remember me is supported
- [x] Update generated `security.yaml`
- [x] Update `security/login.html.twig`

### DIFF
--- a/src/Maker/MakeAuthenticator.php
+++ b/src/Maker/MakeAuthenticator.php
@@ -49,6 +49,7 @@ use Symfony\Component\Security\Http\Authentication\AuthenticationUtils;
 use Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator;
 use Symfony\Component\Security\Http\Authenticator\AbstractLoginFormAuthenticator;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\CsrfTokenBadge;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\RememberMeBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Credentials\PasswordCredentials;
 use Symfony\Component\Security\Http\Authenticator\Passport\Passport;
@@ -65,6 +66,9 @@ final class MakeAuthenticator extends AbstractMaker
 {
     private const AUTH_TYPE_EMPTY_AUTHENTICATOR = 'empty-authenticator';
     private const AUTH_TYPE_FORM_LOGIN = 'form-login';
+
+    private const REMEMBER_ME_TYPE_ALWAYS = 'always';
+    private const REMEMBER_ME_TYPE_CHECKBOX = 'checkbox';
 
     public function __construct(
         private FileManager $fileManager,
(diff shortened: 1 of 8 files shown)


src/Maker/MakeAuthenticator.php
```diff
@@ -194,20 +194,43 @@ function ($answer) {
...
+            );
+
+            if($input->getArgument('support-remember-me')) {
+                $command->addArgument('always-remember-me', InputArgument::REQUIRED);
+                $input->setArgument(
+                    'always-remember-me',
+                    $io->confirm(
+                        'Do you want remember me to be activated via a checkbox or always activated?',
+                        false
+                    )
```
weaverryan, line 214: Hmm, the question isn't very clear. Perhaps the user should choose between 2 options instead of a yes/no:

[0] Activate when the user checks a box
[1] Always activate remember me

src/Resources/skeleton/authenticator/Security52LoginFormAuthenticator.tpl.php
```diff
@@ -39,7 +39,7 @@ public function authenticate(Request $request): PassportInterface
             new UserBadge($<?= $username_field_var ?>),
             new PasswordCredentials($request->request->get('password', '')),
             [
-                new CsrfTokenBadge('authenticate', $request->request->get('_csrf_token')),
+                new CsrfTokenBadge('authenticate', $request->request->get('_csrf_token')),<?= $remember_me_badge ? "\n\t\t\t\tnew RememberMeBadge(),\n" : "\n" ?>
```
weaverryan, line 42: That's ugly, but I can't remember if there is a nicer way to do it :/

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -25,6 +25,29 @@
...
+        Uncomment this section and comment the remember_me.always_remember_me option below your
+        firewall to let the user to opt-out of the remember me feature
+        See https://symfony.com/doc/current/security/remember_me.html#forcing-the-user-to-opt-out-of-the-remember-me-feature
+
+        <div class="checkbox mb-3">
+            <label>
+                <input type="checkbox" name="_remember_me"> Remember me
+            </label>
+        </div>
+    #}
```
weaverryan, line 41: I don't think we need this first section at all: if they choose "always remember me", let's not generate anything in this template.

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -36,6 +59,7 @@
             </label>
         </div>
     #}
```
weaverryan, line 61: Now that we're asking the question explicitly, if the user chooses "no" to remember me, let's not generate anything (so delete the `else` case and the code above)

src/Security/SecurityConfigUpdater.php
```diff
@@ -138,6 +138,30 @@ public function updateForAuthenticator(string $yamlSource, string $firewallName,
...
+                $firewall['remember_me']['lifetime'] ??= 604800;
+                $firewall['remember_me']['path'] ??= '/';
+            }
+
+            if($alwaysRememberMe) {
+                $firewall['remember_me']['always_remember_me'] = true;
+            } else {
+                $firewall['remember_me'][] = $this->manipulator->createCommentLine('always_remember_me: true');
+            }
+        }
```
weaverryan, line 163: Wow, you've done an amazing job being VERY user-friendly here :)

bechir:
Hi @weaverryan !

Thank you for your feedbacks.

Sorry for the late. I just saw that I had a PR in progress.

I updated the code as you suggested me above.

Updated existing tests add added new tests: one for checkbox and an other for `always_remember_me` (It seems to work)

sadikoff:
Hey @bechir 

Sorry for so late ping a lot of time has passed, but will you be able to update your PR and do some last pending fixes?

Cheers!

bechir:
Hi @sadikoff 
I rebased my work and tests passed.
Can you check it please?

weaverryan:
Sorry for how slow this was to get merged, but thank you for your EXCELLENT work on this feature - very good stuff @bechir!

PR #994, opened 2021-10-14
Fix replacing default MERCURE_URL in tests

HypeMC:
The test failing seems to be caused by https://github.com/symfony/recipes/pull/1007


weaverryan:
Thank you @HypeMC - super helpful!

PR #1004, opened 2021-11-03
Fix FQCN of 'security.authentication.success' event

AlexBevilacqua:
can be verified here:

https://github.com/symfony/security-core/blob/5.4/AuthenticationEvents.php
https://github.com/symfony/security-core/blob/6.0/AuthenticationEvents.php


weaverryan:
Thank you @AlexBevilacqua!

PR #1015, opened 2021-11-17
Update ApiTestCase to be compliant with ApiPlatform v3.0

laryjulien:
Following the work in progress in https://github.com/api-platform/core/pull/4556, ApiTestCase will be moved from ApiPlatform\Core\Bridge\Symfony\Bundle\Test to ApiPlatform\Symfony\Bundle\Test. The current file location will be deprecated in 2.7 and become invalid in 3.0 .

As recommended by @dunglas, this implementation is compatible with all versions of API Platform.

### DIFF
--- a/src/Maker/MakeTest.php
+++ b/src/Maker/MakeTest.php
@@ -11,7 +11,8 @@
 
 namespace Symfony\Bundle\MakerBundle\Maker;
 
-use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase;
+use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase as LegacyApiTestCase;
+use ApiPlatform\Symfony\Bundle\Test\ApiTestCase;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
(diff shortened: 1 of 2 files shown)


src/Maker/MakeTest.php
```diff
@@ -103,7 +104,7 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
             );
         }
 
-        if ('ApiTestCase' === $input->getArgument('type') && !class_exists(ApiTestCase::class)) {
+        if ('ApiTestCase' === $input->getArgument('type') && !(class_exists(ApiTestCase::class) || class_exists(LegacyApiTestCase::class))) {
```
dunglas, line 107:
```suggestion
        if ('ApiTestCase' === $input->getArgument('type') && !class_exists(ApiTestCase::class) && !class_exists(LegacyApiTestCase::class)) {
```

dunglas (changes requested):

src/Maker/MakeTest.php
laryjulien: Condition has been changed

weaverryan (approved):

dunglas (approved):

PR #1016, opened 2021-11-20
Fix PHP 8.1 deprecations

derrabus:
This PR fixes two pieces of code that trigger a deprecation error on PHP 8.1. You can see them in the CI output of #1012.

### DIFF
--- a/src/Util/YamlSourceManipulator.php
+++ b/src/Util/YamlSourceManipulator.php
@@ -714,7 +714,7 @@ class YamlSourceManipulator
     private function convertToYaml($data)
     {
         $indent = $this->depth > 0 && isset($this->indentationForDepths[$this->depth])
-            ? $this->indentationForDepths[$this->depth] / $this->depth
+            ? intdiv($this->indentationForDepths[$this->depth], $this->depth)
             : 4;
 
         $newDataString = Yaml::dump($data, 4, $indent);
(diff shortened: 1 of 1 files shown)


src/Util/YamlSourceManipulator.php
```diff
@@ -714,7 +714,7 @@ private function updateContents(string $newContents, array $newData, int $newPos
     private function convertToYaml($data)
     {
         $indent = $this->depth > 0 && isset($this->indentationForDepths[$this->depth])
-            ? $this->indentationForDepths[$this->depth] / $this->depth
```
derrabus, line 717: 💡 This division might evaluate to a float value. The problem is that `$indent` is implicitly cast to an integer later and PHP 8.1 will complain if that implicit cast would result in a loss of precision. Since the loss in precision is actually desired here, I'm performing an integer division instead which will always yield an integer.

src/Util/YamlSourceManipulator.php
```diff
@@ -1132,7 +1132,7 @@ private function removeMetadataKeys(array $data)
                 unset($data[$key]);
             }
 
-            if (0 === strpos($val, self::COMMENT_PLACEHOLDER_VALUE)) {
+            if (null !== $val && 0 === strpos($val, self::COMMENT_PLACEHOLDER_VALUE)) {
```
derrabus, line 1135: 💡 Passing `null` to `strpos()` will trigger a depreaction on PHP 8.1.

weaverryan:
Thank you @derrabus!

PR #1017, opened 2021-11-22
[reset-password] fix missing entity manager di

jrushlow:
DI the `EntityManagerInterface` into the constructor as the private method is ultimately called in https://github.com/symfony/maker-bundle/blob/63dc8ba1089f7ab7ae400b03dbee347f8b631275/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php#L41 as well as the public `reset` method as well.


weaverryan (approved):
👍  We need a test for this, as this shouldn't have worked, but nothing failed. But I'll merge now so we can get this fix in a release

weaverryan:
Thanks Jesse!

PR #1025, opened 2021-11-30
fix make:auth

sinus22:


weaverryan:
Hey @sinus22!

Thanks for opening this :). Unfortunately, the fix isn't quite correct - this method ONLY applied to the old Guard authentication. To basically, we just need to make it NOT explode (or not be called at all) when the user is using the new authenticator system. See #1026.

But, thanks again for opening this - it's a critical bug I want to get fixed!

PR #1028, opened 2021-12-01
Add typed properties for make:reset-password and make:registration

seb-jean:


jrushlow (changes requested):
Because maker-bundle still supports PHP < 7.4, we'll need a conditional for typed properties. 

https://github.com/symfony/maker-bundle/blob/bff3fbbe03d0fbd691d921bb1034c929effd825b/src/Resources/skeleton/authenticator/Security52LoginFormAuthenticator.tpl.php#L25

jrushlow (approved):
Thank you @seb-jean!

PR #1030, opened 2021-12-01
Propose to Wrap the Migration in a transaction 

zairigimad:
Propose to wrap the migration in transaction

### DIFF
--- a/src/Maker/MakeMigration.php
+++ b/src/Maker/MakeMigration.php
@@ -127,7 +127,7 @@ final class MakeMigration extends AbstractMaker implements ApplicationAwareMaker
 
         $io->text([
             sprintf('Next: Review the new migration <info>%s</info>', $migrationName),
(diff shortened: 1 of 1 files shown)


chapterjason:
I'm unsure about that. IIRC The option is about to run multiple migrations in one transaction (all-or-nothing) but in the moment I generate a migration all the migrations before musst already run or I will get some duplicates. (The command checks that and warns you already if you doesn't run all available migrations beforehand)

For me it doesn't make sense to recommend the command with the all-or-nothing option.

sadikoff:
I don't see the sense in changing tip about running migration, it up to user how he will run it and in most cases it will be enough to just run migrations without any additional flags. And in case that it's just a tip about how to run migration it should be pretty straightforward.

just my opinion, Cheers!

jrushlow:
I'm going to vote this down, as @sadikoff said, the user has the ability to run the migrations how they see fit. And I don't really see the benefit of running the migrations as a transaction by default. Thank you though for the idea!

PR #1039, opened 2021-12-18
make:entity Add use of Doctrine types constants in attributes

Geekimo:
Makes use of  `Doctrine\DBAL\Types\Types` in entities using PHP8 attributes instead of string values for column type.


Geekimo:
Hello,
Looks like it's ok from my side. All failing tests doesn't look to be linked to my modifications. If anyone can give me a hint if this is the case.

jrushlow (changes requested):
Thanks @Geekimo for the work on this. Could you `squash` & `rebase` this please? Also github is flagging your commits as `Unverified` due to lack of a commit signature. (Github vigilante mode)

weaverryan:
I've continued the work (with the original commits) over on #1147. Thanks @Geekimo for kicking this off!

Issue #979, opened 2021-09-27
make:entity add PHPDoc to Collection (getBars()) getter method for auto-completion


weaverryan:
https://blog.jetbrains.com/phpstorm/2021/09/phpstorm-2021-3-early-access-program-is-open/#improved_support_for_doctrine_collections

So:

```diff
+/**
+ * @return Collection<int, Bar>
+ */
public function getBars(): Collection
{

}
```

The `@return` will aid auto-completion and is more correct than saying `@return Bar[]`, which is what we do currently: https://github.com/symfony/maker-bundle/blob/884f10da90939612f26989fa6a789e46c1194930/src/Util/ClassSourceManipulator.php#L604-L610

ariashahdadi:
Hi I would like to contribute if its okay. 

```
$this->addGetter( 
     $relation->getPropertyName(), 
     $collectionTypeHint, 
     false, 
     // add @return that advertises this as a collection of specific objects 
     [sprintf('@return Collection<%s,%s>', $collectionTypeHint, $typeHint)] 
 ); 
```
 
 the code should look like this?

PR #1040, opened 2021-12-19
[make:entity] Change getter PHPDoc return type on Collection

mehdibo:
Improves #979


mehdibo:
Oops, still need more fixing

mehdibo:
Let me know if I caused the tests to fail, looks like they are not related to my changes

allan-simon:
Is there something that can be done to help this pull request to move forward,  this will help making the autogenerated code already compliant with phpstan checks , which would be a blessing :)

weaverryan:
Awesome! Thank you Mehdi!

PR #1041, opened 2021-12-28
[tests] handle missing type config value for doctrine

jrushlow:
handles recipe change from https://github.com/symfony/recipes/pull/971 for `DoctrineBundle`


weaverryan:
Thanks Jesse!

PR #1042, opened 2021-12-28
[CSM] fix: Handle enum as values

Geekimo:
This PR aims to handle enums as value.
It's purpose is to prevent the make;entity command to fail because of manual use of Enum in entity attributes.
It may help when Doctrine will support Enums natively.

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -514,7 +514,7 @@ final class ClassSourceManipulator
 
         if (\function_exists('enum_exists')) {
             // do we have an enum ?
-            if (is_object($value) && enum_exists(get_class($value))) {
+            if (\is_object($value) && enum_exists(\get_class($value))) {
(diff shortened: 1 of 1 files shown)


jrushlow (changes requested):
Thanks @Geekimo for the work on this. Could you `rebase` this please? Also github is flagging your commits as `Unverified` due to lack of a commit signature. (Github vigilante mode)

src/Util/ClassSourceManipulator.php
```diff
@@ -493,6 +493,13 @@ private function quoteAnnotationValue($value)
             throw new \Exception('Invalid value: loop before quoting.');
         }
 
+        if (\PHP_VERSION_ID >= 80000) {
```
weaverryan, line 496:
```suggestion
        if (\function_exists('enum_exists')) {
```

jrushlow (approved):
Thank you @Geekimo!

PR #1046, opened 2022-01-08
Exclude inherited embedded class properties

### LINKED ISSUE #1045: [make:entity --regenerate] Inherited embedded classes should be ignored
Classe properties defined in parent classes are ignored by the maker:
https://github.com/symfony/maker-bundle/blob/0b0e9288a6136120d26c50904acfe235da6b9ae9/src/Doctrine/EntityRegenerator.php#L94-L109

It should be the same with embedded classes properties. 
At the moment, if an entity extends a class with embedded properties, the property, getter and setter are added to the child class.

Vincz:
Fix #1045


jrushlow (changes requested):
Awesome! Could you rebase this PR please. We've fixed a number of broken tests since you've opened the PR. Thanks in advanced.

jrushlow (approved):
Awesome! Thank you @Vincz

PR #1058, opened 2022-01-25
[POC] Refactor make:auth to also bootstrap built-in authenticators

### LINKED ISSUE #1049: [Security] `make:auth` should default to *not* implementing custom authenticators
Since Symfony 5.3, we (to be honest: I) have changed the policy from "create your own form login guard" to "use the build in form_login authenticator". This is now also the main content of the security docs: https://symfony.com/doc/current/security.html

I think we should change `make:auth` to provide configuration, controllers and template required to get the build in authenticators working. At the end of the list, we can have a "custom authenticator" option.

_This issue is mostly an RFC/FYI, I've already started working on something like this_

wouterj:
Fixes #1049

This is a very early proof of concept. Please focus on this description, rather than on the code and CI failures :)

Goal
---

1. If you look at https://symfony.com/doc/current/security/#authenticating-users , you can see that almost all built-in authenticator require bootstrapping code: controllers, templates, etc. I would like to replace that whole section with: **"Run `make:auth`, choose the authentication method you want and follow the steps."**
2. The maker bundle currently "forces" you to write a custom authenticator. For most common use-cases, this is not needed and **we should instead encourage people to use the built-in authenticators**.

How do we get there?
---

We should start by asking a few simple questions in order to determine whether a custom authenticator, or built-in authenticators should be generated (solves (2)). After those questions, we can bootstrap the code required to get all authenticators working (solves (1)).

The state of this PR at the moment is that it more or less fixed the 2nd goal, by listing all available built-in authenticators and asking an extra question for form login/custom authenticator:

```
$ symfony console make:auth

 What style of authentication do you want? [Login form]:
  [0] Login form
  [1] JSON API login
  [2] HTTP basic
  [3] Login link
  [4] Pre authenticated
  [5] Custom authenticator
 > 0

 Do you require other fields than user identifier (e.g. email) and password? (yes/no) [no]:
 > no

 # ... configure and bootstrap the built-in authenticator
 # a custom form login authenticator would be bootstrapped when the last answer was "yes"
```

Finally, I implemented the HTTP basic authenticator bootstrapping (for the simple reason that this one doesn't require anything else than a `security.yaml` update).

Please let me know what you think :)

### DIFF
--- a/src/Maker/MakeAuthenticator.php
+++ b/src/Maker/MakeAuthenticator.php
@@ -18,28 +18,16 @@ use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
-use Symfony\Bundle\MakerBundle\Security\InteractiveSecurityHelper;
+use Symfony\Bundle\MakerBundle\Security\Authenticator\AuthenticatorMakerInterface;
 use Symfony\Bundle\MakerBundle\Security\SecurityConfigUpdater;
-use Symfony\Bundle\MakerBundle\Security\SecurityControllerBuilder;
-use Symfony\Bundle\MakerBundle\Str;
-use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
-use Symfony\Bundle\MakerBundle\Util\YamlManipulationFailedException;
 use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
-use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Bundle\SecurityBundle\SecurityBundle;
 use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Question\Question;
-use Symfony\Component\Form\Form;
-use Symfony\Component\HttpKernel\Kernel;
-use Symfony\Component\Security\Guard\Authenticator\AbstractFormLoginAuthenticator;
-use Symfony\Component\Security\Http\Authenticator\AuthenticatorInterface;
-use Symfony\Component\Security\Http\Authenticator\LoginLinkAuthenticator;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
-use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
 use Symfony\Component\Yaml\Yaml;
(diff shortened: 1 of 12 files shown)


javiereguiluz:
Very nice proposal! Thanks!!

-----

I know it's difficult to make something that absolutely everybody can fully understand ... but maybe the first question is a bit intimidating for some:

```
 What style of authentication do you want? [Login form]:
  [0] Login form
  [1] JSON API login
  [2] HTTP basic
  [3] Login link
  [4] Pre authenticated
  [5] Custom authenticator
```

For example, it's easy to not know what "pre authenticated" means. Maybe we could add something like: _"Pre authenticated (e.g. X.509 certificates)"_ (I don't know if this is about certificates, but you get the idea).

Also, the "Custom authenticator" may (wrongly) seem like the best option for some developers. They might think: _"well, this is my app, so I must develop this myself ... I can't chose any of the other because they are "magic" or generate "garbage code" that I need to delete later"_. So maybe we need to be super clear about this --> "you probably don't need custom authenticators at all. It's a super advanced feature for most apps. Pick one of the other options".

wouterj:
Thank you Javier. This type of feedback from the great trainers/documentors in the community is exactly why I opened this PR so early in the process :)

> [...] but maybe the first question is a bit intimidating for some: For example, it's easy to not know what "pre authenticated" means. Maybe we could add something like: "Pre authenticated (e.g. X.509 certificates)" (I don't know if this is about certificates, but you get the idea).

Hmm, yes we can add an example. But this question got me thinking: what if we add more authenticators (which I hope we'll do over the next years). Maybe we should not list all options, or split 

> Also, the "Custom authenticator" may (wrongly) seem like the best option for some developers. They might think: "well, this is my app, so I must develop this myself ... I can't choose any of the other because they are "magic" or generate "garbage code" that I need to delete later". So maybe we need to be super clear about this --> "you probably don't need custom authenticators at all. It's a super advanced feature for most apps. Pick one of the other options".

I like the idea, but I think the text is a bit too long to include in the list. I've added an extra question when choosing the custom authenticator:

<details>
<summary>Example</summary>

```
$ symfony console make:auth

 What style of authentication do you want? [Login form]:
  [0] Login form
  [1] JSON API login
  [2] HTTP basic
  [3] Login link
  [4] Pre authenticated (e.g. Kerebos or certificates)
  [5] Custom authenticator
 > 5

 A custom authenticator is only required in advanced use-cases, are you sure
 your use-case isn't covered by one of the built-in authenticators? (yes/no) [yes]:
 > no

 What style of authentication do you want? [Login form]:
  [0] Login form
  [1] JSON API login
  [2] HTTP basic
  [3] Login link
  [4] Pre authenticated (e.g. Kerebos or certificates)
 > 
```

</details>

wouterj:
I've just pushed a major update to this PR: splitting `MakeAuthenticator` into an "authenticator maker" per authenticator step. Imho, this makes the code much easier to maintain and share between multiple authenticators. Let me know what you think :)

_quick note: I'll be afk for the next 2 weeks, so it'll take some time before I implemented other authenticator types_

PR #1059, opened 2022-01-31
[make:reset-password] Translate exception reasons provided by ResetPasswordBundle

bocharsky-bw:
If there's `Symfony\Contracts\Translation\TranslatorInterface` service available in the project - Maker will generate a slightly different controller's code to inject translator service and translate exception reasons out of the box.

Available after: https://github.com/SymfonyCasts/reset-password-bundle/pull/206

### DIFF
--- a/src/Maker/MakeResetPassword.php
+++ b/src/Maker/MakeResetPassword.php
@@ -44,8 +44,10 @@ use Symfony\Component\Mime\Address;
 use Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+use Symfony\Component\Translation\Translator;
 use Symfony\Component\Validator\Validation;
 use Symfony\Component\Yaml\Yaml;
+use Symfony\Contracts\Translation\TranslatorInterface;
 use SymfonyCasts\Bundle\ResetPassword\Controller\ResetPasswordControllerTrait;
 use SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface;
 use SymfonyCasts\Bundle\ResetPassword\Model\ResetPasswordRequestInterface;
@@ -236,6 +238,18 @@ class MakeResetPassword extends AbstractMaker
             EntityManagerInterface::class,
         ];
 
+        // Namespace for ResetPasswordExceptionInterface was imported above
+        $problemValidateMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE'
+            : "'There was a problem validating your password reset request'";
+        $problemHandleMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_HANDLE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_HANDLE'
+            : "'There was a problem handling your password reset request'";
+
+        if ($isTranslatorAvailable = class_exists(Translator::class)) {
+            $useStatements[] = TranslatorInterface::class;
+        }
+
         $generator->generateController(
             $controllerClassNameDetails->getFullName(),
             'resetPassword/ResetPasswordController.tpl.php',
@@ -253,6 +267,9 @@ class MakeResetPassword extends AbstractMaker
                 'password_hasher_class_details' => ($passwordClassDetails = $generator->createClassNameDetails($passwordHasher, '\\')),
                 'password_hasher_variable_name' => str_replace('Interface', '', sprintf('$%s', lcfirst($passwordClassDetails->getShortName()))), // @legacy see passwordHasher conditional above
                 'use_password_hasher' => UserPasswordHasherInterface::class === $passwordHasher, // @legacy see passwordHasher conditional above
+                'problem_validate_message_or_constant' => $problemValidateMessageOrConstant,
+                'problem_handle_message_or_constant' => $problemHandleMessageOrConstant,
+                'translator_available' => $isTranslatorAvailable,
             ]
         );
 
--- a/tests/Maker/MakeResetPasswordTest.php
+++ b/tests/Maker/MakeResetPasswordTest.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MakerBundle\Tests\Maker;
 
+use Doctrine\ORM\Mapping\Driver\AttributeReader;
 use Symfony\Bundle\MakerBundle\Maker\MakeResetPassword;
 use Symfony\Bundle\MakerBundle\Test\MakerTestCase;
 use Symfony\Bundle\MakerBundle\Test\MakerTestDetails;
@@ -72,6 +73,22 @@ class MakeResetPasswordTest extends MakerTestCase
                 $this->assertSame('App\Repository\ResetPasswordRequestRepository', $resetPasswordConfig['symfonycasts_reset_password']['request_password_repository']);
 
                 $this->runResetPasswordTest($runner, 'it_generates_with_normal_setup.php');
+            }),
+        ];
+
+        yield 'it_generates_with_translator_installed' => [$this->createResetPasswordTest()
+            ->addExtraDependencies('symfony/translation')
+            ->run(function (MakerTestRunner $runner) {
+                $this->makeUser($runner);
+
+                $output = $runner->runMaker([
+                    'App\Entity\User',
+                    'app_home',
+                    'victor@symfonycasts.com',
+                    'SymfonyCasts',
+                ]);
+
+                $this->assertStringContainsString('Success', $output);
             }),
         ];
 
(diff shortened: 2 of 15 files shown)


src/Maker/MakeResetPassword.php
```diff
@@ -236,10 +237,27 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+            : "'There was a problem validating your password reset request'";
+        $problemHandleMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_HANDLE'
+            : "'There was a problem handling your password reset request'";
+
+        $optionalTranslatorParams = [];
+        if ($isTranslatorAvailable = interface_exists(TranslatorInterface::class)) {
+            $optionalTranslatorParams = [
+                'translator_class_details' => $generator->createClassNameDetails(TranslatorInterface::class, '\\'),
+                'translator_variable_name' => '$translator',
```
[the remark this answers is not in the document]
bocharsky-bw: Agree, less variables - simpler template :)

bocharsky-bw:
Thank you for the review, @codedmonkey !

src/Maker/MakeResetPassword.php
```diff
@@ -236,10 +237,24 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+        $problemValidateMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE'
+            : "'There was a problem validating your password reset request'";
+        $problemHandleMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_HANDLE'
+            : "'There was a problem handling your password reset request'";
+
+        $optionalTranslatorParams = [];
+        if ($isTranslatorAvailable = interface_exists(TranslatorInterface::class)) {
+            $optionalTranslatorParams['translator_class_details'] = $generator->createClassNameDetails(TranslatorInterface::class, '\\');
```
bocharsky-bw, line 250: Fairly speaking, I'm not sure we need to create class name details for translator as well. I just did the same way `password_hasher_class_details` is implemented. Should we just hardcode translator class instead?

bocharsky-bw: Thank you for sharing some thoughts about it! I made this change, the logic would be simpler this way unless we really need to use some benefits of that `createClassNameDetails()`

jrushlow (approved):
Awesome thanks @bocharsky-bw - this will pair nicely with the work you did over in ResetPasswordBundle

src/Maker/MakeResetPassword.php
```diff
@@ -236,6 +238,18 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             EntityManagerInterface::class,
         ];
 
+        // Namespace for ResetPasswordExceptionInterface was imported above
+        $problemValidateMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE'
+            : "'There was a problem validating your password reset request'";
+        $problemHandleMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
```
weaverryan, line 245:
```suggestion
        $problemHandleMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_HANDLE')
```

jrushlow (changes requested):
@bocharsky-bw I think the `<?php PHP_EOL ?>` is causing issues on windows (appveyor tests).. try removing those EOL tags and we should be good to go

src/Maker/MakeResetPassword.php
```diff
@@ -111,6 +112,7 @@ public function configureDependencies(DependencyBuilder $dependencies): void
         $dependencies->addClassDependency(Validation::class, 'symfony/validator');
         $dependencies->addClassDependency(SecurityBundle::class, 'security-bundle');
         $dependencies->addClassDependency(AppVariable::class, 'twig');
+        $dependencies->addClassDependency(TranslatorInterface::class, 'symfony/translation', false);
```
weaverryan, line 115: Do we want to add this here? I believe the end result is if you are missing any required dependencies, then the `composer require` will now include `symfony/translation` in the error message. But, the user doesn't NEED to install this package, and I think the intention is just to detect IF they have it, so that we can know how to generate the controller.

src/Maker/MakeResetPassword.php
```diff
@@ -236,6 +238,18 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
 
+        // Namespace for ResetPasswordExceptionInterface was imported above
+        $problemValidateMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE'
+            : "'There was a problem validating your password reset request'";
+        $problemHandleMessageOrConstant = \defined('SymfonyCasts\Bundle\ResetPassword\Exception\ResetPasswordExceptionInterface::MESSAGE_PROBLEM_VALIDATE')
+            ? 'ResetPasswordExceptionInterface::MESSAGE_PROBLEM_HANDLE'
+            : "'There was a problem handling your password reset request'";
+
+        if ($isTranslatorAvailable = interface_exists(TranslatorInterface::class)) {
```
weaverryan, line 249: I think we should check for https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Translation/Translator.php instead (but then we would inject the `TranslatorInterface` for the `use` statement as you have).

The reason is that `TranslatorInterface` is from `symfony/translations-contract`. So currently, this isn't checking to see if you have `symfony/translator` installed, but instead it's checking to see if you have `symfony/translations-contract` installed, which will ALWAYS be the case because that is a dependency of `symfony/validator`.

src/Maker/MakeResetPassword.php
bocharsky-bw: Oh my, thanks!

src/Maker/MakeResetPassword.php
bocharsky-bw: Makes sense, changed

src/Maker/MakeResetPassword.php
bocharsky-bw: @weaverryan The intention is meant to be exactly like you described! If users have a translator installed - they will have translated exceptions. But without this line tests were failing somehow... and I passed `false` as the 3rd argument there that already says this dependency is "optional" and not "required". Is it enough? Or do I understand that 3rd option wrong and it's not about specifying required/optional dependency?

bocharsky-bw:
> I think the <?php PHP_EOL ?> is causing issues on windows (appveyor tests).. try removing those EOL tags and we should be good to go

@jrushlow Thanks for the tip! Hm, I tried to replace PHP_EOL with "\n" to see if tests will pass. But without that new line char at the end the line pops up to the previous in the rendered HTML, i.e. something like:
```php
        if ($form->isSubmitted() && $form->isValid()) {
            return $this->processSendingPasswordResetEmail(
                $form->get('email')->getData(),
                $mailer,                $translator            );
        }
```

Let's see if "\n" helps

bocharsky-bw:
Tests are green! Though this question is still open probably: https://github.com/symfony/maker-bundle/pull/1059#discussion_r807231653

src/Maker/MakeResetPassword.php
weaverryan: Ah, if it was to get the tests working, then try adding `->addExtraDependencies()` in the test case - e.g. https://github.com/symfony/maker-bundle/blob/112ed509fc7753a3d4f555f4e491a04213ee79ac/tests/Maker/MakeEntityTest.php#L81

Normally, when you run a test, it will find all of the dependencies configured in this method and will install those. So probably this solution worked because it caused the `symfony/translator` to be installed. BUT, we don't really want to recommend it as a package, so probably the better thing to do is to *not* include it here, but then hint to the test system that it SHOULD install it.

However, if you didn't include it here AND you didn't include it in `setExtraDependencies()`, in theory, the tests should NOT have failed. In that situation, the test would have been testing the "what happens of `symfony/translator` is not installed situation". We may need to have one test that does NOT include `symfony/translator` and another that DOES include it, and we make sure both work.

tests/Maker/MakeResetPasswordTest.php
```diff
@@ -76,6 +76,49 @@ public function getTestDetails()
...
+                $configFileContents = file_get_contents($runner->getPath('config/packages/reset_password.yaml'));
+
+                // Flex recipe adds comments in reset_password.yaml, check file was replaced by maker
+                $this->assertStringNotContainsString('#', $configFileContents);
+
+                $resetPasswordConfig = $runner->readYaml('config/packages/reset_password.yaml');
+
+                $this->assertSame('App\Repository\ResetPasswordRequestRepository', $resetPasswordConfig['symfonycasts_reset_password']['request_password_repository']);
+
+                $this->runResetPasswordTest($runner, 'it_generates_with_normal_setup.php');
```
bocharsky-bw, line 118: I hope it's OK to reuse this `it_generates_with_normal_setup.php` here because I literally just copy/pasted the `it_generates_with_normal_setup` test completely where just added `addExtraDependencies('symfony/translation')`

src/Maker/MakeResetPassword.php
bocharsky-bw: OK, failed tests were valid due to empty line, fixed in a00d568

And thanks for the tip about `addExtraDependencies()`! I added a new test to cover the case installed Translator in 6e64745

tests/Maker/MakeResetPasswordTest.php
weaverryan: I think we could remove all the assertions inside of `run()`  except for `$this->assertStringContainsString('Success', $output);` and `$this->runResetPasswordTest($runner, 'it_generates_with_normal_setup.php');`. So, just make it a simpler "smoke test" since the earlier test already covers all the details

bocharsky-bw: Makes sense 👍

weaverryan:
Thank you Victor!

PR #1062, opened 2022-02-07
[MakeRegistration] add support for verify email attributes

jrushlow:
- fixes CI related MakeRegistrationFailures
- Adds support for class level attributes, specifically `UniqueEntity` attributes on the `User` entity.
- Adds support for VerifyEmail property attributes

Known Issues:
Adding a class level attribute that utilizes an aliased use statement, e.g. `use CLASS as PREFIX` type alias is not supported in this PR.

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -13,6 +13,13 @@ namespace Symfony\Bundle\MakerBundle\Util;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Embedded;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\ManyToMany;
+use Doctrine\ORM\Mapping\ManyToOne;
+use Doctrine\ORM\Mapping\OneToMany;
+use Doctrine\ORM\Mapping\OneToOne;
 use PhpParser\Builder;
 use PhpParser\BuilderHelpers;
 use PhpParser\Comment\Doc;
@@ -93,7 +100,7 @@ final class ClassSourceManipulator
         $attributes = [];
 
         if ($this->useAttributesForDoctrineMapping) {
-            $attributes[] = $this->buildAttributeNode('ORM\Column', $columnOptions);
+            $attributes[] = $this->buildAttributeNode(Column::class, $columnOptions, 'ORM');
         } else {
             $comments[] = $this->buildAnnotationLine('@ORM\Column', $columnOptions);
         }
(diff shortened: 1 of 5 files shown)


jrushlow:
Tests are needed for CSM

src/Util/ClassSourceManipulator.php
```diff
@@ -900,6 +925,31 @@ public function addUseStatementIfNecessary(string $class): string
...
 
+    /**
+     * Builds a PHPParser attribute node.
+     *
+     * @param ClassNameDetails $attributeClass The attribute class which should be used for the attribute. If the class has a substitute
+     *                                         prefix, the node is built using the prefix. E.g. #[ORM\Column()] Otherwise, the
+     *                                         short name is used. E.g. #[Column()]
+     * @param array            $options        the named arguments for the attribute ($key = argument name, $value = argument value)
+     */
+    public function buildAttributeNode(ClassNameDetails $attributeClass, array $options): Node\Attribute
```
weaverryan, line 936: If I understand correctly, we're using `ClassNameDetails` here so that we can communicate a possible prefix - e.g. `ORM\`. If so, could we instead add that as a 3rd arg?

```suggestion
    public function buildAttributeNode(string $attributeClass, array $options, string $attributePrefix = null): Node\Attribute
```

weaverryan (commented):
This generally looks good to me 👍 

The ClassSourceManipulator is a mess when it comes to attributes vs annotations, but that's a known issue that we need to fix... at some point.

src/Util/ClassSourceManipulator.php
jrushlow: Yes, but also because we need access to `ClassNameDetails::getShortName()`. Otherwise we get a FQCN in the in the attribute.. e.g. `#[Doctrine\Some\Long\Path\MyAttribute(....)]`. 

I suppose we could do something like:
```php
public function buildAttributeNode(string $attributeClass, array $options, string $attributePrefix = null): Node\Attribute
{
    $details = new ClassNameDetails($attributeClass, '\\');
}
```

And I think I was first going with that, except that the attribute method is then making assumptions about the 2nd argument for the CSD constructor - Minor detail which may be irrelevant in the grand scheme of things..

jrushlow: Wait we also have MakerBundles `Str::getShortClassName()` - https://github.com/symfony/maker-bundle/blob/64318ec4a7870904e54f70e76e29ac8bc20e9875/src/Str.php#L123

That would work even better then init'ing a ClassNameDetails object..

src/Util/ClassSourceManipulator.php
```diff
@@ -358,6 +367,17 @@ public function addProperty(string $name, array $annotationLines = [], $defaultV
         $this->addNodeAfterProperties($newPropertyNode);
     }
 
+    public function addAttributeToClass(string $attributeClass, array $options): void
```
jrushlow, line 370: Adding a class level attribute that requires an aliased use statement, e.g. ` use Doctrine\Mapping as ORM` is not possible in this scope without refactoring the `ClassSourceManipulator::addUseStatementIfNecessary()` method. A task which is not required by our current use case and should be handled in a separate PR.

PR #1065, opened 2022-02-08
[ci] allow panther <2.0 in tests

jrushlow:
Panther `>= 2.x` only supports PHP >= `8.0`. We still need to use Panther `1.x` for PHP `7.x` tests until PHP 7 support is dropped in MakerBundle


weaverryan:
Thanks Jesse!

PR #1067, opened 2022-02-14
[ci] handle when@test for doctrine

jrushlow:
DoctrineBundle >= 2.4 Flex Recipe uses `when@test` in the `config/packages/doctrine.yaml` file instead of a separate `config/packages/test/doctrine.yaml` file. 

Prior to this PR we were removing the `test/doctrine.yaml` file because we did not want to use the `_test` database suffix in our tests. Now we remove the `dbname_suffix` property if it isset in the test environment. For BC, we still remove the `test/doctrine.yaml` file. 

We could remove then entire `when@test` property from `doctrine.yaml`, but it would probably be best to do this if something changes to the recipe in the future that warrants the removal. (Something may be added that we don't want to remove in the future as well.)

### DIFF
--- a/src/Test/MakerTestRunner.php
+++ b/src/Test/MakerTestRunner.php
@@ -178,10 +178,21 @@ class MakerTestRunner
 
         // Flex includes a recipe to suffix the dbname w/ "_test" - lets keep
         // things simple for these tests and not do that.
-        $this->removeFromFile(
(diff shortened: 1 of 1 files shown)


jrushlow:
Test failures are related to MakeRegistrationTests and a flex issue specific to Symfony 4.4.

weaverryan:
Thanks Jesse!

PR #1069, opened 2022-02-15
Ignore fake file created after running tests and reorder .gitignore

bocharsky-bw:
After running tests the repo is in a dirty state - the `/fake` file is created in the root dir.

Also, I reordered `.gitignore` file here


jrushlow (approved):
Cool beans! Thanks @bocharsky-bw

Issue #809, opened 2021-02-05
Symfony UX: make Stimulus controller


dunglas:
It would be nice for discoverability to add a command to create a Stimulus controller. I suggest adding a new maker called `make:stimulus-controller` generating a JS file and the corresponding Twig template loading the controller.

Related to https://github.com/symfony/recipes/pull/891 and https://github.com/symfony/recipes/pull/890.

WDYT @weaverryan @tgalopin?

tgalopin:
Does it make sense to create a Twig template? I feel it would make more sense to display a snippet of HTML in the console instead of creating a Twig template, as the controller will probably not be dedicated to a Twig template the same way a Symfony controller is.

dunglas:
As a Stimulus controller is tight to an HTML element, maybe it could make sense to also generate the related Twig template, that will be used as a partial (`include` etc). We could also add a `--no-template` option (the Symfony Controller maker as a similar option).

OskarStark:
Shall we name them make:ux:..... ?

jrushlow:
> We could also add a --no-template option (the Symfony Controller maker as a similar option).

Instead of a flag, it would be better to ask this as question "Do you want to generate the templates...."(or something along those lines)  that defaults to `true`.

> Shall we name them make:ux:..... ? 

I think `make:stimulus` would be a bit more intuitive as that's what we are referring to.. But depending on how many other makers are foreseeable - `make:ux:stimulus` would also be a desirable path forward.

JabriAbdelilah:
@dunglas I made a PR #1075

PR #1075, opened 2022-02-20
[make:stimulus-controller] New Stimulus controller Maker command

JabriAbdelilah:
New Command to generate Stimulus Controller with targets and values.

Stimulus controller Maker command

**test_controller.js**
```
import { Controller } from '@hotwired/stimulus';

/*
* The following line makes this controller "lazy": it won't be downloaded until needed
* See https://github.com/symfony/stimulus-bridge#lazy-controllers
*/
/* stimulusFetch: 'lazy' */
export default class extends Controller {
    static targets = ['errors']
    static values = {
        maxItems: Number,
    }
    // ...
}
```
Tickets #809

### DIFF
--- a/src/Maker/MakeStimulusController.php
+++ b/src/Maker/MakeStimulusController.php
@@ -1,0 +1,240 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Question\Question;
+use Symfony\WebpackEncoreBundle\WebpackEncoreBundle;
+
+/**
+ * @author Abdelilah Jabri <jbrabdelilah@gmail.com>
+ *
+ * @internal
+ */
+final class MakeStimulusController extends AbstractMaker
+{
+    public static function getCommandName(): string
+    {
+        return 'make:stimulus-controller';
+    }
+
+    public static function getCommandDescription(): string
+    {
+        return 'Creates a new Stimulus controller';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig)
+    {
+        $command
+            ->addArgument('name', InputArgument::REQUIRED, 'The name of the Stimulus controller (e.g. <fg=yellow>hello</>)')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeStimulusController.txt'));
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command)
+    {
+        $command->addArgument('extension', InputArgument::OPTIONAL);
+        $command->addArgument('targets', InputArgument::OPTIONAL, '', []);
+        $command->addArgument('values', InputArgument::OPTIONAL, '', []);
+
+        $chosenExtension = $io->choice(
+            'Language (<fg=yellow>JavaScript</> or <fg=yellow>TypeScript</>)',
+            [
+                'js' => 'JavaScript',
+                'ts' => 'TypeScript',
+            ]
+        );
+
+        $input->setArgument('extension', $chosenExtension);
+
+        $includeTargets = $io->confirm('Do you want to include targets?');
+        if ($includeTargets) {
+            $targets = [];
+            $isFirstTarget = true;
+            while (true) {
+                $newTarget = $this->askForNextTarget($io, $targets, $isFirstTarget);
+                $isFirstTarget = false;
+
+                if (null === $newTarget) {
+                    break;
+                }
+
+                $targets[] = $newTarget;
+            }
+
+            $input->setArgument('targets', $targets);
+        }
+
+        $includeValues = $io->confirm('Do you want to include values?');
+        if ($includeValues) {
+            $values = [];
+            $isFirstValue = true;
+            while (true) {
+                $newValue = $this->askForNextValue($io, $values, $isFirstValue);
+                $isFirstValue = false;
+
+                if (null === $newValue) {
+                    break;
+                }
+
+                $values[$newValue['name']] = $newValue;
+            }
+
+            $input->setArgument('values', $values);
+        }
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $controllerName = Str::asSnakeCase($input->getArgument('name'));
+        $chosenExtension = $input->getArgument('extension');
+        $targets = $input->getArgument('targets');
+        $values = $input->getArgument('values');
+
+        $targets = empty($targets) ? $targets : sprintf("['%s']", implode("', '", $targets));
+
+        $fileName = sprintf('%s_controller.%s', $controllerName, $chosenExtension);
+        $filePath = sprintf('assets/controllers/%s', $fileName);
+
+        $generator->generateFile(
+            $filePath,
+            'stimulus/Controller.tpl.php',
+            [
+                'targets' => $targets,
+                'values' => $values,
+            ]
+        );
+
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text([
+            'Next:',
+            sprintf('- Open <info>%s</info> and add the code you need', $filePath),
+            'Find the documentation at <fg=yellow>https://github.com/symfony/stimulus-bridge</>',
+        ]);
+    }
+
+    private function askForNextTarget(ConsoleStyle $io, array $targets, bool $isFirstTarget)
+    {
+        if ($isFirstTarget) {
+            $questionText = 'New target name (press <return> to stop adding targets)';
+        } else {
+            $questionText = 'Add another target? Enter the target name (or press <return> to stop adding targets)';
+        }
+
+        $targetName = $io->ask($questionText, null, function ($name) use ($targets) {
+            if (\in_array($name, $targets)) {
+                throw new \InvalidArgumentException(sprintf('The "%s" target already exists.', $name));
+            }
+
+            return $name;
+        });
+
+        if (!$targetName) {
+            return null;
+        }
+
+        return $targetName;
+    }
+
+    private function askForNextValue(ConsoleStyle $io, array $values, bool $isFirstValue)
+    {
+        if ($isFirstValue) {
+            $questionText = 'New value name (press <return> to stop adding values)';
+        } else {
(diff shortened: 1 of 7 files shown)


dunglas (approved):

zairigimad (approved):

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,81 @@
...
+use Symfony\WebpackEncoreBundle\WebpackEncoreBundle;
+
+/**
+ * @author Abdelilah Jabri <jbrabdelilah@gmail.com>
+ *
+ * @internal
+ */
+final class MakeStimulusController extends AbstractMaker
+{
+    public const SUFFIX = '_controller.js';
```
OskarStark, line 30: Need to be public?

Same for the other constant.

If not and the class is final, we can Remove the @internal annotation I guess

src/Resources/skeleton/stimulus/Controller.tpl.php
```diff
@@ -0,0 +1,7 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+    connect() {
```
zairigimad, line 4:
```js
    // static targets = [ 'fields', 'field', 'addButton' ]
    // static values = {
    //    prototype: String,
    //    maxItems: Number,
    //    itemsCount: Number,
    // }
```
maybe adding this in the template commented ?

JabriAbdelilah: @zairigimad I used the example given in Stimulus bridge doc, to have a minimal template

src/Maker/MakeStimulusController.php
JabriAbdelilah: @OskarStark both constants don't need to be public, I made them private

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,81 @@
...
+
+    public static function getCommandDescription(): string
+    {
+        return 'Creates a new Stimulus controller';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig)
+    {
+        $command
+            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the stimulus controller (e.g. <fg=yellow>hello</>)')
```
94noni, line 47:
```suggestion
            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the Stimulus controller (e.g. <fg=yellow>hello</>)')
```

src/Resources/skeleton/stimulus/Controller.tpl.php
```diff
@@ -0,0 +1,7 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
```
94noni, line 3: what bout adding `/* stimulusFetch: 'lazy' */` ?
Ref a PR of mine on the demo repo https://github.com/symfony/demo/pull/1284/files#r770934137

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,81 @@
...
+use Symfony\WebpackEncoreBundle\WebpackEncoreBundle;
+
+/**
+ * @author Abdelilah Jabri <jbrabdelilah@gmail.com>
+ *
+ * @internal
+ */
+final class MakeStimulusController extends AbstractMaker
+{
+    private const SUFFIX = '_controller.js';
```
94noni, line 30: can it be `.ts`?

src/Resources/skeleton/stimulus/Controller.tpl.php
94noni: what about adding an interactive mode?
then defining this targets and values
and outputting a default html template at the end of the generation ?

it would be 🥇

src/Maker/MakeStimulusController.php
JabriAbdelilah: I think we can give the choice to the user. I will add a new argument `language` or `extension`

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,81 @@
...
+        $command
+            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the stimulus controller (e.g. <fg=yellow>hello</>)')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeStimulusController.txt'));
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $controllerName = $input->getArgument('name');
+        $fileName = $controllerName.self::SUFFIX;
+        $filePath = self::PATH.$fileName;
```
jrushlow, line 55: Same for this one as well.
```suggestion
        $filePath = sprintf('assets/controllers/%s', $fileName);
```

src/Maker/MakeStimulusController.php
jrushlow, line 30:
```suggestion
```

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @author Abdelilah Jabri <jbrabdelilah@gmail.com>
+ *
+ * @internal
+ */
+final class MakeStimulusController extends AbstractMaker
+{
+    private const SUFFIX = '_controller.js';
+
+    private const PATH = 'assets/controllers/';
```
jrushlow, line 32:
```suggestion
```

src/Resources/skeleton/stimulus/Controller.tpl.php
jrushlow: Agreed, we should just include `/* stimulusFetch: 'lazy' */` by default in the template. It's easy enough for the user to remove if they do not want it.

cc/ @weaverryan

src/Resources/skeleton/stimulus/Controller.tpl.php
jrushlow: I'd say 👎 for adding commented out `targets` & `values` in the template. I'm open to having maker ask something along the lines of `Do you want to include targets?...` then adding them into the template if the user says yes.

tests/Maker/MakeStimulusControllerTest.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        yield 'it_generates_stimulus_controller' => [$this->createMakerTest()
+            ->run(function (MakerTestRunner $runner) {
+                $runner->runMaker(
+                    [
+                        // controller name
+                        'custom',
+                    ]);
+
+                $this->assertFileExists($runner->getPath('assets/controllers/custom_controller.js'));
```
jrushlow, line 35: We should also assert the contents of the file match our expectations of what the file should be. e.g. `assertSame(.....)`. We can do this by creating a complete fixture for `make:stimulus` then grab the contents of the generated file and compare.

jrushlow (changes requested):
This is looking real good @JabriAbdelilah! a couple of comments below:

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,141 @@
...
+                $targets[] = $newTarget;
+            }
+
+            $input->setArgument('targets', $targets);
+        }
+    }
+
+    public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
+    {
+        $controllerName = $input->getArgument('name');
```
94noni, line 83: $controllerName perhaps need to be rewriten according to the user input 

if the user write ProductSearch or productSearch it should be product_search at the end i think

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,141 @@
...
+        );
+
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text([
+            'Next:',
+            sprintf('- Open <info>%s</info> and add the code you need', $filePath),
+            'Find the documentation at <fg=yellow>https://github.com/symfony/stimulus-bridge</>',
```
94noni, line 107: Read instead of Find ?
Also perhaps add link to official doc website?

src/Resources/help/MakeStimulusController.txt
```diff
@@ -0,0 +1,5 @@
+The <info>%command.name%</info> command generates new Stimulus Controller.
+
+<info>php %command.full_name% hello</info>
+
+If the argument is missing, the command will ask for the controller name interactively.
```
94noni, line 5: The extension also isnt it?

src/Maker/MakeStimulusController.php
```diff
@@ -0,0 +1,141 @@
...
+            'Language (<fg=yellow>JavaScript</> or <fg=yellow>TypeScript</>)',
+            [
+                'js' => 'JavaScript',
+                'ts' => 'TypeScript',
+            ]
+        );
+
+        $input->setArgument('extension', $chosenExtension);
+
+        $includeTargets = $io->confirm('Do you want to include targets?');
```
94noni, line 62: Allowing adding targets is already very nice,
Do tou think it can be doable to mimic and add values? Name and type
Apart from that this PR is cool :)

src/Maker/MakeStimulusController.php
JabriAbdelilah: Every Maker uses Find not Read:
image


@94noni you mean the [Stimulus official doc](https://stimulus.hotwired.dev/handbook/introduction)? or [Symfony UX](https://symfony.com/doc/current/frontend/ux.html#stimulus-tools-around-the-world)?

src/Maker/MakeStimulusController.php
JabriAbdelilah: Yes it's doable.

src/Maker/MakeStimulusController.php
94noni: I was talking of the official hotwired stimulus doc

But i am unsure if it is ok and perhaps only symfony doc url is wanted (need core maintainer review :) )

And ok for find

src/Resources/help/MakeStimulusController.txt
JabriAbdelilah: @94noni in fact, the extension is interactive, no need to add it in here

src/Maker/MakeStimulusController.php
JabriAbdelilah: @94noni I added Str::asSnakeCase

src/Maker/MakeStimulusController.php
JabriAbdelilah: @94noni I finished the possibility to add values, can you re-check please?

src/Resources/skeleton/stimulus/Controller.tpl.php
```diff
@@ -0,0 +1,18 @@
...
+* The following line makes this controller "lazy": it won't be downloaded until needed
+* See https://github.com/symfony/stimulus-bridge#lazy-controllers
+*/
+/* stimulusFetch: 'lazy' */
+export default class extends Controller {
+<?= $targets ? "    static targets = $targets\n" : "" ?>
+<?php if ($values) { ?>
+    static values = {
+<?php foreach ($values as $value): ?>
+        <?= $value['name'] ?>: <?= $value['type'] ?>,
```
94noni, line 13: just I think the last `,` is to much no? I mean after the latest value type is written

zairigimad (approved):

src/Resources/skeleton/stimulus/Controller.tpl.php
AbdelilahJabri: That's a trailing comma, it's a good practice in Javascript

JabriAbdelilah:
@jrushlow this PR is ready for review

maxhelias (approved):

jrushlow (approved):
Thank you @JabriAbdelilah for the new maker!

PR #1076, opened 2022-02-22
[make:registration-form] Translate reasons for VerifyEmailBundle if translator available

bocharsky-bw:
Exception reasons were translated in https://github.com/SymfonyCasts/verify-email-bundle/pull/103 that allows us to translate them with Maker bundle out of the box if Translator is available.

This PR is similar to #1059 but for VerifyEmailBundle


jrushlow (approved):
I love it! Thanks @bocharsky-bw

PR #1077, opened 2022-02-23
[ci] fix overriding entire route config

jrushlow:
Previously we were replacing the `config/routes.yaml` file generated by recipes with our own. This broke several tests when https://github.com/symfony/recipes/issues/1030 was resolved in Symfony 6. 

To better test against a "real" environment, we use the generated `routes.yaml` file and then append our static routes to it.


weaverryan (approved):

weaverryan:
Good one - thanks Jesse!

PR #1085, opened 2022-03-29
[RFC][WIP] add scaffolding system (`make:scaffold`)

kbond:
This adds a `make:scaffold ...<scaffold>` command. This is an alternative to what I'm calling _single-use makers_ like `make:register`/`make:user`. We could potentially deprecate these other makers.

A _scaffold_ consists of:
1. files that are copied to your project to create the feature (_**including tests that are immediately runnable and pass!**_)
2. required Composer packages (_these are auto-installed_)
3. required JS packages (_these are auto-installed_)
4. dependent scaffolds (most current scaffolds depend on others, ie `register` requires `auth`)
5. A little configuration system to allow more advanced file manipulations (ie update an exiting yaml file)

Some additional notes about this feature:
- Scaffolds are grouped by Symfony major versions with 6.0 being the only supplied version currently.
- These are meant to be run on new projects. Running them on an established project is possible but since the supplied files override existing files, you'll have to be careful and use git to help with conflicts.
- Copying files from a scaffold is currently "dumb" and they just replace existing files without confirming. It could be possible to make this smarter.
- Scaffolds could be grouped into "starter-kits" like what's provided with Laravel Breeze/Jetstream.
- I don't really want a wizard for these scaffolds. I think these add a huge amount of complexity. If we want a variation on a scaffold, we'd create a new scaffold.
- A future iteration could provide styling options? tailwind/bootstrap versions of each scaffold
- The generated code (especially the tests) is opinionated and use several of my personal libraries (zenstruck/browser|foundry). This is really "How Kevin Writes Symfony Apps" [:)](https://www.slideshare.net/kriswallsmith/how-kris-writes-symfony-apps). I don't have an issue if we want to make changes to code/tests to make them "more generic". 
- `bin/console make:scaffold` (no arguments) lists the available scaffolds to choose from.
- `bin/console make:scaffold x y z` installs scaffolds `x`, `y` and `z`.

This PR provides the following scaffolds:
1. `homepage`: basic homepage controller, template, functional test for other scaffolds to use
2. `user`: user entity/repository, foundry factory and unit test
3. `auth`: form login with functional tests
6. `register`: registration form with functional tests
7. `change-password`: change password form with functional tests
8. `profile`: user profile management form with functional tests
9. `reset-password`: using symfonycasts/password-reset-bundle
10. `bootstrapcss`: installs/configures bootstrap with encore

Included is also a `starter-kit`. This is just another scaffold but it includes all the above scaffolds (styled with bootstrap) and an application shell. The shell has the following features:
- navbar on every page with:
  - login & register links if not logged in
  - manage profile, change password & logout user menu
- forgot password link on login page

Future ideas for scaffolds:
- `reset-password:stateless`: using signed urls instead of db token table
- `register:verification`: similar to register but with email verification
- `change-email`: change email system for verified emails
- `tailwindcss`: similar to the current bootstrapcss scaffold but for Tailwind CSS

Future ideas for starter-kits:
- `starter-kit:verifcation`: similar to the current starter kit but with email verification
- `starter-kit:tailwind`: similar to the current starter kit but for Tailwind CSS
- `starter-kit:verification:tailwind`: combination of the above two starter kits

Test this PR:
```
symfony new project
cd project
composer config repositories.maker vcs git@github.com:kbond/maker-bundle
composer require symfony/maker-bundle:dev-scaffolding
composer require process # required for this maker
composer require log # (optional) to surpress logs in console

bin/console make:scaffold starter-kit

# run unit tests
DATABASE_URL="sqlite:///%kernel.project_dir%/var/test.db" bin/phpunit

# test in browser!
echo DATABASE_URL="sqlite:///%kernel.project_dir%/var/data.db" > .env.local

symfony server:start -d
```

Some TODOS:
- [x] add better descriptions/help for each scaffold, post install notes ("what's next")
- [x] deprecate some existing maker's that are replaced by scaffolds
- [x] _smarter overwrites_: if contents is the same, skip, if contents are different, ask user to confirm.
- [x] ~Adjust password-reset to flash the "email sent" message instead of a separate template.~ Decided to leave the email sent template as there is more detail than we'd probably want in a flash message.
- [ ] base.html.twig patch with flashes...
- [x] cleanup configuration system, add some helper methods to `FileManager` (ie `modifyYaml`, `modifyJson`, `appendToFile`, `replaceInFile`).
- [x] suffix the scaffold files with `.tpl` to avoid your IDE from suggesting
- [x] instead of using `Filesystem::mirror()` to copy scaffold files, use `FilesystemManager` - this will allow the tests to lint the php/template files.
- [x] replace require package "aliases" with real packages to improve perf


kbond:
Some Symfony/Recipe PR's that would help simplify this:
- https://github.com/symfony/recipes/pull/1071 (makes running the out of the box test suites and testing the starter kit in your browser simpler)
- https://github.com/symfony/recipes/pull/1070 (would remove the need to list the flashes in the `homepage` scaffold)
- https://github.com/symfony/symfony/pull/41274 (would remove the need for LoginUser that's part of the `auth` scaffold.

src/Resources/scaffolds/6.0/reset-password/tests/Functional/ResetPasswordTest.php
```diff
@@ -0,0 +1,387 @@
...
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Test\ResetDatabase;
+use Zenstruck\Mailer\Test\InteractsWithMailer;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class ResetPasswordTest extends KernelTestCase
+{
+    use HasBrowser, Factories, ResetDatabase, InteractsWithMailer;
```
weaverryan, line 19: This test might be a bit TOO thorough :). Maybe we just include the first test?

src/Resources/scaffolds/6.0/reset-password/tests/Functional/ResetPasswordTest.php
```diff
@@ -0,0 +1,387 @@
...
+
+    /**
+     * @test
+     */
+    public function can_reset_password(): void
```
weaverryan, line 24: This style of naming with `@test` is not, I believe, the "standard".

weaverryan, line 24: This style of naming with `@test` is not, I believe, the "standard".

src/Resources/scaffolds/6.0/auth.php
```diff
@@ -0,0 +1,55 @@
...
+        $security = new YamlSourceManipulator($files->getFileContents('config/packages/security.yaml'));
+        $data = $security->getData();
+        $data['security']['providers'] = [
+            'app_user_provider' => [
+                'entity' => [
+                    'class' => 'App\Entity\User',
+                    'property' => 'email',
+                ],
+            ],
+        ];
```
weaverryan, line 31: It seems like this change should go into the "user" scaffold, no? We currently add this in `make:user`.

weaverryan, line 31: It seems like this change should go into the "user" scaffold, no? We currently add this in `make:user`.

src/Resources/scaffolds/6.0/user.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+return [
+    'description' => 'Create a basic user and unit test.',
+    'packages' => [
+        'orm' => 'all',
```
weaverryan, line 6: The problem with using Flex aliases here is that these are ALWAYS installed. What I mean is: the `isPackageInstalled()` method always returns false for this. So if you, for example, ran `make:scaffold user` and then `make:scaffold register`, the latter will (of course) re-run the `user` scaffold... but that `user` scaffold is slow because it tries to reinstall `orm`, `security` and `phpunit`.

weaverryan, line 6: The problem with using Flex aliases here is that these are ALWAYS installed. What I mean is: the `isPackageInstalled()` method always returns false for this. So if you, for example, ran `make:scaffold user` and then `make:scaffold register`, the latter will (of course) re-run the `user` scaffold... but that `user` scaffold is slow because it tries to reinstall `orm`, `security` and `phpunit`.

src/Resources/scaffolds/6.0/user/src/Entity/User.php
```diff
@@ -0,0 +1,111 @@
...
+    private $email;
+
+    #[ORM\Column(type: 'string', length: 255)]
+    private $name;
+
+    #[ORM\Column(type: 'json')]
+    private $roles = [];
+
+    #[ORM\Column(type: 'string')]
+    private $password;
```
weaverryan, line 28: You prefer no `plainPassword`?

weaverryan, line 28: You prefer no `plainPassword`?

src/Resources/scaffolds/6.0/change-password/src/Form/User/ChangePasswordFormType.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace App\Form\User;
```
weaverryan, line 3: That's a deep directory/namespace. `ChangePasswordFormType` and `RegistrationFormType` are both pretty obvious to me that they relate to users. So I'd prefer not to have the extra directory. Same with `Controller/User`.

weaverryan, line 3: That's a deep directory/namespace. `ChangePasswordFormType` and `RegistrationFormType` are both pretty obvious to me that they relate to users. So I'd prefer not to have the extra directory. Same with `Controller/User`.

src/Resources/scaffolds/6.0/change-password/tests/Functional/User/ChangePasswordTest.php
```diff
@@ -0,0 +1,178 @@
...
+use Zenstruck\Browser\Test\HasBrowser;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Test\ResetDatabase;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class ChangePasswordTest extends KernelTestCase
+{
+    use HasBrowser, Factories, ResetDatabase;
```
weaverryan, line 17: This is another test that's GREAT... but maybe too great (big) for giving to users :)

weaverryan, line 17: This is another test that's GREAT... but maybe too great (big) for giving to users :)

src/Resources/scaffolds/6.0/profile/tests/Functional/User/ProfileTest.php
```diff
@@ -0,0 +1,73 @@
...
+            ->visit('/user')
+            ->assertFieldEquals('Name', 'Mary Edwards')
+            ->fillField('Name', 'John Smith')
+            ->click('Save')
+            ->assertOn('/')
+            ->assertSuccessful()
+            ->assertSeeIn('.alert', 'You\'ve successfully updated your profile.')
+        ;
+
+        $this->assertSame('John Smith', $user->getName());
```
weaverryan, line 38: Maybe just include this first test

weaverryan, line 38: Maybe just include this first test

src/Resources/scaffolds/6.0/register/tests/Functional/User/RegisterTest.php
```diff
@@ -0,0 +1,159 @@
...
+            ->assertSuccessful()
+            ->fillField('Name', 'Madison')
+            ->fillField('Email', 'madison@example.com')
+            ->fillField('Password', '1234')
+            ->click('Register')
+            ->assertOn('/register')
+            ->assertSee('Your password should be at least 6 characters')
+            ->use(Authentication::assertNotAuthenticated())
+        ;
+    }
```
weaverryan, line 158: There's a lot of tests here too. What about a "success test" and a "validation error test" only? We would only be able to cover, for example, that "name" is required, "email" is an email and password requires 6 chars, but that's probably good enough.

weaverryan, line 158: There's a lot of tests here too. What about a "success test" and a "validation error test" only? We would only be able to cover, for example, that "name" is required, "email" is an email and password requires 6 chars, but that's probably good enough.

src/Resources/scaffolds/6.0/user.php
kbond: Agreed! I had this thought as well.

src/Resources/scaffolds/6.0/user/src/Entity/User.php
kbond: What do you mean? A non-persisted `$plainPassword` property that's cleared with `eraseCredentials()`. Yes, personally, as it cuts down on code but don't have a problem if you want to do that instead.

src/Resources/scaffolds/6.0/auth.php
kbond: I struggled with this a bit. Don't really need it in user because you haven't made the decision yet on how it will be authenticated...

src/Resources/scaffolds/6.0/change-password/src/Form/User/ChangePasswordFormType.php
kbond: Ok, yeah, easy enough for users to refactor themselves.

seb-jean:
This PR looks really amazing  :muscle:

src/Resources/scaffolds/6.0/change-password/tests/Functional/User/ChangePasswordTest.php
kbond: I removed one test but since this deals with passwords, I oped to leave the others. Let me know if you'd like to prune further.

src/Resources/scaffolds/6.0/profile/tests/Functional/User/ProfileTest.php
kbond: I kept all the tests but renamed the second one `testValidation` so user's can expand on this.

seb-jean:
According to Twig best practices, we should `use lower cased and underscored variable names`:
https://twig.symfony.com/doc/3.x/coding_standards.html

For instance:

```twig
{{ form_start(changePasswordForm) }}
```

becomes

```twig
{{ form_start(change_password_form) }}
```

And the same in the other templates.

kbond:
Hard to tell but my interpretation of that is for variables set _in the template_.

The templates are actually through twigcs as part of the tests and I fixed several issues yesterday.

src/Resources/scaffolds/6.0/reset-password/src/Controller/ResetPasswordController.php.tpl
```diff
@@ -0,0 +1,188 @@
...
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            return $this->processSendingPasswordResetEmail(
+                $form->get('email')->getData(),
+                $mailer
+            );
+        }
+
+        return $this->render('reset_password/request.html.twig', [
```
weaverryan, line 53:
```suggestion
        return $this->renderForm('reset_password/request.html.twig', [
```

... and then no `->createView()`

src/Resources/scaffolds/6.0/reset-password/src/Controller/ResetPasswordController.php.tpl
```diff
@@ -0,0 +1,188 @@
...
+        return $this->render('reset_password/check_email.html.twig', [
+            'resetToken' => $resetToken,
+        ]);
+    }
+
+    /**
+     * Validates and process the reset URL that the user clicked in their email.
+     */
+    #[Route('/reset/{token}', name: 'reset_password')]
+    public function reset(Request $request, LoginUser $login, UserPasswordHasherInterface $userPasswordHasher, string $token = null): Response
```
weaverryan, line 79:
```suggestion
    public function reset(Request $request, LoginUser $loginUser, UserPasswordHasherInterface $userPasswordHasher, string $token = null): Response
```

src/Resources/scaffolds/6.0/reset-password/src/Controller/ResetPasswordController.php.tpl
```diff
@@ -0,0 +1,188 @@
...
+
+            // programmatic user login
+            $login($user, $request);
+
+            $this->addFlash('success', 'Your password was successfully reset, you are now logged in.');
+
+            return $this->redirectToRoute('homepage');
+        }
+
+        return $this->render('reset_password/reset.html.twig', [
```
weaverryan, line 126:
```suggestion
        return $this->renderForm('reset_password/reset.html.twig', [
```

src/Resources/scaffolds/6.0/reset-password/src/Entity/ResetPasswordRequest.php.tpl
```diff
@@ -0,0 +1,39 @@
...
+    use ResetPasswordRequestTrait;
+
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: 'integer')]
+    private $id;
+
+    #[ORM\ManyToOne(targetEntity: User::class)]
+    #[ORM\JoinColumn(nullable: false)]
+    private $user;
```
weaverryan, line 22:
```suggestion
    private User $user;
```

Then we don't need `targetEntity` No need for "nullable" in this case, since we set it in the constructor

src/Resources/scaffolds/6.0/reset-password/src/Entity/ResetPasswordRequest.php.tpl
```diff
@@ -0,0 +1,39 @@
...
+
+    public function __construct(object $user, \DateTimeInterface $expiresAt, string $selector, string $hashedToken)
```
weaverryan, line 24:
```suggestion
    public function __construct(User $user, \DateTimeInterface $expiresAt, string $selector, string $hashedToken)
```

We know it'll be a concrete `User` instance.

src/Resources/scaffolds/6.0/reset-password/src/Form/ResetPasswordFormType.php.tpl
```diff
@@ -0,0 +1,51 @@
...
+                    'label' => 'New password',
+                ],
+                'second_options' => [
+                    'attr' => ['autocomplete' => 'new-password'],
+                    'label' => 'Repeat Password',
+                ],
+                'invalid_message' => 'The password fields must match.',
+                // Instead of being set onto the object directly,
+                // this is read and encoded in the controller
+                'mapped' => false,
```
weaverryan, line 42: I don't think this is needed

src/Resources/scaffolds/6.0/reset-password/src/Form/ResetPasswordFormType.php.tpl
```diff
@@ -0,0 +1,51 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Validator\Constraints\Length;
+use Symfony\Component\Validator\Constraints\NotBlank;
+
+class ResetPasswordFormType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('plainPassword', RepeatedType::class, [
```
weaverryan, line 18: Same comment about RepeatedType: is that still used much?

weaverryan (commented):
Also can we batch the "composer requires" together (at least within a single scaffold)? Same with adding JS packages

Thanks for working on this - it's super fun (especially when you style with the `bootstrapcss` scaffold).

src/Resources/scaffolds/6.0/auth.php
```diff
@@ -0,0 +1,66 @@
...
+                    'enable_csrf' => true,
+                ],
+                'logout' => [
+                    'path' => 'logout',
+                    'target' => 'homepage',
+                ],
+                'remember_me' => [
+                    'secret' => '%kernel.secret%',
+                    'secure' => 'auto',
+                    'samesite' => 'lax',
```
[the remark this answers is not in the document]
kbond: I copied from one of my apps. The default for `secure` is `false` so I think changing to `auto` makes sense? Not sure about `samesite`. I don't have much knowledge on the effect of these values.

src/Resources/scaffolds/6.0/auth/src/Controller/SecurityController.php.tpl
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Http\Authentication\AuthenticationUtils;
+
+class SecurityController extends AbstractController
+{
+    #[Route(path: '/login', name: 'login')]
+    public function login(AuthenticationUtils $authenticationUtils, Request $request): Response
+    {
+        if ($this->isGranted('IS_AUTHENTICATED_FULLY')) {
+            return new RedirectResponse($request->query->get('target', $this->generateUrl('homepage')));
```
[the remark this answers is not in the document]
kbond: It allows creating login links on every page (ie in the navbar) that will redirect back to the page the user was on when clicking login. Is there a better way to do this? I don't mind dropping this either.

src/Resources/scaffolds/6.0/bootstrapcss.php
```diff
@@ -0,0 +1,44 @@
...
+    'configure' => function (FileManager $files) {
+        // add bootstrap form theme
+        $files->manipulateYaml('config/packages/twig.yaml', function (array $data) {
+            $data['twig']['form_themes'] = ['bootstrap_5_layout.html.twig'];
+
+            return $data;
+        });
+
+        // add bootstrap to app.css
+        $files->dumpFile('assets/styles/app.css', "@import \"~bootstrap/dist/css/bootstrap.css\";\n");
```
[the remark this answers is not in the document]
kbond: I agree in general, the reason I chose to override in this case is the default style from the recipe adds a sample background which makes the starter kit look crappier... what do you think? Another option is to include the file in the scaffold - this would trigger the confirmation (once this feature is added).

src/Resources/scaffolds/6.0/change-password/src/Controller/ChangePasswordController.php.tpl
```diff
@@ -0,0 +1,54 @@
...
+use Symfony\Component\Security\Core\User\UserInterface;
+
+#[Route('/user/change-password', name: 'change_password')]
+class ChangePasswordController extends AbstractController
+{
+    public function __invoke(
+        Request $request,
+        UserPasswordHasherInterface $userPasswordHasher,
+        UserRepository $userRepository,
+        ?UserInterface $user = null,
```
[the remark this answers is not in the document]
kbond: Hmm not sure, the better way would be to use the `CurrentUser` parameter attribute to inject the actual entity. I believe this causes problems when using the framework-extra-bundle (it would try and load from db).

src/Resources/scaffolds/6.0/homepage/src/Controller/HomepageController.php.tpl
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+#[Route('/', name: 'homepage')]
```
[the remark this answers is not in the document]
kbond: It can be above the `__invoke()`.

src/Resources/scaffolds/6.0/homepage/templates/index.html.twig.tpl
```diff
@@ -0,0 +1,15 @@
...
+{% block body %}
+    <h1>Home</h1>
+
+    {% for type, messages in app.flashes %}
+        {% for message in messages %}
+            <div class="alert alert-{{ type|replace({error: 'danger'}) }}">
+                {{ message }}
+            </div>
+        {% endfor %}
+    {% endfor %}
```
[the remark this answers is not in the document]
kbond: Another option would be to include out own base.html.twig in the homepage scaffold. User would get the confirmation when overriding. Downside would be keeping it in-sync with the recipe...

src/Resources/scaffolds/6.0/user/src/Entity/User.php.tpl
```diff
@@ -0,0 +1,111 @@
...
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: 'integer')]
+    private $id;
+
+    #[ORM\Column(type: 'string', length: 180, unique: true)]
+    private $email;
+
+    #[ORM\Column(type: 'string', length: 255)]
+    private $name;
```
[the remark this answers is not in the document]
kbond: I'll do this for the other properties as well.

src/Resources/scaffolds/6.0/reset-password/src/Entity/ResetPasswordRequest.php.tpl
kbond: > Then we don't need targetEntity

Cool, didn't know that.

> No need for "nullable" in this case, since we set it in the constructor

I believe the column is nullable in the db by default. Or does typehinting the non-nullable User on the property set it as non-nullable?

src/Resources/scaffolds/6.0/auth.php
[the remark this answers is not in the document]
wouterj: These values default to `framework.session`. I think it's fine to not include them here (the `framework.session` is set to secure=auto and samesite=lax in the recipe)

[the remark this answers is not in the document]
kbond: Perfect, thanks @wouterj!

src/Resources/scaffolds/6.0/auth/src/Controller/SecurityController.php.tpl
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Http\Authentication\AuthenticationUtils;
+
+class SecurityController extends AbstractController
+{
+    #[Route(path: '/login', name: 'login')]
+    public function login(AuthenticationUtils $authenticationUtils, Request $request): Response
+    {
+        if ($this->isGranted('IS_AUTHENTICATED_FULLY')) {
```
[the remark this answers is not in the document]
kbond: Don't think this is right. We want a way to fully authenticate only _remembered_ users.

src/Resources/scaffolds/6.0/change-password/src/Form/ChangePasswordFormType.php.tpl
```diff
@@ -0,0 +1,57 @@
...
+                    'label' => 'New password',
+                ],
+                'second_options' => [
+                    'attr' => ['autocomplete' => 'new-password'],
+                    'label' => 'Repeat Password',
+                ],
+                'invalid_message' => 'The password fields must match.',
+                // Instead of being set onto the object directly,
+                // this is read and encoded in the controller
+                'mapped' => false,
```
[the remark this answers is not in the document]
kbond: I copied this from the password reset maker so... not entirely sure.

src/Resources/scaffolds/6.0/change-password/src/Form/ChangePasswordFormType.php.tpl
```diff
@@ -0,0 +1,57 @@
...
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('currentPassword', PasswordType::class, [
+                'constraints' => [
+                    new UserPassword(['message' => 'This is not your current password.']),
+                ],
+                'mapped' => false,
+            ])
+            ->add('plainPassword', RepeatedType::class, [
```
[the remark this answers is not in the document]
kbond: Ok, remove the repeated field for all scaffolds? `register`, `change-password` & `reset-password`?

src/Resources/scaffolds/6.0/reset-password/src/Entity/ResetPasswordRequest.php.tpl
kbond: Maybe you just meant "no need for `?User`". I also added `onDelete: 'CASCADE'` to the join column.

src/Resources/scaffolds/6.0/bootstrapcss.php
```diff
@@ -0,0 +1,41 @@
...
+        });
+
+        // add bootstrap to app.js
+        $appJs = $files->getFileContents('assets/app.js');
+
+        if (str_contains($appJs, "require('bootstrap');")) {
+            return;
+        }
+
+        $files->dumpFile('assets/app.js', $appJs."require('bootstrap');\n");
```
weaverryan, line 39: I'm thinking:

```js
// activates all of Bootstrap's JavaScrip functionality
import 'bootstrap';
```

src/Resources/scaffolds/6.0/bootstrapcss/assets/styles/app.css.tpl
```diff
@@ -0,0 +1 @@
+@import "~bootstrap/dist/css/bootstrap.css";
```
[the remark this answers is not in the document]
weaverryan: This is a spot where, like with `app.js`, I'd love to patch instead of replace. There's a practical reason. Right now, when you run the bootstrapcss scaffold, (A) the recipe adds `bootstrap.css` and then (B) the scaffold tries to replace it... so it immediately asks us if we want to overwrite `app.css`... which is weird :p.

src/Resources/scaffolds/6.0/user/src/Repository/UserRepository.php.tpl
```diff
@@ -0,0 +1,64 @@
...
+     * @throws ORMException
+     * @throws OptimisticLockException
+     */
+    public function remove(User $entity, bool $flush = true): void
+    {
+        $this->_em->remove($entity);
+        if ($flush) {
+            $this->_em->flush();
+        }
+    }
```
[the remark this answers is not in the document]
kbond: What do you suggest here? I copied these methods from the `make:entity` maker.

src/Resources/scaffolds/6.0/bootstrapcss.php
kbond: Does import here have the same effect as require?

weaverryan: It does, I tested it locally to be absolutely sure.

src/Resources/scaffolds/6.0/user/src/Repository/UserRepository.php.tpl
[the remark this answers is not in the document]
weaverryan: oooooh, I forgot we added those. Nevermind then!

src/Resources/scaffolds/6.0/bootstrapcss/assets/styles/app.css.tpl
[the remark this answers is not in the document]
weaverryan: I was wondering that too. Ignore it? As I, keep it there (and just add the import for bootstrap). It doesn't really hurt anything. I added that background just so there was SOMETHING, and then you'd think "hey, where does that come from". I think it still accomplishes that.

src/Resources/scaffolds/6.0/user/src/Repository/UserRepository.php.tpl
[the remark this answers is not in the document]
jrushlow: In #1087 those methods do not `flush()` by default anymore.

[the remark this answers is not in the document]
kbond: > In https://github.com/symfony/maker-bundle/pull/1087 those methods do not flush() by default anymore.

Ok, I'll change to match.

ChrisRiddell:
Would it be worth also using the [_Undocumented/Only documented in API_](https://symfony.com/doc/6.1/reference/configuration/security.html#csrf-parameter) logout CSRF protection?

```
logout:
        path: logout
        csrf_parameter: token
        csrf_token_generator: security.csrf.token_manager
```

```
<a href="{{ path('logout', {'token' : csrf_token('logout')}) }}">Logout</a>
```

kbond:
> Would it be worth also using the Undocumented/Only documented in API logout CSRF protection?

@wouterj?

wouterj:
>  > Would it be worth also using the Undocumented/Only documented in API logout CSRF protection?
>
> @wouterj?

Strictly speaken, I think you can get away with only having CSRF on login (or only on log-out). But it doesn't hurt that much to enable logout CSRF as well (and it keeps applications secure when they move away from the default form_login config provided by the scaffold).

I would recommend using the approach that I submitted to the demo application (using the `logout_path()` helper): https://github.com/symfony/demo/pull/1312 This makes enabling/disabling logout CSRF protection a config-only change, and it's a pretty cool helper overall :)

src/Maker/MakeScaffold.php
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        $io->text("Installing <info>{$name}</info> Scaffold...");
+
+        // install required packages
+        foreach ($scaffold['packages'] ?? [] as $package => $env) {
+            if (!$this->isPackageInstalled($package)) {
+                $io->text("Installing Composer package: <comment>{$package}</comment>...");
+
+                $command = [$this->composerBin(), 'require', '--no-scripts', 'dev' === $env ? '--dev' : null, $package];
```
wouterj, line 180: I think we should also require packages using `--no-update` and do a `composer update package1 package2 ...` after the loop. Doing a single update will speed up the process for slower internet connections.

wouterj (commented):
Hi! Thank you for all the work on this PR. It's a great DX to see how quickly you can get a working system using these scaffolds, and being static the scaffolds also appear easier to maintain in this bundle :)

I've been testing this a bit in a new project, and left some comments on the scaffold code (mostly with some optional tips and tricks that we might want to apply). Besides this, I also have two more high-level questions:

- The reset password and change password scaffolds are really similar, aren't they? Do we want to provide these almost similar scaffolds, or do we rather want to use the SymfonyCastsPasswordResetBundle on the profile edit page as well? (easier to maintain & less chance of introducing vulnerabilities)
- Should the scaffolds create & run database migrations? (with a confirmation question) I think we should either do this automatically, or add a message with "next steps" that show `make:migrate` and `doctrine:migration:migrate`.

src/Resources/scaffolds/6.0/user/src/Entity/User.php.tpl
```diff
@@ -0,0 +1,117 @@
...
+use App\Repository\UserRepository;
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
+use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ORM\Entity(repositoryClass: UserRepository::class)]
+#[UniqueEntity(fields: ['email'], message: 'There is already an account with this email')]
+class User implements UserInterface, PasswordAuthenticatedUserInterface
```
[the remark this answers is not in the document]
kbond: What about `users`? I typically pluralize my table names.

src/Resources/scaffolds/6.0/auth.php
```diff
@@ -0,0 +1,64 @@
...
+        'homepage',
+    ],
+    'packages' => [
+        'symfony/web-profiler-bundle' => 'dev',
+        'symfony/stopwatch' => 'dev',
+    ],
+    'configure' => function (FileManager $files) {
+        // add LoginUser service
+        $files->manipulateYaml('config/services.yaml', function (array $data) {
+            $data['services']['App\Security\LoginUser']['$authenticator'] = '@security.authenticator.form_login.main';
```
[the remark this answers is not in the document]
kbond: I'm sort of thinking to make the first scaffolds 6.1+ - this way we can use the `Autowire` parameter attribute and remove the need to modify this yaml file.

src/Resources/scaffolds/6.0/auth/src/Security/LoginUser.php.tpl
```diff
@@ -0,0 +1,25 @@
...
+    public function __construct(
+        private UserAuthenticatorInterface $userAuthenticator,
+        private AuthenticatorInterface $authenticator,
+    ) {
+    }
+
+    public function __invoke(User $user, Request $request): void
+    {
+        $this->userAuthenticator->authenticateUser($user, $this->authenticator, $request);
+    }
```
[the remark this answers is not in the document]
kbond: It's actually used in two scaffolds (`reset-password` and `register`). This is sort of a stopgap solution until https://github.com/symfony/symfony/pull/41274 is available (🤞🏻 6.2!)

seb-jean:
This PR should be taken into account: https://github.com/symfony/maker-bundle/pull/1114 in particular `UserRepository.php`

seb-jean:
This PR should be taken into account: https://github.com/symfony/maker-bundle/pull/1114 in particular `UserRepository.php`

src/Resources/scaffolds/6.0/change-password/src/Form/ChangePasswordFormType.php.tpl
```diff
@@ -0,0 +1,43 @@
...
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('currentPassword', PasswordType::class, [
+                'constraints' => [
+                    new UserPassword(['message' => 'This is not your current password.']),
+                ],
+            ])
+            ->add('plainPassword', PasswordType::class, [
```
seb-jean, line 23: Change PasswordType::class to RepeatedType::class ?

src/Resources/scaffolds/6.0/reset-password/src/Form/ResetPasswordFormType.php.tpl
```diff
@@ -0,0 +1,37 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Validator\Constraints\Length;
+use Symfony\Component\Validator\Constraints\NotBlank;
+
+class ResetPasswordFormType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('plainPassword', PasswordType::class, [
```
seb-jean, line 17: Change `PasswordType::class` to `RepeatedType::class` ?
And add `'type' => PasswordType::class,`

src/Resources/scaffolds/6.0/user/src/Entity/User.php.tpl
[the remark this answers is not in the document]
OskarStark: Lets please go with the pluralized version 👍

seb-jean:
Would it be interesting to add `Form/Model`?
For example, the `ChangePassword` Model:

```php
// src/Form/Model/ChangePassword.php
namespace App\Form\Model;

use Symfony\Component\Security\Core\Validator\Constraints as SecurityAssert;

class ChangePassword
{
    #[SecurityAssert\UserPassword(
        message: 'This is not your current password.',
    )]
    protected $currentPassword;


    [...]
}
```

The idea is to put the validation constraints in the Model and remove it in the Form classes. The Form class would then be easier to read. It's better to organize. What do you think ?

seb-jean:
@wouterj 

> * The reset password and change password scaffolds are really similar, aren't they? [...]

They look the same except there is an extra field to enter the current password (`currentPassword`).

ahmedyakoubi:
hey @kbond , is this PR going to be ready before the release of symfony 6.2 ?

kbond:
Hey, no, I don't think so. It's a bit stalled right now as we've been discussing the possibility of making this feature part of flex (the recipe system) itself. I'm hoping to nail down what direction we want to go at SymfonyCon (in Novemeber).

kbond:
Closing as we aren't going to go in this direction.

PR #1098, opened 2022-04-21
drop symfony 4.4 support bump minimum Symfony version to 5.4.7

jrushlow:
- MakerBundle Now supports Symfony >= `5.4.7`
- Minimum PHP Version >= `7.2.5`
- Removes legacy internal code
- Reduces some Doctrine complexity (Leaves the door open to continue this effort in future PR's)
- Fixes #1090

PR #1099, opened 2022-04-22
Hotfix #1090

### LINKED ISSUE #1090: Issue with doctrine/orm 2.12.x

5.4.7

### Description

Symfony make entity command results in Exception

`symfony console make:entity`

`[critical] Error thrown while running command "make:entity". Message: "Cannot access protected property Doctrine\ORM\Mapping\Driver\AttributeDriver::$classNames"`

### How to reproduce

1. Set up new Symfony project
2. Use doctrine/orm 2.12.x-dev package
3. run make:entity

### Possible Solution

_No response_

### Additional Context

_No response_

Tofandel:
A hotfix for #1090 so that users upgrading packages don't encounter this bug until a more appropriate fix can be worked

Fixes #1090

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,5 +51,8 @@
         "branch-alias": {
             "dev-main": "1.0-dev"
         }
+    },
+    "conflict": {
+        "doctrine/orm": "2.12.0"
     }
 }


jrushlow (changes requested):
Howdy @Tofandel - thanks for creating this PR. Unfortunately I do not think we are going to cut a hotfix for this issue as a fix is being introduced in #1098 that is nearly complete. We should have the fixed merged and released shortly. Thanks again!

PR #1102, opened 2022-04-23
Lower symfony/finder & symfony/yaml requirements

bobvandevijver:
Because 5.4.7 simply doesn't exist for those packages, as Symfony no longer tags the packages when no changes have occured.


jrushlow (approved):
Good Catch! Thank you @bobvandevijver

PR #1104, opened 2022-04-23
[make:auth] drop guard support and legacy code cleanup 

### LINKED ISSUE #1082: make:auth generates deprecated twig code
There is a line in the file `login.html.twig` : `You are logged in as {{ app.user.username }}`
that should be changed to `You are logged in as {{ app.user.userIdentifier }}`

I think this command is not maintained any more because symfony doesn't need an authenticator for form login by default as symfony 6

jrushlow:
- drops support for the Guard Security System
- deprecates `Generator::getControllerForBaseClass`
- fixes #1082


src/Generator.php
```diff
@@ -246,8 +244,14 @@ public function generateTemplate(string $targetPath, string $templateName, array
         );
     }
 
+    /**
+     * @deprecated MakerBundle only supports AbstractController::class. This method will be removed in the future.
+     * @legacy Remove method in the next major MakerBundle release.
```
weaverryan, line 249: I think we can just use the first line. I know WE use `@lgacy` to track which things can be removed, but they kind of serve 2 different purposes: (A) `@deprecated` means "Hey user stop using this" while (B) `@legacy` means "Hey future maintainers, don't forget to delete this code someday.

When something is `@deprecated`, let's only put that. This means that, when pruning old features, we just need to look for both tags.

src/Maker/MakeAuthenticator.php
```diff
@@ -123,15 +125,9 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
         if (self::AUTH_TYPE_FORM_LOGIN === $input->getArgument('authenticator-type')) {
-            $neededDependencies = [TwigBundle::class => 'twig'];
-            $missingPackagesMessage = 'Twig must be installed to display the login form.';
-
-            if (Kernel::VERSION_ID < 40100) {
-                $neededDependencies[Form::class] = 'symfony/form';
-                $missingPackagesMessage = 'Twig and symfony/form must be installed to display the login form';
-            }
+            $neededDependencies = [TwigBundle::class => 'twig', Form::class => 'symfony/form'];
+            $missingPackagesMessage = $this->addDependencies($neededDependencies, 'Twig and symfony/form must be installed to display the login form.');
```
weaverryan, line 129: I don't remember (and it's a bit old and hard to track down) the reason for the `if (Kernel::VERSION_ID < 40100) {` statement above, but I think you have it backwards. Previously, if the user's version was LESS than 4.1, `symfony/form` was installed. So, if 4.1 or greater, it is NOT required.

src/Maker/MakeAuthenticator.php
```diff
@@ -161,13 +157,6 @@ function ($answer) {
 
         $command->addOption('entry-point', null, InputOption::VALUE_OPTIONAL);
 
-        if (!$this->useSecurity52) {
-            $input->setOption(
-                'entry-point',
-                $interactiveSecurityHelper->guessEntryPoint($io, $securityData, $input->getArgument('authenticator-class'), $firewallName)
```
weaverryan, line 167: Is `guessEntryPoint()` used at all anymore?

weaverryan, line 167: Is `guessEntryPoint()` used at all anymore?

src/Resources/skeleton/authenticator/login_form.tpl.php
```diff
@@ -11,7 +11,7 @@
 <?php if ($logout_setup): ?>
     {% if app.user %}
         <div class="mb-3">
-            You are logged in as {{ app.user.username }}, <a href="{{ path('app_logout') }}">Logout</a>
+            You are logged in as {{ app.user.getUserIdentifier }}, <a href="{{ path('app_logout') }}">Logout</a>
```
weaverryan, line 14:
```suggestion
            You are logged in as {{ app.user.userIdentifier }}, <a href="{{ path('app_logout') }}">Logout</a>
```

weaverryan (commented):
Minor comments - it'll be great to remove all this old complexity!

src/Maker/MakeAuthenticator.php
jrushlow: good catch!

jrushlow: nope

src/Generator.php
jrushlow: good point - done

src/Maker/MakeAuthenticator.php
jrushlow: i cant remember either and grepping for `symfony/form` was - well.. ya.. haha removed, we'll see what the tests say..

src/Generator.php
```diff
@@ -246,8 +244,13 @@ public function generateTemplate(string $targetPath, string $templateName, array
         );
     }
 
+    /**
+     * @deprecated MakerBundle only supports AbstractController::class. This method will be removed in the future.
+     */
     public static function getControllerBaseClass(): ClassNameDetails
     {
+        trigger_deprecation('symfony/maker-bundle', 'v1.41.0', 'MakerBundle only supports AbstractController::class. This method will be removed in the future.');
```
weaverryan, line 252:
```suggestion
        trigger_deprecation('symfony/maker-bundle', 'v1.41.0', 'MakerBundle only supports AbstractController. This method will be removed in the future.');
```

weaverryan:
Thanks Jesse!

PR #1106, opened 2022-04-25
[make:*] use sorted use statements

jrushlow:
- removes manual use statements from templates, use statements are generated by `UseStatementGenerator`
- remove legacy code and logic from `make:command`
- adds the ability to use statements with aliases needed in 1107. e.g. `use \Symfony\Something as Else;` 
- removes some legacy code from `make:registration`. Remainder to be removed in #1107


src/Maker/MakeCommand.php
```diff
@@ -16,18 +16,29 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  * @author Ryan Weaver <weaverryan@gmail.com>
  */
 final class MakeCommand extends AbstractMaker
 {
+    private $phpCompatUtil;
```
OskarStark, line 35: Typehint?

src/Maker/AbstractMaker.php
```diff
@@ -22,6 +23,9 @@
  */
 abstract class AbstractMaker implements MakerInterface
 {
+    /** @var array Fully Qualified Classnames for Use Statements */
+    protected $useStatements = [];
```
OskarStark, line 27: Typhint?

src/Maker/MakeCommand.php
jrushlow: It's inferred from the constructor

src/Maker/AbstractMaker.php
jrushlow: we still have support for PHP 7.2 until Symfony 5.4 is EOL

OskarStark: Check, ok

src/Maker/AbstractMaker.php
```diff
@@ -22,6 +23,9 @@
  */
 abstract class AbstractMaker implements MakerInterface
 {
+    /** @var array Fully Qualified Classnames for Use Statements */
```
weaverryan, line 26:
```suggestion
    /** @var string[] fully qualified class names for use statements */
```

src/Maker/MakeCommand.php
```diff
@@ -58,13 +69,26 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
         );
 
+        $this->useStatements = [
+            Command::class,
+            InputArgument::class,
+            InputInterface::class,
+            InputOption::class,
+            OutputInterface::class,
+            SymfonyStyle::class,
+        ];
```
weaverryan, line 79: Hmm. I don't love this whole "setting on a property" idea. It's very common for us to generate multiple files from a single maker. Would we re-use this property for all the different files we generate, being careful to "clear" it between generating? It's messy :/. 

If we want some nice helper for this, let's create a model class - e.g .`UseStatements`:

```
$useStatements = new UseStatements([...]);
$useStatements->add(...);

echo $useStatements; // returns the final string with line breaks and in the right order.
```

src/Maker/MakeSerializerNormalizer.php
```diff
@@ -35,25 +39,38 @@ public static function getCommandDescription(): string
...
             'Serializer\\Normalizer\\',
             'Normalizer'
         );
 
+        $useStatements = new UseStatementGenerator([
+            NormalizerInterface::class,
+            ObjectNormalizer::class,
+        ]);
+
+        if ($cacheable = interface_exists(CacheableSupportsMethodInterface::class)) {
```
weaverryan, line 63: This will always exist, right? This interface exists at least as far back as 4.4.

tests/Maker/MakeCommandTest.php
```diff
@@ -26,8 +26,12 @@ protected function getMakerClass(): string
     public function getTestDetails(): \Generator
     {
         yield 'it_makes_a_command_no_attributes' => [$this->createMakerTest()
-            ->addRequiredPackageVersion('symfony/console', '<5.3')
             ->run(function (MakerTestRunner $runner) {
+                if (\PHP_MAJOR_VERSION < 8) {
+                    // @legacy Remove when PHP 7 is now longer supported.
+                    $this->markTestSkipped('Test only runs on PHP < 8');
+                }
```
weaverryan, line 33: There is a `setRequiredPhpVersion()` - would it work here?

src/Maker/MakeSerializerNormalizer.php
jrushlow: good catch

tests/Maker/MakeCommandTest.php
jrushlow: good catch again

PR #1107, opened 2022-04-25
[make:user] Legacy <= 5.3 & Doctrine Cleanup

### LINKED ISSUE #990: make:user & password_hashers: skip password_hashers update if possible
Hi!

The latest `security-bundle` recipe comes with a security.yaml file containing this out of the box:

```yaml
security:
    # ...

    # https://symfony.com/doc/current/security.html#c-hashing-passwords
    password_hashers:
        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
```

And so, when `make:user` is executed (with "yes" for password), we do not need to update the password hasher. In other words, what we do now is this:

```diff
security:
    password_hashers:
        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
+        App\Entity\User:
+            algorithm: auto
```

Those 2 new lines should not be added if `PasswordAuthenticatedUserInterface` is already present, as they are just not needed.

jrushlow:
- `TemplateComponentGenerator::generateUseStatements()` can now handle use statements with aliases.
- Add `doctrine/orm` conflict w/ `<2.10`
- drop legacy Symfony < `5.4` Doctrine && User conditionals
- remove legacy `UserBuilder` tests
- fixes #990 

- [x] Depends on #1104


composer.json
```diff
@@ -28,7 +28,7 @@
     "require-dev": {
         "composer/semver": "^3.0",
         "doctrine/doctrine-bundle": "^2.4",
-        "doctrine/orm": "^2.3",
+        "doctrine/orm": "^2.10.0",
```
jrushlow, line 31: potentially controversial.. need to double check dependency dependencies to ensure this is possible... Reasoning being the bump to 10 / conflict w/ <2.9 is primarily related to the new exception classes introduced in 2.10 that will become interfaces in 3.0..

src/Security/SecurityConfigUpdater.php
```diff
@@ -185,34 +183,33 @@ private function updateProviders(UserClassConfiguration $userConfig, string $use
         $this->manipulator->setData($newData);
     }
 
-    private function updatePasswordHashers(UserClassConfiguration $userConfig, string $userClass, string $keyName = 'password_hashers'): void
+    private function updatePasswordHashers(UserClassConfiguration $userConfig, string $userClass): void
     {
         $newData = $this->manipulator->getData();
-        if ('password_hashers' === $keyName && isset($newData['security']['encoders'])) {
-            // fallback to "encoders" if the user already defined encoder config
-            $this->updatePasswordHashers($userConfig, $userClass, 'encoders');
```
jrushlow, line 189: Re for situations where a user still has `security.encoders` set - we could do any of:
A) Throw an exception  e.g. "MakerBundle does not support password encoders.."
B) Remove the existing `security.encoders` and continue with setting `security.password_hashers`

The problem with the latter is we dont have a way to tell the user that we're removing the existing encoders.. I suppose we _could_ start linking `make:user`'s IO to `SecurityConfigUpdater` or something to make this "async" but that just feels like a headache w/ not alot of gain. 

If we just throw an exception, what action do we tell the user to take to correct the problem?

composer.json
weaverryan: My initial thought is: it's ok. ORM 2.10 was released Oct 3rd, 2021. Maker now requires Symfony 5.4, which was released at the end of November 2021. So, if you have 5.4, you've been upgrading already. And bumping to 2.10 is a minor release bump.

src/DependencyInjection/CompilerPass/SetDoctrineManagerRegistryClassPass.php
```diff
@@ -22,9 +22,9 @@ class SetDoctrineManagerRegistryClassPass implements CompilerPassInterface
...
     public function process(ContainerBuilder $container): void
     {
-        if ($container->hasAlias(ManagerRegistry::class)) {
-            $definition = $container->getDefinition('maker.entity_class_generator');
-            $definition->addMethodCall('setMangerRegistryClassName', [ManagerRegistry::class]);
-        }
+//        if ($container->hasAlias(ManagerRegistry::class)) {
+//            $definition = $container->getDefinition('maker.entity_class_generator');
+//            $definition->addMethodCall('setMangerRegistryClassName', [ManagerRegistry::class]);
+//        }
```
weaverryan, line 28: This isn't needed anymore? Can we delete?

src/Maker/MakeUser.php
```diff
@@ -169,13 +169,22 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
         // C) Generate a custom user provider, if necessary
         if (!$userClassConfiguration->isEntity()) {
             $userClassConfiguration->setUserProviderClass($generator->getRootNamespace().'\\Security\\UserProvider');
+
+            $useStatements = [
+                UnsupportedUserException::class,
+                UsernameNotFoundException::class,
```
weaverryan, line 175: I think this is backwards: `UsernameNotFoundException` is deprecated and `UserNotFoundException` is the new one.

src/Security/SecurityConfigUpdater.php
weaverryan: If `encoders` is set, let's throw a clear exception that explains to change to `password_hashers`.

weaverryan (commented):
Very complex, nice work! Minor comments

src/Security/SecurityConfigUpdater.php
```diff
@@ -165,34 +163,33 @@ private function updateProviders(UserClassConfiguration $userConfig, string $use
...
 
+        // The security-bundle recipe sets the password hasher via Flex. If it exists, move on...
+        if (isset($newData['security']['password_hashers'][PasswordAuthenticatedUserInterface::class])) {
```
jrushlow, line 171: merged in via #1110

src/DependencyInjection/CompilerPass/SetDoctrineManagerRegistryClassPass.php
jrushlow: yup yup

src/Maker/MakeUser.php
jrushlow: good catch

src/Security/SecurityConfigUpdater.php
jrushlow: sounds like a plan

weaverryan (approved):

weaverryan:
HUGE work - thanks Jesse!

PR #1108, opened 2022-04-30
[make:test] Removal of the condition requiring a php version < 8.1

### LINKED ISSUE #1092: ApiTestCase generator uses wrong path

6.0.x

### Description

Generating a new `ApiTestCase` via cli results in failing tests. It seems as the generator uses the wrong class path.

### How to reproduce

run

```shell
symfony new --dir .
composer require symfony/maker-bundle symfony/test-pack symfony/http-client --dev
php bin/console make:test
# > ApiTestCase
# > Example
composer require api
php bin/phpunit
```

results in

> PHP Fatal error:  Uncaught Error: Class "ApiPlatform\Symfony\Bundle\Test\ApiTestCase" not found in [...]/tests/ExampleTest.php:7

### Possible Solution

In the resulting `tests/ExampleTest.php`, cange the import:

```diff
<?php

namespace App\Tests;

-use ApiPlatform\Symfony\Bundle\Test\ApiTestCase;
+use ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase;

class ExampleTest extends ApiTestCase
{
    public function testSomething(): void
    {
        $response = static::createClient()->request('GET', '/');

        $this->assertResponseIsSuccessful();
        $this->assertJsonContains(['@id' => '/']);
    }
}
```

### Additional Context

PHP 8.1.3
Symfony CLI version 5.4.7

mdoutreluingne:
Fixes #1092 

Here I do not see why we must have a version of php < 8.1 . This is what was causing the problem it seems to me, the condition passed in the else, so users with a version of php > 8.1 and API platform 2.6 had the wrong class path (`ApiPlatform\Symfony\Bundle\Test\ApiTestCase`).

I think here we just need to know whether the class exists or not. Future versions of API platform will warn about deprecation of this wrong class path `ApiPlatform\Core\Bridge\Symfony\Bundle\Test\ApiTestCase` from version 2.7 and become invalid in 3.0, as shown [here ](https://github.com/symfony/maker-bundle/pull/1015).

### DIFF
--- a/src/Maker/MakeTest.php
+++ b/src/Maker/MakeTest.php
@@ -149,7 +149,7 @@ final class MakeTest extends AbstractMaker implements InputAwareMakerInterface
             [
                 'web_assertions_are_available' => trait_exists(WebTestAssertionsTrait::class),
                 'use_legacy_container_property' => $this->useLegacyContainerProperty(),
-                'api_test_case_fqcn' => \PHP_VERSION_ID < 80100 && !class_exists(ApiTestCase::class) ? LegacyApiTestCase::class : ApiTestCase::class,
+                'api_test_case_fqcn' => !class_exists(ApiTestCase::class) ? LegacyApiTestCase::class : ApiTestCase::class,
             ]
         );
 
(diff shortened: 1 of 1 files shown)


jrushlow (approved):
Thank you @mdoutreluingne - I too cannot see why we have PHP 8.1 apart of this conditional..

PR #1110, opened 2022-05-02
[make:user] Don't add to passwork_hashers if default recipe is in use

nicolas-grekas:
Not needed since https://github.com/symfony/recipes/pull/981


jrushlow (approved):
Thanks @nicolas-grekas!

PR #1114, opened 2022-05-09
[make:entity] _em will be private in ORM 3.0

jrushlow:


src/Resources/skeleton/doctrine/Repository.tpl.php
```diff
@@ -19,27 +19,21 @@ public function __construct(ManagerRegistry $registry)
...
     }
 
-    /**
-     * @throws ORMException
-     * @throws OptimisticLockException
-     */
     public function add(<?= $entity_class_name ?> $entity, bool $flush = false): void
     {
-        $this->_em->persist($entity);
+        ($em = $this->getEntityManager())->persist($entity);
```
weaverryan, line 24: That seems more fancy than it needs to be. I'd prefer `$this->getEntityManager()` in both places.

weaverryan (commented):
Just one small comment

weaverryan (approved):

PR #1115, opened 2022-05-09
[make:crud] fix broken controller with custom repository

### LINKED ISSUE #1113: [make:crud] Missing use statement for EntityManagerInterface
Hi!

After #1106, I think we lost the `EntityManagerInterface` use statement needed for `make:crud`. Well, actually, I think the problem might be that it's generating `EntityManagerInterface` arguments instead of my proper custom repo - e.g. `ProductRepository`: https://github.com/symfony/maker-bundle/blob/main/src/Resources/skeleton/crud/controller/Controller.tpl.php#L38-L42

The problem is that we removed the `repository_full_class_name` variable in #1106, but it's still being referenced in the template.

Cheers!

jrushlow:
fixes #1113


weaverryan:
Thanks Jesse Rushlow!

Issue #921, opened 2021-07-04
[Console] Bug with multiple entity_managers


inzanty:
**Symfony version(s) affected**: 5.3.3

**Description**
I need to define two entity managers and my config looks similar this:
```yaml orm:
        default_entity_manager: default
        entity_managers:
            default:
                connection: default
                mappings:
                    Web:
                        is_bundle: false
                        type: annotation
                        dir: '%kernel.project_dir%/src/Entity/Web'
                        prefix: 'App\Entity\Web'
                        alias: Web
            stats:
                connection: stats
                mappings:
                    Stats:
                        is_bundle: false
                        type: annotation
                        dir: '%kernel.project_dir%/src/Entity/Stats'
                        prefix: 'App\Entity\Stats'
                        alias: Stats
```

**How to reproduce**
If execute `php bin/console make:entity` command, i will get this error:
```[ERROR] Only annotation mapping is supported by make:entity, but the
         <info>App\Entity\User</info> class uses a different format. If you would
         like this command to generate the properties & getter/setter methods, add
         your mapping configuration, and then re-run this command with the
         <info>--regenerate</info> flag.
```

**Possible Solution**
Exactly don't know how to fix this problem.

**Additional context**
![Screenshot_11](https://user-images.githubusercontent.com/53940590/124377605-606ed680-dcb5-11eb-95be-e3757fc20dac.png)

SimonMarx:
To create entities the maker bundle needs the mapping information which are provided by doctrine.
When you are using 2 Entity Managers and have two mappings leading to `App\Entity\Web` and `App\Entity\Stats`, doctrine do not have any mapping for `App\Entity` (could be that some default mapping exists but without `type: annotation`, idk).

In case you are not using the FQCN, the maker-bundle should not use automatically which of your both mapping should be used to create the `User` entity.

I think a better error message would be something like `Creation of entity for the namespace "App\Entity\" failed. No doctrine mapping found for that namespace. Configured namespaces are "App\Entity\Web", "App\Entity\Stats", "...". Create a new mapping for namespace "App\Entity\" or enter the FQCN of the entity you want to create/update`.

Or we could provide a choice input in case we found multiple entity managers, where the user can select a namespace to create the entity for, in case no FQCN is provided in his input :)?

Issue #783, opened 2021-01-11
make:entity with multiple mappings fails 


pkly:
Using bundle v. 1.26.1 with symfony 4.4
```
doctrine:
    dbal:
        url: '%env(resolve:DATABASE_URL)%'

        # IMPORTANT: You MUST configure your server version,
        # either here or in the DATABASE_URL env var (see .env file)
        #server_version: '13'
    orm:
        auto_generate_proxy_classes: true
        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
        auto_mapping: true
        mappings:
            AppCommon:
                is_bundle: false
                type: annotation
                dir: '%kernel.project_dir%/src/Common/Entity'
                prefix: 'App\Common\Entity'
                alias: AppCommon
            AppApi:
                is_bundle: false
                type: annotation
                dir: '%kernel.project_dir%/src/Api/Entity'
                prefix: 'App\Api\Entity'
                alias: AppApi
            AppDb:
                is_bundle: false
                type: annotation
                dir: '%kernel.project_dir%/src/Db/Entity'
                prefix: 'App\Db\Entity'
                alias: AppDb
```
The following configuration is a valid one, but make:entity does not allow for files to be generated in those mappings, falling back to an unconfigured one which exists by default.
It creates new folders under src/ (entity and repository) for any class I type in
Something like bin/console make:entity AppDb\Something also fails.

It should generate the entity in src/Db/Entity and it just makes it in /src

David-Baron:
The src / Entity is the default path. I think your config file is misconfigured. 
The doc is here: https://symfony.com/doc/current/doctrine/multiple_entity_managers.html
- make your config like example:
orm:
        default_entity_manager: default
        entity_managers:
            default:
                connection: default
                mappings:
                    Main:
                        is_bundle: false
                        type: annotation
                        dir: '%kernel.project_dir%/src/Entity/Main'
                        prefix: 'App\Entity\Main'
                        alias: Main
            customer:
                connection: customer
                mappings:
                    Customer:
                        is_bundle: false
                        type: annotation
                        dir: '%kernel.project_dir%/src/Entity/Customer'
                        prefix: 'App\Entity\Customer'
                        alias: Customer

- Call maker:
 maker:entity Customer\YourEntity

pkly:
> The src / Entity is the default path. I think your config file is misconfigured.

Yes, you're correct. It's **the default.**
That doesn't mean it's the only possible configuration.

[...]

Issue #636, opened 2020-06-23
make:entity with multiple entity managers


jacraade:
having multiple connections / entity managers like this:

```yaml
    orm:
        default_entity_manager: default
        entity_managers:
            default:
                connection: default
                mappings:
                    Default:
                        is_bundle: false
                        type: annotation
                        dir: '%kernel.project_dir%/src/Entity/Default'
                        prefix: 'App\Entity\Default'
                        alias: Default
            sync:
                connection: sync
                mappings:
                    Sync:
                        is_bundle: false
                        type: annotation
                        dir: '%kernel.project_dir%/src/Entity/Sync'
                        prefix: 'App\Entity\Sync'
                        alias: Sync
```

gets ignored by make:entity and for that reason it cant find existing entities and new entites are created in the root folder and need manual adjustments (move entity into folder, adjust namespace, and the same for the repository)

would be great if it detects multiple connections / entity managers to ask which to use first

version 1.19.0 with symfony 4.4.10

velreine:
php bin/console make:entity Default\MyNewEntity
php bin/console make:entity Sync\MyOtherNewEntity

jacraade:
> php bin/console make:entity Default\MyNewEntity
> php bin/console make:entity Sync\MyOtherNewEntity

i guess that kinda works. yet it still generates some strange things like:

`@ORM\Entity(repositoryClass="App\Repository\Repository")(repositoryClass=TestRepository::class)`

when i create a Sync\Test entity. where it should be:

`@ORM\Entity(repositoryClass="App\Repository\Sync\TestRepository")`

jenyak:
@velreine is there a way to generate in custom directory like `/src/MyNewEntity` ?

velreine:
I'm not associated with the bundle, the commands i supplied are just the one i use which works for me. I use Symfony 5.1 -- when i execute these commands Entities are automatically put into src/Entity/<dir>/<entityName>

when i execute php bin/console make:entity <dir>\<entityname>

I don't know why you owuld want to put your Entities into the root of src though that seems like a bad practice as your project grows.

jenyak:
@velreine thanks for advice. I did not want to move my other entities to from `src/Entity/` to `src/Entity/Default`.
Solve it by rearranging managers in config
```
doctrine:
    orm:
        entity_managers:
            my_new_entities:
            default:
```

[...]

PR #1119, opened 2022-05-15
[WIP][POC]: make:entity - adding support for multiple entity managers and mappings

SimonMarx:
### WIP

Currently not all cases are covered and tests are missing. May someone else find some issues or improvements

#### Missing stuff
1) FQCN are not resolved when passing them directly as argument. We should resolve them like the Drivers. Find the closest match to given namespace by mapping prefixes of doctrine config.
2) Autocompletion of entity names should be based on selected manager and mapping (removing the prefix path) - Still missing, need to check side effects with other stuff using the class strings. Currently the autocomplete already returns all Entities with full namespaces, expect the entities located in `src/Entity`
3) Manage relation autocomplete. Doctrine does not support relations between entity managers (as far as i know), but relations between mappings within the same entity managers should be possible. Prevent autocomplete for other entitymanagers while field generation

Feel free to participate in this PR ❤️ 

### Description

Working with multiple entity manager does not allow customizable namespaces and will always generate entities and repositories in `src/Entity` and `src/Repository`, also when the Prefix configured in the doctrine configure is different.

This PR will add support for the `make:entity` command and allows developers to choose a configured entity manager and mapping when there are multiple configured.

When only one Entity Manager and / or Mapping is defined, this will be taken by default instead of force the user to select this values.

### Related Issues:
https://github.com/symfony/maker-bundle/issues/636
https://github.com/symfony/maker-bundle/issues/783
https://github.com/symfony/maker-bundle/issues/921

#### Doctrine config
image

#### Creating entity for default manager

image

#### Creating entity for another entity manager with multiple mappings

image

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -306,17 +306,108 @@ final class DoctrineHelper
         $lowestCharacterDiff = null;
         $foundDriver = null;
 
+        $mappedDrivers = [];
+
         foreach ($this->mappingDriversByPrefix ?? [] as $mappings) {
             foreach ($mappings as [$prefix, $driver]) {
-                $diff = substr_compare($namespace, $prefix, 0);
-
-                if ($diff >= 0 && (null === $lowestCharacterDiff || $diff < $lowestCharacterDiff)) {
-                    $lowestCharacterDiff = $diff;
-                    $foundDriver = $driver;
-                }
+                $mappedDrivers[$prefix] = $driver;
             }
         }
 
-        return $foundDriver;
+        return $this->findByNamespace($namespace, $mappedDrivers);
+    }
+
+    public function hasMultipleManagers(): bool
+    {
+        return count($this->registry->getManagers()) > 0;
+    }
+
+    public function getEntityNamespaces(?string $managerName = null): array
+    {
+        $namespaces = [];
+
+        /** @var EntityManagerInterface $em */
+        foreach ($this->getRegistry()->getManagers() as $name => $em) {
+            $namespaces[$name] = $em->getConfiguration()->getEntityNamespaces();
+        }
+
+        if ($managerName) {
+            if (!\array_key_exists($managerName, $namespaces)) {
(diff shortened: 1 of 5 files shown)


SimonMarx:
@weaverryan before I continue here I would like to know what you think about it? Do you see any problems that could arise from this right away? Maybe there are / were reasons why the Maker Bundle does not include the doctrine config (managers and mappings) when creating entities?

src/Doctrine/DoctrineHelper.php
```diff
@@ -306,17 +306,108 @@ private function getMappingDriverForNamespace(string $namespace): ?MappingDriver
...
-                if ($diff >= 0 && (null === $lowestCharacterDiff || $diff < $lowestCharacterDiff)) {
-                    $lowestCharacterDiff = $diff;
-                    $foundDriver = $driver;
-                }
+        return $this->findByNamespace($namespace, $mappedDrivers);
+    }
+
+    public function hasMultipleManagers(): bool
+    {
+        return count($this->registry->getManagers()) > 0;
```
OskarStark, line 322:
```suggestion
        return count($this->registry->getManagers()) > 1;
```
?

gmorisseau-sigma:
Hello, this would be a nice feature, but maybe not yet enough popular. Nice job @SimonMarx and @OskarStark 

❓For now, should we use the workaround :
```
php bin/console make:entity Main\\Grumpy
```
which generates entity / repo in the right folder
```
src/Entity/Main/Grumpy.php
src/Entity/Main/GrumpyRepository.php
```
Also, it is OK for a ```make:migration```.

```php
public function up(Schema $schema): void
    {
        // this up() migration is auto-generated, please modify it to your needs
        $this->addSql('CREATE SEQUENCE grumpy_id_seq INCREMENT BY 1 MINVALUE 1 START 1');
        $this->addSql('CREATE TABLE grumpy (id INT NOT NULL, name VARCHAR(255) NOT NULL, angry VARCHAR(255) NOT NULL, PRIMARY KEY(id))');
    }
```

PR #1118, opened 2022-05-13
Fix errors when enable_authenticator_manager is not set

l-vo:
Fix errors when enable_authenticator_manager is not set and true is the default value (from Symfony 6)

Related to symfony/recipes#1085

PR #1120, opened 2022-05-16
[make:controller] Return a JsonResponse instead of a Response with --no-template

l-vo:
Return a JsonResponse instead of a Response when json method is used in controller.

Failing tests not related, see #1118 for a fix


jrushlow (approved):
Thank you @l-vo

PR #1122, opened 2022-05-17
drop PHP 7.x support

jrushlow:
- removes legacy typed property conditionals
- removes legacy attribute conditions
- reduces test complexity
- implements constructor property promotion where possible in generated templates and internal classes
- always use typed properties internally
- most of the doctrine related PHP 7.x code will be removed in a separate PR.
- does not fully remove `PhpCompatUtil::class` - that will be handled under a separate PR.


src/Doctrine/DoctrineHelper.php
```diff
@@ -145,7 +145,7 @@ public function doesClassUsesAttributes(string $className): bool
 
     public function isDoctrineSupportingAttributes(): bool
     {
-        return $this->isDoctrineInstalled() && $this->attributeMappingSupport && $this->phpCompatUtil->canUseAttributes();
+        return $this->isDoctrineInstalled() && $this->attributeMappingSupport;
     }
```
jrushlow, line 150: We will handle bringing doctrine related makers up to PHP8 standards in a separate PR to avoid too much bloat in this one.

src/Maker/MakeEntity.php
```diff
@@ -877,10 +872,6 @@ private function doesEntityUseAnnotationMapping(string $className): bool
     /** @legacy Drop when Annotations are no longer supported */
     private function doesEntityUseAttributeMapping(string $className): bool
     {
-        if (!$this->phpCompatUtil->canUseAttributes()) {
-            return false;
-        }
-
```
jrushlow, line 883: Handled under a separate PR dedicated to doctrine.

appveyor.yml
```diff
@@ -27,7 +27,7 @@ environment:
     TEST_DATABASE_DSN: mysql://root:Password12!@127.0.0.1:3306/test_maker
     matrix:
       - dependencies: highest
-        php_ver_target: 7.2.5
+        php_ver_target: 8.0.0
```
weaverryan, line 30: Tests failing, looks like we may need to bump this to 8.0.SOMETHING that actually works, which is fine.

weaverryan (approved):
🔥  just need to get the test to pass

PR #1124, opened 2022-05-18
WIP - implement types and constructor property promotion for utilities

jrushlow:


src/Doctrine/EntityClassGenerator.php
```diff
@@ -30,13 +30,8 @@
  */
 final class EntityClassGenerator
 {
-    private $generator;
-    private $doctrineHelper;
-
-    public function __construct(Generator $generator, DoctrineHelper $doctrineHelper)
+    public function __construct(private Generator $generator, private DoctrineHelper $doctrineHelper)
```
OskarStark, line 33:
```suggestion
    public function __construct(
    private Generator $generator,
    private DoctrineHelper $doctrineHelper,
    )
```

src/Security/SecurityConfigUpdater.php
```diff
@@ -23,11 +23,8 @@
...
 {
-    /** @var YamlSourceManipulator */
-    private $manipulator;
-
-    /** @var Logger|null */
-    private $ysmLogger;
+    private ?YamlSourceManipulator $manipulator;
+    private ?Logger $ysmLogger;
 
     public function __construct(Logger $ysmLogger = null)
```
OskarStark, line 29:
```suggestion
    public function __construct(?Logger $ysmLogger = null)
```

src/Security/UserClassConfiguration.php
```diff
@@ -18,21 +18,11 @@
...
-    private $identityPropertyName;
-
-    private $hasPassword;
-
-    private $useArgon2 = false;
-
-    private $userProviderClass;
-
-    public function __construct(bool $isEntity, string $identityPropertyName, bool $hasPassword)
+    public function __construct(private bool $isEntity, private string $identityPropertyName, private bool $hasPassword)
```
OskarStark, line 24:
```suggestion
    public function __construct(
    private bool $isEntity,
    private string $identityPropertyName,
    private bool $hasPassword,
    )
```

src/Util/ClassNameDetails.php
```diff
@@ -15,13 +15,13 @@
...
-    private $suffix;
+    private ?string $suffix;
 
-    public function __construct(string $fullClassName, string $namespacePrefix, string $suffix = null)
-    {
-        $this->fullClassName = $fullClassName;
+    public function __construct(
+        private string $fullClassName,
+        private string $namespacePrefix,
+        string $suffix = null
```
OskarStark, line 23:
```suggestion
        private ?string $suffix = null
```

and the member and assignment can be removed?

OskarStark:
I really think all property constructer promotions should be on a new line like:
```suggestion
    public function __construct(
        private LoggerInterface $logger = null,
    ) {
    }
```

it increases readability a lot. Not sure what you think @weaverryan

src/Util/UseStatementGenerator.php
```diff
@@ -30,9 +28,9 @@ final class UseStatementGenerator implements \Stringable
      *
      * @param string[]|array<string, string> $classesToBeImported
```
jrushlow, line 29: @OskarStark - over in symfony, do you guys use any of the #[ArrayShape] attributes instead of the param notations?

Something like https://blog.jetbrains.com/phpstorm/2020/10/phpstorm-2020-3-eap-4/#arrayshape or similar?

src/Doctrine/EntityDetails.php
```diff
@@ -21,14 +21,9 @@
...
-    private $metadata;
-
-    /**
-     * @param ClassMetadata|LegacyClassMetadata $metadata
-     */
-    public function __construct($metadata)
-    {
-        $this->metadata = $metadata;
+    public function __construct(
+        private ClassMetadata|LegacyClassMetadata $metadata
```
jrushlow, line 25: Note to self: we may be able to remove the `LegacyClassMetedata` - depends on doctrine version constraints...

src/Doctrine/EntityRelation.php
```diff
@@ -21,26 +21,18 @@ final class EntityRelation
...
-    private $isSelfReferencing = false;
-
-    private $orphanRemoval = false;
-
-    private $mapInverseRelation = true;
-
-    public function __construct(string $type, string $owningClass, string $inverseClass)
-    {
+    private bool $isNullable = false;
+    private bool $isSelfReferencing;
```
weaverryan, line 27:
```suggestion
    private bool $isSelfReferencing = false;
```

src/Util/UseStatementGenerator.php
OskarStark: Can't remember

src/Security/SecurityControllerBuilder.php
```diff
@@ -22,11 +22,9 @@
  */
 final class SecurityControllerBuilder
 {
-    private $phpCompatUtil;
-
-    public function __construct(PhpCompatUtil $phpCompatUtil)
-    {
-        $this->phpCompatUtil = $phpCompatUtil;
+    public function __construct(
+        private PhpCompatUtil $phpCompatUtil
```
OskarStark, line 26: We should always end with a `,`, it makes future diffs easier to read, also there is a PHP-CS-Fixer rule

src/Util/UseStatementGenerator.php
OskarStark: Thank you Wouter 😃

jrushlow: Thanks Wouter! I'm getting curious about using `#ArrayShape` from PHPStorm, or "cloning" it in MakerBundle's NS so we can use it here. We have quite a few multi-dimensional arrays that I've run across when refactoring the code over the past couple of weeks - I'm having to jump back and forth to figure out what the array is supposed to look like. 

On the flip side, this could be a sign we should be using a simple object representation instead of a complex array in those cases...

src/Security/SecurityControllerBuilder.php
jrushlow: Good catch! thanks Oskar. Done in #1127

PR #1127, opened 2022-05-20
update php-cs-fixer ruleset for php8

jrushlow:
- adds the PHP 8 && PHPUnit 8.4 migration rulesets

- forces trailing commas in multiline parameter lists to improve diff readability. e.g.
```php
public function __construct(
    private int $number,
    private string $word,
)
```

I also added trailing line commas in multiline _argument_ lists and then reverted the commit. See commit https://github.com/symfony/maker-bundle/pull/1127/commits/352f951250cd7fda3c63996c8839cb8778e66e7a

If this we _should_ keep multiline trailing commas for args - I'll drop the commit that reverts it. Thoughts @weaverryan @OskarStark @wouterj ?

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -15,7 +15,8 @@ $finder = (new PhpCsFixer\Finder())
 
 return (new PhpCsFixer\Config())
     ->setRules(array(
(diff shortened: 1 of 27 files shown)


OskarStark (approved):

weaverryan:
Thanks Jesse!

PR #1128, opened 2022-05-24
WIP

jrushlow:


src/Doctrine/EntityRegenerator.php
```diff
@@ -38,7 +38,7 @@ public function regenerateEntities(string $classOrNamespace): void
     {
         try {
             $metadata = $this->doctrineHelper->getMetadata($classOrNamespace);
-        } catch (MappingException|LegacyCommonMappingException|PersistenceMappingException $mappingException) {
+        } catch (MappingException|LegacyCommonMappingException|PersistenceMappingException) {
```
jrushlow, line 41: `Doctrine\Common` namespaced imports e.g. `LegacyCommonMappingException` will be removed in a separate PR as there are a lot of changes related to outdated legacy classes that need to be made based on a git grep

src/Maker/MakeUser.php
```diff
@@ -143,14 +138,17 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+        if ($isEntity && !$entityUsesAttributes) {
+            throw new \RuntimeException('MakeUser only supports attribute mapping with doctrine entities.');
+        }
 
         // B) Implement UserInterface
         $manipulator = new ClassSourceManipulator(
             sourceCode: $this->fileManager->getFileContents($classPath),
             overwrite: true,
-            useAttributesForDoctrineMapping: $useAttributesForDoctrineMapping
+            useAttributesForDoctrineMapping: $entityUsesAttributes
```
jrushlow, line 151: For make user we can gen a user that _is not_ an entity - in which case we don't add any attributes to the generated class.. But if the user _is_ an entity, we only generate it using attributes, otherwise we complain. Is my thinking correct on this?

weaverryan (commented):
Just minor stuff!

src/Generator.php
```diff
@@ -23,26 +23,23 @@
...
+    ) {
         $this->twigHelper = new GeneratorTwigHelper($fileManager);
         $this->namespacePrefix = trim($namespacePrefix, '\\');
 
-        if (null === $phpCompatUtil) {
-            $phpCompatUtil = new PhpCompatUtil($fileManager);
-
-            trigger_deprecation('symfony/maker-bundle', '1.25', 'Initializing Generator without providing an instance of PhpCompatUtil is deprecated.');
+        if (null !== $phpCompatUtil) {
+            trigger_deprecation('symfony/maker-bundle', '1.25', 'Initializing Generator while providing an instance of PhpCompatUtil is deprecated.');
```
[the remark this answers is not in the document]
jrushlow: Correct and good catch!

PR #1129, opened 2022-05-24
[tests] bring test suite up to PHP8 standards

jrushlow:


src/Test/MakerTestDetails.php
```diff
@@ -70,6 +66,8 @@ public function addExtraDependencies(string ...$packages): self
 
     public function setRequiredPhpVersion(int $version): self
     {
+        @trigger_deprecation('symfony/maker-bundle', 'v1.44.0', 'setRequiredPhpVersion() is no longer used and will be removed in a future version.');
```
weaverryan, line 69: Are we sure we won't need this again? :)

src/Test/MakerTestEnvironment.php
```diff
@@ -405,7 +398,7 @@ private function determineMissingDependencies(): array
         ');
 
         $process = MakerTestProcess::create('php dep_runner.php', $this->path)->run();
-        $data = json_decode($process->getOutput(), true);
+        $data = json_decode($process->getOutput(), true, 512, \JSON_THROW_ON_ERROR);
         if (null === $data) {
             throw new \Exception('Could not determine dependencies');
         }
```
weaverryan, line 404: I think we don't need this exception anymore, right?

src/Test/MakerTestRunner.php
```diff
@@ -64,6 +66,8 @@ public function copy(string $source, string $destination)
 
     /**
      * When using an authenticator "fixtures" file, this adjusts it to support Symfony 5.2/5.3.
+     *
+     * @deprecated legacy PassportInterface support will be dropped in a future version
```
weaverryan, line 70: Indeed - we can remove this right now (in another PR, as this applies to 5.2 and 5.3 only).

weaverryan (commented):
minor comments, but looks good 👍

src/Test/MakerTestEnvironment.php
jrushlow: good catch

src/Test/MakerTestDetails.php
jrushlow: Nope. But _we_ are not using it within the code base at the moment. Because the class isn't internal - I figured we'd throw the deprecation. And if/when we need to use this (PHP `9`ish) - we can either remove the deprecation or add a replacement `internal` method that gives us a bit more freedom. 

I _think_ we are going with a similar strategy w/ the `PHPCompatUtil` - we don't have a need for now (probably wont until PHP 9ish). So when we introduced the util in non-internal classes, we added a deprecation that "hey, you need to supply this util.. it'll be mandatory in the next version..." Now that we _don't_ need it in say MakeWhatever::class, I changed the wording on the deprecation "hey, you dont need to supply this..."

That strategy kinda sucks, but I can't think of a better way w/ non internal classes. Any ideas?

src/Test/MakerTestRunner.php
jrushlow: I wiped the code within the method, but we can't remove it until the next major because the class isn't internal right?

weaverryan: I see your point. But as this is a tool for testing... and the method only applies to versions that MakerBundle no longer supports, I think we can get away with it.

jrushlow: works for me.. done

src/Test/MakerTestDetails.php
weaverryan: works for me 👍

weaverryan:
Thanks Jesse!

PR #1136, opened 2022-06-07
use method add() instead of [] in collection adder-method

HKandulla:
use method add() instead of [] in collection adder-method


lyrixx (approved):

jrushlow (approved):
Awesome! Thank you @HKandulla

PR #1137, opened 2022-06-14
Fix for #1090 while keeping the symfony to 4.4

spyridonas:
LTS should be supported until the end of it's life.
This change allows for maker bundle to work with an updated Symfony 4.4 installation, fixes issue #1090

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -177,13 +177,20 @@ final class DoctrineHelper
      */
     public function getMetadata(string $classOrNamespace = null, bool $disconnected = false)
     {
-        $classNames = (new \ReflectionClass(AnnotationDriver::class))->getProperty('classNames');
-        $classNames->setAccessible(true);
-
-        // Invalidating the cached AnnotationDriver::$classNames to find new Entity classes
         foreach ($this->mappingDriversByPrefix ?? [] as $managerName => $prefixes) {
             foreach ($prefixes as [$prefix, $annotationDriver]) {
-                if (null !== $annotationDriver) {
+                if ($annotationDriver === null) {
+                    continue;
+                }
+                if (class_exists(AnnotationDriver::class)) {
+                    $classNames = (new \ReflectionClass(AnnotationDriver::class))->getProperty('classNames');
+                }
+                if ($annotationDriver instanceof AttributeDriver) {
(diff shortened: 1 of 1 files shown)


jrushlow:
`symfony/symfony` _does_ support `4.4` until it's end of life. However, MakerBundle is a bit more bleeding edge and does not need to continue to support older versions Symfony - e.g. `4.4`. In all practicality - users of MakerBundle tend to be developing newer applications, e.g. Symfony 6. The majority of our users are beyond actively developing in Symfony 4.4 other than maintenance. 

Dropping older versions allows us to progress MakerBundle quicker and helps relieve the burden of maintenance. Thank you for the PR, but MakerBundle no longer supports annotations.

sergiodinizoswald:
@jrushlow don't know if it's the right place to discuss this or not, but here it goes. If you prefer another way let me know.

Allow me to disagree, And I'm sure I'm not the only one.

> MakerBundle is a bit more bleeding edge and does not need to continue to support older versions Symfony - e.g. 4.4

I would say that this completely incorrect, you should support at least the LTS versions.

> In all practicality - users of MakerBundle tend to be developing newer applications, e.g. Symfony 6. The majority of our users are beyond actively developing in Symfony 4.4 other than maintenance.

You are making a huge assumption here, I think this is completely far from the truth. Why do you think LTS versions exist? It's not for them to be discarded after a new version exists, it's because the stability they provide. I'm sure there is a lot of people developing new projects on 6. As well there are many people maintaining projects on 4.4 and 5.4 ( like myself ), and won't even touch version 6 until there is an LTS version for it.

> Dropping older versions allows us to progress MakerBundle quicker and helps relieve the burden of maintenance. Thank you for the PR, but MakerBundle no longer supports annotations.

I agree, but since this a bundle for Symfony, you should support at least the LTS versions of Symfony, like someone said,
what image does it give to everybody when you install an LTS of Symfony and it simply gives an error on one of it's main packages.

Hope you guys reconsider including this PR.

weaverryan:
If it's really as simple as merging this PR to fix the bug in 4.4, I would support tagging a 4.4 bug fix release. Though, we would need a new PR - as the source branch was deleted.

spyridonas:
@weaverryan i hope you can merge it, i've remade it on https://github.com/symfony/maker-bundle/pull/1163

spyridonas:
@weaverryan any plans to merge the other pr ?

PR #1139, opened 2022-06-15
[make:entity] Improve uid support

HypeMC:
1. Currently `uuid` is offered as the default type when the field name is uuid, which breaks when the `symfony/uid` component is not installed. Now `guid` is used when `uuid` is not available.

2. Getters & setters for `uuid`/`ulid` types now have type hints added when generated.


jrushlow (approved):
Thank you @HypeMC

PR #1140, opened 2022-06-16
Erase password by default

nicolas-grekas:
While the password is hashed, we don't need it in the session storage, so we can erase it, isn't it?

### DIFF
--- a/src/Security/UserClassBuilder.php
+++ b/src/Security/UserClassBuilder.php
@@ -34,7 +34,7 @@ final class UserClassBuilder
 
         $this->addPasswordImplementation($manipulator, $userClassConfig);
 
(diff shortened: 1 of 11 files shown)


lyrixx (approved):

weaverryan (commented):
I think... this doesn't work, unfortunately:

1) If we erase the password after successful auth, there's a risk something will `$em->flush()` during the request and save the empty string to the database.

2) Even if we did this, on the next request, a fresh `User` would be loaded with their hashed password... and then the hashed password would be stored in the session at the end of the request anyways.

iirc, we also rely on the hashed password of the serialized User and the "fresh" user to match, otherwise we log you out (this is the mechanism to log out other sessions of their user when they change their password).

PR #1145, opened 2022-07-05
[ci] fix blocked plugins

jrushlow:
Ref'd in https://github.com/composer/composer/issues/10925

Since Composer `2.3.9` unless you explicitly allow/disallow a plugin (`flex` in our case), an error is thrown (halts our CI run) before any tests are actually ran.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -139,7 +139,7 @@ jobs:
               run: composer diagnose
 
             - name: "Require symfony/flex"
-              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex
+              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex -vvv
 
             - name: "Composer install"
               uses: "ramsey/composer-install@v1"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -135,6 +135,9 @@ jobs:
             - name: "Add PHPUnit matcher"
               run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
 
+            - name: "No Plugins"
+              run: composer global config --no-plugins allow-plugins.symfony/flex false
```
weaverryan, line 139: I think we DO want Flex enabled - iirc, we install it to get a performance boost.

Seldaek:
Note the allow-plugins could simply always be in composer.json. And the flex require could also perhaps be in require-dev always but maybe not, not sure why you add it dynamically, I'll leave that up to you :)

jrushlow:
Thanks Jordi - as it turns out we don't need to require flex anymore. This was actually an artifact from the old TravisCI days where doing a global require flex provided a perf boost in CI. 

Just for reference in the future; if we were to explicitly add flex, or any other package, to our allow-plugins in composer.json, would this mean that when a user added _our_ package to their composer project - the plugins that _we_ allowed here would be inherited and in their allow list? Or would they also have to explicitly add flex (or xyz package) to their allow-plugins as well?

Not sure if that makes sense. Either way glad we could help identify a bug on your end and remove some legacy config on our end.

Seldaek:
No every project is fully in charge of which plug-ins are run to ensure a dependency can't sneak one on you.

PR #1147, opened 2022-07-11
[make:entity] Property types, Types:: constant & type guessing

weaverryan:
Hi!

This expands on the excellent work in #1039. This adds to `make:entity`:

A) Property types - e.g. `private ?int $id = null`. Most property types are nullable and default to null, effectively making them typed, but identical to how things worked before (before types, properties were naturally nullable and defaulted to null).

B) Removes *most* `type:` options in `ORM\Column` as these are now "guessed" since doctrine/orm 2.9 from the property types.

C) For the few `type:` options that remain, we use a `Types::` constant.

Cheers!

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Doctrine;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\ORM\Mapping\MappingException as ORMMappingException;
@@ -24,6 +25,8 @@ use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
 use Doctrine\Persistence\Mapping\MappingException as PersistenceMappingException;
 use Symfony\Bundle\MakerBundle\Util\ClassNameDetails;
+use Symfony\Component\Uid\Uuid;
+use Symfony\Component\Uid\Ulid;
 
 /**
  * @author Fabien Potencier <fabien@symfony.com>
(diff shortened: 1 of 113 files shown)


weaverryan:
This is ready! The Symfony 6.2.x-dev failures are from an upstream bug - https://github.com/symfony/symfony/issues/46916

src/Doctrine/DoctrineHelper.php
```diff
@@ -230,6 +233,64 @@ public function isClassAMappedEntity(string $className): bool
...
+    public static function getTypeConstant(string $columnType): ?string
+    {
+        $reflection = new \ReflectionClass(Types::class);
+        $constants = array_flip($reflection->getConstants());
+
+        if (!isset($constants[$columnType])) {
+            return null;
+        }
+
+        return 'Types::'.$constants[$columnType];
```
jrushlow, line 291: For consistency...
```suggestion
        return sprintf('Types::%s', $constants[$columnType]);
```

src/Doctrine/EntityClassGenerator.php
```diff
@@ -48,7 +47,7 @@ public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $
 
         $useStatements = new UseStatementGenerator([
             $repoClassDetails->getFullName(),
-            [Mapping::class => 'ORM'],
+            ['Doctrine\\ORM\\Mapping' => 'ORM'],
```
jrushlow, line 50: Is this for code readability?

src/Resources/skeleton/doctrine/Entity.tpl.php
```diff
@@ -17,8 +17,8 @@ class <?= $class_name."\n" ?>
 {
     #[ORM\Id]
     #[ORM\GeneratedValue]
-    #[ORM\Column(type: 'integer')]
-    private int $id;
+    #[ORM\Column()]
+    private ?int $id = null;
```
jrushlow, line 21: Side note - I think with the minimum versions we use for Doctrine now, we no longer need properties to be nullable. As I recall, this was _required_ in earlier versions otherwise doctrine would through an `exception` when initiating a class. E.g. 
```php
SomeEntity
{
    private int $id;

    public function getId(): int
    {
        return $this->id;
    }
    ...
}
```

This is no longer the case and the above works without issue. But it depends on which side of the coin we are looking at - for the purposes of Doctrine, `null` properties are not needed we are just adding unnecessary types (especially in terms of auto-generated data). In the terms of the codebase, nullable private properties _could_ make sense if a property is not init'd via the constructor. 

Personally, I lean more towards the doctrine train of thought - only make a property nullable if it is intended to be nullable.. Thoughts?

src/Util/ClassSourceManipulator.php
```diff
@@ -47,6 +47,8 @@ final class ClassSourceManipulator
     private const CONTEXT_CLASS = 'class';
     private const CONTEXT_CLASS_METHOD = 'class_method';
 
+    private const DEFAULT_VALUE_NONE = '__default_value_none';
```
jrushlow, line 50:
```suggestion
    private const CONTEXT_CLASS_METHOD = 'class_method';
    private const DEFAULT_VALUE_NONE = '__default_value_none';
```

jrushlow (commented):
Shweet! Couple of minor comments...

src/Resources/skeleton/doctrine/Entity.tpl.php
weaverryan: I decided to make them nullable, not for Doctrine, but to avoid the PHP `must not be accessed before initialization` error that you'd get if you accessed the `$id` property (or any other property) before setting it / saving. I'm not positive it's the best approach, what I did is the most conservative: the code works exactly like the untyped version of the code. But, it does have "looser" types, so it's not *as* awesome.

jrushlow (approved):
Awesome Work! Thanks @Geekimo && @weaverryan

Geekimo:
Thanks @weaverryan for updating my work !

icedevelopment:
@Geekimo @weaverryan first off, nice work! I noticed this PR explicitely adds ` = null` after each property which doesn't seem necessary as they should already default to null or maybe I'm missing something?

Geekimo:
@icedevelopment Hello, thanks.
A property never defaults to null unless it is explicitely specified.
If you do not set a default value, it won't be accessible unless initialized via an affectation.

jorismak:
phpstan now always flags the properties as incorrect, because the database says 'not null' and the PHP code says 'nullable'.

I know that you can't access an uninitialized property, but forcing it to null... is that right?

We've worked for years with non-nullable properties this way. You set them in the constructor if you want safe, or other coding standards..

I'm now in doubt if I need to ignore the errors in phpstan, or adjust the scaffolding from make:entity to make the nullable types match the database layout.

Maybe from my C/C++ background, but an uninitialized property (and the dangers of it :P) sound perfectly fine in my head :).

PR #1153, opened 2022-07-16
[make:registration-form] Fix escape text-strings

### LINKED ISSUE #1151: generator doesn't full escape text-strings when required

6.1.0

### Description

registration form generator does not escape strings where applicable meaning it's very easy to generate invalid PHP.

### How to reproduce

1) execute `console make:registration-form`

2) fill in FROM address for email as something with an apostrophe in it.  e.g. "Xalior's Awesome Webapp"

Results in following code:

`

            $this->emailVerifier->sendEmailConfirmation('app_verify_email', $user,

                (new TemplatedEmail())

                    ->from(new Address('noreply@example.com', 'Xalior's Awesome Webapp'))

                    ->to($user->getEmail())

                    ->subject('Please Confirm your Email')

                    ->htmlTemplate('registration/confirmation_email.html.twig')`

### Possible Solution

Escape the string before generating PHP; or don't permit single quotes in entryfield.

### Additional Context

_No response_

mdoutreluingne:
Fixes #1151 

Need reviews to know if it is correct to use the [`addslashes()`](https://www.php.net/manual/fr/function.addslashes.php) function for fix escape text-strings here🙏

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -160,13 +160,13 @@ final class MakeRegistrationForm extends AbstractMaker
             $this->emailGetter = $interactiveSecurityHelper->guessEmailGetter($io, $this->userClass, 'email');
 
             $this->fromEmailAddress = $io->ask(
-                'What email address will be used to send registration confirmations? e.g. mailer@your-domain.com',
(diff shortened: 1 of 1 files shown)


jrushlow (approved):
Thanks @mdoutreluingne - `addslashes` works for me. We could use `filter_var()` w/ the `FILTER_SANITIZE_ADD_SLASHES` but this would ultimately call `addSlashes()` anyways.

Issue #1166, opened 2022-08-05
ClassSourceManipulator::addSetter(): Argument #2 ($type) must be of type string, null given


mklewitz:
https://github.com/symfony/maker-bundle/commit/ffc80c4d9b4eab6f86d07c33b90e6b9d4f030359 added ```string``` as type hint for ```Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator::addSetter()```'s ```$type```.

However, ```ClassSourceManipulator::addEntityField()``` will call ```addSetter()``` with ```$type == null``` for all custom defined types and even for builtin type ```blob```.

That results in a type error when (re)generating entities: ```TypeError: Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator::addSetter(): Argument #2 ($type) must be of type string, null given```

sandoba:
Can confirm, this creates problems at the moment for the types "binary", "blob" and "ascii_string". All of these will crash the entity creation process. All other types seem to be ok.

The error message above continues:
`called in ...\vendor\symfony\maker-bundle\src\Util\ClassSourceManipulator.php on line 144`

vladdnepr:
Same issue

jrushlow:
Fixed in #1204

PR #1167, opened 2022-08-05
Fix type error in ClassSourceManipulator::addSetter()

mklewitz:
Fixes https://github.com/symfony/maker-bundle/issues/1166


jrushlow:
Thanks for the fix @mklewitz - we actually just merged #1204 which addresses this.

Issue #1138, opened 2022-06-14
Repository template - rename add method to save 


MrYamous:
Hi, 

Few times ago, PR #939 added new `add` and `remove` methods in repository template

I recently had the opportunity to use them, and when i made a method to edit a resource in my controller i was like "where is the method to edit ?" and then i realize `add` can do the job but i think it's a confusing name.

What would you think about rename this `add` method to something more generic like `save` ?

MrYamous:
Hi @bdaler 
As contributor of the concerned PR, do you have any advice about this ?

bdaler:
Hi @MrYamous 
Sounds reasonable and I don't mind. Fill free to make PR

Issue #788, opened 2021-01-18
Improve controller generation in make:crud command


garak:
**Description**  

Currently, the `make:crud` maker command generates actions that rely on explicit calls to Doctrine's manager.
I know that is already possible for user to change that code and improve it, but my point is about giving good suggestions (and I think current implementation fails on that).
Example of a generated code:
```php
// in "new" action
$entityManager = $this->getDoctrine()->getManager();
$entityManager->persist($tinyChef);
$entityManager->flush();
// in "edit" action
$this->getDoctrine()->getManager()->flush();
// in "delete" action
$entityManager = $this->getDoctrine()->getManager();
$entityManager->remove($tinyChef);
$entityManager->flush();
```
This is breaking the law of Demeter, making controllers know too much about Doctrine internals.

My proposal is to add two methods to generated repository class: `add` and `remove`. The first method would be responsible to add/edit entity and the second one to delete it. With such methods (of course, by injecting repository in "new"/"edit"/"delete" actions), previous lines can be replaced with:

```php
// in "new" action
$tinyChefRepository->add($tinyChef);
// in "edit" action
$tinyChefRepository->add($tinyChef);
// in "delete" action
$tinyChefRepository->remove($tinyChef);
```

This is an example of generated code for `add` method in repository class:

```php
public function add(TinyChef $tinyChef, bool $flush = true): void
{
    $this->getEntityManager()->persist($tinyChef);
    if ($flush) {    // I think it's good to keep the opportunity to avoid flushing
        $this->getEntityManager()->flush();
    }
}
```

The `remove` method would be pretty the same (with `remove` instead of `persist`).

With such change, controllers wouldn't even know that they're handling Doctrine (except for namespaces), they would just rely on repositories.
I guess that, in a future, this could be also leading to deprecation/removal of `getDoctrine()` helper in the ControllerTrait.

xabbuh:
@garak FYI, I have moved the issue to the MakerBundle repository.

garak:
> @garak FYI, I have moved the issue to the MakerBundle repository.

Ooops, sorry about that, I was convinced that this bundle was in the symfony monorepo

jrushlow:
Howdy @garak. I actually love the idea - I do this currently in my projects anytime I need to leverage the repository itself. @weaverryan and I were talking about getting away from leaning on calling the entity manager directly in cases where we don't need to explicitly for other parts of maker:bundle. 

[...]

PR #1169, opened 2022-08-20
[make:crud] Rename `add` method of repository to `save`.

bdaler:
- rename `add` method of repository to `save`
- https://github.com/symfony/maker-bundle/issues/1138

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -74,8 +74,8 @@ jobs:
                 image: dunglas/mercure
                 env:
                     SERVER_NAME: :1337
-                    MERCURE_PUBLISHER_JWT_KEY: '!ChangeMe!'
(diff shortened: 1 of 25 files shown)


jrushlow:
@bdaler I just merged in a few CI fixes, can you do a rebase please?

jrushlow (changes requested):
Looks good to me, we just need passing tests that are related to this PR.

weaverryan:
Ping @garak who originally proposed the methods in #788. I don't really care what the methods are called - but I don't like changing them over and over again. So if we are going to make this change, we need some agreement that this is how it should be named.

garak:
IMHO "add" is more consistent with the concept of collection, but I don't have anything against "save"

jrushlow:
The commit history looks a bit mashed up in this PR. Closing in favor of #1204 which contains cherry-picked changes from this PR.

PR #1176, opened 2022-09-10
[make:entity] Fix error while making blob in entity

### LINKED ISSUE #1174: Error while making blob in entity
I am making an entity using `php bin/console make:entity`. all other fields work fine but when I make a field `blob` maker bundle return this error.
```
Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator::addSetter(): Argume  
  nt #2 ($type) must be of type string, null given, called in /home/arashdeep  
  /www/raagfinderweb/vendor/symfony/maker-bundle/src/Util/ClassSourceManipula  
  tor.php on line 144  
```

mdoutreluingne:
Fixes #1174

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -266,7 +266,7 @@ final class ClassSourceManipulator
         $this->addCustomGetter($propertyName, $methodName, $returnType, $isReturnTypeNullable, $commentLines);
     }
 
-    public function addSetter(string $propertyName, string $type, bool $isNullable, array $commentLines = []): void
+    public function addSetter(string $propertyName, ?string $type, bool $isNullable, array $commentLines = []): void
     {
         $builder = $this->createSetterNodeBuilder($propertyName, $type, $isNullable, $commentLines);
         $builder->addStmt(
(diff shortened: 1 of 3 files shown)


weaverryan:
This looks good - but tests are red. If you rebase with `main`, that might help - @jrushlow fixed most of the test suite a few days ago.

mdoutreluingne:
Rebase done. It seems that my test is not working, I will look at it tomorrow to try to correct it.

mdoutreluingne:
It seems that the red tests are not related to my PR. However, I want to correct the tests but I can't, help would be welcome :)

weaverryan (approved):
Some upstream changes, once again, broke the tests (MakerBundle's test suite, by design, is sensitive to recipe changes... which is great, but also annoying). We'll put a PR soon to fix that. But this is ready to go 👍

jrushlow:
@mdoutreluingne I just merged in a few CI fixes. Can you rebase this PR?

mdoutreluingne:
Rebase done. Unfortunately the tests are always red with the same message `Failed asserting that two strings are identical.`. I also re-tested locally by deleting my changes (to see if it is related to my PR) but the tests are also red.

jrushlow:
Thanks @mdoutreluingne - I'm going to push a commit or two to this PR to get the passing tests. It looks like we'll need to add another test fixture to handle the change...

No need to push any additional changes to the PR.

jrushlow (approved):
Thanks @mdoutreluingne for the fix!

PR #1197, opened 2022-09-15
Fix methodName as string when adding method

Vincz:
`$method->name` is an instance of `PhpParser\Node\Identifier` and must be cast to string for the following operations.  Related error:

````
[TypeError]
  Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator::methodExists(): Argument #1 ($methodName) must be of type string, PhpParser\Node\Identifier given, called in /project/vendor/symfony/maker-bundle/src/Util/ClassSourceManipulator.php on line 1036
```

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -1026,7 +1026,7 @@ final class ClassSourceManipulator
     private function addMethod(Node\Stmt\ClassMethod $methodNode): void
     {
         $classNode = $this->getClassNode();
-        $methodName = $methodNode->name;
+        $methodName = (string) $methodNode->name;
         $existingIndex = null;
         if ($this->methodExists($methodName)) {
             if (!$this->overwrite) {


sadikoff:
Hey @Vincz 

Just a gentle ping, can you provide more information here? 

Cheers!

Vincz:
Hi guys! Sorry for the late reply. I think it was a problem on my side. My editor automatically adds `declare(strict_types=1)` in the header of my PHP files when I saved them. I think I saved by mistake the file, the strict_types was added and the code failed.
The auto-casting to `string` of `stringable objects` seems to be only allowed if `strict_types` is not enabled. 

The following code doesn't work for example (but works if we remove the strict_types).

```php
<?php

declare(strict_types=1);

class Identifier
{
    public function __toString()
    {
        return 'coucou';
    }
}

function test(string $name)
{
    echo $name;
}

$identifier = new Identifier();
test($identifier);
```

I think you can close this PR. Sorry for the noise.

jrushlow:
No worries & sorry for the late reply :) I'm tempted to merge this, but as we do not use `strict_types` anywheres within MakerBundle, and CSM is internal, I'm not. Don't want to start a precedent of supporting `strict_types` just yet. Thanks!

PR #1199, opened 2022-09-20
[make:entity] fix compatibility with api-platform 3.0

### LINKED ISSUE #1198: MakeEntity is not compatible with api-platform 3.0
Api-platform 3.0 use ApiPlatform\Metadata\ApiResource class instead ApiPlatform\Core\Annotation\ApiResource, but MakerEntity still use it.

yobrx:
I fixed #1198, but I don't know how launch tests suites easily, I don't have found documentation for that.

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -318,14 +318,14 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
     public function configureDependencies(DependencyBuilder $dependencies, InputInterface $input = null): void
     {
         if (null !== $input && $input->getOption('api-resource')) {
-            if (class_exists(LegacyApiResource::class)) {
+            if (class_exists(ApiResource::class)) {
                 $dependencies->addClassDependency(
-                    LegacyApiResource::class,
+                    ApiResource::class,
                     'api'
                 );
             } else {
                 $dependencies->addClassDependency(
-                    ApiResource::class,
+                    LegacyApiResource::class,
                     'api'
                 );
             }
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -11,7 +11,8 @@
 
 namespace Symfony\Bundle\MakerBundle\Maker;
 
-use ApiPlatform\Core\Annotation\ApiResource;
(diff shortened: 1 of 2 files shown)


weaverryan:
We need to push a fix for the test suite - a recipe change upstream broke things. 

But, to run tests locally, you can run `composer install` then `./vendor/bin/simple-phpunit`. You will likely also need to tweak the `TEST_DATABASE_DSN` setting in `phpunit.xml.dist` to match your local settings. The test suite is HUGE and slow - so I usually target the specific tests that I want to run :)

yobrx:
Thanks ! I have failures but not (normally) related to my dev.

jrushlow:
@yobrx - I've just merged in a few fixes to our CI. Doing a rebase against `main` should fix some of the CI failures we're seeing here

yobrx:
It's green, except one test on [6.2.x](https://github.com/symfony/maker-bundle/actions/runs/3092586683/jobs/5004024578)

weaverryan (approved):
The 6.2 failure looks weird and probably a real failure - but certainly unrelated to this PR 👍

src/Maker/MakeEntity.php
```diff
@@ -317,10 +318,17 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
     public function configureDependencies(DependencyBuilder $dependencies, InputInterface $input = null): void
     {
         if (null !== $input && $input->getOption('api-resource')) {
-            $dependencies->addClassDependency(
-                ApiResource::class,
-                'api'
-            );
+            if (class_exists(LegacyApiResource::class)) {
```
dunglas, line 321: I would test for `ApiResource` first because in 2.7 both classes exist, and we prefer using the non-deprecated attribute.

dunglas (changes requested):

src/Maker/MakeEntity.php
yobrx: Hi Kévin, I fixed that.

dunglas (approved):
Thank you Yoann!!

jrushlow (approved):
Awesome! Thank you @yobrx

PR #1201, opened 2022-09-20
[ci] handle reset-password 8.1.10 version constraint

jrushlow:


src/Test/MakerTestDetails.php
```diff
@@ -73,6 +75,22 @@ public function setRequiredPhpVersion(int $version): self
         return $this;
     }
 
+    /**
+     * Skip a test from running between a range of PHP Versions.
+     *
+     * @param int $lowerLimit Versions below this value will be allowed
+     * @param int $upperLimit Version above this value will be allowed
```
jrushlow, line 82:
```suggestion
     * @param int $upperLimit Versions above this value will be allowed
```

src/Test/MakerTestDetails.php
```diff
@@ -118,7 +136,22 @@ public function getDependencyBuilder(): DependencyBuilder
...
+        $hasPhpVersionConstraint = $this->blockedPhpVersionLower > 0 && $this->blockedPhpVersionUpper > 0;
+        $isSupported = false;
+
+        if (!$hasPhpVersionConstraint) {
+            $isSupported = true;
+        }
+
+        if (\PHP_VERSION_ID > $this->blockedPhpVersionUpper) {
+            $isSupported = true;
+        }
```
weaverryan, line 148: This is what makes sense to me:

```php
if ($hasPhpVersionConstraint && \PHP_VERSION_ID >= $this->blockedPhpVersionLower && \PHP_VERSION_ID <= $this->blockedPhpVersionUpper) {
    return false;
}
```

It says: 

> If the a blocked range of php versions was specified AND the current php version is inside of that range (greater than or equal to the lower and less than or equal to the upper), return false: the current php version is not supported.

PR #1202, opened 2022-09-21
[reset-password] use higher level "options" in ChangePasswordFormType.tpl.php

seb-jean:
I add [options ](https://symfony.com/doc/current/reference/forms/types/repeated.html#options) instead of writing to first_options and second_options.


jrushlow (approved):
Thank you! @seb-jean

PR #1204, opened 2022-09-22
[make:crud] use save instead of add repository methods

### LINKED ISSUE #788: Improve controller generation in make:crud command

### LINKED ISSUE #1138: Repository template - rename add method to save

jrushlow:
Instead of using an `add()` method in the repository, it's been renamed to `save()`. Logic behind the change is to not confuse with "add()" and item to a collection - we're "save()"ing an entity to persistence. 

Original work done by @bdaler in #1169 
fixes #1138 
fixes #788

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -29,7 +29,7 @@ class <?= $class_name ?> extends AbstractController
 <?php endif ?>
 
 <?= $generator->generateRouteForControllerMethod('/new', sprintf('%s_new', $route_name), ['GET', 'POST']) ?>
-<?php if (isset($repository_full_class_name) && $generator->repositoryHasAddRemoveMethods($repository_full_class_name)) { ?>
+<?php if (isset($repository_full_class_name) && $generator->repositoryHasSaveAndRemoveMethods($repository_full_class_name)) { ?>
     public function new(Request $request, <?= $repository_class_name ?> $<?= $repository_var ?>): Response
 <?php } else { ?>
     public function new(Request $request, EntityManagerInterface $entityManager): Response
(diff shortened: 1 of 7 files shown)


weaverryan (approved):
ok with me 👍

PR #1209, opened 2022-09-26
[make:crud] use save instead of add in `upgradePassword`

seb-jean:
The `save` function has been forgotten for `upgradePassword`


jrushlow (approved):
Awesome! Good Catch! Thanks @seb-jean

PR #1211, opened 2022-09-26
[make:twig-extension] Change folder for Twig Extension

seb-jean:
I think it would be interesting to put the Twig Extensions in a `src/Twig/Extension` folder rather than `src/Twig`. The idea is to do like the Twig Components.

### DIFF
--- a/src/Maker/MakeTwigExtension.php
+++ b/src/Maker/MakeTwigExtension.php
@@ -51,7 +51,7 @@ final class MakeTwigExtension extends AbstractMaker
     {
         $extensionClassNameDetails = $generator->createClassNameDetails(
             $input->getArgument('name'),
-            'Twig\\',
+            'Twig\\Extension',
             'Extension'
         );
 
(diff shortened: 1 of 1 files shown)


src/Maker/MakeTwigExtension.php
```diff
@@ -51,7 +51,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
     {
         $extensionClassNameDetails = $generator->createClassNameDetails(
             $input->getArgument('name'),
-            'Twig\\',
+            'Twig\\Extension',
```
OskarStark, line 54:
```suggestion
            'Twig\\Extension\\',
```
?

jrushlow (approved):
Thanks @seb-jean!

weaverryan:
Devil's advocate: what else do you have in the `src/Twig/` directory? Same thing with forms - I like putting them in `src/Form` instead of `src/Form/Type` because, most of the time, they will be the only thing in `src/Form` :)

seb-jean:
> Devil's advocate: what else do you have in the `src/Twig/` directory? Same thing with forms - I like putting them in `src/Form` instead of `src/Form/Type` because, most of the time, they will be the only thing in `src/Form` :)

Now, I have a folder `src/Twig/Components` and files `src/Twig/...Extension.php`

weaverryan:
haha, that's a good point 😆 . Ok then 👍

PR #1214, opened 2022-09-28
inline createForm -> HandleRequest

BASAKSemih:


jrushlow (changes requested):
I'm not in favor of this change. While this _is_ something I would probably do in my own code, from a maintainers point of view, it makes the template a bit more harder to read.

Thoughts @weaverryan

weaverryan:
I don't feel too strongly either way. But Maker should generate code like the docs - so a change like this should be proposed there first :).

Cheers!

PR #1215, opened 2022-09-29
WIP [make:crud] replace renderForm by render method (AbstractController)

BASAKSemih:
Since Symfony 6.2, renderForm is deprecated


OskarStark:
Thank you for the PR.

@jrushlow @weaverryan I am not sure how we deal with such things as this bundle can still be used with an older Symfony version, or am I missing sth.?

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -55,9 +55,9 @@ public function new(Request $request, EntityManagerInterface $entityManager): Re
 <?php } ?>
 
 <?php if ($use_render_form) { ?>
-        return $this->renderForm('<?= $templates_path ?>/new.html.twig', [
+        return $this->render('<?= $templates_path ?>/new.html.twig', [
```
jrushlow, line 58: We'll need to do this conditionally. E.g. if Symfony >6.1 `render()` else `renderForm()`

It looks like in 6.2+, the `render()` method is able to set the appropriate response based on `FormInterface` parameters. (`isSubmitted()` && `isValid()`). As far as I know, that logic does _not_ exist in `render()` in `5.4.`.

Adding a test to ensure this logic works correctly regardless of the version would be awesome!

src/Resources/skeleton/crud/controller/Controller.tpl.php
```diff
@@ -55,9 +55,9 @@ public function new(Request $request, EntityManagerInterface $entityManager): Re
...
             '<?= $entity_twig_var_singular ?>' => $<?= $entity_var_singular ?>,
-            'form' => $form,
+            'form' => $form->createView(),
```
jrushlow, line 60: This would need to be conditional as well...

jrushlow (changes requested):

src/Resources/skeleton/crud/controller/Controller.tpl.php
BASAKSemih: I will update my PR soon :D

PR #1221, opened 2022-10-10
[make:voter] Set type for subject in Voter template

N-M:
Setting type for $subject in Voter template so phpstan issue can be covered by default

```
 ------ ----------------------------------------------------------------- 
  15     Method App\Security\Voter\TenantVoter::supports() has parameter  
         $subject with no type specified.                                 
 ------ ----------------------------------------------------------------- 
```


94noni (approved):

weaverryan (approved):

jrushlow (approved):
Thank you @N-M!

PR #1227, opened 2022-10-23
[make:registration] Make router optional in MakeRegistrationForm constructor

odolbeau:
In a console application, I have the following configuration:
```yaml
framework:
  router: false
```

As the `MakeRegistrationForm` class need a `RouterInterface` to be constructed, I get the following error: `Symfony\Bundle\MakerBundle\Maker\MakeRegistrationForm::__construct(): Argument #3 ($router) must be of type Symfony\Component\Routing\RouterInterface, null given`

This PR aim to remove this dependency.

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -85,8 +85,8 @@ final class MakeRegistrationForm extends AbstractMaker
     public function __construct(
         private FileManager $fileManager,
         private FormTypeRenderer $formTypeRenderer,
-        private RouterInterface $router,
         private DoctrineHelper $doctrineHelper,
+        private ?RouterInterface $router = null,
     ) {
     }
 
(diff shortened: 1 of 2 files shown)


src/Maker/MakeRegistrationForm.php
```diff
@@ -111,6 +111,10 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
     {
         $interactiveSecurityHelper = new InteractiveSecurityHelper();
 
+        if (null === $this->router) {
+            throw new RuntimeCommandException('Router have been explicitely disabled in your configuration. This command needs to use the router.');
```
odolbeau, line 115: Help welcome to choose a clearer message.

weaverryan:
Ah, so you can't run maker at all in your console app? That makes sense - I don't normally think about the router service not being available, but that seems legal to me.

src/Maker/MakeRegistrationForm.php
```diff
@@ -85,8 +85,8 @@ final class MakeRegistrationForm extends AbstractMaker
     public function __construct(
         private FileManager $fileManager,
         private FormTypeRenderer $formTypeRenderer,
-        private RouterInterface $router,
         private DoctrineHelper $doctrineHelper,
+        private ?RouterInterface $router = null,
```
weaverryan, line 89: Class is internal, so no problems changing args 👍

src/Maker/MakeRegistrationForm.php
weaverryan: Looks good enough to me - especially for this edge-case

weaverryan (approved):

jrushlow (approved):
Thank you @odolbeau

PR #1239, opened 2022-11-21
Improve error messages to show PHP & XML configurations are not supported

### LINKED ISSUE #1231: make:registration-form only works with YAML config (not PHP, XML)
When running `php bin/console make:registration-form`, I'm getting:

> [ERROR] The file "config/packages/security.yaml" does not exist. This command needs that file to accurately build your 
         registration form.

I did the configuration in PHP, so I do have `config/packages/security.php`.

ThomasLandauer:
Closes #1231
As suggested in https://github.com/symfony/maker-bundle/issues/1231#issuecomment-1321891623


jrushlow (approved):
Thanks @ThomasLandauer

PR #1242, opened 2022-11-28
[make:*] use static return type instead of self for setters

### LINKED ISSUE #1236: Entity generation: switching from return types :self to :static in fluent setters
I figured that with PHP8.0+ we have `: static` as a return type , which is the best option for return type in fluent setters in entities. Can this be updated in place of the `: self` that is currently in use ?

jrushlow:
fixes #1236


florentdestremau:
nice !

PR #1243, opened 2022-11-28
[make:registration] drop guard authentication support

jrushlow:
A wee bit late - but this pr drops the use of the old `guard` authenticators from `4.x`

## DX

- login after registration: 
Previously we asked which firewall to use -> which authenticator, etc... Now we just grab all of the authenticators available and let the user choose in one question.

internally we iterate through all of the `security.firewalls` to looking for a built in authenticator or a `custom_authenticator`. If multiple choices are found, we ask the user which one to use. e.g. 

image

If only 1 authenticator was found, obviously don't ask which one to use...

we support 
```yaml 
custom_authenticator: App\My\Authenticator
``` 

&&

```yaml
custom_authenticator:
    - App\My\Authenticator
```

Both of which appear to be valid way to declare a custom authenticator.

## Under the hood

- an `Authenticator` value object && a `AuthenticatorType` enum are introduced to get us away from passing around array's of array's of array's... Goal is to bring better visibility at a glance when reviewing code && get us closer to implementing static analysis without headaches..

- we generate the code to login a user by calling `Security::login($user, 'name_of_authenticator', 'name_of_firewall')` in the template. Most users do not need the 3rd `firewall` parameter, but I'm pulling the "keep maintainers lives easier by writing less code" card. If this bug's folks, later, we can add in additional logic && conditionals to determine _if_ there are multiple firewalls.

refs: https://github.com/SymfonyCasts/verify-email-bundle/issues/126


tests/fixtures/make-registration-form/expected/RegistrationControllerNoLogin.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    #[Route('/register', name: 'app_register')]
+    public function register(Request $request, UserPasswordHasherInterface $userPasswordHasher, EntityManagerInterface $entityManager): Response
+    {
+        $user = new User();
+        $form = $this->createForm(RegistrationFormType::class, $user);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            // encode the plain password
```
weaverryan, line 24:
```suggestion
            // hash the plain password
```

tests/fixtures/make-registration-form/expected/RegistrationControllerNoLogin.php
```diff
@@ -0,0 +1,43 @@
...
+
+            $entityManager->persist($user);
+            $entityManager->flush();
+            // do anything else you need here, like send an email
+
+            return $this->redirectToRoute('app_login');
+        }
+
+        return $this->render('registration/register.html.twig', [
+            'registrationForm' => $form->createView(),
```
weaverryan, line 40: It might not be worth the bother... cause of the complexity, but in Symfony 6.2, you don't need to call `->createView()` when using `->render()`: it happens automatically.

tests/fixtures/make-registration-form/expected/RegistrationControllerFormLogin.php
```diff
@@ -0,0 +1,48 @@
...
+                    $form->get('plainPassword')->getData()
+                )
+            );
+
+            $entityManager->persist($user);
+            $entityManager->flush();
+
+            // do anything else you need here, like send an email
+
+            $security->login($user);
```
weaverryan, line 39:
```suggestion
            return $security->login($user);
```

The authenticators doesn't HAVE to return a `Response`... but whatever authenticator you're using as your "log in" for registration, I think, would return a response. And we want to use that response.

src/Maker/MakeRegistrationForm.php
```diff
@@ -209,17 +212,36 @@ private function interactAuthenticatorQuestions(InputInterface $input, ConsoleSt
...
-                    $authenticatorClasses
-                );
+        // get list of authenticators
+        $authenticators = $interactiveSecurityHelper->getAuthenticatorClasses($firewallsData[$firewallName]);
+
+        $canLoginAfterRegistration = $this->canSecurityBundleLogin;
+
+        if (!$this->canSecurityBundleLogin) {
+            foreach ($authenticators as $authenticator) {
+                if (!$authenticator->isNative()) {
```
weaverryan, line 222: I don't like this `isNative()`: we're hardcoding valid "authenticators". Can't we allow the user to choose from *all* possible if there are multiple? Could we get the list of "authenticator names" the same way that the `Security` class does - e.g. however this `authenticators` property is set?

src/Security/InteractiveSecurityHelper.php
```diff
@@ -142,26 +143,36 @@ public function guessPasswordField(SymfonyStyle $io, string $userClass): string
         );
     }
 
+    /**
+     * @return AuthenticatorType[]
+     */
     public function getAuthenticatorClasses(array $firewallData): array
```
weaverryan, line 149: This doesn't really return authenticator "classes" anymore, does it? Let's rename

src/Maker/MakeRegistrationForm.php
```diff
@@ -209,17 +212,36 @@ private function interactAuthenticatorQuestions(InputInterface $input, ConsoleSt
...
-                1 === \count($authenticatorClasses) ? $authenticatorClasses[0] : $io->choice(
-                    'Which authenticator\'s onAuthenticationSuccess() should be used after logging in?',
-                    $authenticatorClasses
-                );
+        // get list of authenticators
+        $authenticators = $interactiveSecurityHelper->getAuthenticatorClasses($firewallsData[$firewallName]);
+
+        $canLoginAfterRegistration = $this->canSecurityBundleLogin;
+
+        if (!$this->canSecurityBundleLogin) {
```
weaverryan, line 220: What about calling the property something like `supportsLoginByAuthenticatorName()`

tests/fixtures/make-registration-form/expected/RegistrationControllerNoLogin.php
jrushlow: i'd say it's safe to remove that now...

src/Security/InteractiveSecurityHelper.php
```diff
@@ -140,22 +143,57 @@ public function guessPasswordField(SymfonyStyle $io, string $userClass): string
...
+                    continue;
+                }
+
+                // custom_authenticators can be set as a string or an array in security.yaml
+                if (\is_string($configData)) {
+                    $configData = [$configData];
+                }
+
+                foreach ($configData as $customAuthenticatorClass) {
+                    //                    if (!class_exists($customAuthenticatorClass)) {
```
jrushlow, line 180: I'm not sure if we _should_ check if the authenticator exists and that it implements `AuthenticatorInterface::class`. Doing so would be a bonus? but it would also require refactoring the test suite to implement a concrete `CustomAuthenticatorFixture::class` Thoughts?

src/Maker/MakeRegistrationForm.php
```diff
@@ -339,14 +339,13 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
                     'from_email' => $this->fromEmailAddress,
                     'from_email_name' => addslashes($this->fromEmailName),
                     'email_getter' => $this->emailGetter,
-                    'authenticator_class_name' => $this->autoLoginAuthenticator ? Str::getShortClassName($this->autoLoginAuthenticator) : null,
-                    'authenticator_full_class_name' => $this->autoLoginAuthenticator,
-                    'firewall_name' => $this->firewallName,
                     'redirect_route_name' => $this->redirectRouteName,
-                    'password_hasher_class_details' => $generator->createClassNameDetails(UserPasswordHasherInterface::class, '\\'),
+                    'password_hasher_class_details' => $hasherDetails = $generator->createClassNameDetails(UserPasswordHasherInterface::class, '\\'),
+                    'password_hasher_variable_name' => sprintf('$%s', lcfirst($hasherDetails->getShortName())),
```
weaverryan, line 344: Does this stuff still need to be dynamic?

src/Security/InteractiveSecurityHelper.php
weaverryan: I'm actually a bit lost in this method. We're looping over the keys under a firewall, right? And the first if statement with `tryFrom` will exit unless the key matches the specific set we have in `AuthenticatorType`, right? So then, `$configData` is represents the values beneath those keys. So if I have:

```
firewalls:
    main:
        form_login:
            path: /login
```

Won't `$configData` at this point be `['path' => '/login']`? Or am I totally misreading the method?

weaverryan (changes requested):

src/Security/InteractiveSecurityHelper.php
```diff
@@ -196,6 +180,96 @@ public function guessIdGetter(SymfonyStyle $io, string $userClass): string
...
+     *
+     * @param array<string, mixed> $firewallConfig
+     *
+     * @return Authenticator[]
+     */
+    private function getAuthenticatorsFromConfigData(array $firewallConfig, string $firewallName): array
+    {
+        $authenticators = [];
+
+        foreach ($firewallConfig as $potentialAuthenticator => $configData) {
```
jrushlow, line 226: ahh... lost the review comment on push...

@weaverryan:
I'm actually a bit lost in this method. We're looping over the keys under a firewall, right? And the first if statement with tryFrom will exit unless the key matches the specific set we have in AuthenticatorType, right? So then, $configData is represents the values beneath those keys. So if I have:
```yaml
firewalls:
    main:
        form_login:
            path: /login
```

Won't $configData at this point be ['path' => '/login']? Or am I totally misreading the method?

jrushlow: no I think you're thinking 1 level too deep..

We pass in `security.firewalls` -> iterate over each key in `firewalls` to give us:
```yaml
main:
    path: '/some-path'
    form_login:
        - option
        ....
    custom_authenticators:
        - some authenticator    
anotherFirewall:
dev:
test:
```

`tryFrom($potentialAuthenticator)` checks if the "key" is `form_login`, 'custom_authenticator`, or `path`. If it _is_ path, `tryFrom()` returns `null` and we `continue`.

Now `$potentialAuthenticator` is either `form_login` or `custom_authenticator` and `$configData` is the array of either of their stuff..

If its `form_login`, we create the value object and add it to `authenticators[]` otherwiser we now know that `$potentialAuthenticator === 'custom_authenticator'` and `configData` is either `(string) MyCustom` or [MyCustom, NotSureWhyYouCanHave2ButYouCan]

src/Security/InteractiveSecurityHelper.php
```diff
@@ -196,6 +180,96 @@ public function guessIdGetter(SymfonyStyle $io, string $userClass): string
         );
     }
 
+    /**
+     * @param array<string, array<string, mixed>> $firewalls Config data from security.firewalls
+     *
+     * @return Authenticator[]
+     */
+    public function getAuthenticatorsFromConfig(array $firewalls): array
```
jrushlow, line 188: `getAuthenticatorsInFirewallsConfig`?

src/Maker/MakeRegistrationForm.php
jrushlow: Dunno, I thought it was a good idea 2 years ago :D I'll check haha

src/Security/InteractiveSecurityHelper.php
weaverryan: Yup, it looks clear now 👍

weaverryan (approved):
If things are working, let's roll with this 👍

PR #1246, opened 2022-11-30
[make:security:json-login] WIP

jrushlow:


src/Resources/test/.php_cs.test
```diff
@@ -8,6 +8,7 @@ return (new PhpCsFixer\Config())
         '@Symfony' => true,
         '@Symfony:risky' => true,
         'native_function_invocation' => false,
+        'blank_line_before_statement' => ['statements' => ['break', 'case', 'continue', 'declare', 'default', 'do', 'exit', 'for', 'foreach', 'goto', 'if', 'include', 'include_once', 'phpdoc', 'require', 'require_once', 'return', 'switch', 'throw', 'try', 'while', 'yield', 'yield_from']],
```
jrushlow, line 11: we need `phpdoc` to be added... TODO: check whats included in symfony rules set and only add the one we need.

jrushlow: im using are test config here because, well lazy is cool, but thinking we either duplicate this for generating templates or rename / move this away from "test" and still use it for both ci and generating templates

src/Maker/Security/AbstractSecurityMaker.php
```diff
@@ -0,0 +1,92 @@
...
+        protected FileManager $fileManager,
+        protected SecurityConfigUpdater $securityConfigUpdater,
+        protected SecurityControllerBuilder $securityControllerBuilder,
+    ) {
+    }
+
+    protected function runFixer(string $templateFilePath): void
+    {
+        $fixerPath = \dirname(__DIR__, 2).'/bin/php-cs-fixer-v3.13.0.phar';
+        $configPath = \dirname(__DIR__, 2).'/Resources/test/.php_cs.test';
```
weaverryan, line 57: We're probably going to want to use the user's php cs config immediately, if it's available. For starters, we could look in the root dir for whatever the common filenames are. Then fallback to using our version.

Also, why not start doing this for *all* maker commands instead of security? It's a bit opinionated, but I think it's a good idea.

Finally, this would definitely be done better as an independent service, though I think I see why you avoided doing that: it's difficult to inject a service into *all* makers. It makes me wonder (even though I'm the king of "don't add extension points), if we might be better served dispatching an event at the end of running a maker - like right here - https://github.com/symfony/maker-bundle/blob/7c9e6968895c05c2757c08e63240eaba9e2b882e/src/Command/MakerCommand.php#L93 - then, we could ideally fetch whatever files were modified from the `Generator`. If those files end in `.php`, we php-cs-fix them.

jrushlow: done in #1251

tests/fixtures/security/make-form-login/expected/LoginController.php
```diff
@@ -15,13 +15,10 @@ public function login(AuthenticationUtils $authenticationUtils): Response
         // get the login error if there is one
         $error = $authenticationUtils->getLastAuthenticationError();
 
-        // last username entered by the user
+        /** last username entered by the user */
         $lastUsername = $authenticationUtils->getLastUsername();
```
jrushlow, line 19: php-cs-fixer does not have the capability to add a space *above* a comments (`//`) but it can do it with docblocks (`/** */`)

Without changing the comments over, we would end up with:
```php
        // get the login error if there is one
        $error = $authenticationUtils->getLastAuthenticationError();
        // last username entered by the user
        $lastUsername = $authenticationUtils->getLastUsername();
```

which could be alright?

src/Maker/Security/AbstractSecurityMaker.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->ysm = new YamlSourceManipulator($this->fileManager->getFileContents(self::SECURITY_CONFIG_PATH));
+        $securityData = $this->ysm->getData();
+
+        $securityHelper = new InteractiveSecurityHelper();
+        $this->firewallToUpdate = $securityHelper->guessFirewallName($io, $securityData);
+        $this->userClass = $securityHelper->guessUserClass($io, $securityData['security']['providers']);
+        $this->userNameField = $securityHelper->guessUserNameField($io, $this->userClass, $securityData['security']['providers']);
+        $this->willLogout = $io->confirm('Do you want to generate a \'/logout\' URL?');
+    }
+}
```
jrushlow, line 83: We could probably find more repeated code between `make:json` & `make:form` logins if we looked hard enough...

src/Maker/Security/MakeJsonLogin.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $userClassDetails = new ClassNameDetails($this->userClass, '');
+
+        $useStatements = new UseStatementGenerator([
+            $userClassDetails->getFullName(),
+            \Symfony\Bundle\FrameworkBundle\Controller\AbstractController::class,
+            \Symfony\Component\HttpFoundation\JsonResponse::class,
+            \Symfony\Component\HttpFoundation\Response::class,
+            \Symfony\Component\Routing\Annotation\Route::class,
+            \Symfony\Component\Security\Http\Attribute\CurrentUser::class,
```
jrushlow, line 59: These should be imports...

tests/fixtures/security/make-json-login/LoginTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ */
+class LoginTest extends WebTestCase
+{
+    public function testJsonLogin(): void
+    {
+        $client = static::createClient();
+        $client->xmlHttpRequest('POST', '/api/login');
```
jrushlow, line 16:
```suggestion
        $client->jsonRequest('POST', '/api/login');
```

src/Maker/Security/AbstractSecurityMaker.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        if (!$this->fileManager->fileExists(self::SECURITY_CONFIG_PATH)) {
+            throw new RuntimeCommandException(sprintf('The file "%s" does not exist. PHP & XML configuration formats are currently not supported.', self::SECURITY_CONFIG_PATH));
+        }
+
+        $this->securityControllerName = $io->ask(
+            'Choose a name for the controller class (e.g. <fg=yellow>ApiLoginController</>)',
+            'ApiLoginController',
+            [Validator::class, 'validateClassName']
+        );
```
weaverryan, line 72: This question and the suggestion of `ApiLoginController` are in the abstract security maker?

src/Maker/Security/AbstractSecurityMaker.php
```diff
@@ -0,0 +1,83 @@
...
+        protected FileManager $fileManager,
+        protected SecurityConfigUpdater $securityConfigUpdater,
+        protected SecurityControllerBuilder $securityControllerBuilder,
+    ) {
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies): void
+    {
+        $dependencies->addClassDependency(SecurityBundle::class, 'security');
+        $dependencies->addClassDependency(Process::class, 'process');
```
weaverryan, line 57: Is this actually needed?

src/Maker/Security/MakeJsonLogin.php
```diff
@@ -0,0 +1,106 @@
...
+
+        if (!$controllerExists) {
+            $generator->generateController(
+                $controllerNameDetails->getFullName(),
+                'EmptyController.tpl.php',
+                [
+                    'use_statements' => $useStatements,
+                    'controller_name' => $controllerNameDetails->getShortName(),
+                ]
+            );
```
weaverryan, line 76: This could be a new `generateEmptyController()` on `Generator` 🤔

src/Maker/Security/MakeJsonLogin.php
```diff
@@ -0,0 +1,106 @@
...
+
+        $generator->dumpFile(self::SECURITY_CONFIG_PATH, $securityData);
+        $generator->dumpFile($controllerPath, $manipulator->getSourceCode());
+
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+
+        $io->text([
+            'Next: Make a <info>POST</info> request to <info>/api/login</info> with a <info>username</info> and <info>password</info> to login.',
```
weaverryan, line 101: The user also needs to set `Content-Type: application/json` iirc. This is sometimes a "gotcha", so it would be nice to mention it here.

src/Resources/skeleton/security/formLogin/_LoginMethodBody.tpl.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
weaverryan, line 1: The `_L` (capital `L`) looks weird, but I can see you did that on purpose. What was the reason?

src/Security/SecurityControllerBuilder.php
```diff
@@ -80,4 +87,45 @@ public function addLogoutMethod(ClassSourceManipulator $manipulator): void
...
     }
+
+    public function addFormLoginMethod(ClassSourceManipulator $manipulator, string $loginTemplatePath): void
+    {
+        $methodBuilder = $manipulator->createMethodBuilder('login', Response::class, false);
+        $methodBuilder->addAttribute($manipulator->buildAttributeNode(Route::class, ['path' => '/login', 'name' => 'app_login']));
+
+        $this->addUseStatements($manipulator, [Route::class]);
+
+        $methodBuilder->addParam((new Param('authenticationUtils'))->setType('AuthenticationUtils'));
```
weaverryan, line 98: Will there already be a `use` statement for this? In `addLoginMethod()`, we explicitly add the use statement.

src/Security/SecurityControllerBuilder.php
```diff
@@ -80,4 +87,45 @@ public function addLogoutMethod(ClassSourceManipulator $manipulator): void
...
+
+        $methodBuilder->addAttribute($manipulator->buildAttributeNode(Route::class, ['path' => '/api/login', 'name' => 'app_api_login']));
+
+        $this->addUseStatements($manipulator, [Route::class, $userClass->getFullName(), CurrentUser::class]);
+
+        $methodBuilder->addParam(
+            (new Param('user'))
+                ->setType(new NullableType($userClass->getShortName()))
+                ->addAttribute(new Attribute(new Name('CurrentUser')))
+        );
```
weaverryan, line 118: This probably needs to change to `$this->getUser()` for better "portability" across all versions of Symfony, DoctrineBundle, FWEXtra, etc?

weaverryan (commented):
Looking great!

src/Security/SecurityControllerBuilder.php
jrushlow: Should we bother making that conditional based on the Symfony Version? Or perhaps we do that later when say 6.4 or 7 come out for less of a headache for us

PR #1247, opened 2022-12-01
Remove deprecated test maker 

BASAKSemih:


BASAKSemih:
deprecated since MakerBundle 1.29, use Symfony\Bundle\MakerBundle\Maker\MakeTest instead.

jrushlow (changes requested):
Howdy @BASAKSemih - Thank you for taking the time to create this PR. Unfortunately, we can do this just yet. To remove either of these methods would be a BC break as they're not marked as `@internal`. We have to wait to do this in `v2.x` of MakerBundle.

PR #1251, opened 2022-12-06
[make:*] use php-cs-fixer to style/lint all generated php templates

jrushlow:
We use `php-cs-fixer` to lint generated php files created by all makers. We use the following precedents for determining which php-cs-fixer `config` and `binary` to use:

## PHP CS Fixer Config
1) `MAKER_PHP_CS_FIXER_CONFIG_PATH` env is set, we use it
2) If `root-project-path/.php-cs-fixer.dist.php` exists, we use it 
3) We use the bundled config from `Resources/config/php-cs-fixer.config.php`

## PHP CS Fixer Binary
1) `MAKER_PHP_CS_FIXER_BINARY_PATH` env is set, we use it
2) We use the bundled php-cs-fixer phar located in `src/Bin/php-cs-fixer-v3.13.0.phar`

An exception is thrown if either of the env var's above are set, and the file does not exist.

Passing `-v` to the command will show which `php-cs-fixer` binary and configuration file maker is using.

Internally, we no longer run `php-cs-fixer` in every `make:*` test. Doing so now would basically test `php-cs-fixer` itself is working - not our business... Lets save time in each test...


README.md
```diff
@@ -25,3 +25,22 @@ B) The generated code itself may change between minor releases. This
...
 [3]: https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html
 [4]: https://symfony.com/doc/current/contributing/code/bc.html
+
+## Linting Generated Templates
+
+### PHP-CS-FIXER
+
+MakerBundle uses php-cs-fixer to enforce coding standards when generating `.php`
+files. If you have `friendsofphp/php-cs-fixer` added to your project, we'll
+use the `bin/php-cs-fixer` binary and `.php-cs-fixer.dist.php` configuration file
```
weaverryan, line 35:
```suggestion
look for and use the `bin/php-cs-fixer` binary and `.php-cs-fixer.dist.php` configuration file
```

weaverryan: Wouldn't/shouldn't it be `vendor/bin/php-cs-fixer`?

Also, since installing php-cs-fixer into your composer.json isn't a best practice, I'm not sure we should support this. We *could* look to see if there is a global `php-cs-fixer` binary available and use that.

In all cases, if we're using their binary or their config file (regardless of if it's automatic or because they set env vars), it might be a good idea to print a small note at the top of the command's output about this.

README.md
```diff
@@ -25,3 +25,22 @@ B) The generated code itself may change between minor releases. This
...
+automatically.
+
+Otherwise, we use the `src/Bin/php-cs-fixer-v3.13.0.phar` and
```
weaverryan, line 38:
```suggestion
Otherwise, we use a `php-cs-fixer` version that is packaged with this bundle.

No need to mention the version.

Also `php-cs-fixer` probably shouldn't be in `src/Bin` but just maybe `bin/` in our repo :). Or maybe `src/Resources/... something`
```

README.md
```diff
@@ -25,3 +25,22 @@ B) The generated code itself may change between minor releases. This
...
+`src/Resources/config/php-cs-fixer.config.php` that are included with MakerBundle.
+
+You can explicitly set a custom path to the `php-cs-fixer` binary and/or a 
+php-cs-fixer config file by explicitly setting their respective environment
```
weaverryan, line 42:
```suggestion
php-cs-fixer config file by explicitly their respective environment
```

src/Generator.php
```diff
@@ -69,6 +70,8 @@ public function generateClass(string $className, string $templateName, array $va
 
         $this->addOperation($targetPath, $templateName, $variables);
 
+        $this->generatedPhpFiles[] = $targetPath;
+
```
weaverryan, line 74: We put this here, but not in, for example, `generateFile()`? Perhaps this should be done in `writeChanges()`? Each time we process a pending operation, we add it to the array. We could filter it later for files that end in `.php`.

src/Util/TemplateLinter.php
```diff
@@ -0,0 +1,92 @@
...
+
+        $defaultBinaryPath = 'bin/php-cs-fixer';
+
+        if (file_exists($defaultBinaryPath)) {
+            $this->phpCsFixerBinaryPath = $defaultBinaryPath;
+
+            return;
+        }
+
+        $this->phpCsFixerBinaryPath = \dirname(__DIR__).'/Bin/php-cs-fixer-v3.13.0.phar';
```
weaverryan, line 62: Any thoughts on how we could not forget to routinely update this?

weaverryan (commented):
Cool 😎

README.md
jrushlow: When you install via composer e.g. (composer require --dev ......), php-cs-fixer installs the binary in the `path-to-project/bin` folder https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/a2bdba33ddb7df1cedce536c3e9c3db8d73ddd3e/composer.json#L65

Looking at their readme, https://github.com/PHP-CS-Fixer/PHP-CS-Fixer#installation we could check for the binary in `tools/php-cs-fixer` instead/in addition to the above.

I didn't even think about php-cs-fixer installed globally - I'm wondering if we should add this bit in after the fact due to the different ways that the binary can be installed (manually, via composer, phive, and brew). I'd have to do some playing around to see if there are any differences in the `PATH` vars for us to use. e.g. installing manually (globally) the user has to:

- Download composer -> move it to `/usr/local/bin` on linux -> ensure thats in their system `PATH`

But you can also install globally via composer by `composer global install.... `

I'm thinking we could have a more robust method of checking for the binary using the autoloader -> checking the global path(s). 

Thoughts?

> In all cases, if we're using their binary or their config file (regardless of if it's automatic or because they set env vars), it might be a good idea to print a small note at the top of the command's output about this.

Agree'd

README.md
jrushlow: moved to `Resources/bin`

src/Util/TemplateLinter.php
jrushlow: post-it

jrushlow: Aside from post it notes, thinking we have a new ci job the runs daily to check for the latest version and compares it to our `php-cs-fixer.VERSION.php` bin file name... I don't think we want this running on every PR/push as that we result in a bunch of rebases over time?

README.md
jrushlow: We also need to figure out how to write tests for the global installs as I don't think that is something we can do right out of the box now - but I could be mistaken.

weaverryan: I think we can simply use this class: https://github.com/symfony/process/blob/6.2/ExecutableFinder.php

If we find `php-cs-fixer`, cool. Use it. Else, don't. It's so simple I don't think we'd need any additional tests.

weaverryan: > php-cs-fixer installs the binary in the path-to-project/bin folder https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/a2bdba33ddb7df1cedce536c3e9c3db8d73ddd3e/composer.json#L65 (

I think that means "install `php-cs-fixer` into whatever Composer's `bin` directory is", which is `vendor/bin` by default.

src/Util/TemplateLinter.php
weaverryan: LOL

Maybe a monthly CI job would be awesome if it's not too much work - one that could create a PR with the change would be ideal.

README.md
jrushlow: Executable finder works for me...

image

src/Generator.php
jrushlow: Good call and done - this will make it easier to implement the twig linter in a future pr.

src/Generator.php
```diff
@@ -204,7 +189,7 @@ public function writeChanges()
 
             $this->fileManager->dumpFile(
                 $targetPath,
-                $this->getFileContentsForPendingOperation($targetPath, $templateData)
+                $this->getFileContentsForPendingOperation($targetPath)
```
jrushlow, line 192: Unused var unrelated to this pr

tests/Maker/MakeMigrationTest.php
```diff
@@ -113,7 +113,7 @@ public function getTestDetails(): \Generator
 
                 $this->assertStringContainsString('You have 1 available migrations to execute', $output);
                 $this->assertStringContainsString('Success', $output);
-                $this->assertCount(14, explode("\n", $output), 'Asserting that very specific output is shown - some should be hidden');
+                $this->assertCount(17, explode("\n", $output), 'Asserting that very specific output is shown - some should be hidden');
```
jrushlow, line 116: We've added additional lines to the output. This test should be changed in a separate PR to assert exactly what we're looking for rather than total number of lines. - Todo added on a post-it

jrushlow: https://github.com/symfony/maker-bundle/pull/1254

src/Resources/doc/index.rst
```diff
@@ -43,6 +43,21 @@ optional arguments and options. Check them out with the ``--help`` option:
...
+
+MakerBundle uses php-cs-fixer to enforce coding standards when generating ``.php``
+files. When running a ``make`` command, MakerBundle will use a ``php-cs-fixer``
+version and configuration that is packaged with this bundle.
+
+You can explicitly set a custom path to a php-cs-fixer binary and/or configuration
+file by their respective environment variables:
+
+- ``MAKER_PHP_CS_FIXER_BINARY_PATH`` e.g. /path/to/project/php-cs-fixer.php
+- ``MAKER_PHP_CS_FIXER_CONFIG_PATH`` e.g. /path/to/project/php-cs-fixer.config.php
```
weaverryan, line 58: Can I set this to just a path in my app? That'd be much more useful - e.g.

```
MAKER_PHP_CS_FIXER_CONFIG_PATH=.php-cs-fixer.dist.php
```

Same for the binary - can I just say `MAKER_PHP_CS_FIXER_BINARY_PATH=php-cs-fixer` to use a globally-installed one?

Also, I think using `.php-cs-fixer.dist.php` in the example would be more realistic.

src/Util/TemplateLinter.php
```diff
@@ -0,0 +1,127 @@
...
+            $process = Process::fromShellCommandline(sprintf('php %s --config=%s --using-cache=no fix %s', $this->phpCsFixerBinaryPath, $this->phpCsFixerConfigPath, $filePath));
+
+            $process->run();
+        }
+    }
+
+    public function getLinterUserMessage(): array
+    {
+        $message = [];
+        $message[0] = 'Linting Generated Files With:'.\PHP_EOL;
```
weaverryan, line 66: Do we need the EOL? I think if we pass this to `$output->writeln()` then each item in the array becomes a new line.

src/Command/MakerCommand.php
```diff
@@ -90,13 +95,17 @@ protected function interact(InputInterface $input, OutputInterface $output): voi
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
+        $output->write($this->linter->getLinterUserMessage());
```
weaverryan, line 98: My gut reaction at this moment is: let's only show this if `$output` is in verbose mode or higher. Will be a lot of noise... for something where I think 90+% of people will be happy enough with the output.

weaverryan (commented):
Minor comments - looks awesome!

src/Command/MakerCommand.php
jrushlow: agree'd - unfortunately I can test for this... `$output->isVerbose()` is always true during the tests. I think this may having something to do with bridge and how we get deprecations? But im not sure.. Didn't spend a whole lot of time digging into it. 

Locally, this works as expected.. pass `-v` and we get the message...

weaverryan (approved):

k0d3r1s:
Hey. Any reason this is not in any tagged version yet?

weaverryan:
We're just behind on MakerBundle right now due to some "in real life" stuff. Hopefully we can catch up soon :)

PR #1253, opened 2022-12-12
[make:migration] Add link to new migration files

nicolas-grekas:
Makes the output of make:migration more consistent with other commands.

Before:
image

After:
image


jrushlow (changes requested):
Thank you @nicolas-grekas - Any chance you could update the message assertion to get the related tests to pass: 
https://github.com/symfony/maker-bundle/blob/main/tests/Maker/MakeMigrationTest.php#L65

I think if we remove the `"` from `migration/version.......php` we should be all set. 

Side note: `MakeMigrationTest::it_asks_previous_migration_question` failure is unrelated to this PR

jrushlow (approved):

PR #1260, opened 2022-12-28
EntityRegenerator - no need to set annotations in tests

jrushlow:
We removed entity annotation support in https://github.com/symfony/maker-bundle/pull/1126

### DIFF
--- a/src/Util/ClassDetails.php
+++ b/src/Util/ClassDetails.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Bundle\MakerBundle\Util;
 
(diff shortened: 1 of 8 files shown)


src/Util/ClassDetails.php
```diff
@@ -61,6 +63,7 @@ public function getPath(): string
      * An imperfect, but simple way to check for the presence of an annotation.
      *
      * @param string $annotation The annotation - e.g. @UniqueEntity
+     * @deprecated We dont support anntations anymore..
```
OskarStark, line 66: Typo in annotations

jrushlow: Story of my life :smile: Thanks!

PR #1265, opened 2022-12-29
[make:crud] Make sensio/framework-extra-bundle an optional dependency

acrobat:
The `sensio/framework-extra-bundle` is abandoned and replaced by built-in feature of symfony components. So this dependency is not required when using symfony 6.2+.

I've tested the crud maker on a new symfony 6.2 project without the framework-extra-bundle and the controller typehints work correctly

See: https://symfony.com/blog/new-in-symfony-6-2-built-in-cache-security-template-and-doctrine-attributes


jrushlow (approved):
Thank you @acrobat - that's one less bit of bloat in new apps using maker!

PR #1267, opened 2023-01-03
[make:registration-form] use route attributes for test routes

jrushlow:
IIRC we used a `yaml` router config in tests to avoid recipe changes / `doctrine/annotations` problems - this should no longer be a problem as we don't support annotations in maker. So let's be a bit more realistic in our tests and just use route attributes...


jrushlow:
Test failures are as expected...

PR #1270, opened 2023-01-06
[make:authenticator] Core\Security or SecurityBundle\Security - Avoid deprecations in 6.2

nacorp:
When using make:authenticator on a Symfony 6.2 project the generated code is deprecated.

In the generated code, `use Symfony\Component\Security\Core\Security;` should be replace by  `use Symfony\Bundle\SecurityBundle\Security;`.

However, in order to respect the compatibility with 5.4 I used a workaround.


weaverryan (approved):

welcoMattic (approved):

weaverryan:
Thanks Nicolas!

PR #1272, opened 2023-01-13
[make:crud] render or renderForm - Avoid deprecations in 6.2

nacorp:
The old `use_render_form` is always true, in both Symfony 5.4 and 6.2 so we always end un with a `$this->renderForm() ` that has been deprecated in 6.2.

This (ugly, but nevertheless working) hack tests if `getDoctrine` exists in AbstractController allows us to determine Symfony's version and if we should generate a `$this->render()` or a `$this->renderForm()`.

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -70,7 +70,8 @@ final class MakeCrud extends AbstractMaker
     {
         $command
             ->addArgument('entity-class', InputArgument::OPTIONAL, sprintf('The class name of the entity to create CRUD (e.g. <fg=yellow>%s</>)', Str::asClassName(Str::getRandomTerm())))
-            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeCrud.txt'));
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeCrud.txt'))
+        ;
 
         $inputConfig->setArgumentAsNonInteractive('entity-class');
     }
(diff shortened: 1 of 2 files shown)


src/Maker/MakeCrud.php
```diff
@@ -179,7 +179,8 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
                     'entity_var_singular' => $entityVarSingular,
                     'entity_twig_var_singular' => $entityTwigVarSingular,
                     'entity_identifier' => $entityDoctrineDetails->getIdentifier(),
-                    'use_render_form' => method_exists(AbstractController::class, 'renderForm'),
+                    // @legacy - Remove when support for Symfony <6 is dropped - hack to determine version
+                    'use_render_form' => method_exists(AbstractController::class, 'getDoctrine'),
```
welcoMattic, line 183: Please use `Kernel::VERSION` to determine version

weaverryan: Agreed. We normally like to check directly for the feature we're using. But in this case, as you @nacorp mentioned, that's not possible. So let's just check the kernel version :)

nacorp: Here we go !

welcoMattic (approved):

src/Maker/MakeCrud.php
```diff
@@ -179,7 +180,8 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
                     'entity_var_singular' => $entityVarSingular,
                     'entity_twig_var_singular' => $entityTwigVarSingular,
                     'entity_identifier' => $entityDoctrineDetails->getIdentifier(),
-                    'use_render_form' => method_exists(AbstractController::class, 'renderForm'),
+                    // @legacy - Remove when support for Symfony <6 is dropped
+                    'use_render_form' => Kernel::VERSION_ID < 60000,
```
weaverryan, line 184: I believe it should be:

```suggestion
                    'use_render_form' => Kernel::VERSION_ID < 62000,
```

MakerBundle supports 5.4 and higher only, and `renderForm()` was added in 5.3. So:

* 5.4 - 6.1, we should use `renderForm()`
* In 6.2 & higher, use `render()`, always without `createView()`

Issue #1252, opened 2022-05-27
[Serializer] Custom Normalizer broken after upgrading to 6.1


alexchuin:
### Symfony version(s) affected

6.1

### Description

My custom Normalizer is broken since i've upgraded to Symfony 6.1.
This problem seems to be related to the new Serializer Profiler.

Maybe I forgot something?

Thanks,
Alex

### How to reproduce

Just use this example :

https://symfony.com/doc/6.1/serializer/custom_normalizer.html

### Additional Context

`FileNormalizer::__construct(): Argument symfony/symfony#2 ($normalizer) must be of type Symfony\\Component\\Serializer\\Normalizer\\ObjectNormalizer, Symfony\\Component\\Serializer\\Debug\\TraceableNormalizer given`

ivoba:
@lyrixx your link does not link a comment :)

I followed this: https://github.com/symfony/maker-bundle/issues/1252#issuecomment-1342478104

    App\Serializer\CustomDenormalizer:
        autowire: false
        arguments:
            $denormalizer: '@serializer.normalizer.object

So i suppose current documentation is wrong?

kissifrot:
Well disabling autowiring can be quite cumbersome if there are many other services injected in the custom normalizer, is there a workaround for for workaround? :stuck_out_tongue:

mtarld:
I think you still can let autowiring enabled and override only the denormalizer argument

lyrixx:
This works well
https://github.com/symfony/maker-bundle/issues/1252#issuecomment-1342477712 

Or use the NormalizeAwareInterface

ivoba:
@lyrixx i think that NormalizeAwareInterface is *not*  working is the actual problem.
Seems that the autowiring is broken with a circular reference problem.

This is why the most comments here recommed to disable autowire as a solution.

lyrixx:
For me it's was OK, since it's a setter.

aiibileme:
When ı added a comment at collect_serializer_data on web_profiler.yaml, it worked.

`when@dev:
    web_profiler:
        toolbar: true
        intercept_redirects: false

    framework:
        profiler:
            only_exceptions: false
            collect_serializer_data: false
`

gansky-alexander:
Hello, all ! 

I tried all workarounds and them are not working for me.
Do you have any information about this bug when are you going to fix it ?

What issues i faced:
- circular reference on service registration
- endless loading of the controller

mtarld:
[...]

PR #1273, opened 2023-01-19
[make:serializer:normalizer] Inject a NormalizerInterface instead of an ObjectNormalizer

mtarld:
Related to https://github.com/symfony/maker-bundle/issues/1252

This PR aims to do not rely anymore on constructor argument injection and use `NormalizerAwareInterface`/`NormalizerAwareTrait` instead.

Moreover, this allows to not inject explicitly the `ObjectNormalizer` anymore which can be considered as a bad practice as it is a concrete implementation of the `NormalizerInterface`.

### DIFF
--- a/src/FileManager.php
+++ b/src/FileManager.php
@@ -194,6 +194,7 @@ class FileManager
         }
 
         $finalPath = implode('/', $finalParts);
+
         // Normalize: // => /
         // Normalize: /./ => /
         return str_replace(['//', '/./'], '/', $finalPath);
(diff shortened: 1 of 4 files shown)


mtarld:
Actually, after talking with @chalasr, we went to the conclusion that it is better to inject the `ObjectNormalizer` through the constructor.

Therefore, this PR was updated to:
- inject a `NormalizerInterface` in the normalizer constructor
- update the service definition in order to inject explicitly the `ObjectNormalizer`

chalasr (approved):
👍 NormalizerAwareTrait seems to not be much appreciated by the community, and I tend to agree - it feels complicated. Also using it implies to deal with some special context flag to workaround an infinite loop. Regular DI looks good!

jrushlow (approved):
This looks good! Thank you @mtarld

PR #1286, opened 2023-02-08
Changed the package names from alias Flex to the full name from composer on error messages.

bdaler:
Changed the package names from alias Flex to the full name from composer. Requires a thorough code review :)

### DIFF
--- a/src/Util/DependencyPackages.php
+++ b/src/Util/DependencyPackages.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Created by PhpStorm.
+ * User: Daler B. (@bdaler)
+ * Date: 08.02.2023
+ * Time: 21:18.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+class DependencyPackages
+{
+    public const SECURITY_BUNDLE = 'symfony/security-bundle';
+    public const YAML = 'symfony/yaml';
+    public const CONSOLE = 'symfony/console';
+    public const ROUTING = 'symfony/routing';
+    public const FORM = 'symfony/form';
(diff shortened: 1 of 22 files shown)


src/Util/DependencyPackages.php
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Created by PhpStorm.
+ * User: Daler B. (@bdaler)
+ * Date: 08.02.2023
+ * Time: 21:18.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Util;
+
+class DependencyPackages
```
bdaler, line 14: "I just thought, maybe it could be a new enum in PHP 8.1?"

stof: this would be a bad usage of an enum. Please using those constants need a string.

Btw, I'm not even sure those constants provide much benefit.

bdaler: @stof Hi. The essence of using constants is that, for example, the `orm` package is used in 4 or 5 files. In the future (of course, this is unlikely, but nevertheless), if it is necessary to change this, it will have to be done in 4 or 5 different places. But with constants, you can make the change in one place and it will apply to all.

sadikoff: Really Enum is not a variant, however class is not a good way too I think, it will be better to use `Interface` as constant keeper, and also it will be better to keep original package names (aliases used in code) not real packages names.

jrushlow:
Howdy @bdaler! I'm not sure this is a good idea in this particular use case. the string's used are passed to `composer` directly, & while I'm generally all for `const`'s & `enums`; using them here _reduces_ the clarity (dx) in the packages we're actually requiring. 

In short, its adds an extra step for maintainers to figure out exactly what packages we're grabbing from composer for tests and their makers.

PR #1290, opened 2023-02-22
[make:registration-form] avoid using internal method get from Request

Romaixn:
When launching the `make:registration-controller`, there is a `$request->get('id')` in the generated controller.

Replace `$request->get('id')` with `$request->query->get('id')` in the RegistrationController template to avoid using the Symfony internal method.


weaverryan (commented):
Definitely.  I'm not sure about the tests - they're broken right now - so it's hard to know if you need to update any tests. But definite 👍 for this change

PR #1293, opened 2023-03-15
[make:entity] don't set array field default value for nullable column

Rootie:
Setting a default value will lead to a wrong (not nullable) return value of the getter method.

Also setting a default value on a nullable field somehow cancels out the intention of nullable (except for setting it explicitly).

I'm not sure if this should be treated as a bc break. but since no existing properties are overwritten i think it should be fine.

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -110,7 +110,7 @@ final class ClassSourceManipulator
         $attributes[] = $this->buildAttributeNode(Column::class, $columnOptions, 'ORM');
 
         $defaultValue = null;
-        if ('array' === $typeHint) {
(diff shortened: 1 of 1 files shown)


sadikoff:
Hi @Rootie,

Can you please update the description and PR so it can be reviewed?

Cheers!

Rootie:
Hi @sadikoff,

i updated the description.

I also wanted to note that if keeping the default value for array fields is wanted, then it is also possible to just modify the addGetter call in Util\ClassSourceManipulator::addEntityField

weaverryan:
This makes sense to me. Could you rebase so we can check the tests?

Rootie:
Hi @weaverryan,

i updated my branch

weaverryan:
Thanks @Rootie!

PR #1295, opened 2023-03-23
[make:authenticator] Add missing RememberMeBadge

matthieumota:
Hi,

I see that RememberMeBadge is missing on LoginForm guard that is generated with `make:authenticator`. It can be usefull to add badge if checkbox is uncommented and checked.


matthieumota:
I think #986 has a better version. No need to check parameter, it is always checked with Badge.

weaverryan:
Other PR is now merged - thanks!

PR #1302, opened 2023-04-27
MakeEntity: Removing parentheses in field types overview

ThomasLandauer:
Reason: In `date (or date_immutable)`, the second option looks inferior/exotic. So I changed it to look more equal: `date or date_immutable`


weaverryan:
That works for me - thanks Thomas!

PR #1308, opened 2023-05-23
[make:user] add the void return type to eraseCredentials()

jrushlow:
In 6.3 tests - handle:
```
Method "Symfony\Component\Security\Core\User\UserInterface::eraseCredentials()" might add "void" 
as a native return type declaration in the future. Do the same in implementation "App\Entity\User" now
 to avoid errors or add an explicit @return annotation to suppress this message.
```


weaverryan:
Thanks Jesse!

PR #1309, opened 2023-05-23
Apply `get_class_to_class_keyword` PHP-CS-Fixer rule

seb-jean:


jrushlow (changes requested):
Shweet! To be consistent, we should do the same in https://github.com/symfony/maker-bundle/blob/main/src/Resources/skeleton/security/UserProvider.tpl.php#L49

seb-jean:
> Shweet! To be consistent, we should do the same in https://github.com/symfony/maker-bundle/blob/main/src/Resources/skeleton/security/UserProvider.tpl.php#L49

Done

jrushlow (approved):
Thanks @seb-jean

PR #1312, opened 2023-05-23
[make:entity] better repo methods

jrushlow:
- repositories are generated with `persist(...)`, `persistAndFlush(...)`, `remove(...)`, `removeAndFlush(...)` & `flush()` methods.

- legacy `save(...)` & `remove(...)` methods _may_ be generated instead by passing the `--legacy-methods` option to `make:entity`

The legacy option is used to provide consistency in projects that already have repositories with the `save()` && `remove()` methods. However, we strongly encourage projects not to use `--legacy-methods` as it is deprecated and will be removed in a future version of MakerBundle.

The legacy option is _not_ available with `make:user` or `make:reset-password` to improve code base maintainability.

### DIFF
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -35,7 +35,7 @@ final class EntityClassGenerator
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false): string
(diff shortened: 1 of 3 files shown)


jrushlow:
We could generate these methods in a `RepositoryMethodsTrait::class` and import them in each repo. This would keep the generated repos DRY and essentially empty of code. The downside would be we xor the user would have to handle any static analysis complaints. The method signatures would be something like `RepositoryMethodsTrait::persistAndFlush(object $entity): void`.

Thoughts?

sadikoff:
what if propose this trait for `DoctrineBundle`? I'm not sure that generating these methods is a good plan?

PR #1313, opened 2023-05-23
Fix crash on user class without properties

jaroslavtyc:
End builder properly when no user class properties as choices are available.

---

I had this User class

```php
class EmptyUser implements UserInterface
{
    public function getRoles(): array
    {
        return [];
    }

    public function eraseCredentials()
    {
        // TODO: Implement eraseCredentials() method.
    }

    public function getUserIdentifier(): string
    {
        return '';
    }

}
```

and `make:reset-password` ended with generic critical error

```
 What is the User entity that should be used with the "forgotten password" feature? (e.g. App\Entity\User) []:
 > \Gamecon\Symfony\Entity\EmptyUser

[critical] Error thrown while running command "make:reset-password". Message: "Choice question must have at least 1 choice available."

In ChoiceQuestion.php line 36:
                                                          
  Choice question must have at least 1 choice available.                                                        
```

Reason is a missing check that given user class has some properties, used later as choices.

### DIFF
--- a/src/Security/InteractiveSecurityHelper.php
+++ b/src/Security/InteractiveSecurityHelper.php
@@ -89,6 +89,17 @@ final class InteractiveSecurityHelper
             return 'username';
         }
 
+        $classProperties = $this->getClassProperties($userClass);
+
+        return $io->choice(
+            sprintf('Which field on your <fg=yellow>%s</> class will people enter when logging in?', $userClass),
+            $classProperties,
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks for this! And sorry for the slow review. With this fix, the user would still encounter an error, correct? It would just be a better error (`No properties were found in "%s" entity`)?

jaroslavtyc:
> Thanks for this! And sorry for the slow review. With this fix, the user would still encounter an error, correct? It would just be a better error (`No properties were found in "%s" entity`)?

Exactly

PR #1317, opened 2023-06-05
Add cascade annotation option support in ClassSourceManipulator

trislem:
Allow adding the **cascade** annotation option for OneToMany, ManyToOne and ManyToMany relation in ClassSourceManipulator.
The OneToOne remains forced to ['persist', 'remove']


trislem:
Is the CI error related to the commit ?

PR #1324, opened 2023-06-13
Fix(Doctrine Repository template)/Avoid potential double call in save method

Mano-Lis:
Store the entity manager in a variable rather than calling getEntityManager() twice in case of persist + flush


Mano-Lis:
@dunglas

dunglas (approved):

weaverryan:
Ok for me! FYI - we are planning to remove these methods, but until we do, this change is good. Thanks Emmanuel!

PR #1328, opened 2023-06-24
[WCM] Deleting `save` and `remove` methods from repositories

### LINKED ISSUE #1327: Do not flush in the repository
ATM, the generated code is:
```php
    public function add(Message $entity, bool $flush = true): void
    {
        $this->_em->persist($entity);
        if ($flush) {
            $this->_em->flush();
        }
    }
```

This is a bad idea:
* it's confusing: one may thing it will flush only the new message (it's not!)
* flush should be done in the most outside layer of the application (controller, command, message handler) to avoid flushing invalid data

mdoutreluingne:
Fix #1327

In this PR I've removed the `save` and `remove` methods from the repositories, for the reasons explained here #1327.

( The add method has been renamed to save, here #1204 )

Reviews welcome!

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -163,6 +163,7 @@ final class MakeCrud extends AbstractMaker
             Request::class,
             Response::class,
             Route::class,
+            EntityManagerInterface::class,
         ]);
 
         $generator->generateController(
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -163,8 +163,11 @@ final class MakeCrud extends AbstractMaker
             Request::class,
             Response::class,
             Route::class,
-            EntityManagerInterface::class,
         ]);
+
+        if (EntityManagerInterface::class !== $repositoryClassName) {
+            $useStatements->addUseStatement(EntityManagerInterface::class);
+        }
 
         $generator->generateController(
             $controllerClassDetails->getFullName(),
@@ -257,6 +260,10 @@ final class MakeCrud extends AbstractMaker
 
             if ($usesEntityManager) {
                 $useStatements->addUseStatement(EntityRepository::class);
+            }
+
+            if (EntityManagerInterface::class !== $repositoryClassName) {
+                $useStatements->addUseStatement(EntityManagerInterface::class);
             }
 
             $generator->generateFile(
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -29,30 +29,18 @@ class <?= $class_name ?> extends AbstractController
 <?php endif ?>
 
 <?= $generator->generateRouteForControllerMethod('/new', sprintf('%s_new', $route_name), ['GET', 'POST']) ?>
-<?php if (isset($repository_full_class_name) && $generator->repositoryHasSaveAndRemoveMethods($repository_full_class_name)) { ?>
-    public function new(Request $request, <?= $repository_class_name ?> $<?= $repository_var ?>): Response
-<?php } else { ?>
     public function new(Request $request, EntityManagerInterface $entityManager): Response
-<?php } ?>
     {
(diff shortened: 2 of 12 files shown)


mdoutreluingne:
It seems that the tests in red are linked to my PR. However, I want to correct the tests but I am unable to correct them.

Locally I still have this error `Compile Error: Cannot use Doctrine\ORM\EntityManagerInterface as EntityManagerInterface because the name is already in use`  that I can't correct.

Can you please help me?

src/Maker/MakeCrud.php
```diff
@@ -163,6 +163,7 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             Request::class,
             Response::class,
             Route::class,
+            EntityManagerInterface::class,
```
weaverryan, line 166: Further above, there is a `$repositoryClassName`. In some cases, this is equal to `EntityManagerInterface`. So we should check if it is and, if so, avoid adding `EntityManagerInterface`. Or, hmm, perhaps better: we could improve `UseStatementGenerator` to ignore duplicates.

weaverryan (commented):
Thanks for taking this on! I just dropped a hint about the failed test

src/Maker/MakeCrud.php
mdoutreluingne: Thank you for your, it was so simple and right in front of my eyes but I hadn't even seen...

I'll make a note to ignore duplications in UseStatementGenerator for another PR, or in this PR if I have time

mdoutreluingne:
It looks like the other failed test came from `it_generates_basic_crud.php`, I'll try to fix that tomorrow

mdoutreluingne:
Finally, the remaining failed test came from `it_generates_crud_using_custom_repository` on the assert `assertFileEquals` because the files `WithCustomRepository.php` and `SweetFoodController.php` were not identical due to the renderForm methods in `WithCustomRepository.php` for the route names `app_sweet_food_new` and `app_sweet_food_edit`.

When I ran the tests locally, I didn't get this error again. It's strange because after committing and pushing on the repository, the `it_generates_crud_using_custom_repository` test still fails, whereas locally I don't get any more errors concerning this test.

I noticed that this test also failed on the last PR(#1330).

Would you have any suggestions for correcting this test?

weaverryan:
The tests can be sensitive. That's partly by design, but they can be a pain... especially when it comes to different Symfony versions. Here's what I'm thinking:

A) Duplicate `WithCustomRepository.php` to create a new `WithCustomRepositoryLegacy.php` that includes the `renderForm()` version.

B) In MakeCrudTest, dynamically choose `WithCustomRepository.php` vs `WithCustomRepositoryLegacy.php` in the `assertFileEquals()` by checking `Kernel::VERSION_ID < 60200`. Something like:

```diff
- sprintf('%s/fixtures/%s', \dirname(__DIR__), 'make-crud/expected/WithCustomRepository.php'),
+ sprintf('%s/fixtures/make-crud/expected/WithCustomRepository%s.php', \dirname(__DIR__), Kernel::VERSION_ID < 60200 ? 'Legacy' : ''),
```

Let me know if that helps :)

mdoutreluingne:
Unfortunately, the tests failed globally (1 test failed when installing symfony/console), I don't know why it compares `render()` and `renderForm()` when `WithCustomRepository.php` is chosen dynamically.

**One thing to note:** the last 2 tests failed on the `index` and `show` route names, while the other 2 failed on the `new` and `edit` route names.

weaverryan:
My code suggestion for checking the Symfony version was slightly wrong - I just pushed the fix for it. Let's see what the tests think now :)

weaverryan:
> I didn't understand why you replaced renderForm() by render() for the route name index and show in WithCustomRepositoryLegacy.php, which should normally be the version with renderForm()

Even back when `renderForm()` existed and was the correct solution, there was no reason to call `renderForm()` if you weren't passing a form into the template. So `index` and `show` have always generated as `render()`.

Anyway, tests green now! Woo!

weaverryan:
Thanks Maxime!

Issue #1037, opened 2021-12-17
make:entity make:entity --uuid or --uild as option 


jaures-kano:

pakolkf:
Hi, I have implemented this option but it is impossible for me to dump it to the main branch since it does not pass the tests. Any ideas? It is possible to speak Spanish @javiereguiluz

jrushlow:
Sorry for the slow response, @pakolkf - without a PR, it's impossible to tell what the problems could be.

Closing as a duplicate of #745

Issue #1033, opened 2021-12-06
Doctrine - Optional feature $id as UUID instead of Integer when creating new entity


skocdopolet:
### Description

I am working on a new project in Symfony 5.3. I am using this command bin/console make:entity for creating entities.

This wizard automatically creates an entity with $id as primary key of type integer. I am prefering UUID instead of integer.

How I should change settings to get Entity like example?

### Example

namespace App\Entity;
    
use App\Repository\EventRepository;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
    
/**
 * @ORM\Entity(repositoryClass=EventRepository::class)
 */
class Event
{
    /**
     * @ORM\Id
     * @ORM\Column(type="uuid", unique=true)
     * @ORM\GeneratedValue(strategy="CUSTOM")
     * @ORM\CustomIdGenerator(class=UuidGenerator::class)
     */
    private $id;
    
     ...
}

skocdopolet:
Hello, is there please some progress?  Thanks

jrushlow:
Duplicate of #745 - we'll have PR #1329 merge / released soon!

Issue #745, opened 2020-11-24
leverage Symfony's UID component in makers


jrushlow:
If Symfony's UID component is available, why not use it when creating new entities, users, etc.. PR's welcome

OskarStark:
`RFPR` Request for Pull Request 😄

lyrixx:
If we do that, it should be optional. I almost always UUID in my application, but I don't use object for it

PR #1329, opened 2023-06-27
Add support for generating UUID id fields in entities

Coffee2CodeNL:
Related issues: #1037 #1033 #745


flkasper:
Suggestion for improving this PR:

- https://github.com/symfony/maker-bundle/blob/79c710ba27a38e6e0986e0c6524db3f3f345bb86/src/Maker/MakeEntity.php#L380 If uuid enabled, then uuid instead of integer

Coffee2CodeNL:
@flkasper good idea

flkasper:
Another suggestion for improvement that I had forgotten:

https://github.com/symfony/maker-bundle/blob/79c710ba27a38e6e0986e0c6524db3f3f345bb86/src/DependencyInjection/Configuration.php#L17

New project based config in config/packages/dev/maker.yaml `entity.uuid_ids: boolean` to globally enable uuid generation in the project and evaluate it in make:entity.

src/Doctrine/EntityClassGenerator.php
```diff
@@ -59,6 +60,10 @@ public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $
             $useStatements->addUseStatement(class_exists(ApiResource::class) ? ApiResource::class : \ApiPlatform\Core\Annotation\ApiResource::class);
         }
 
+        if ($useUUIDIdentifier) {
```
jrushlow, line 63: Perhaps `$useUid` instead of `$useUUID` as I would assume uuid or ulid's would both be possible.

src/Maker/MakeUser.php
```diff
@@ -114,6 +115,12 @@ class_exists(DoctrineBundle::class)
 
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator): void
     {
+        if ($input->getOption('uuid_id')) {
```
jrushlow, line 118: Let's move this conditional to the `interact` method - it's a better dx when we know right away if we're missing a key package.

jrushlow (changes requested):
This looks pretty good @Coffee2CodeNL! I have not looked to closely at the tests yet, if you can `rebase` this PR - we can work on getting those tests green! Thanks in advance...

Issue #364, opened 2019-02-19
Add command for mongo odm


nicodmf:
As mongo odm is a good tool, this will be a good point to generate document and form dependents from mongodb

pascalwacker:
How would you handle function signatures? AFAIK Mongo DB doesn't have an equivalent to nullable/non nullable, therefore everything would need `?string`, `?int`, etc. as return types, or am I missing something?

thedomeffm:
+1 
Would be nice to have a command `make:document`.

MongoODM don't have a command for that (anymore).

See: https://github.com/doctrine/DoctrineMongoDBBundle/issues/498#issuecomment-434442953

Issue #320, opened 2018-11-16
how to generate a document in symfony 4.1 with maker command


christiankasse:
Hello everyone, I am unable to generate documents with symfony 4.1 using mongodb because there are no commands for that, someone can help me please. thx

bestog:
Wanted to ask the same thing. @weaverryan Is there a plan for that? 👍

weaverryan:
No plan, but it doesn’t mean it can’t haopen. The make:entity command is super complex and took a long time to develop. Making one for th ODM, which would be great, is just much lower on the priority list because of the comparative number of users. Someone can contribute it of course.

Cheers!

christiankasse:
And for the time what's can we do ? For moment a generate manually the getters and setters

weaverryan:
Yep - for now, you’ll need to do things by hand like always - if you want getters and setters, using your IDE is a nice way (it sounds like you already are).

SamuelTS:
> Hello everyone, I am unable to generate documents with symfony 4.1 using mongodb because there are no commands for that, someone can help me please. thx

I made a [composer package](https://packagist.org/packages/samuel-tallet/symfony-mongodb-document-maker) to handle this task. Project is still very young but basic features are there. 😉

PR #1330, opened 2023-06-28
New Command: `make:document`

constantable:
New Command: `make:document` to interactively generate document classes for `doctrine/mongodb-odm-bundle`

Related issues #320 , #364


constantable:
Need help with the MongoDB and ext-mongodb PHP extention setup for AppVeyor.

constantable:
> Need help with the MongoDB and ext-mongodb PHP extention setup for AppVeyor.

Fixed AppVeyor

OskarStark:
cc @alcaeus

src/Doctrine/DoctrineODMHelper.php
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    private function isInstanceOf($object, string $class): bool
+    {
+        if (!\is_object($object)) {
+            return false;
+        }
+
+        return $object instanceof $class;
+    }
```
alcaeus, line 296: `$value instanceof $class` works even when `$value` is not an object. Not sure why you'd need this helper.

src/Doctrine/DoctrineODMHelper.php
```diff
@@ -0,0 +1,352 @@
...
+
+    public function getPotentialCollectionName(string $className): string
+    {
+        $documentManager = $this->getRegistry()->getManager();
+
+        if (!$documentManager instanceof DocumentManager) {
+            throw new \RuntimeException('ObjectManager is not an DocumentManager.');
```
alcaeus, line 303:
```suggestion
            throw new \RuntimeException('ObjectManager is not a DocumentManager.');
```

src/Doctrine/DoctrineODMHelper.php
```diff
@@ -0,0 +1,352 @@
...
+
+        return $className;
+    }
+
+    public function isKeyword(string $name): bool
+    {
+        /** @var Connection $connection */
+        $connection = $this->getRegistry()->getConnection();
+
+        return $connection->getDatabasePlatform()->getReservedKeywordsList()->isKeyword($name);
```
alcaeus, line 323: There are no such things as reserved keywords in MongoDB.

src/Doctrine/DocumentRegenerator.php
```diff
@@ -0,0 +1,236 @@
...
+
+            $manipulator = $this->createClassManipulator($classPath);
+            $operations[$classPath] = $manipulator;
+
+            $embeddedClasses = [];
+
+            foreach ($classMetadata->getEmbeddedFieldsMappings() as $fieldName => $mapping) {
+                if (str_contains($fieldName, '.')) {
+                    continue;
+                }
```
alcaeus, line 77: While not recommended, field names in MongoDB can contain dots. Why is this excluded here?

src/Doctrine/DocumentRegenerator.php
```diff
@@ -0,0 +1,236 @@
...
+                        $relation = (new ReferenceMany(
+                            propertyName: $mapping['fieldName'],
+                            targetDocument: $mapping['targetDocument'],
+                            targetPropertyName: $mapping['mappedBy'],
+                            orphanRemoval: $mapping['orphanRemoval'],
+                        ));
+
+                        $manipulator->addOneToManyReference($relation);
+
+                        break;
```
alcaeus, line 142: Associations in ODM come in two variants: references and embedded documents. You assume that any association is a reference, which is incorrect. You need to differentiate on `$mapping['embedded']` to distinguish between references and embedded documents.

alcaeus (changes requested):
I've left some code feedback below.

What I'm completely missing in this PR is support for embedded documents. For one, this should support creating embedded document classes (which use a different document attribute, have no repository, and can be final), as well as creating embedded relationships in the maker itself. Currently the maker assumes all relationships to be references, which isn't true for ODM. In fact, embedded documents are a core concept in MongoDB, and as such I believe they need to be included in the builder for it to be considered.

constantable:
@alcaeus Thank you for code review and suggestions!

Implemented support for embedded documents. 
An EmbeddedDocument could be created form scratch if `--embedded` option specified. 

When embeding a class that has the `ODM\Document` PHP attribute maker will ask should it be transformed to `ODM\EmbeddedDocument`. Please leave a word what you think about this approach.

src/Doctrine/DoctrineODMHelper.php
constantable: DoctrineODMHelper is mostly based on DoctrineHelper as you noted. I have removed this method as minimum PHP version is 8.0 and `instanceof` acts exactly the same as the isInstanceOf method.
Am I supposed to make corresponding changes in DoctrineHelper in this PR?

alcaeus:
> Implemented support for embedded documents. An EmbeddedDocument could be created form scratch if `--embedded` option specified.

Sounds good!

> When embeding a class that has the `ODM\Document` PHP attribute maker will ask should it be transformed to `ODM\EmbeddedDocument`. Please leave a word what you think about this approach.

The problem is that anything mapped as document can't be embedded, and anything mapped as embedded document can't be stored as root document in a collection. IMO the smartest choice would be to print an error message and suggest an alternative (e.g. a mapped superclass containing common properties, extended by two separate classes that then add the `Document` and `EmbeddedDocument` mappings). Unfortunately, changing a `Document` to an `EmbeddedDocument` will break existing functionality if the document is already used in a non-embedded context somewhere. I believe it's fine to explain this problem to the user and letting them solve it themselves before allowing to embed the class.

constantable:
> 

Made changes to print the error message.
image

weaverryan:
Hi there!

I think this is a very cool thing. I have avoided reviewing it because... to be fully honest, I don't want to have to maintain all of the code to make this work. `make:entity` requires a lot of code and complexity, but I'm ok with it because I use it. But I won't use this command (nothing bad about the ODM! It's just not in my stack).

I would prefer to have this as a 3rd party, independent bundle and to add any new APIs to MakerBundle to make that possible. I would even be cool with adding a fake` make:document` to MakerBundle that tells you what `composer require` command to run to get the bundle that *actually* adds it.

Let me know what you think - and sorry for not talking with you about this sooner - it's always very busy.

Cheers!

jrushlow:
Howdy @constantable - this looks awesome! But I have to agree with @weaverryan on the maintenance burden this would add to maker-bundle. If you need help on getting this going as a 3rd party bundle, please don't hesitate to create an issue, we'll be glad to help.

We love new ideas and the PR's that are created from those ideas, so keep them coming! Thanks...

TracKer:
@constantable It would be really cool if you could organize your code into a separate maker-bundle package

src/Maker/MakeDocument.php
```diff
@@ -0,0 +1,917 @@
...
+use Symfony\Bundle\MakerBundle\Doctrine\ODMDependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\FileManager;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputAwareMakerInterface;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Bundle\MakerBundle\Util\ClassDetails;
+use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
+use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
```
GromNaN, line 33: This classes are marked as internal, in order to create a 3rd party package, we would have to duplicate them.

src/Doctrine/ODMDependencyBuilder.php
```diff
@@ -0,0 +1,42 @@
...
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Doctrine;
```
GromNaN, line 12: In order to ease maintenance, all classes related to MongoDB ODM should be in a specific namespace: `Symfony\Bundle\MakerBundle\Doctrine\MongoDB`

GromNaN (commented):
Due to the rejection of this PR, we will create a new bundle for Doctrine MongoDB ODM support with MakerBundle.
https://github.com/doctrine/mongodb-maker-bundle

constantable:
@GromNaN you can use https://github.com/constantable/odm-document-maker if you find it useful.

PR #1088, opened 2022-04-03
Add `@extends` tag to Doctrine repository template.

hhamon:
Adding the `@extends` phpdoc tag on generated Doctrine repository classes will prevent phpstan from generating an error when it's installed in the project.

PR #1332, opened 2023-07-06
Add conditional @implements tag to Doctrine repository template

gremo:
Adding the @implements phpdoc tag on generated Doctrine repository classes will help IDE and prevent tools like phpstan from generating errors.

@implements is only added if `$with_password_upgrade` is set to true, avoiding useless phpdocs.

Immagine 2023-07-06 120413

Similar to #1088


weaverryan:
Thanks @gremo!

PR #1340, opened 2023-07-20
Add `--standalone` to `make:controller`

adrienlucas:
Allows to create controllers without base class [as described in the documentation](https://symfony.com/doc/current/controller/service.html).

[See output examples](https://gist.github.com/adrienlucas/5f832d2aef49012572f6924d5538d611).

PS : Not sure about the `standalone` wording tho, the word itself doesn't appear in the said documentation. Actually I only want it to be an affirmative word, not `--without-base` or something, but I did not find any better alternative.

### DIFF
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -26,7 +26,9 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Attribute\AsController;
 use Symfony\Component\Routing\Annotation\Route;
+use Twig\Environment;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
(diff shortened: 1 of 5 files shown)


weaverryan:
This looks like a solid PR. But I'm not sold on the need. Invokable controllers are a very valid way to do things. Not extending the base controller is also valid - but not really a mainstream use-case. I'm a bit 👎 on this for that reason.

adrienlucas:
@weaverryan I see it as a way to introduce this concept to developers, you could say it's advertising the  `AsController` attribute.

In any case, no worries if it doesn't get merged.

adrienlucas:
I understand it is a too-specific use-case and for this reason it cannot be merged.
For the few people who may want this, here is a PoC of extending the MakerBundle to add this feature in your own applications : https://gist.github.com/adrienlucas/a6c5bfd22efb4b1a89bba3e2b6e0c934

Feel free to close this request, thank you for the feedback!

jrushlow:
Howdy @adrienlucas! I'm with @weaverryan on this. Love the idea, and can see a couple use cases for this. But, using `_invoke()` for a controller isn't main stream enough to outweigh the added complexity (maintenance burden) in maker bundle.

Keep the PR's coming! And feel free to raise this again in the future for a revisit... Thanks!

PR #1342, opened 2023-07-21
fix missing blank space in Repository.tpl.php

OleksiiBulba:
Added extra space character to keep the code style consistent.

Without this space, the new UserRepository is generated like this:
image


Xbirdfr (approved):

jrushlow (approved):
Thanks @OleksiiBulba for catching this!

Issue #1336, opened 2023-07-18
Add support for `--formatted` option in `make:migration` command


lyrixx:
It would be nice to have the same option on the make bundle

```
 symfony console doc:mi:diff -h
[...]
Options:
      --formatted                                          Format the generated SQL.
[...]
```

jrushlow:
added in #1345

PR #1345, opened 2023-08-14
[make:migration] Format the generated migration sql by passing `--formatted` to the command

StevenRenaux:
Hi,

Related to this #1336 I tried to work on this PR.

Now you can run `make:migration --formatted` to get a formatted SQL migration file.

### DIFF
--- a/src/Maker/MakeMigration.php
+++ b/src/Maker/MakeMigration.php
@@ -70,6 +70,9 @@ final class MakeMigration extends AbstractMaker implements ApplicationAwareMaker
                 ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection name')
             ;
         }
+
+        $command
+            ->addOption('formatted', null, InputOption::VALUE_NONE, 'Format the generated SQL');
     }
 
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator)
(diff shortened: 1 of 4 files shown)


src/Maker/MakeMigration.php
```diff
@@ -88,6 +91,10 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
         }
         // end 2.x support
 
+        if ($input->hasOption('formatted') && null !== $input->getOption('formatted')) {
```
mdoutreluingne, line 94: There's a problem with this condition, for me it always `true` without having added the `--formatted` option. Here you just need to know if the option has been added, or not. So like this :

```suggestion
        if ($input->getOption('formatted')) {
```

StevenRenaux: you right ;)

mdoutreluingne (approved):

src/Maker/MakeMigration.php
```diff
@@ -70,6 +70,9 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
                 ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection name')
             ;
         }
+
```
ktherage, line 73: the above help must certainly be updated. line 62.

skigun:
This would be a nice to have 👍

tests/Maker/MakeMigrationTest.php
```diff
@@ -131,5 +131,16 @@ public function getTestDetails(): \Generator
                 $this->assertStringNotContainsString('Success', $output);
             }),
         ];
+
+        yield 'it_generate_a_formatted_migration' => [$this->createMakeMigrationTest()
```
MaximePinot, line 135:
```suggestion
        yield 'it_generates_a_formatted_migration' => [$this->createMakeMigrationTest()
```

jrushlow (approved):
thank you @StevenRenaux!

PR #1355, opened 2023-09-07
Create empty-ready construct on command generate

stipic:
I understand that this was not put on purpose, but I believe that it would be useful for most people, because when generating a lot of commands, in 99% of cases I have to manually write this generic code to be able to do dependency injection

### DIFF
--- a/src/Resources/skeleton/command/Command.tpl.php
+++ b/src/Resources/skeleton/command/Command.tpl.php
@@ -10,6 +10,11 @@ namespace <?= $namespace; ?>;
 )]
 class <?= $class_name; ?> extends Command
 {
+    public function __construct()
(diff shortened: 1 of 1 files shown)


src/Resources/skeleton/command/Command.tpl.php
```diff
@@ -10,6 +10,11 @@
 )]
 class <?= $class_name; ?> extends Command
 {
+    public function __construct()
```
weaverryan, line 13: Should we have this?

```suggestion
    public function __construct(string $name = null)
```

weaverryan (approved):

weaverryan:
Thanks @stipic!

PR #1360, opened 2023-09-15
Fix UseStatementGenerator usage in MakeFixtures.php

beesofts:
UseStatement is not used as it should

### DIFF
--- a/src/Maker/MakeFixtures.php
+++ b/src/Maker/MakeFixtures.php
@@ -55,8 +55,8 @@ final class MakeFixtures extends AbstractMaker
         );
 
         $useStatements = new UseStatementGenerator([
-            Fixture::class => null,
-            ObjectManager::class => null,
+            Fixture::class,
+            ObjectManager::class,
         ]);
 
         $generator->generateClass(


src/Maker/MakeFixtures.php
```diff
@@ -55,8 +55,8 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
         );
 
         $useStatements = new UseStatementGenerator([
```
OskarStark, line 57: Maybe we can tell static code analysis, that it expects a list?

beesofts: in fact UserStatementGenerator can be used in different ways and the phpdoc seems good :
`/**
     * For use statements that contain aliases, the $classesToBeImported array
     * may contain an array(s) like [\Some\Class::class => 'ZYX']. The generated
     * use statement would appear as "use Some\Class::class as 'ZXY'". It is ok
     * to mix non-aliases classes with aliases.
     *
     * @param string[]|array<string, string> $classesToBeImported
     */
`

OskarStark (approved):

weaverryan:
Thanks Benoît!

Issue #1301, opened 2023-04-27
Add "make:listener" command and alias "make:subscriber" to it 


nicolas-grekas:
Instead of generating a implementation of `EventSubscriberInterface`, it might be great to generate a class with `#[AsEventListener]` attributes.

jrushlow:
Completed in #1366 - Thanks for the idea Nicolas!

PR #1366, opened 2023-10-03
[make:listener] Add global command for listener and subscriber

StevenRenaux:
Related to #1301 I also suggest to add a global maker for both, Listener and Subscriber.

I worked with the based class `MakeSubscriber` and updated it to make this solution.

In the terminal you can run `make:event` and a question will be displayed like that to know which event class you want to create.

```
$ php bin/console make:event FooBar kernel.request


 What event class to generate(enter ? to see all types) [Subscriber]:
 > ?

Event Types Class
  * Subscriber
  * Listener

 What event class to generate(enter ? to see all types) [Subscriber]:
 > 
```

And if you write well what you want as `FooBarSubscriber` or `FooBarListener` and the event to listen, your file will be generated as Listener or Subscriber.

```php
namespace App\EventListener;

use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
use Symfony\Component\HttpKernel\Event\RequestEvent;
use Symfony\Component\HttpKernel\KernelEvents;

final class FooBarListener
{
    #[AsEventListener(event: KernelEvents::REQUEST)]
    public function onKernelRequest(RequestEvent $event): void
    {
        // ...
    }
}
```

### DIFF
--- a/src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
+++ b/src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
@@ -63,6 +63,14 @@ class MakeCommandRegistrationPass implements CompilerPassInterface
                 }
             }
 
+            /*
+             * @deprecated remove this block when removing make:subscriber
+             */
+            if (method_exists($class, 'getCommandAlias')) {
+                $alias = $class::getCommandAlias();
+                $commandDefinition->addTag('console.command', ['command' => $alias, 'description' => 'Deprecated alias of "make:listener"']);
+            }
+
             $container->setDefinition(sprintf('maker.auto_command.%s', Str::asTwigVariable($class::getCommandName())), $commandDefinition);
         }
     }
--- a/src/DependencyInjection/MakerExtension.php
+++ b/src/DependencyInjection/MakerExtension.php
@@ -15,7 +15,6 @@ use Symfony\Bundle\MakerBundle\DependencyInjection\CompilerPass\MakeCommandRegis
 use Symfony\Bundle\MakerBundle\MakerInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Loader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 
@@ -26,27 +25,11 @@ use Symfony\Component\HttpKernel\DependencyInjection\Extension;
  */
 class MakerExtension extends Extension
 {
-    /**
-     * @deprecated remove this block when removing make:unit-test and make:functional-test
-     */
-    private const TEST_MAKER_DEPRECATION_MESSAGE = 'The "%service_id%" service is deprecated, use "maker.maker.make_test" instead.';
-
     public function load(array $configs, ContainerBuilder $container): void
     {
         $loader = new Loader\XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('services.xml');
         $loader->load('makers.xml');
-
-        /**
-         * @deprecated remove this block when removing make:unit-test and make:functional-test
-         */
-        $deprecParams = method_exists(Definition::class, 'getDeprecation') ? ['symfony/maker-bundle', '1.29', self::TEST_MAKER_DEPRECATION_MESSAGE] : [true, self::TEST_MAKER_DEPRECATION_MESSAGE];
-        $container
-            ->getDefinition('maker.maker.make_unit_test')
-            ->setDeprecated(...$deprecParams);
-        $container
-            ->getDefinition('maker.maker.make_functional_test')
-            ->setDeprecated(...$deprecParams);
 
         $configuration = $this->getConfiguration($configs, $container);
         $config = $this->processConfiguration($configuration, $configs);
--- a/src/Maker/MakeListener.php
+++ b/src/Maker/MakeListener.php
@@ -1,0 +1,230 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\EventRegistry;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Bundle\MakerBundle\Util\UseStatementGenerator;
+use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\Question;
+use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Ryan Weaver <weaverryan@gmail.com>
+ * @author Steven Renaux <steven.renaux8000@gmail.com>
+ */
+final class MakeListener extends AbstractMaker
+{
+    private const ALL_TYPES = ['Listener', 'Subscriber'];
+    private bool $isSubscriber = false;
+
+    public function __construct(private readonly EventRegistry $eventRegistry)
+    {
+    }
+
+    public static function getCommandName(): string
+    {
+        return 'make:listener';
+    }
+
+    /**
+     * @deprecated remove this method when removing make:subscriber
+     */
+    public static function getCommandAlias(): string
+    {
+        return 'make:subscriber';
+    }
+
+    public static function getCommandDescription(): string
+    {
+        return 'Creates a new event subscriber class or a new event listener class';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig): void
+    {
+        $command
+            ->addArgument('name', InputArgument::OPTIONAL, 'Choose a class name for your event listener or subscriber (e.g. <fg=yellow>ExceptionListener</> or <fg=yellow>ExceptionSubscriber</>)')
+            ->addArgument('event', InputArgument::OPTIONAL, 'What event do you want to listen to?')
+            ->setHelp(file_get_contents(__DIR__.'/../Resources/help/MakeListener.txt'))
+        ;
+
+        $inputConfig->setArgumentAsNonInteractive('event');
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
+    {
+        /* @deprecated remove the following block when removing make:subscriber */
+        $this->handleDeprecatedMakerCommands($input, $io);
+
+        $io->writeln('');
+
+        $name = $input->getArgument('name');
+
+        if (!str_ends_with($name, 'Subscriber') && !str_ends_with($name, 'Listener')) {
+            $question = new ChoiceQuestion('Do you want to generate an event listener or subscriber?', self::ALL_TYPES, 0);
+            $classToGenerate = $io->askQuestion($question);
+
+            $input->setArgument('name', $name.$classToGenerate);
+        }
+
+        if (str_ends_with($input->getArgument('name'), 'Subscriber')) {
+            $this->isSubscriber = true;
+        }
+
+        if (!$input->getArgument('event')) {
+            $events = $this->eventRegistry->getAllActiveEvents();
+
+            $io->writeln(' <fg=green>Suggested Events:</>');
+            $io->listing($this->eventRegistry->listActiveEvents($events));
+            $question = new Question(sprintf(' <fg=green>%s</>', $command->getDefinition()->getArgument('event')->getDescription()));
+            $question->setAutocompleterValues($events);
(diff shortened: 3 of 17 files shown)


src/Resources/config/makers.xml
```diff
@@ -51,6 +51,11 @@
                 <tag name="maker.command" />
             </service>
 
+            <service id="maker.maker.make_event" class="Symfony\Bundle\MakerBundle\Maker\MakeEvent">
```
Jean-Beru, line 54: I think we have to keep the `maker.maker.make_subscriber` to avoid breaking changes if it a service decorates it.

tests/Maker/MakeEventTest.php
```diff
@@ -0,0 +1,202 @@
...
+                );
+
+                self::assertFileExists($runner->getPath('src/EventSubscriber/FooSubscriber.php'));
+
+                $content = file_get_contents($runner->getPath('src/EventSubscriber/FooSubscriber.php'));
+
+                self::assertStringContainsString(
+                    'KernelEvents::REQUEST => \'onKernelRequest\'',
+                    $content
+                );
```
Jean-Beru, line 41: Maybe we can compare generated code with a file in `Tests/Maker/fixtures/`, WDYT ?

src/Resources/config/makers.xml
StevenRenaux: Revert deletion and add deprecation

src/Maker/MakeEvent.php
```diff
@@ -23,42 +23,78 @@
...
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\KernelEvents;
 
 /**
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  * @author Ryan Weaver <weaverryan@gmail.com>
+ * @author Steven Renaux <steven.renaux8000@gmail.com>
  */
-final class MakeSubscriber extends AbstractMaker
+final class MakeEvent extends AbstractMaker
```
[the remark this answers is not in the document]
StevenRenaux: Fixed with this [commit](https://github.com/symfony/maker-bundle/pull/1366/commits/3c1a24137ab17249d259f4bea5a1fa09c243c167)

tests/Maker/MakeEventTest.php
StevenRenaux: Updated if you want to check ;) [9c96bbf](https://github.com/symfony/maker-bundle/pull/1366/commits/9c96bbf350b7484cc8650e3f86bacde96c09ec6c)

src/Maker/MakeEvent.php
```diff
@@ -23,42 +23,78 @@
...
+    private bool $isSubscriber = false;
+
+    public function __construct(private readonly EventRegistry $eventRegistry)
     {
     }
 
     public static function getCommandName(): string
     {
-        return 'make:subscriber';
+        return 'make:event';
```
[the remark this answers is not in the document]
StevenRenaux: I don't know because for me it's also sounds like disturbing to me to write `make:subscriber` to get a listener at the end.

Maybe let see what the others reviewers will think about the naming. It's not written in stone :wink:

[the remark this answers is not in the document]
Jean-Beru: IMO, a subscriber is no more than a listener which brings its configuration (event => method). `make:event-listener --as-subscriber` could be an alternative.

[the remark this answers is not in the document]
maelanleborgne: It might be too complex, but here is a suggestion :
With `make:event-listener` as the command name, we could have some logic that checks :
- If there's a `--as-subscriber` or `--as-listener` option. If so, we stick with that.
- Else, try to parse the classname's suffix. If we can determine what type of class the user wants to generate, stick with that.
- Else, ask a question, with `EventListener` as default

src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
```diff
@@ -63,6 +63,14 @@ public function process(ContainerBuilder $container): void
                 }
             }
 
+            /*
+             * @deprecated remove this block when removing make:subscriber
+             */
+            if (method_exists($class, 'getCommandAlias')) {
+                $alias = $class::getCommandAlias();
+                $commandDefinition->addTag('console.command', ['command' => $alias, 'description' => 'Deprecated alias of "make:event"']);
+            }
```
weaverryan, line 72: Should/could we leverage the `getCommandAliases()` referenced above since that logic already exists?

src/Maker/MakeEvent.php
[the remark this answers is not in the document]
weaverryan: Yea, if we call it `make:event-listener`... you can still run it with `make:event` :). But I probably like `make:listener` best. We think of "event listeners" in Symfony, JavaScript, etc. "listener" is a dominate name that has a lot of meaning. 

> If there's a --as-subscriber or --as-listener option. If so, we stick with that.

This is a nice idea, but I don't think we need it. I'm cool with looking at the suffix of that class and, if it's not `Listener` or `Subscriber`, asking them which they want. Or, maybe better, just using a listener, as they are so nice nowadays thanks to `AsEventListener`.

src/Maker/MakeEvent.php
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /**
+     * @deprecated remove this method when removing make:subscriber
+     */
+    public static function getCommandAlias(): string
+    {
+        return 'make:subscriber';
+    }
```
weaverryan, line 55: Wait, will this mean that `make:subscriber` will now call THIS maker? If so, let's not do this. Instead, let's keep `make:subscriber` as it is, but deprecate it.

src/Maker/MakeEvent.php
```diff
@@ -0,0 +1,257 @@
...
+
+    public static function getCommandDescription(): string
+    {
+        return 'Creates a new event subscriber class or a new event listener class';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig): void
+    {
+        $command
+            ->addArgument('name', InputArgument::OPTIONAL, 'Choose a class name for your event subscriber or listener (e.g. <fg=yellow>ExceptionSubscriber</> or <fg=yellow>ExceptionListener</>)')
```
weaverryan, line 65:
```suggestion
            ->addArgument('name', InputArgument::OPTIONAL, 'Choose a class name for your event subscriber or listener (e.g. <fg=yellow>ExceptionListener</> or <fg=yellow>ExceptionSubscriber</>)')
```

weaverryan (commented):
This is a cool idea 👍

src/Maker/MakeEvent.php
StevenRenaux: I made a deprecation and followed the tricks as it was done for `make:unit-test` and `make:functional-test`.
When `make:subscriber` is called it run my maker and alert the user.

https://github.com/symfony/maker-bundle/blob/ab025cd05b93e9b3955e7416cefede5e87fb9deb/src/Maker/MakeTest.php#L60

weaverryan: Hmm, ok fair enough. We are free to change what these commands due between minor releases. So let's give them the better experience.

StevenRenaux:
@weaverryan  Command renamed make:listener  with listener as default type. 

To resume:
If Listener or Subscriber was not found in the suffix

```
$ php bin/console make:listener foo


 What event class to generate(enter ? to see all types) [Listener]:
 > 
```

If Listener or Subscriber was  found in the suffix we display the originals events to choose, as the legacy command did.

```
$ php bin/console make:listener FooListener

 Suggested Events:
 * Generator::class ()
 * Symfony\Component\Mailer\Event\MessageEvent (Symfony\Component\Mailer\Event\MessageEvent)
 * Symfony\UX\TwigComponent\Event\PostRenderEvent (Symfony\UX\TwigComponent\Event\PostRenderEvent)
 * Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent (Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent)
/* ...*/

  What event do you want to listen to?:
 > 
```

And if no name was filled, a question is displayed:

```
$ php bin/console make:listener

 Choose a class name for your event listener or subscriber (e.g. ExceptionListener or ExceptionSubscriber):
 > 

```
I didn't really understand your comment, if I was right to notify a deprecation about make:subscriber :
#discussion_r1353072629

If it good for you, maybe we would need to create a new Issue for the next major version to clean all of this deprecation as:

- [MakerExtension](https://github.com/StevenRenaux/maker-bundle/blob/f04136d0b5dfbf2d2cba95e30a1981c887fbf122/src/DependencyInjection/MakerExtension.php#L36)
- [MakerExtension](https://github.com/StevenRenaux/maker-bundle/blob/f04136d0b5dfbf2d2cba95e30a1981c887fbf122/src/DependencyInjection/MakerExtension.php#L58)
- [MakeListener](https://github.com/StevenRenaux/maker-bundle/blob/f04136d0b5dfbf2d2cba95e30a1981c887fbf122/src/Maker/MakeListener.php#L52)
- [MakeListener](https://github.com/StevenRenaux/maker-bundle/blob/f04136d0b5dfbf2d2cba95e30a1981c887fbf122/src/Maker/MakeListener.php#L76)
- [MakeListener](https://github.com/StevenRenaux/maker-bundle/blob/f04136d0b5dfbf2d2cba95e30a1981c887fbf122/src/Maker/MakeListener.php#L244)
- [MakeCommandRegistrationPass](https://github.com/StevenRenaux/maker-bundle/blob/f04136d0b5dfbf2d2cba95e30a1981c887fbf122/src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php#L69)

PS: not related to this PR but we also have deprecated code about make:functinal-test  and make:unit-test .

src/Maker/MakeListener.php
```diff
@@ -0,0 +1,257 @@
...
+
+        $name = $input->getArgument('name');
+
+        if (!str_ends_with($name, 'Subscriber') && !str_ends_with($name, 'Listener')) {
+            $classToGenerate = null;
+
+            while (null === $classToGenerate) {
+                $defaultEvent = 'Listener';
+                $question = new Question('What event class to generate(enter <comment>?</comment> to see all types)', $defaultEvent);
+                $question->setAutocompleterValues(self::ALL_TYPES);
```
weaverryan, line 88: This makes more sense as a choice question: either choose `Listener` or `Susbcriber`.

src/Maker/MakeListener.php
```diff
@@ -0,0 +1,257 @@
...
+        $io->writeln('');
+
+        $name = $input->getArgument('name');
+
+        if (!str_ends_with($name, 'Subscriber') && !str_ends_with($name, 'Listener')) {
+            $classToGenerate = null;
+
+            while (null === $classToGenerate) {
+                $defaultEvent = 'Listener';
+                $question = new Question('What event class to generate(enter <comment>?</comment> to see all types)', $defaultEvent);
```
weaverryan, line 87: > Do you want to generate an event listener or subscriber?

weaverryan:
> I didn't really understand your comment, if I was right to notify a deprecation about make:subscriber:

It looks correct what you did 👍 

> If it good for you, maybe we would need to create a new Issue for the next major version to clean all of this deprecation as:

Yes - though no issue needed, this will naturally happen. I don't know when we might release a new major version, but the main thing we will do when that happens is hunt down all of the deprecated code paths and remove them. Assuming we've done a good job marking this with `@deprecated` or `trigger_deprecation()`, they should be easy to find.

src/Maker/MakeListener.php
StevenRenaux: Really better, thanks for the tips :wink:

src/DependencyInjection/MakerExtension.php
```diff
@@ -48,6 +52,14 @@ public function load(array $configs, ContainerBuilder $container): void
             ->getDefinition('maker.maker.make_functional_test')
             ->setDeprecated(...$deprecParams);
 
+        /**
+         * @deprecated remove this block when removing make:subscriber
+         */
+        $deprecParamsSubscriber = method_exists(Definition::class, 'getDeprecation') ? ['symfony/maker-bundle', '1.51', self::SUBSCRIBER_MAKER_DEPRECATION_MESSAGE] : [true, self::SUBSCRIBER_MAKER_DEPRECATION_MESSAGE];
+        $container
+            ->getDefinition('maker.maker.make_subscriber')
+            ->setDeprecated(...$deprecParamsSubscriber);
```
weaverryan, line 61: Actually, `getDeprecation()` has now existed for a long time and exists in every version that Maker supports. So can you simplify this - and while you're can you simplify the code above also for the test stuff?

weaverryan (commented):
Don't forget to check the php-cs CI job for some coding standards stuff :)

StevenRenaux:
Need help @weaverryan , I don't understand why it failed in the appveyor. It only failed when Listener are generated. But in local or during the other tests in the CI it work.
Example on local, on the left side the file generated and in the right the one compare with. I don't understant why we loose an indention during appveyor test.

image

By the way I updated the deprecated messages and finally removed it from MakerExtension.php and added it in the maker.xml

src/Resources/skeleton/event/Listener.tpl.php
```diff
@@ -0,0 +1,14 @@
+<?= "<?php\n" ?>
+
+namespace <?= $namespace; ?>;
+
+<?= $use_statements; ?>
+
+final class <?= $class_name ?>
```
weaverryan, line 7: Add a `\n` on the end - see https://github.com/symfony/maker-bundle/blob/main/src/Resources/skeleton/message/Message.tpl.php#L5

If `?>` is the last thing on a line, the PHP sucks up the whitespace before the next char. Not sure why we're seeing this on Windows only, but I bet it'll fix it (maybe this odd behavior is windows-only)

StevenRenaux: It's okay about my generated files in appveyor, thanks @weaverryan :wink: 

The two errors left are about `TemplateLinterTest`

src/DependencyInjection/CompilerPass/MakeCommandRegistrationPass.php
squrious: +1, I found it a little bit weird to have methods like `getCommandAliases` whose existence leads to messages for very specific classes.

Maybe something like this?
```php
if (method_exists($class, 'getDeprecatedCommandAliases')) {
    foreach ($class::getDeprecatedCommandAliases() as $alias) {
        $commandDefinition->addTag('console.command', ['command' => $alias, 'description' => \sprintf('Deprecated alias of "%s"', $class::getCommandName())]);
    }
}
```

weaverryan:
Thank you Steven!

PR #1368, opened 2023-10-07
Use TwigComponent HTML syntax

smnandre:
Use the new default HTML syntax 

( `<twig:Alert />` instead of  `{{ component('Alert') }}` )

cc @weaverryan


smnandre:
Failures seem unrelated

weaverryan:
Thanks Simon!

Issue #1365, opened 2023-10-02
Validator entityExists function not working correctly


Antarian:
On this line Validator is checking if Entity exists if it is defined as abolute path namespace and class.
https://github.com/symfony/maker-bundle/blob/0890fd3cf1e2a5221f9b3c6ee1769c537aef683d/src/Validator.php#L215
If condition just after this one will fail anyway, because namespace is absolute and thus it is not in array of entities returned by Doctrine, as they are not defined as absolute namespaces (without prepending slash).

I have discovered this trying to use make:crud with absolute namespaced entity.

Antarian:
If I change 219 to this:
`if (!\in_array($className, $entities) && !\in_array(ltrim($className, '\\'), $entities)) {`
crud generator is able to generate my crud controller for entity which is not in usual App\ namespace.

weaverryan:
PR welcome :). Your solution could, in theory, allow the user to pass an invalid `\Full\Entity\ClassName` that is allowed just because there is a `\Real\Entity\ClassName` that has the same ending (`ClassName`)... but I think that's ok. Better to allow an edge-case invalid situation vs stopping legitimate use-cases like yours.

Antarian:
@weaverryan `ltrim` should cut only first (back)slash or nothing at all. Stopping at reaching first character not in mask.
Also absolute namespacing is only required because `Entity\\` and `Repository\\` prefixes are hardwired in the CRUD generator.

PR #1370, opened 2023-10-11
Entity Validator is accepting absolute namespaces

Antarian:
Resolving issue when Absolute namespace for Entity is not passing through validation even if it should.
https://github.com/symfony/maker-bundle/issues/1365#issuecomment-1757938060


weaverryan:
Thanks Peter!

Issue #1372, opened 2023-10-13
[make:form] Add association detection to generate EntityType in form


maelanleborgne:
Hi, I'd be happy to work on a new feature to enhance form generation so that associations will be rendered with the correct EntityType inputs. I've had quick look in the issue and couldn't find any discussion about this, but if it's already been covered let me know.

PR #1373, opened 2023-10-13
Add relations to entity form generation

maelanleborgne:
PR for #1372 

When generating a form for an entity (through `make:form` or `make:crud`), this feature adds an EntityType input for each compatible relations in the generate form.

- [x] Add tests

### DIFF
--- a/src/Doctrine/EntityDetails.php
+++ b/src/Doctrine/EntityDetails.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Doctrine;
 
 use Doctrine\Persistence\Mapping\ClassMetadata;
+use Symfony\Bridge\Doctrine\Form\Type\EntityType;
 
 /**
  * @author Sadicov Vladimir <sadikoff@gmail.com>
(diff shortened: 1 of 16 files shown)


src/Doctrine/EntityDetails.php
```diff
@@ -55,17 +56,21 @@ public function getFormFields(): array
...
-            }
-        }
-
         $fieldsWithTypes = [];
         foreach ($fields as $field) {
             $fieldsWithTypes[$field] = null;
         }
 
+        foreach ($this->metadata->associationMappings as $fieldName => $relation) {
+            if (\Doctrine\ORM\Mapping\ClassMetadata::ONE_TO_MANY !== $relation['type']) {
```
weaverryan, line 65: It's been awhile since I looked in the code :). So, does this mean that if a field is a OneToMany to another entity, it would add an `EntityType`? Like, if we're building a `CategoryForm` and it has a `products` OneToMany property?

If so, do we need `multiple => true`? And why not add the other fields as well? Or maybe I'm missing the flow - it happens frequently ;)

weaverryan: Ha! Thank you for pointing out the `!` I missed on the `!==` :). So no, leave OneToMany off 👍

weaverryan (commented):
This seems good to me - I'd love to see a test and also the `multiple => true` when needed

weaverryan:
Thanks @maelanleborgne!

PR #1376, opened 2023-10-18
[make:auth] Use LogoutRouteLoader for logout route when available

### LINKED ISSUE #1375: [make:auth] Adapt command to use the upcoming LogoutRouteLoader
This PR has been merged a few days ago : https://github.com/symfony/symfony/pull/50946
We now don't need to make a controller to match the logout path, we just need to register a route loader that deals with loading a route for each firewall that has a logout.path.

It would be nice to have this change reflected in the make:auth command.

I've started to work on this, but I'm not sure if there are any plan to release a new version of the maker-bundle with minimum compatibility with symfony 6.4 / 7, or if I should test if the new `LogoutRouteLoader` service exists and change the behaviour accordingly.

maelanleborgne:
Closes #1375


src/Maker/MakeAuthenticator.php
```diff
@@ -253,6 +255,20 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+                $yamlData = Yaml::parse($this->fileManager->getFileContents($path)) ?? [];
+            }
+            if (!(\array_key_exists('_symfony_logout', $yamlData) && 'security.route_loader.logout' !== $yamlData['_symfony_logout']['ressource'] ?? null)) {
+                $yamlData['_symfony_logout'] = [
+                    'resource' => 'security.route_loader.logout',
+                    'type' => 'service',
+                ];
+            }
+            $generator->dumpFile($path, Yaml::dump($yamlData));
+        }
```
weaverryan, line 270: To follow on Wouter's comment. I would:

A) Check if `config/routes/security.yaml` exists
B) If it does, do nothing. We will assume they have the latest recipe from https://github.com/symfony/recipes/pull/1244
C) If it does not exist, we print a message:

> Be sure to add the "logout" route to config/routes/security.yaml or upgrade the symfony/security-bundle recipe.

src/Maker/MakeAuthenticator.php
```diff
@@ -253,13 +255,28 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+            }
+            if (!(\array_key_exists('_symfony_logout', $yamlData) && 'security.route_loader.logout' !== $yamlData['_symfony_logout']['ressource'] ?? null)) {
+                $yamlData['_symfony_logout'] = [
+                    'resource' => 'security.route_loader.logout',
+                    'type' => 'service'
+                ];
+            }
+            $generator->dumpFile($path, Yaml::dump($yamlData));
+        }
+
```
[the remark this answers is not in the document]
maelanleborgne: Yes :cry:  I tried (to no avail) to find a way to check for the presence of the loader, and ended up using Ryan's suggestion. If users moved the definition elsewhere they should now how to interpret the message.

src/Maker/MakeAuthenticator.php
maelanleborgne: I couldn't find a way to check for the presence of the loader, so I ended up using using your suggestion. If users moved the definition elsewhere they should now how to interpret the message.

Issue #1380, opened 2023-10-21
Handle "forbiden" or dangerous/ambiguous class names in make:entity


olberger:
It seems that "Collection" for instance, would be a very bad name for an user entity generated by make:entity: Collection, per Doctrine\Common\Collections\Collection is already used for OneToMany relations, for instance... better avoid ambiguity.

However, it seems that there isn't any warning of show-stopper if a user wants to manage a set of collectibles and chose Collection for a Doctrine model entity :-/

I guess it would make sense to have a set of "reserved" names to prevent issues as early as possible.

Hope this makes sense

weaverryan:
PR welcome here also to tighten this up :)

PR #1389, opened 2023-10-31
Add collection as reserved keywords

StevenRenaux:
Related to #1380 

In the `generate method`  we call the `Generator class` and the method  `createClassNameDetails`.

```
        $entityClassDetails = $generator->createClassNameDetails(
            $input->getArgument('name'),
            'Entity\\'
        );
```

And in this method we call the static method `validateClassName` of the `Validator class`.
In it we already have a check about the reserved keywords.

```
        $reservedKeywords = ['__halt_compiler', 'abstract', 'and', 'array',
            'as', 'break', 'callable', 'case', 'catch', 'class',
            'clone', 'const', 'continue', 'declare', 'default', 'die', 'do',
            'echo', 'else', 'elseif', 'empty', 'enddeclare', 'endfor',
            'endforeach', 'endif', 'endswitch', 'endwhile', 'eval',
            'exit', 'extends', 'final', 'finally', 'fn', 'for', 'foreach', 'function',
            'global', 'goto', 'if', 'implements', 'include',
            'include_once', 'instanceof', 'insteadof', 'interface', 'isset',
            'list', 'match', 'namespace', 'new', 'or', 'print', 'private',
            'protected', 'public', 'readonly', 'require', 'require_once', 'return',
            'static', 'switch', 'throw', 'trait', 'try', 'unset',
            'use', 'var', 'while', 'xor', 'yield',
            'int', 'float', 'bool', 'string', 'true', 'false', 'null', 'void',
            'iterable', 'object', '__file__', '__line__', '__dir__', '__function__', '__class__',
            '__method__', '__namespace__', '__trait__', 'self', 'parent', 'collection',
        ];
```

Is this the best solution to avoid these dangerous/ambiguous class names?
Because this method is also called by the other makers.

Or do we need to add an additional verification in the entity maker?

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -43,7 +43,7 @@ final class Validator
             'use', 'var', 'while', 'xor', 'yield',
             'int', 'float', 'bool', 'string', 'true', 'false', 'null', 'void',
             'iterable', 'object', '__file__', '__line__', '__dir__', '__function__', '__class__',
-            '__method__', '__namespace__', '__trait__', 'self', 'parent',
+            '__method__', '__namespace__', '__trait__', 'self', 'parent', 'collection',
         ];
 
         foreach ($pieces as $piece) {


weaverryan:
Yea... I see what you mean. I'm ok with putting it here. I have a hard time imagining someone trying to create a class called `Collection` with any maker other than `make:entity`.

weaverryan:
Thanks Steven!

PR #1397, opened 2023-11-21
[CS] Fix php-cs-fixer errors

smnandre:
That will make ONE check green in the CI .. 👼


weaverryan:
Thanks - merged #1398 because it had one other fix :)

PR #1398, opened 2023-11-21
Fix tests CS violations

maelanleborgne:
Lately CI checks fail because of an [upgrade of the default phpunit version used in phpunit-bridge](https://github.com/symfony/symfony/pull/50760). This PR fixes this in the bundle's tests. And also a couple of minor cs fixes.

(EDIT: just noticed the CS part was fixed in #1397 just earlier)


weaverryan:
Thanks @maelanleborgne!

PR #1403, opened 2023-12-06
Doctrine make:fixtures - Remove extra semicolon

MrYamous:
Using the command `make:fixtures` I noticed an extra semicolon in the generated file 

```
use Doctrine\Bundle\FixturesBundle\Fixture;
use Doctrine\Persistence\ObjectManager;
;
```


jrushlow (approved):
Thanks @MrYamous!

PR #1405, opened 2023-12-08
[make:reset-password] increase password strength & check for comprimised password

Spomky:
For the password reset process:
* [x] Adds the `NotCompromisedPassword` and `PasswordStrength` constraints to the password reset form
* [x] Set a minimal password length to 12 instead of 6

### DIFF
--- a/src/Resources/skeleton/resetPassword/ChangePasswordFormType.tpl.php
+++ b/src/Resources/skeleton/resetPassword/ChangePasswordFormType.tpl.php
@@ -22,11 +22,13 @@ class <?= $class_name ?> extends AbstractType
                             'message' => 'Please enter a password',
                         ]),
                         new Length([
-                            'min' => 6,
+                            'min' => 12,
                             'minMessage' => 'Your password should be at least {{ limit }} characters',
                             // max length allowed by Symfony for security reasons
                             'max' => 4096,
                         ]),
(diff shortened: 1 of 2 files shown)


src/Resources/skeleton/resetPassword/ChangePasswordFormType.tpl.php
```diff
@@ -22,11 +22,13 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                             'message' => 'Please enter a password',
                         ]),
                         new Length([
-                            'min' => 6,
+                            'min' => 12,
                             'minMessage' => 'Your password should be at least {{ limit }} characters',
                             // max length allowed by Symfony for security reasons
                             'max' => 4096,
                         ]),
+                        new PasswordStrength(),
```
94noni, line 30: I like this 👍🏻

94noni:
perhaps split into 2 PRs ?
one related to login password and other to throttling

Spomky:
> perhaps split into 2 PRs ? one related to login password and other to throttling

Yes indeed. I will change it.
The throttling part is not complete as I should add the symfony/rate-limiter as a dependency.

jrushlow (approved):
Awesome! thank you @Spomky for this PR

PR #1406, opened 2023-12-08
[make:auth] Add throttling support to authenticator maker

Spomky:
In the MakeAuthenticator class, a new optional argument 'support-throttling' was added. This allows the user to specify if they want to enable throttling protection during the authentication generation process. If 'support-throttling' is enabled, the LimiterInterface dependency is added and the 'throttling' node is set in the firewall configuration.


jrushlow:
Closing - See https://github.com/symfony/maker-bundle/pull/1523 

`make:auth` is being deprecated - new features should be implemented in one of the `make:security:*` commands

Thank you & looking forward to your next PR...

Issue #1002, opened 2021-10-29
Allow doctrine custom types to register their php class to makerbundle for generating Type Hint


sarim:
Right now its hardcoded in https://github.com/symfony/maker-bundle/blob/f76abb1e4facd712a5cdbf35aff998e87c416801/src/Util/ClassSourceManipulator.php#L1116-L1163

It would be useful to allow a configuration or any other way to allow user code to extend this type hint mappings. Right now for any fields of doctrine custom types, maker bundle does not generate any type hint for getter and setter methods. Which is expected as `Doctrine\DBAL\Types\Type` does not have any way to convey php object's type. Its cumbersome to manually add type hint to getter and setter, and its pointless as when regenerating entity (make:entity --regenerate --overwrite) its wiped out anyways.

PR #1021, opened 2021-11-29
Add support to custom type hint

IndraGunawan:
Fix #1002

PR #1413, opened 2024-01-01
[DoctrineHelper] handle property type for custom doctrine type

IndraGunawan:
fixed: #1002
replace: #1021

sample custom type

```php
<?php

declare(strict_types=1);

namespace App\Doctrine\DBAL\Types;

use Cake\Chronos\Chronos;
use DateTimeImmutable;
use Doctrine\DBAL\Platforms\AbstractPlatform;
use Doctrine\DBAL\Types\ConversionException;
use Doctrine\DBAL\Types\DateTimeImmutableType;

class ChronosDateTimeType extends DateTimeImmutableType
{
    final public const CHRONOS_DATETIME = 'chronos_datetime';

    /**
     * {@inheritDoc}
     */
    public function convertToPHPValue($value, AbstractPlatform $platform): ?Chronos
    {
        if (null === $value) {
            return null;
        }

        $dateTime = parent::convertToPHPValue($value, $platform);

        return Chronos::instance($dateTime);
    }

    /*
     * ....
     */
}
```

this change will get `convertToPHPValue` function return type and set it as property type
if no return type defined on `convertToPHPValue` function then the property will not have any type hint

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -273,18 +273,22 @@ final class DoctrineHelper
             default => null,
         };
 
-        if (null === $propertyType && Type::getTypeRegistry()->has($columnType)) {
-            $reflection = new \ReflectionClass(\get_class(Type::getTypeRegistry()->get($columnType)));
-            if ($reflection->hasMethod('convertToPHPValue') && $returnType = $reflection->getMethod('convertToPHPValue')->getReturnType()) {
-                if ($returnType->isBuiltin()) {
-                    $propertyType = $returnType->getName();
-                } else {
-                    $propertyType = '\\'.$returnType->getName();
-                }
-            }
+        if (null !== $propertyType || !($registry = Type::getTypeRegistry())->has($columnType)) {
+            return $propertyType;
         }
 
-        return $propertyType;
+        $reflection = new \ReflectionClass(($registry->get($columnType))::class);
+
+        $returnType = $reflection->getMethod('convertToPHPValue')->getReturnType();
+
+        /*
+         * we do not support union and intersection types
+         */
+        if (!$returnType instanceof \ReflectionNamedType) {
+            return null;
+        }
+
+        return $returnType->isBuiltin() ? $returnType->getName() : '\\'.$returnType->getName();
     }
 
     /**
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Doctrine;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Types\Type;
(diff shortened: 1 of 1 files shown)


IndraGunawan:
CI error was not related

IndraGunawan:
@jrushlow @weaverryan any chance to review this?

src/Doctrine/DoctrineHelper.php
```diff
@@ -271,6 +272,19 @@ public static function getPropertyTypeForColumn(string $columnType): ?string
             'ulid' => '\\'.Ulid::class,
             default => null,
         };
+
+        if (null === $propertyType && Type::getTypeRegistry()->has($columnType)) {
+            $reflection = new \ReflectionClass(\get_class(Type::getTypeRegistry()->get($columnType)));
```
jrushlow, line 277:
```suggestion
            $reflection = new \ReflectionClass(($registry->get($columnType))::class);
```

src/Doctrine/DoctrineHelper.php
```diff
@@ -271,6 +272,19 @@ public static function getPropertyTypeForColumn(string $columnType): ?string
             'ulid' => '\\'.Ulid::class,
             default => null,
         };
+
+        if (null === $propertyType && Type::getTypeRegistry()->has($columnType)) {
```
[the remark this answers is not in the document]
jrushlow: Lets exit early to avoid a bunch of nested conditionals

src/Doctrine/DoctrineHelper.php
```diff
@@ -271,6 +272,19 @@ public static function getPropertyTypeForColumn(string $columnType): ?string
...
+            $reflection = new \ReflectionClass(\get_class(Type::getTypeRegistry()->get($columnType)));
+            if ($reflection->hasMethod('convertToPHPValue') && $returnType = $reflection->getMethod('convertToPHPValue')->getReturnType()) {
```
jrushlow, line 278: We should be able to drop this conditional all together. `getType()` always returns a `Type` and abstract `Type` has the convert method.

https://github.com/doctrine/dbal/blob/f3defbb4f88a7472314746f122f6d4d84615d424/src/Types/TypeRegistry.php#L41

src/Doctrine/DoctrineHelper.php
```diff
@@ -271,6 +272,19 @@ public static function getPropertyTypeForColumn(string $columnType): ?string
...
+                if ($returnType->isBuiltin()) {
+                    $propertyType = $returnType->getName();
+                } else {
+                    $propertyType = '\\'.$returnType->getName();
+                }
```
jrushlow, line 283: We could do something like `return $returnType->isBuitin() ? $returnType->getName() : ......`

jrushlow (changes requested):
This looks good. A couple minor comments to improve readability.

jrushlow (approved):
Thank you @IndraGunawan!

PR #1414, opened 2024-01-05
Proper annotation for generated validator constraint

zorn-v:
Currently, if you are using phpstan in your project you'll get

> Access to an undefined property Symfony\Component\Validator\Constraint::$message

error on any new generated validator


drieschel (commented):
I tested it and it works. :thumbsup:

Do you think it makes sense also to add the type to the `$message` property in `src/Resources/skeleton/validator/Constraint.tpl.php` in this mr?

`    public string $message = 'The value "{{ value }}" is not valid.';`

weaverryan (approved):

weaverryan:
Thanks @zorn-v!

PR #1416, opened 2024-01-07
Adding `ORM\JoinTable` attribute to the `make:entity` command's `ManyToMany` field type

ramity:
You can add two ManyToMany relations to the same target entity, but the generated code will be invalid. You'll hit a "Table already exists" error because of how the two relations will share the same join table name under the hood. I whipped up these changes to give the user the power to manually specify the join table name during the prompt.

This PR adds an additional check during the make:entity command's ManyToMany field type steps by asking the user if they'd like to specify a join table, allows them to do so, and adds the ORM\JoinTable attribute automagically.


src/Doctrine/EntityRelation.php
```diff
@@ -27,6 +27,7 @@ final class EntityRelation
     private bool $isSelfReferencing = false;
     private bool $orphanRemoval = false;
     private bool $mapInverseRelation = true;
+    private $joinTableName;
```
OskarStark, line 30:
```suggestion
    private ?string $joinTableName;
```

OskarStark (commented):
Can you add a test with a join table?

src/Doctrine/EntityRelation.php
ramity: I added a test case with a custom join table and added the above suggested change.

src/Maker/MakeEntity.php
```diff
@@ -619,6 +619,20 @@ function ($name) use ($targetClass) {
             $relation->setMapInverseRelation($mapInverse);
         };
 
+        $askJoinTableName = function (EntityRelation $relation) use ($io) {
+            $joinTableDecision = $io->confirm(
+                'Do you want to specify a join table? You may want to do this if you plan on having multiple many-to-many relations to the same entity.',
```
ramity, line 624: Open to suggestions on rewording

src/Maker/MakeEntity.php
```diff
@@ -619,6 +619,20 @@ function ($name) use ($targetClass) {
...
+                false
+            );
+
+            if ($joinTableDecision) {
+                $relation->setJoinTableName($io->ask(
+                    'What should the join table be named?',
```
ramity, line 630: Open to suggestions on rewording

ramity:
I reverted the previous commit since it was made under the impression I broke something, but it seems like something related to nikic/php-parser. I'll revisit ensuring test coverage for these changes once that's fixed up.

src/Maker/MakeEntity.php
weaverryan: In the current placement, this question will be asked for all relation types, right? But we only want it for `ManyToMany`

For the wording, what about just one question:

> A join table will be created for the relationship. You can name this, or let Doctrine choose a name for you.
> Join table name (default: set automatically): 

If they hit enter (leave it blank), then we know to name it automatically.

ramity: This area just defines and stores the function as a variable. The logic within isn't actually ran until the `$askJoinTableName($relation);` statement is executed later down in the code [here](https://github.com/symfony/maker-bundle/pull/1416/files#diff-dc199e4a5d406f1d06cbb6b936027bccbbaf2e4e1ec7be7d5f54bd6393fa1955R725).

As much as I'd prefer the lower complexity of one question, we'll want to give the user the ability to say no to the addition of a join table. Cases like [bidirectional many-to-many](https://www.doctrine-project.org/projects/doctrine-orm/en/2.17/reference/association-mapping.html#many-to-many-bidirectional) and [self referencing many-to-many](https://www.doctrine-project.org/projects/doctrine-orm/en/2.17/reference/association-mapping.html#many-to-many-self-referencing) require the use of a ManyToMany attribute without a JoinTable attribute on one side.

PR #1417, opened 2024-01-08
Compatibility to dependency php parser v5

drieschel:
This PR adds compatibility for dependency `nikic/php-parser` v5. 

Not sure if it works without additional changes, lets see if the tests will succeed.

**Edit:** Looks like a new major (or at least minor) version is required in order to make it compatible with v5, because v4 will not be compatible anymore.

**Edit2:** I did some refactoring and now it is working with v4 and v5.

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -23,11 +23,11 @@ use Doctrine\ORM\Mapping\OneToMany;
 use Doctrine\ORM\Mapping\OneToOne;
 use PhpParser\Builder;
 use PhpParser\BuilderHelpers;
-use PhpParser\Lexer;
 use PhpParser\Node;
 use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitor;
 use PhpParser\Parser;
+use PhpParser\ParserFactory;
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\Doctrine\BaseCollectionRelation;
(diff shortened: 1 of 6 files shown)


src/Util/ClassSourceManipulator.php
```diff
@@ -64,14 +65,21 @@ public function __construct(
...
     ) {
-        $this->lexer = new Lexer\Emulative([
-            'usedAttributes' => [
-                'comments',
-                'startLine', 'endLine',
-                'startTokenPos', 'endTokenPos',
-            ],
-        ]);
-        $this->parser = new Parser\Php7($this->lexer);
+        if (class_exists(PhpVersion::class)) {
```
weaverryan, line 68: For the parts of the code that are for v4 of the parser, could you add a little:

```
/* @legacy Support for nikic/php-parser v4 */
```

weaverryan (commented):
This is fantastic! Thank you for taking on this important chore - huge help!

src/Util/ClassSourceManipulator.php
drieschel: Should I also add this comment to the property `Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator::$lexer`? Because it will be not needed anymore after v4 support is dropped.

weaverryan (approved):

derrabus:
@weaverryan Can we please release this change? 🙏🏻

jrushlow:
> @weaverryan Can we please release this change? 🙏🏻

done

derrabus:
Thanks!

PR #1418, opened 2024-01-13
Determine events map dynamically

HypeMC:
The current events map is missing some events, such as the `command.signal` event. It also references some legacy event classes, like the `GetResponseEvent` class, which are no longer necessary since the minimum required version does not support these classes anymore.

This PR modifies the `EventRegistry` so that the events map is determined dynamically. The added benefit of this approach is that only the events that are actually available are listed.


weaverryan:
This is so much better - thanks @HypeMC!

PR #1419, opened 2024-01-23
[make:controller] remove CodeExtension::getFileLink() call when generating twig templates

### LINKED ISSUE #1415: Replace or remove "file_link" filter ?
As `file_link` is now considered internal, using it in the generated template of `make:controller` (via [GeneratorTwigHelper](https://github.com/symfony/maker-bundle/blob/main/src/GeneratorTwigHelper.php#L84)) has been linked to an issue: https://github.com/symfony/symfony/issues/53431 

If my first feeling is correct, we should stop using it from userland (even in dev / temporary mode).. so how would you replace it ? 

Just waiting a confirmation before attacking the PR :)

jrushlow:
TwigBridge's `CodeExtension` was made `@internal` via https://github.com/symfony/twig-bridge/commit/76b43934b95308231fc82a4905fb10eeac31b310

- removes file link anchor in generated template to the controller and template
- deprecates `GeneratorTwigHelper::getFileLink()`

fixes #1415

### DIFF
--- a/src/Resources/skeleton/controller/twig_template.tpl.php
+++ b/src/Resources/skeleton/controller/twig_template.tpl.php
@@ -11,8 +11,8 @@
 
     This friendly message is coming from:
     <ul>
-        <li>Your controller at <code><?= $helper->getFileLink("$root_directory/$controller_path", "$controller_path"); ?></code></li>
-        <li>Your template at <code><?= $helper->getFileLink("$root_directory/$relative_path", "$relative_path"); ?></code></li>
+        <li>Your controller at <code><?= $root_directory ?>/<?= $controller_path ?></code></li>
+        <li>Your template at <code><?= $root_directory ?>/<?= $relative_path ?></code></li>
     </ul>
 </div>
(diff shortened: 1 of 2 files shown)


src/Resources/skeleton/controller/twig_template.tpl.php
```diff
@@ -11,8 +11,8 @@
 
     This friendly message is coming from:
     <ul>
-        <li>Your controller at <code><?= $helper->getFileLink("$root_directory/$controller_path", "$controller_path"); ?></code></li>
-        <li>Your template at <code><?= $helper->getFileLink("$root_directory/$relative_path", "$relative_path"); ?></code></li>
+        <li>Your controller at <code><?= $root_directory ?>/<?= $controller_path ?></code></li>
+        <li>Your template at <code><?= $root_directory ?>/<?= $relative_path ?></code></li>
```
weaverryan, line 15: Minor, but I suppose we should deprecate `GeneratorTwigHelper::getFileLink()`

PR #1420, opened 2024-01-23
[ci] handle postgres 16 recipe support in .env

jrushlow:
See - https://github.com/symfony/recipes/pull/1284


jrushlow:
8.2 (7.0) CI failure handled in #1419

Issue #1371, opened 2023-10-12
Generator example from DocBlock will not work


Antarian:
https://github.com/symfony/maker-bundle/blob/0f060e9ee6fbc79626fef16004a0bb8b0bee32ab/src/Generator.php#L137

In current implementation name either needs to be absolute, or to make it work generator code needs to change.
E.g. change line 150
https://github.com/symfony/maker-bundle/blob/0f060e9ee6fbc79626fef16004a0bb8b0bee32ab/src/Generator.php#L150
to:
```
try {
    Validator::classDoesNotExist($name);
    $className = rtrim($fullNamespacePrefix, '\\').'\\'.Str::asClassName($name, $suffix);
} catch (RuntimeCommandException $e) {
    $className = $name;
}
```
This would have edge case if there is match with class even if $name value is only partial. But that should be very rare.
Like `Prototype\\DummyEntity` may exist and we really want "App\\Entity\\Prototype\\DummyEntity".

weaverryan:
I'm open to a PR!

PR #1422, opened 2024-01-23
Generator verifies if class exists before altering class name

Antarian:
Generator was changing namespace of the class with Prefixed namespace, even in the case it was absolute.
Now if class exists, even if it is not starting with backslash `\`, it would be accepted.

https://github.com/symfony/maker-bundle/issues/1371


jrushlow (approved):
Thank you @Antarian!

GPHemsley:
This seems to have caused regressions:
* #1479
* #1566

PR #1424, opened 2024-01-25
[ci] windows tests on GitHub actions

jrushlow:
- tests that use `symfony/panther` are skipped on windows. The windows runner does not chrome/firefox/etc pre-installed.

- fixes regex that gets a list of files create or updated by MakerBundle (used to test those files were actually created). the file path in the output differs on linux, github actions linux images, macos, & windows. The new regex pattern handles all of those cases. This regex is only used in Maker tests

- adds a test for that regex to protect from hidden test failures in the future.

### DIFF
--- a/src/Test/MakerTestEnvironment.php
+++ b/src/Test/MakerTestEnvironment.php
@@ -22,6 +22,8 @@ use Symfony\Component\Process\InputStream;
  */
 final class MakerTestEnvironment
 {
+    public const GENERATED_FILES_REGEX = '#(?:created|updated):\s(?:.*\\\\)*(.*\.[a-z]{3,4}).*(?:\\\\n)?#ui';
+
     private Filesystem $fs;
     private bool|string $rootPath;
     private string $cachePath;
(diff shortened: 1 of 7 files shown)


src/Test/MakerTestEnvironment.php
```diff
@@ -217,9 +217,14 @@ public function getGeneratedFilesFromOutputText(): array
...
+        //        preg_match_all('#(created|updated): (]8;;[^]*\\\)?(.*?)(]8;;\\\)?\n#iu', $output, $matches, \PREG_PATTERN_ORDER);
+        preg_match_all('#(created|updated):.*(src.*\.php)#iu', $output, $matches, \PREG_PATTERN_ORDER);
+        //        dump($output, $matches);
+        $result = array_map('trim', $matches[2]);
 
-        return array_map('trim', $matches[3]);
+        //        dump($result);
+
+        return $result;
     }
```
jrushlow, line 228: note to self: this avoids windows / linux file path differences / conflicts?

tests/RegexTest.php
```diff
@@ -0,0 +1,61 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Tests;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\MakerBundle\Test\MakerTestEnvironment;
+
+class RegexTest extends TestCase
```
jrushlow, line 17: Calling this `MakeTestEnvironmentTest` was not compatible with PHPUnit `--filter` (test wouldn't run) and I'm not sure it was running w/ just `simple-phpunit` && no flags either...

weaverryan (approved):
👏

PR #1425, opened 2024-01-26
[dx] require >= PHPUnit 9.6, update phpunit schema, deprecate skipOnSymfony7() 

jrushlow:
- deprecates `MakerTestDetails::skipOnSymfony7()` methods that were meant to be used internally.

- add ability to skip tests locally (e.g. mercure tests can be skipped by setting `MAKER_SKIP_MERCURE_TEST` to `true` in `phpunit.xml`. Previously that var had to be set in the `$_SERVER` super global.) This plays into Windows tests on GitHub actions.

- fix outdated / deprecated `phpunit.xml.dist` schema / attributes

- require PHPUnit 9.6+ in tests (Only have to worry about 1 phpunit.xml schema. )

- adds a `MakerTestDetails::skipTest()` method & logic to replace the now deprecated `skipOnSymfony7()` methods. Ideally, we'll keep the new method for future use even if we're not using it now. `skipTest()` is marked `@internal` for the time being.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
         "composer/semver": "^3.0",
         "doctrine/doctrine-bundle": "^2.5.0",
         "doctrine/orm": "^2.10.0",
-        "phpunit/phpunit": "^9.6",
         "symfony/http-client": "^6.4|^7.0",
         "symfony/phpunit-bridge": "^6.4.1|^7.0",
         "symfony/security-core": "^6.4|^7.0",
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "composer/semver": "^3.0",
         "doctrine/doctrine-bundle": "^2.5.0",
         "doctrine/orm": "^2.10.0",
+        "phpunit/phpunit": "^9.6",
         "symfony/http-client": "^6.4|^7.0",
         "symfony/phpunit-bridge": "^6.4.1|^7.0",
         "symfony/security-core": "^6.4|^7.0",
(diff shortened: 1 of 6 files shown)


tests/Maker/MakeEntityTest.php
```diff
@@ -37,12 +37,26 @@ private function createMakeEntityTest(bool $withDatabase = true): MakerTestDetai
...
+            // This test is skipped, don't worry about persistence
+            return $this->createMakerTest()
+                ->skipTest('MAKER_SKIP_MERCURE_TEST set to true')
+            ;
+        }
+
+        // @legacy - MakeEntity uses ux-turbo-mercure (archived), it needs to use ux-turbo (mercure built in) for Symfony 7.0
+        if ('7.0.x-dev' === $_SERVER['SYMFONY_VERSION']) {
+            return $this->createMakerTest()
+                ->skipTest('symfony/ux-turbo-mercure is not supported on Symfony 7.')
```
jrushlow, line 50: Chicken or the egg... `MakeEntity` needs to be refactored to use `ux-turbo` and at first glance doesn't appear to be a "1 for 1" swap with `ux-turbo-mercure`. To keep the entire test suite from blowing up due to a deprecation - this skip conditional was moved here for now.

composer.json
```diff
@@ -30,6 +30,7 @@
         "composer/semver": "^3.0",
         "doctrine/doctrine-bundle": "^2.5.0",
         "doctrine/orm": "^2.10.0",
+        "phpunit/phpunit": "^9.6",
```
weaverryan, line 33: Does `phpunit-bridge` read / respect this? We run `vendor/bin/simple-phpunit`, which downloads its own phpunit. Does this signal to download 9.6? I thought the version was done with an env var.

tests/Maker/MakeEntityTest.php
weaverryan: Ok for now for me

weaverryan (changes requested):

composer.json
jrushlow: good catch! added the `SYMFONY_PHPUNIT_VERSION` with `9.6` to `phpunit.xml.dist` instead....  `simple-phpunit` picks up on that locally and on ci when doing the install.

PR #1433, opened 2024-02-03
[make:crud] Remove extra dot

annechko:
Hi! Noticed an extra dot after a question mark in tests question

```
Do you want to generate tests for the controller?. [Experimental] (yes/no) [no]:
```


jrushlow (approved):
Good catch! Thank you @annechko

Issue #1437, opened 2024-02-03
make:user ends with "Undefined constant Doctrine\DBAL\Types\Types::ARRAY" error


Dreadnoth:
Hello

During starting a blank new SF 7.0 project , first action that I've did was to secure the newly created app using [this link](https://symfony.com/doc/current/security.html). After running ```php bin/console make:user``` and selecting defaults I've received error:
![error_maker](https://github.com/symfony/maker-bundle/assets/15017947/aa9d77b9-62a7-49c3-874e-95d01a1ae43d) (for SEO: " Undefined constant Doctrine\DBAL\Types\Types::ARRAY")

I'm attaching project [composer.json](https://github.com/symfony/maker-bundle/files/14154446/composer.json)

jpmmartin:
I just installed a symfony 6.4 LTS project and am getting the same error,
Execution of command `php bin/console make:entity`
My php version is 8.3
My composer.json:

```
{
    "type": "project",
    "license": "proprietary",
    "minimum-stability": "stable",
    "prefer-stable": true,
    "require": {
        "php": ">=8.1",
        "ext-ctype": "*",
        "ext-iconv": "*",
        "api-platform/core": "^3.2",
        "doctrine/doctrine-bundle": "^2.11",
        "doctrine/doctrine-migrations-bundle": "^3.3",
        "doctrine/orm": "^3.0",
        "nelmio/cors-bundle": "^2.4",
        "phpdocumentor/reflection-docblock": "^5.3",
        "phpstan/phpdoc-parser": "^1.25",
        "symfony/asset": "6.4.*",
        "symfony/console": "6.4.*",
        "symfony/dotenv": "6.4.*",
        "symfony/expression-language": "6.4.*",
        "symfony/flex": "^2",
        "symfony/framework-bundle": "6.4.*",
        "symfony/property-access": "6.4.*",
        "symfony/property-info": "6.4.*",
        "symfony/runtime": "6.4.*",
        "symfony/security-bundle": "6.4.*",
        "symfony/serializer": "6.4.*",
        "symfony/twig-bundle": "6.4.*",
        "symfony/uid": "6.4.*",
        "symfony/validator": "6.4.*",
        "symfony/yaml": "6.4.*"
    },
    "config": {
        "allow-plugins": {
            "php-http/discovery": true,
            "symfony/flex": true,
            "symfony/runtime": true
        },
        "sort-packages": true
    },
    "autoload": {
        "psr-4": {
[...]

PR #1438, opened 2024-02-03
Remove mentions of undefined Doctrine constants Types::ARRAY and Types::OBJECT

njutn95:
Fixes https://github.com/symfony/maker-bundle/issues/1437

This PR removes the use of `Types::ARRAY` and `Types::OBJECT` as they were removed in Doctrine v4.0 (see https://github.com/doctrine/dbal/commit/763b4ad34134bbdb3a8bade640e33f7df2ecfa73).

### DIFF
--- a/src/Doctrine/DoctrineHelper.php
+++ b/src/Doctrine/DoctrineHelper.php
@@ -259,14 +259,14 @@ final class DoctrineHelper
     {
         return match ($columnType) {
             Types::STRING, Types::TEXT, Types::GUID, Types::BIGINT, Types::DECIMAL => 'string',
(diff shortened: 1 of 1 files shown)


derrabus:
What if someone still uses DBAL 3 and relies on those deprecated types?

stof:
I suggest putting the type name as a string for those instead of removing them entirely for those legacy types (as done for the `uuid` and `ulid` types)

njutn95:
You're right! I've added backward compatibility. I just thought that adding the BC for something that has already been deprecated in the past is not really worth it in the maker bundle.

derrabus (approved):

weaverryan:
Thanks @njutn95! We're working on the test failures - it looks like the started exploding with the new Doctrine releases 🤔

PR #1439, opened 2024-02-05
[make:*] improve support for ORM 3.x / DBAL 4.x

jrushlow:
- bumps the required `doctrine/orm` version to `^2.15|^3`
- bumps minimum `doctrine/doctrine-bundle` to `^2.10` 
- handles new Mapping classes for entites. (`FieldMapping`, `ManyToOneMapping`, etc...)
- adds a `MAKER_PROCESS_TIMEOUT` env that can be set to `null` or a float. This overrides the default `Process::timeout` value of `60` for shell commands in test. (Allows step debugging w/ XDebug)
- tmp disable Mercure tests on CI. `ux/turbo` uses `Doctrine\Common\Util\ClassUtils` which no longer exists. (Fails all of our tests) See comment in review.
- adds legacy tests for ORM 2.x entity attributes that  "reversed" property order compared to ORM 3.x
- `ORM\..\DisconnectedClassMetadataFactory` is removed in `orm 3.0` -> set the `StaticReflectionService::class` manually when getting metadata for disconnected entities. (e.g. xml mapping) -  https://github.com/doctrine/orm/pull/9704


phpunit.xml.dist
```diff
@@ -19,6 +19,8 @@
         <env name="SYMFONY_PHPUNIT_VERSION" value="9.6" />
         <env name="MAKER_SKIP_MERCURE_TEST" value="false"/>
         <env name="MAKER_SKIP_PANTHER_TEST" value="false" />
+<!--    Overrides process timeout when step debugging -->
+<!--    <env name="MAKER_PROCESS_TIMEOUT" value="null" /> -->
```
jrushlow, line 23: Allows for debugging `Process::fromShellCommandLine()` which sets the process `$timeout` to 60 seconds

src/Util/ClassSource/Model/ClassProperty.php
```diff
@@ -0,0 +1,98 @@
...
+
+use Doctrine\ORM\Mapping\FieldMapping;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+
+/**
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+final class ClassProperty
```
jrushlow, line 22: We'll be able to do additional refactorings in CSM that will further utilize value object.. e.g. `addProperty()`

.github/workflows/ci-linux.yaml
```diff
@@ -11,6 +11,7 @@ on:
 env:
     PHPUNIT_FLAGS: "-v"
     SYMFONY_PHPUNIT_DIR: "$HOME/symfony-bridge/.phpunit"
+    MAKER_SKIP_MERCURE_TEST: 1
```
jrushlow, line 14: `ux-turbo` uses `Doctrine\Commom\Util\ClassUtils::class` which no longer exists. As such, 2 `MakeEntity` tests fail that use Mercure. Keep this here && create an maker bundle issue to fix? Or live with (for now) the red x's on all of our tests?

weaverryan: Keep this for now. Then we'll fix it in ux-turbo at some point.

weaverryan (approved):
This looks like very detailed work & solid 🦾. If any details were missed, the tests should have caught them.

PR #1445, opened 2024-02-08
[make:entity] use ux-turbo instead of ux-turbo-mercure

jrushlow:


docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+services:
```
jrushlow, line 1: for local dev's to be able to run mercure tests

tests/Maker/MakeEntityTest.php
```diff
@@ -44,18 +44,12 @@ private function createMakeEntityTestForMercure(): MakerTestDetails
...
-            ;
-        }
-
         return $this->createMakeEntityTest()
             ->preRun(function (MakerTestRunner $runner) {
                 // installed manually later so that the compatibility check can run first
-                $runner->runProcess('composer require symfony/ux-turbo-mercure');
+                $runner->runProcess('composer require symfony/ux-turbo');
             })
+            ->addExtraDependencies('mercure', 'twig')
```
jrushlow, line 52: `ux-turbo` installs `stimulus-bundle` which requires `twig`. But the twig recipe doesn't get generated in the `tmp/cache` project. adding it as an extra depend fixes that.

`mercure` is needed obviously to pub/sub changes

weaverryan (approved):
Well done

PR #1449, opened 2024-02-15
[tests] fix broken assignment operator - addExtraDependencies()

jrushlow:
`+=` is for arithmetic  and doesn't merge 2 arrays as intended.

https://3v4l.org/FL7cV 

_broke out into separate PR for changelog entry. fixes tests in #1445_


OskarStark:
We should add a testcase here

PR #1450, opened 2024-02-15
housekeeping - add rector rules - one step closer to static analysis

jrushlow:
- `StrContainsRector` && `ChangeSwitchToMatchRector` are purposely disabled - codesmell / bugs exist in CSM that are outside the scope of this PR

- Running rector on `tests/` is very possible, but needs to be done in a separate PR (breaks some tests)

- once the above is done, we can remove all of the `rules([...])` declarations and enable just `sets([LevelSetList::UP_TO_PHP_81])`

Rectors Applied:
- Class Property Assignment to Constructor Promotion
```diff
class
{
-    private string $property
    public function __constructor(
+        private string $property
    )
}
```
- Closure to Arrow Functions
```diff
- function(static function () { return strtoupper() })
+ function(static fn () => strtoupper())
```
- First class callables
```diff
- $io->ask(validation: [Validator::class, 'notBlank'])
+ $io->ask(validation: Validator::notBlank(...))
```
- Remove unused variables in catch
```diff
- } catch(\RuntimeException $e) {};
+ } catch(\RuntimeException) {};
```

- Restore Default null to nullable type property
```diff
- private ?string $var
+ private ?string $var = null
```

- StrEndsWith
```diff
- $x = 'howdy' === substr($var, -5)
+ $x = str_ends_with($var, 'howdy')
```


rector.php
```diff
@@ -0,0 +1,52 @@
...
+
+    $config->sets([
+//        LevelSetList::UP_TO_PHP_81,
+    ]);
+
+    $config->skip([
+        NullToStrictStringFuncCallArgRector::class,
+        ReadOnlyPropertyRector::class,
+    ]);
+};
```
weaverryan, line 52: Do we actually need rector in the project? Or can you just use it locally this one time, and going forward, we rely on something like php-cs-fixer or phpstan?

weaverryan (commented):
I'm cool with all these changes

rector.php
jrushlow: I think having it in the project will be beneficial in the long run. `rector.php` is just the config file (unfortunately rector doesnt pickup on `rector.dist.php` by default) - similar to `.php-cs-fixer.dist.php`. Rector itself lives in `tools/rector` like `tools/php-cs-fixer` does - it doesnt cause composer dependency headaches.

The `php-cs-fixer` rulesets for PHP are pretty lacking in terms of upgrading the codebase to a language level. Rector is for more robust in this aspect. The same holds true for `psalm` - its "auto fix" is limited in scope leading to more manual work. && `phpstan` auto fixing appears to be a paid feature.  I would however like to get `phpstan` working against the codebase (in a future pr). I've seen a handful of codesmells that a static analyser would have flagged for us long ago..

having the config file in the codebase at a minimum will be a leg up if/when we enable this in CI as a sniff tests..

weaverryan: Unless having rector in the project is standard elsewhere (and I'm not aware of anything), I'm 👎  on including it. The process should be:

A) A one-time rector usage to fix the project
B) Some tool to fail in CI if we're violating standards. I don't care as much about the autofixing aspect. If php-cs-fixer is missing a check - like for property promotion - I think we just track that manually in PR's.

In short, I don't want to do anything non-standard or extra-forward in this repo. I feel like these files will get forgotten and just get out of date. Sorry :)

OskarStark (approved):

PR #1451, opened 2024-02-15
[make:entity] Use `const` for `ASC` in ORDER BY

ThomasLandauer:
I haven't really tested this - waiting for some feedback first...

### DIFF
--- a/src/Resources/skeleton/doctrine/Repository.tpl.php
+++ b/src/Resources/skeleton/doctrine/Repository.tpl.php
@@ -47,7 +47,7 @@ class <?= $class_name; ?> extends ServiceEntityRepository<?= $with_password_upgr
 //        return $this->createQueryBuilder('<?= $entity_alias; ?>')
 //            ->andWhere('<?= $entity_alias; ?>.exampleField = :val')
 //            ->setParameter('val', $value)
-//            ->orderBy('<?= $entity_alias; ?>.id', 'ASC')
+//            ->orderBy('<?= $entity_alias; ?>.id', Criteria::ASC)
 //            ->setMaxResults(10)
 //            ->getQuery()
 //            ->getResult()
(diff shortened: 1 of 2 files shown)


src/Resources/skeleton/doctrine/Repository.tpl.php
```diff
@@ -47,7 +47,7 @@ public function upgradePassword(<?= sprintf('%s ', $password_upgrade_user_interf
 //        return $this->createQueryBuilder('<?= $entity_alias; ?>')
 //            ->andWhere('<?= $entity_alias; ?>.exampleField = :val')
 //            ->setParameter('val', $value)
-//            ->orderBy('<?= $entity_alias; ?>.id', 'ASC')
+//            ->orderBy('<?= $entity_alias; ?>.id', Criteria::ASC)
```
jrushlow, line 50: in general terms, I'm not opposed to using a `const` or `enum` in place of strings. My only concern would be, are `doctrine/common` / `doctrine/common/collections` classes on the chopping block in `dbal` 4||5. E.g. where we would have to revert this behavior in the immediate future.

 I took a quick peak in `dbal/src/QueryBuilder` and did not see any evidence of such a possibility. But did not dig through either of the repos for open prs/issues on the matter.

If that's not an issue, green tests - I'm all for it.

jrushlow (changes requested):

ThomasLandauer:
Using strings there is about to get deprecated, in favor of a new enum, see https://github.com/doctrine/collections/pull/389
So I'm closing here for now - I'll come up with a new PR once Doctrine has released it.
The main question here will then be: In which version should the switch be done?

jrushlow:
Awesome! In the new PR - we'd have todo some sort of conditional to support the old strings `if(!exists(DoctrineEnum))` Looking forward to the upcoming changes.

PR #1455, opened 2024-02-18
[make:crud] Removal of the `createView()` method, as it is no longer required to render the form

### LINKED ISSUE #1453: [BUG] [Make:registration-form] Form errors are not handled
# How to reproduce

Use attribute "no validate" in the registration form.

```
{{ form_start(registrationForm, {
    'attr': {
        'novalidate': 'novalidate'
    }
}) }}

{# ... #}

{{ form_end(registrationForm) }}
```

Nothing appears when clicking "S'inscrire" (no violations or form errors in the debug bar either):

Capture d’écran 2024-02-16 à 17 00 36

Expected behavior :

Capture d’écran 2024-02-16 à 17 03 08

# How to fix

Replace :
```
return $this->render('registration/register.html.twig', [
    'registrationForm' => $form->createView(),
]);
```
with :
```
return $this->render('registration/register.html.twig', [
    'registrationForm' => $form,
]);
```

mdoutreluingne:
Fixes #1453

### DIFF
--- a/src/Resources/skeleton/registration/RegistrationController.tpl.php
+++ b/src/Resources/skeleton/registration/RegistrationController.tpl.php
@@ -58,7 +58,7 @@ class <?= $class_name; ?> extends AbstractController
         }
 
         return $this->render('registration/register.html.twig', [
-            'registrationForm' => $form->createView(),
+            'registrationForm' => $form,
         ]);
     }
 <?php if ($will_verify_email): ?>
--- a/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
+++ b/src/Resources/skeleton/resetPassword/ResetPasswordController.tpl.php
@@ -33,7 +33,7 @@ class <?= $class_name ?> extends AbstractController
         }
 
         return $this->render('reset_password/request.html.twig', [
-            'requestForm' => $form->createView(),
+            'requestForm' => $form,
         ]);
     }
 
@@ -109,7 +109,7 @@ class <?= $class_name ?> extends AbstractController
         }
 
         return $this->render('reset_password/reset.html.twig', [
-            'resetForm' => $form->createView(),
+            'resetForm' => $form,
         ]);
     }
 


tests/fixtures/make-registration-form/tests/it_generates_registration_with_entity_and_authenticator.php
```diff
@@ -54,7 +54,7 @@ public function testRegistrationValidationError()
         $form['registration_form[plainPassword]'] = 'foo';
         $client->submit($form);
 
-        $this->assertSame(200, $client->getResponse()->getStatusCode());
+        $this->assertSame(422, $client->getResponse()->getStatusCode());
```
jrushlow, line 57: This smells funny - there should have been a test failure before _this_ PR to show that `422` was the expected status code. Makes me think there is something up with the test suite... 

Can you push a commit that changes this back to `200` & then push _another_ commit to change this back to `422` so we can see the tests fail in CI? Thank you in advance

jrushlow (changes requested):
Awesome! One comment though...

tests/fixtures/make-registration-form/tests/it_generates_registration_with_entity_and_authenticator.php
mdoutreluingne: It's actually quite strange, I myself didn't understand how the tests could work with a status code of 200 in the case of an invalid form.

But I didn't notice straight away that the status of the code was wrong when I pushed my first commit, it's thanks to the CI tests that failed and told me that it was expecting a code with status 422 in the event of an invalid form.

jrushlow: I've been chewing on this all morning and _think_ the reason why we didn't see this fail before your PR has to do with an exception that was caught in the test but didn't bubble up. MakerBundles test setup is pretty complex. when we run `vendor/bin/simple-phpunit` each `Make*Test` actually creates a new "project" -> runs the maker -> runs tests within the project -> reports back to the original `Make*Test`... sometimes things get lost in translation... Anyhow, thats out of scope for this PR.. :)

jrushlow (approved):
Thank you @mdoutreluingne!

PR #1460, opened 2024-02-21
[make:user] handle ORM\Column.unique deprecation - use ORM\UniqueConstrain

maelanleborgne:
As a result of https://github.com/doctrine/dbal/pull/5656, we should stop using the `unique` property of the `ORM\Column` attribute in the `make:user` command and prefer the `ORM\UniqueConstraint`.

This PR should solve this problem 🙂

### DIFF
--- a/src/Security/UserClassBuilder.php
+++ b/src/Security/UserClassBuilder.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Security;
 
 use PhpParser\Node;
+use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Util\ClassSource\Model\ClassProperty;
 use Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator;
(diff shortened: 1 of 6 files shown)


src/Security/UserClassBuilder.php
```diff
@@ -259,4 +261,17 @@ private function addEraseCredentials(ClassSourceManipulator $manipulator): void
 
         $manipulator->addMethodBuilder($builder);
     }
+
+    private function addUniqueConstraint(ClassSourceManipulator $manipulator, UserClassConfiguration $userClassConfig): void
+    {
+        if ($userClassConfig->isEntity()) {
```
jrushlow, line 267: Let's exit early when possible.
```suggestion
        if (!$userClassConfig->isEntity()) {
            return;
        }
        
```

jrushlow (changes requested):
Awesome! a minor comment

jrushlow (approved):
Awesome! Thank you @maelanleborgne

PR #1463, opened 2024-02-27
[make:serializer:normalizer] stop using CacheableSupportsMethodInterface

jrushlow:
- `CacheableSupportsMethodInterface` is deprecated since Symfony 6.3 https://github.com/symfony/symfony/blob/3e296f1a924af7ceccaac96d1279c583d32c6ff1/src/Symfony/Component/Serializer/Normalizer/CacheableSupportsMethodInterface.php#L23

- fix normalizer property param name - refs #1273 

- removes legacy test dependency related to symfony <5.4 composer conflict

- use `[#Autowire()]` attribute instead of modifying `services.yaml` 

- import entity class instead of using FQCN in template

### DIFF
--- a/src/Maker/MakeSerializerNormalizer.php
+++ b/src/Maker/MakeSerializerNormalizer.php
@@ -17,11 +17,10 @@ use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Util\UseStatementGenerator;
-use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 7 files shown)


src/Maker/MakeSerializerNormalizer.php
```diff
@@ -94,34 +88,17 @@ public function configureDependencies(DependencyBuilder $dependencies): void
...
     }
 
-    private function generateNormalizer(string $className, Generator $generator): void
+    private function generateNormalizer(ClassNameDetails $className, Generator $generator): void
     {
         $useStatements = new UseStatementGenerator([
             NormalizerInterface::class,
-            CacheableSupportsMethodInterface::class,
+            Autowire::class,
+            sprintf('App\Entity\%s', str_replace('Normalizer', '', $className->getShortName())),
```
weaverryan, line 96: What if we use `class_exists` to see if this is actually a class that exists? If it is, we do what we do now. If it is NOT, we leave the related methods entirely empty in the normalizer and allow the user to fill them in.

weaverryan (changes requested):
Just one suggestion :)

src/Maker/MakeSerializerNormalizer.php
jrushlow: good idea! && done

PR #1466, opened 2024-02-28
[make:*] interactivly install composer dependencies

### LINKED ISSUE #1277: make:user complains missing symfony/security package, but package is abandoned
I want to use `php bin/console make:user` as in https://symfony.com/doc/current/security.html#the-user.

My composer.json:

````
{
  "require": {
    "symfony/runtime": "^6.2",
    "symfony/http-kernel": "^6.2",
    "symfony/framework-bundle": "^6.2",
    "symfony/console": "^6.2"
  },
  "require-dev": {
    "phpunit/phpunit": "^9.5",
    "symfony/maker-bundle": "^1.48"
  },
  "autoload": {
    "psr-4": {
      "App\\": "src/"
    }
  },
  "config": {
    "allow-plugins": {
      "symfony/runtime": true
    }
  }
}

````

When I make 

    php bin/console make:user

I recieve 

```
[ERROR] Missing package: to use the make:user command, run:                                                            
                                                                                                                        
         composer require security
```

When I do 

    composer require security

I receive 

````
Could not find package security.  
                                    
  Did you mean one of these?        
      symfony/security              
      symfony/security-acl          
      symfony/security-core         
      symfony/security-http         
      symfony/security-csrf
````

When I do 

    composer require symfony/security

I receive 

````
- Installing symfony/security (2.0.6): Extracting archive
2 package suggestions were added by new dependencies, use `composer suggest` to see details.
Package symfony/security is abandoned, you should avoid using it. No replacement was suggested.
Generating autoload files
56 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
Found 4 security vulnerability advisories affecting 1 package.
````

And I can still not run `php bin/console make:user`.

When I do 

    composer suggest 

I receive

````
phpunit/phpunit suggests:
 - ext-soap: *

symfony/console suggests:
 - symfony/lock
 - symfony/process

symfony/framework-bundle suggests:
 - ext-apcu: For best performance of the system caches
 - symfony/form: For using forms
 - symfony/property-info: For using the property_info service
 - symfony/serializer: For using the serializer service
 - symfony/validator: For using validation
 - symfony/web-link: For using web links, features such as preloading, prefetching or prerendering

symfony/http-kernel suggests:
 - symfony/browser-kit

symfony/security suggests:
 - symfony/class-loader: >=2.0
 - symfony/form: >=2.0

9 additional suggestions by transitive dependencies can be shown with --all
````

Now what should I install for being able to use `php bin/console make:user`?

- Tried also to install `symfony/security-core`, still complaining
- Tried also to install `doctrine/orm`, still complaining

### PR DESCRIPTION
**Pull request by jrushlow**
- Instead of throwing an exception if `$isLiveComponent` but `ux-live-component` isn't installed, let's ask if they want us to run composer for them...

On Success (keep it clean with no output from the process):

image

On Failure (we dump composer output):

image

fixes: #1277

### DIFF
--- a/src/Dependency/Model/AbstractClassDependency.php
+++ b/src/Dependency/Model/AbstractClassDependency.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Dependency\Model;
+
+/**
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+abstract class AbstractClassDependency
+{
+    public function __construct(
+        public string $className,
+        public string $composerPackage,
+        public bool $installAsRequireDev = false,
+    ) {
+    }
+}
--- a/src/Dependency/Model/AbstractClassDependency.php
+++ b/src/Dependency/Model/AbstractClassDependency.php
@@ -22,6 +22,7 @@ abstract class AbstractClassDependency
         public string $className,
         public string $composerPackage,
         public bool $installAsRequireDev = false,
+        public ?string $preInstallMessage = null,
     ) {
     }
 }
--- a/src/Maker/AbstractMaker.php
+++ b/src/Maker/AbstractMaker.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MakerBundle\Maker;
 
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\Dependency\DependencyManager;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
 use Symfony\Bundle\MakerBundle\MakerInterface;
 use Symfony\Component\Console\Command\Command;
@@ -53,5 +54,16 @@ abstract class AbstractMaker implements MakerInterface
             static::getCommandName(),
             $message
         );
+    }
+
+    public function configureComposerDependencies(DependencyManager $dependencyManager): void
+    {
+        // @TODO - method here in abstract prevents BC with signature added to `MakerInterface::class`
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies)
+    {
+        // @TODO - do we deprecate this method in favor of the one above. then remove in 2.x
+        // @TODO - still have plenty of work todo to determine if thats possible or a good idea...
     }
 }
--- a/src/Maker/MakeTwigComponent.php
+++ b/src/Maker/MakeTwigComponent.php
@@ -89,10 +89,6 @@ final class MakeTwigComponent extends AbstractMaker
         $name = $input->getArgument('name');
         $live = $input->getOption('live');
 
-        if ($live && !class_exists(AsLiveComponent::class)) {
-            throw new \RuntimeException('You must install symfony/ux-live-component to create a live component (composer require symfony/ux-live-component)');
-        }
-
         $factory = $generator->createClassNameDetails(
             $name,
             'Twig\\Components',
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: src/MakerInterface.php [hunk 1]
```diff
@@ -11,13 +11,15 @@
...
 
+use Symfony\Bundle\MakerBundle\Dependency\DependencyManager;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 
 /**
  * Interface that all maker commands must implement.
  *
  * @method static string getCommandDescription()
+ * @method        void   configureComposerDependencies(DependencyManager $dependencyManager)
```
jrushlow, line 22: Not a fan of touching the interface... better ideas are welcome..

src/Dependency/Model/AbstractClassDependency.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+namespace Symfony\Bundle\MakerBundle\Dependency\Model;
+
+/**
+ * @author Jesse Rushlow <jr@rushlow.dev>
+ *
+ * @internal
+ */
+abstract class AbstractClassDependency
```
jrushlow, line 19: Originally went with a single class that had `bool $isRequired` to determine if the dependency was required or optional. I went with an abstract base model and a `RequiredClassDependency` && `OptionalClassDependency` for better DX / code review readability... Atleast that was the intent..

src/Maker/AbstractMaker.php
```diff
@@ -54,4 +55,15 @@ protected function addDependencies(array $dependencies, ?string $message = null)
...
+    public function configureComposerDependencies(DependencyManager $dependencyManager): void
+    {
+        // @TODO - method here in abstract prevents BC with signature added to `MakerInterface::class`
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies)
+    {
+        // @TODO - do we deprecate this method in favor of the one above. then remove in 2.x
+        // @TODO - still have plenty of work todo to determine if thats possible or a good idea...
+    }
```
jrushlow, line 68: Not a fan of touching the abstraction - but it keeps BC in check w/ the interface change.

src/Maker/MakeTwigComponent.php
```diff
@@ -47,9 +51,37 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+        if (!$input->getOption('live')) {
+            $input->setOption('live', $io->confirm('Make this a live component?'));
+        }
+
+        if (!$input->getOption('live')) {
+            return;
+        }
+
+        // @TODO - with the dependencyManager in `Command` -> we can't use it outside of configure dependencies.....
+        $this->dependencyManager->installOptionalDependencies();
```
jrushlow, line 84: Not sure how to go about using the manager in other maker methods (aside from configureDepends...) Gonna play around with this a bit more. but other ideas are welcome.. Thought about some sort of setter in `AbstractMaker` or a constructor param in the individual makers that needs this...

src/Maker/MakeTwigComponent.php
```diff
@@ -47,20 +51,44 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
         ;
     }
 
-    public function configureDependencies(DependencyBuilder $dependencies): void
+    public function configureComposerDependencies(DependencyManager $dependencyManager): void
```
weaverryan, line 54: I'm wondering if we can simplify. A few things:

* Could we keep `DependencyBuilder`? It already records the required dependencies, which allows us to know if we need to throw an error. Now we're just going to use this to run `composer require` instead of throwing an error.
* In `DependencyBuilder`, we have this idea of an "optional dependency". I'm wondering if we can scrap that. It seems rarely used and the usages look questionable. For example, in `MakeForm`, we mark `validator` as optional. We should be more opinionated: validation is required. We also mark `orm` as optional in `MakeForm`. That probably *should* be optional. But why record that as an optional dependency at all? I'd just not mention it or handle it at all.

The tl;dr is 

A) Can we keep using `DependencyBuilder` but not worry about this optional dependencies idea anymore?
B) For packages that we discover that we need only during the interactive phase (e.g. `ux-live-component` after answering the "live" question), could we just make some new service available (that maker commands could DI into their constructor) that allows you to trigger the install? e.g.

```php
if ($input->getOption('live')) {
    $this->dependencyDownloader->installPackage('symfony/ux-live-component', $io);
}
```

This could interactively ask them if they want to install and do it. Else throw an exception to exit the command. Or maybe we go further and just `composer require` packages that are needed and just notify the user as we're doing it.

PR #1469, opened 2024-02-29
[make:form] fix generated field options_code indentation

IndraGunawan:
when generating form with `EntityType`, the indentation of `options_code` looks wrong

below example is `Foo <-- ManyToMany --> Bar`
Before

```php
<?php

class BarType extends AbstractType
{
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder
            ->add('name')
            ->add('Foo', EntityType::class, [
                'class' => Foo::class,
'choice_label' => 'id',
'multiple' => true,
            ])
        ;
    }

    // ....
}
```

After 

```php
<?php

class BarType extends AbstractType
{
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder
            ->add('name')
            ->add('Foo', EntityType::class, [
                'class' => Foo::class,
                'choice_label' => 'id',
                'multiple' => true,
            ])
        ;
    }

    // ....
}
```

### DIFF
--- a/src/Util/TemplateLinter.php
+++ b/src/Util/TemplateLinter.php
@@ -59,7 +59,7 @@ final class TemplateLinter
             $templateFilePath = [$templateFilePath];
         }
 
-        $ignoreEnv = str_contains(strtolower(\PHP_OS), 'win') ? 'set PHP_CS_FIXER_IGNORE_ENV=1&' : 'PHP_CS_FIXER_IGNORE_ENV=1 ';
+        $ignoreEnv = str_starts_with(strtolower(\PHP_OS), 'win') ? 'set PHP_CS_FIXER_IGNORE_ENV=1& ' : 'PHP_CS_FIXER_IGNORE_ENV=1 ';
 
         $cmdPrefix = $this->needsPhpCmdPrefix ? 'php ' : '';
 
--- a/src/Resources/config/php-cs-fixer.config.php
+++ b/src/Resources/config/php-cs-fixer.config.php
@@ -15,6 +15,7 @@ return (new PhpCsFixer\Config())
         '@Symfony:risky' => true,
         'native_function_invocation' => false,
         'blank_line_before_statement' => ['statements' => ['break', 'case', 'continue', 'declare', 'default', 'do', 'exit', 'for', 'foreach', 'goto', 'if', 'include', 'include_once', 'phpdoc', 'require', 'require_once', 'return', 'switch', 'throw', 'try', 'while', 'yield', 'yield_from']],
+        'array_indentation' => true,
     ])
     ->setRiskyAllowed(true)
 ;


src/Doctrine/EntityDetails.php
```diff
@@ -77,11 +77,18 @@ public function getFormFields(): array
             }
             $fieldsWithTypes[$fieldName] = [
                 'type' => EntityType::class,
-                'options_code' => sprintf('\'class\' => %s::class,', $relation['targetEntity']).\PHP_EOL.'\'choice_label\' => \'id\',',
+                'options_code' => <<<EOF
```
jrushlow, line 80: This doesn't feel right. The bundled `php-cs-fixer` is run after generating templates and _should_  clean up any indentation issues. If it is not doing that, we need to figure out why. Having "indentation corrections" outside of a template file will be error prone.

jrushlow (changes requested):

src/Doctrine/EntityDetails.php
jrushlow: One way to tell if the indentation is related to something `php-cs-fixer` is _not_ doing, override the bundled fixer w/ a local install. Then run the command again and see if there's any change. 
https://github.com/symfony/maker-bundle/blob/8182644ba29b9545e1186edbe5a339f26afe69a3/src/Resources/config/services.xml#L44

The same would hold true for the `php-cs-fixer` config:
https://github.com/symfony/maker-bundle/blob/8182644ba29b9545e1186edbe5a339f26afe69a3/src/Resources/config/services.xml#L45

_If_ the results are same as with the bundled php-cs-fixer, it would suggest that php-cs-fixer is not running in this case for some reason...

IndraGunawan: rule added

src/Util/TemplateLinter.php
```diff
@@ -59,7 +59,7 @@ public function lintPhpTemplate(string|array $templateFilePath): void
             $templateFilePath = [$templateFilePath];
         }
 
-        $ignoreEnv = str_contains(strtolower(\PHP_OS), 'win') ? 'set PHP_CS_FIXER_IGNORE_ENV=1&' : 'PHP_CS_FIXER_IGNORE_ENV=1 ';
+        $ignoreEnv = str_starts_with(strtolower(\PHP_OS), 'win') ? 'set PHP_CS_FIXER_IGNORE_ENV=1& ' : 'PHP_CS_FIXER_IGNORE_ENV=1 ';
```
IndraGunawan, line 62: i also made changes on this as these changes didn't seem right https://github.com/symfony/maker-bundle/pull/1435/files#diff-63196606ee883ed9470b1dfceed2cc28540986c0180afdb82e274812fa4690daR59

using `str_contains`, it will affect `Darwin` as well

IndraGunawan, line 62: @jrushlow i also made changes on this as these changes didn't seem right https://github.com/symfony/maker-bundle/pull/1435/files#diff-63196606ee883ed9470b1dfceed2cc28540986c0180afdb82e274812fa4690daR59

using str_contains, it will affect Darwin as well

jrushlow (approved):
Awesome, Thank you @IndraGunawan

PR #1474, opened 2024-03-05
[make:entity] Prevent entity name from having an accent

### LINKED ISSUE #1379: Prevent entity with accentuated names in make:entity
It seems that one can use accents (french ones for instance) in entity names, when using make:entity.

Even though PHP allows it, there are problems afterwards (with Doctrine, SQLite for instance).

It would probably be wise to filter input and refuse problematic characters in class/entity names.

Hope this helps.

And yes, I know... who in their right mind would be designing model entities with french vocabulary... newbees, obviously ;-)

Fan2Shrek:
Fixes #1379 

Entity name cannot longer have accent :)

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -103,6 +103,10 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
     public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
     {
         if ($input->getArgument('name')) {
+            if (!$this->verifyEntityName($input->getArgument('name'))) {
+                throw new \InvalidArgumentException('An entity can only have ASCII letters');
+            }
+
             return;
         }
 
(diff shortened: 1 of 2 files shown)


94noni (approved):

jrushlow (approved):
Thank you @Fan2Shrek

vsyrovat:
Bad MR. I had to downgrade to 1.56 because of it.

Firstly, the error message is inaccurate. The ASCII character set contains more than just A-Z range, see ascii-code dot com for details. So correct message for current regexp should be "Only a-zA-Z allowed" not "ASCII". (Actually, the result error message should not be this, I describe this point later).

Secondly, why only A-Z? Where is the digits? I want to make the `OAuth2ClientProfile` but with maker bundle 1.57 I cannot do this anymore. The maker bundle is broken now.

Thirdly, in the description of the issue exists phrase: "It would probably be wise to filter input and **refuse problematic characters** in class/entity names." Not "refuse all characters except of good" but "refuse problematic". In meaning, this is closer to a blacklist than a whitelist. Creating of a whitelist requires found ALL of the non-problematic characters before adding em to the whitelist. It means more attention and accuracy than blacklist.

Fourthly, this MR changes behavior of maker bundle significantly. Denying a programmer to use some character is bold, unreasonably bold. It is better to **notify** the programmer about problematic characters presented in typed entity name and let him decide whether to leave them. "The entity name you entered has some non-ASCII characters which are potentially problematic with SQLite. It is recommended to use only ASCII characters in entity names. Would you keep entered name? [no]".

So, I consider 1.57.0 is broken and see two possible solutions:
- fast: add digits and underscore to the regexp,
- full: use recommendations instead of prohibitions.

Issue #929, opened 2021-07-20
New options for strict_types


upyx:
Hi

Using of `declare(strict_types=1);` was offered in #80 and then declined. However, many things have changed since 2017.  The most notable ones are: ending support of [PHP 5](https://www.php.net/supported-versions.php) and accepting the [PSR-12](https://www.php-fig.org/psr/psr-12/) standard that requires strict types. So new Symfony-based projects mostly use PSR and strict types.

It is easy to format code in IDE to any code styles' requirements, but it is so annoying to add `declare(strict_types=1);` again and again... and again...

I suggest adding command options for turning strict types on and off and adding an optional configuration for choosing default behavior.

| Options | Description |
|------------|-----------------|
|`--strict-types`| Add `declare(strict_types=1);` to head of PHP file |
|`--no-strict-types`| Don't add `declare(strict_types=1);` to head of PHP file |

Example of usage:
```
bin/console make:command --strict-types app:make-me-happy
```

Example of configuration:
```yaml
# config/packages/dev/maker.yaml
maker:
    # Add `declare(strict_types=1);` to head of generated PHP file by default
    # default: false
    strict_types: true
```

---

I'm ready to implement that :wink:

jrushlow:
Howdy @upyx - I do not believe we'll implement `strict_types` in MakerBundle. Adding `declare(strict_types=1)` is purely internal and we already have / are implementing strong type hints where possible - both in the generated templates and internally in maker itself. 

Because strict typing is on a per file basis, I would estimate that 95% of the users would not truly utilize this feature. E.g. Symfony does not declare `strict_types` anywhere in the code base. So if MakerBundle added `declare(strict_types=1)` to say the `ResetPasswordController::class` - the _file_ calling a method within the `ResetPasswordController::class` would be deciding whether to use strict typing regardless if `strict_types` was declared in the `ResetPasswordController::class` file.

See https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict for more information.

Because this was already shot down some time ago in Symfony for the same reasons I'm stating here, it wouldn't make sense for us to be the "exception to the rule" so to speak.  https://github.com/symfony/symfony/issues/28423 

>accepting the PSR-12 standard that requires strict types

I've reviewed PSR-12 in the past, while PSR-12 show examples of `declare(strict_types=1)` usage and provides guidance on the placement of the declaration if it is used - it does not explicitly state that it `MUST` be used. 

[...]

PR #1478, opened 2024-03-08
Update Class.tpl.php: Adding ` declare(strict_types=1);`

ThomasLandauer:
Well, a few years have passed since the latest suggestion to add `strict_types` https://github.com/symfony/maker-bundle/issues/929, so I thought it's about time for a new attempt ;-)

It was rejected cause Symfony isn't doing it either in their codebase: https://github.com/symfony/maker-bundle/issues/929#issuecomment-889711173
However: Symfony's argument was that they don't want to have it in their **internal** classes: https://github.com/symfony/symfony/issues/28423#issuecomment-466927534 But here in Maker Bundle, we're talking about pre-generated classes for the user.

Two arguments for adding it:
* It's good practice nowadays.
* It's easier to *delete* it (for those users who dont' want it), than it is to *add* it (for those who do want it).

### DIFF
--- a/src/Resources/skeleton/Class.tpl.php
+++ b/src/Resources/skeleton/Class.tpl.php
@@ -1,4 +1,4 @@
-<?= "<?php\n" ?>
+<?= "<?php declare(strict_types=1);\n" ?>
 
 namespace <?= $namespace; ?>;
 


weaverryan:
Idk, I've actually never used this before. And I don't have a good feeling of how many people want this vs don't want this.

Since we can't make everyone happy, I'd like to continue to *not* include this. PHP-cs-fixer has a `declare_strict_types` option to add it back in. Or, if enough people did want this, it would be a new config option. Also, the PR would be a lot bigger to add this everywhere - e.g. https://github.com/symfony/maker-bundle/blob/main/src/Resources/skeleton/doctrine/Fixtures.tpl.php - many classes generated from other skeleton files.

nicolas-grekas:
Same opinion as Ryan. Also, see https://github.com/Girgias/unify-typing-modes-rfc, which is a good analysis of why strict types doesn't improve anything.

weaverryan:
Cool, so while I know this won't please everyone, I'm closing. 

Cheers!

Issue #1479, opened 2024-03-15
[make:entity] Cannot create an entity called "Locale"


madflow:
I cannot create an entity called "Locale" on a freshly bootstrapped Symfony project.

```
symfony new app --version="7.0.*"
cd app
composer require symfony/maker-bundle
composer require symfony/orm-pack
./bin/console make:entity Locale
```

```
Your entity already exists! So let's add some new fields!
[critical] Error thrown while running command "make:entity Locale". Message: "Notice: file_get_contents(): Read of 8736 bytes failed with errno=21 Is a directory"

In FileManager.php line 111:

  Notice: file_get_contents(): Read of 8736 bytes failed with errno=21 Is a directory


make:entity [-a|--api-resource] [-b|--broadcast] [--regenerate] [--overwrite] [--] [<name>]
```

I would argue, that this should be possible.

I can think of a name collision with the global "Locale" class. https://www.php.net/manual/de/class.locale.php

smnandre:
It seems to be a side effect of https://github.com/symfony/maker-bundle/pull/1422

smnandre:
I think #1422 should be reverted, as it contredict the documentation of the method :/ 

```
     * @param string $name            The short "name" that will be turned into the class name
     * @param string $namespacePrefix Recommended namespace where this class should live, but *without* the "App\\" part
     * @param string $suffix          Optional suffix to guarantee is on the end of the class
```


I started something using [ReflectionClass::isInternal()](https://www.php.net/manual/en/reflectionclass.isinternal.php) , that could help in this specific case. But overall i'm not sure we should change the original behaviour. For the absolute path, it would require using `\\` as before, but that was more self-explanatory than today :|

GPHemsley:
See also:
* #1566
* #1480
* #1482
* #1676
* #1681

PR #1480, opened 2024-03-15
[make:entity] fix: add failing tests

madflow:
- Adds failing tests for #1479 
- I suspect there is no easy fix, maybe not a desired one.
- I would need some direction ;)


jrushlow (changes requested):
Thanks for the test @madflow - the problem is in `Generator` and ultimately in _how_ we validate a class against the autoloader. I've created a PR #1482 with a similar test case and I'm working on a fix that, hopefully, doesnt require reverting #1422.

We'll more than likely get this merged in about the same time I'm able to work out a fix in #1482

GPHemsley:
See also:
* #1676
* #1681

GromNaN:

PR #1483, opened 2024-03-16
do not include dev files in release

jrushlow:
Excludes the following from being "installed" by the user in a project
- https://github.com/symfony/maker-bundle/blob/main/phpunit.xml.dist
- https://github.com/symfony/maker-bundle/blob/main/docker-compose.yml

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
 .* export-ignore
 *.md export-ignore
-/appveyor.yml
-/phpunit.xml.dist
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,6 +1,6 @@
 .* export-ignore
 *.md export-ignore
-/appveyor.yml
```
jrushlow, line 3: `appveyor` is no longer used

PR #1487, opened 2024-03-22
[make:schedule] a new command for creating recurring Symfony Schedules

jrushlow:
- adds a new `make:schedule` command

- creates stateful schedulers

- If `Symfony/Scheduler` is not already installed - rather than using an exit early strategy - we `composer install symfony/scheduler`. This is the "shift" MakerBundle is making to handling missing dependencies.

- We check any files in `src/Message` -> if files exist -> ask the user to select a message class to use with the new schedule. The "suggested name" of the new schedule is the filename of the message w/ the `Message` suffix removed.
image

- The user also has the ability to select "Empty Message" to create a new schedule regardless if any message were found. This will create a "empty" schedule.

- If no existing messages exist -> ask the user the name of the "empty" schedule we'll create
image

Empty Schedule:
image

Schedule from message:
image


src/Maker/MakeSchedule.php
```diff
@@ -0,0 +1,144 @@
...
+
+            if ('Empty Schedule' !== $selectedMessage) {
+                $this->message = $selectedMessage;
+
+                // We don't want SomeMessageSchedule, so remove the "Message" suffix to give us SomeSchedule
+                $scheduleNameHint = sprintf('%sSchedule', Str::removeSuffix($selectedMessage, 'Message'));
+            }
+        }
+
+        // Otherwise call `make:message` to create a new message
```
jrushlow, line 100: This could be cool todo. Idea is in the choice list we have:
0 -> Empty Schedule
1 -> New Schedule
2 -> ...

But I can't remember if a) this is actually dooable to run a maker inside a maker without duplicating code && b) if there _arn't_ any messages, or if `symfony/messenger` is _not_ installed - what should we do...

jrushlow: We're going to skip this for now. Revisit at a later time...

PR #1491, opened 2024-03-25
[make:webhook] Add new command for Symfony's Webhook Component

maelanleborgne:
Add a maker that would help create a webhook :
- Create a RequestParser extending `AbstractRequestParser` and interactively define the RequestMatcher(s) to use.
- Create a WebHookConsumer implementing `ConsumerInterface` with the `#[AsRemoteEventConsumer]` attribute.
- Add an entry to config/packages/webhook.yaml with the name of the webhook + the RequestParser service name + empty secret with an instruction on how to set it (as a comment).

That is an easy scaffolding command but since Webhook is a quite recent component and the documentation is pretty scarce right now, I think it would make sense to have a maker to help setting things up.

image

Regex: https://regex101.com/r/S3BWkx/1

Bumps test process timeout from `10` to `30` seconds - composer is a bit slow on Windows. 

https://github.com/symfony/maker-bundle/blob/30cdf17f8a9eb602bf0cecf8484e0b7671ae7a1b/src/Test/MakerTestEnvironment.php#L339


src/Maker/MakeWebhook.php
```diff
@@ -0,0 +1,285 @@
...
+
+        $this->generator->writeChanges();
+        $this->fileManager->dumpFile(self::WEBHOOK_CONFIG_PATH, $this->ysm->getContents());
+    }
+
+    private function addToYamlConfig(string $webhookName, ClassNameDetails $requestParserDetails): void
+    {
+        if (!$this->fileManager->fileExists(self::WEBHOOK_CONFIG_PATH)) {
+            $yamlConfig = Yaml::dump(['framework' => ['webhook' => ['routing' => []]]], 4, 2);
+        } else {
```
jrushlow, line 162: If we can eliminate this `else` that would be awesome. Something like 
```php
$yamlConfig = YamlDump([.......])
if ($this->fileManager...... {
    $yamlConfig = .....
```

src/Maker/MakeWebhook.php
```diff
@@ -0,0 +1,285 @@
...
+            $yamlConfig = $this->fileManager->getFileContents(self::WEBHOOK_CONFIG_PATH);
+        }
+        $this->ysm = new YamlSourceManipulator($yamlConfig);
```
jrushlow, line 165: Are we calling `$this->ysm` outside of this method? If not, let's remove the class property and keep this a local variable

src/Maker/MakeWebhook.php
```diff
@@ -0,0 +1,285 @@
...
+                $questionText = 'Add a RequestMatcher (press <return> to skip this step)';
+            } else {
+                $questionText = 'Add another RequestMatcher? Enter the RequestMatcher name (or press <return> to stop adding matchers)';
+            }
+
+            $choices = array_diff($availableMatchers, $addedMatchers);
+            $question = new ChoiceQuestion($questionText, array_values(['<skip>'] + $choices), 0);
+            $matcherName = $io->askQuestion($question);
+            if ('<skip>' === $matcherName) {
+                return null;
```
jrushlow, line 249: Do the same for other parts of this maker: Let's add so more spacing within the code. grouping multiple single line variable declarations is A-OK. But using that enter button between multi-line declarations, conditionals, etc... improves readability..

```suggestion
            $choices = array_diff($availableMatchers, $addedMatchers);
            $question = new ChoiceQuestion($questionText, array_values(['<skip>'] + $choices), 0);
            $matcherName = $io->askQuestion($question);
            
            if ('<skip>' === $matcherName) {
                return null;
```

jrushlow (changes requested):
This is looking pretty nifty @maelanleborgne - a few suggestions and I think we'll be in good shape..

maelanleborgne:
> This is looking pretty nifty @maelanleborgne - a few suggestions and I think we'll be in good shape..

Thank you for the review. I'm at SFlive right now, I'll try to push the changes on my way back. Just curious about the bit on unit testing the regex : testing invalid cases seems pretty hard with the MakerTestCase, should I make the method public? Use reflection method and invoke maybe?

jrushlow:
The regex test can be added to: 

https://github.com/symfony/maker-bundle/blob/74a8918c15ea2ce6fe5b5233e0b46eb57745e213/tests/RegexTest.php#L17

I'll have to do some minor refactoring on that class to make it "multi-purpose" (ill get a pr up and merged here shortly) - but you should be able to add another dataProvider to that test and be all set. 

As to the regex itself - move the regex string to a constant like:
https://github.com/symfony/maker-bundle/blob/74a8918c15ea2ce6fe5b5233e0b46eb57745e213/src/Test/MakerTestEnvironment.php#L25

Then you can call that regex expression from within the class and within the test without too much trouble.

jrushlow (approved):
Awesome! Thank you @maelanleborgne

Issue #1492, opened 2024-03-26
[make:entity] addGetter on boolean field that starts with "is"


ClemRiviere:
Hi devs, 
 
## Context

Here is an exemple of an Entity created with a "isFooProp" boolean field : 

```bash
> symfony console make:entity Test
 created: src/Entity/Test.php
 created: src/Repository/TestRepository.php
 
 Entity generated! Now let's add some fields!
 You can always add more fields later manually or by re-running this command.

 New property name (press <return> to stop adding fields):
 > isFooProp

 Field type (enter ? to see all types) [boolean]:
 > 

 Can this field be null in the database (nullable) (yes/no) [no]:
 > 

 updated: src/Entity/Test.php
```

## Current behavior

This is the resulting entity : 

```php
#[ORM\Entity(repositoryClass: TestRepository::class)]
class Test
{
   ...
    #[ORM\Column]
    private ?bool $isFooProp = null;
    ...

    public function isIsFooProp(): ?bool
    {
        return $this->isFooProp;
    }

    public function setIsFooProp(bool $isFooProp): static
    ...
}
```

## Proposed behavior

```php
#[ORM\Entity(repositoryClass: TestRepository::class)]
class Test
{
   ...
    #[ORM\Column]
    private ?bool $isFooProp = null;
    ...

    public function isFooProp(): ?bool
    {
        return $this->isFooProp;
    }

    public function setIsFooProp(bool $isFooProp): static
    ...
}
```

I already worked on this modification and should be able to propose a PR, I just wanted to have your feedback on this topic before. All comments are welcome :)

jrushlow:
Howdy @ClemRiviere - I'm not opposed to the idea. I've actually run into the same issue in the past - though in practice I tend to lean towards "non-verb" naming conventions for class properties. Looking forward to reviewing your PR

PR #1493, opened 2024-03-26
[make:entity] managing keyword prefixes (is, has) for boolean properties getters

ClemRiviere:
- Refactoring the getter prefixes  for `make:entity`
- Introducing exceptions for boolean fields : not adding the prefix to getter name if field begins with "is" or "has"

Related to #1492

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -282,10 +282,12 @@ final class ClassSourceManipulator
         if ('bool' !== $returnType) {
             return 'get' . Str::asCamelCase($propertyName);
         }
+        
         // exclude is & has from getter definition if already in property name
-        if (!strncasecmp($propertyName, 'is', 2) && !strncasecmp($propertyName, 'has', 3)) {
+        if (0 !== strncasecmp($propertyName, 'is', 2) && 0 !== strncasecmp($propertyName, 'has', 3)) {
             return 'is' . Str::asCamelCase($propertyName);
         }
+
         return Str::asLowerCamelCase($propertyName);
     }
 
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -273,7 +273,7 @@ final class ClassSourceManipulator
         $methodName = $this->getGetterName($propertyName, $returnType);
         $this->addCustomGetter($propertyName, $methodName, $returnType, $isReturnTypeNullable, $commentLines);
     }
-
+    
     private function getGetterName(string $propertyName, $returnType): string
     {
         if ('bool' !== $returnType) {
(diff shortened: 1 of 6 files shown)


src/Util/ClassSourceManipulator.php
```diff
@@ -270,27 +270,23 @@ public function addAccessorMethod(string $propertyName, string $methodName, $ret
...
-        $methodName = $prefix ? $prefix.Str::asCamelCase($propertyName) : Str::asLowerCamelCase($propertyName);
+        $methodName = $this->getGetterName($propertyName, $returnType);
         $this->addCustomGetter($propertyName, $methodName, $returnType, $isReturnTypeNullable, $commentLines);
     }
 
     /**
      * @return string|null : getter prefix or null if no prefix to append
      */
-    private function getGetterPrefix($propertyName, $returnType): ?string
+    private function getGetterName($propertyName, $returnType): ?string
```
nicolas-grekas, line 280:
```suggestion
    private function getGetterName($propertyName, $returnType): string
```

nicolas-grekas (commented):
This change makes sense to me. It's consistent with the autodetection of the type (boolean) when asking for the type.
I'd suggest doing the same for setters, but only for the "is" prefix: isInternational property => setInternational. (but hasColor => setHasColor)

src/Util/ClassSourceManipulator.php
```diff
@@ -270,10 +270,27 @@ public function addAccessorMethod(string $propertyName, string $methodName, $ret
...
     {
-        $methodName = ('bool' === $returnType ? 'is' : 'get').Str::asCamelCase($propertyName);
+        $methodName = $this->getGetterName($propertyName, $returnType);
         $this->addCustomGetter($propertyName, $methodName, $returnType, $isReturnTypeNullable, $commentLines);
     }
 
+    /**
+     * @return string|null : getter prefix or null if no prefix to append
+     */
+    private function getGetterName($propertyName, $returnType): ?string
```
jrushlow, line 280: `$propertyName` should by typed with `string`. And as @nicolas-grekas suggested, the return type should not be `null`able

`$returnType` can be left without a type - adding a type is out of scope in this PR (requires changes elsewhere)

jrushlow (changes requested):
This looks pretty good @ClemRiviere! - 1 minor comment. Also if you're up for doing the same for `addSetter()` as @nicolas-grekas suggested, that would be even more awesome!

jrushlow (approved):
Thanks @ClemRiviere!

stof:
This change does not make sense to me as it makes those setter incompatible with the property-access component (and so with the form component that relies on it)

toby-griffiths:
A link wasn't added automatically (maybe becaues it was a discussion, not an issue), but this is now causing property access issues as discussed in https://github.com/symfony/symfony/discussions/54885#discussioncomment-9814172.

ClemRiviere:
@stof I understand, I think we have two potential solutions for this : 
- rollback my work on setters so `isInternational` property becomes `isInternational()` getter and `setIsInternational(bool $isInternational)` setter in maker-bundle
- adapt property-access component to manage `is` and `has` prefixes

I would personally go for the first solution, @jrushlow or @nicolas-grekas do you guys have any advices on this topic ?

PR #1496, opened 2024-03-30
[make:entity] confirm to allow non-ascii char's in entity names

### LINKED ISSUE #1556: [make:entity] Entity Names Should Be Allowed to Have Digits (i.e. v1)
Steps to reproduce:
- Use make:entity with a namespaced classname that includes a digit (i.e., Foo\v1\Bar)
- Observe ASCII error

https://github.com/symfony/maker-bundle/blob/6960c1cc31e8e4f8f1de6b8a139b7867092e5c13/src/Maker/MakeEntity.php#L832

Replace regex in line 832
`
Old = ^[a-zA-Z\\\\]+$
`
`
New = ^[a-zA-Z\\\d]+$
`

Removes duplicated \\ (escaped "\") and adds digit as allowable character.

Fan2Shrek:
Following the message https://github.com/symfony/maker-bundle/pull/1474#issuecomment-2027716072, I made some modifications

---
edited by @jrushlow 
fixes #1556

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -819,6 +819,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         return $io->askQuestion($question);
     }
 
+    /** @return string[] */
     private function verifyEntityName(string $entityName): array
     {
         preg_match('/([^\x00-\x7F]+)/u', $entityName, $matches);
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -107,11 +107,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
 
     public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
     {
-        if ($input->getArgument('name')) {
-            if (!$this->verifyEntityName($input->getArgument('name'))) {
-                throw new \InvalidArgumentException('An entity can only have ASCII letters');
-            }
-
+        if (($entityClassName = $input->getArgument('name')) && empty($this->verifyEntityName($entityClassName))) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/Maker/MakeEntity.php
```diff
@@ -820,9 +819,11 @@ private function askRelationType(ConsoleStyle $io, string $entityClass, string $
         return $io->askQuestion($question);
     }
 
-    private function verifyEntityName(string $entityName): bool
+    private function verifyEntityName(string $entityName): array
     {
-        return preg_match('/^[a-zA-Z\\\\]+$/', $entityName);
+        preg_match('/[^\x00-\x7F]/u', $entityName, $matches);
```
jrushlow, line 824:
```suggestion
        preg_match('/([^\x00-\x7F]+)/u', $entityName, $matches);
```

jrushlow: We just need to know if the entity name contains any non-ascii's chars. Otherwise, we'd ask the questions multiple times which isnt very DX friendly...

src/Maker/MakeEntity.php
```diff
@@ -131,10 +127,13 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
 
         $argument = $command->getDefinition()->getArgument('name');
         $question = $this->createEntityClassQuestion($argument->getDescription());
-        $entityClassName = $io->askQuestion($question);
+        $entityClassName ??= $io->askQuestion($question);
+
+        while ($dangerous = $this->verifyEntityName($entityClassName)) {
+            if ($io->confirm(sprintf('The %s character is non-ASCII, which are potentially problematic with some database. It is recommended to use only ASCII characters in entity names. Would you keep entered name ?', $dangerous[0]), false)) {
```
jrushlow, line 133:
```suggestion
            if ($io->confirm(sprintf('"%s" contains one or more non-ASCII characters, which are potentially problematic with some database. It is recommended to use only ASCII characters for entity names. Continue anyway?', $entityClassName), false)) {
```

jrushlow (approved):
Thank you @Fan2Shrek - I've suggested a couple minor changes which I'll go ahead and commit so we can merge.

src/Maker/MakeEntity.php
```diff
@@ -820,9 +819,11 @@ private function askRelationType(ConsoleStyle $io, string $entityClass, string $
         return $io->askQuestion($question);
     }
 
-    private function verifyEntityName(string $entityName): bool
+    private function verifyEntityName(string $entityName): array
```
jrushlow, line 822:
```suggestion
    /** @return string[] */
    private function verifyEntityName(string $entityName): array
```

PR #1498, opened 2024-04-02
[ci] implement phpstan for static analysis of the codebase

jrushlow:
- uses PHPStan to help catch bugs

We'll start with the scope below and slowly open it up as the code base is updated to the desired PHPStan level. 

### Scope
- `src/Maker`
- `tests/Command`
- `tests/Docker`
- `tests/Maker`

--- 
## Optional Maker Dependencies - `tools/phpstan/includes`

In our root `composer.json`, we don't `require --dev` packages that are not specifically needed for MakerBundle to function. e.g. `api-platform`. But we _do_ still need those packages for phpstan. for now, the `composer.json` in the `includes` dir will satisfy that need and gives MakerBundle dev's the added benefit of code completion in the IDE.

We currently do not require those "maker specific" packages in our root `composer.json` because it would have adverse side effects in some of our tests. I _think_ in MakerBundle 2.x and/or #1466 - we'll change how we manage dependencies.

### DIFF
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -13,49 +13,6 @@ env:
     SYMFONY_PHPUNIT_DIR: "$HOME/symfony-bridge/.phpunit"
 
 jobs:
-    coding-standards:
-        name: "Coding Standards (${{ matrix.php-version }})"
-
-        runs-on: ubuntu-latest
-
-        strategy:
-            fail-fast: false
-            matrix:
-                php-version:
-                    - '8.1'
-
-        steps:
-            -
-                name: Checkout code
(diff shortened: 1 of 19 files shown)


.github/workflows/ci-linux.yaml
```diff
@@ -137,4 +94,4 @@ jobs:
               run: vendor/bin/simple-phpunit --version
 
             - name: Run tests
-              run: vendor/bin/simple-phpunit ${{ env.PHPUNIT_FLAGS }}
+              run: vendor/bin/simple-phpunit ${{ env.PHPUNIT_FLAGS }} --version
```
jrushlow, line 97: - Re-enable tests here and in windows ci

PR #1512, opened 2024-04-11
[make:webhook] Rephrase comments to be more explicit + minor improvement of generated code

maelanleborgne:
Previously, the generated code of `FooRequestParser::doParse` started with `// Implement your own logic to validate and parse the request, and return a RemoteEvent object.`, followed by an empty line, then some code. I think it may be confusing for the developers, some may think the custom logic should live right bellow the comment and that the following code should not be messed with.  
I think that changing this to `// TODO: Adapt or replace the content of this method to fit your need.` might help clear things up.

Additionally I removed an unnecessary check for null  in the generated code + fixed the payload conversion method in the fixture.


jrushlow (approved):
Thank you @maelanleborgne! I like the simplification...

PR #1514, opened 2024-04-17
replace Bundle with the new AbstractBundle class

jrushlow:
Move the bundle from the old "Bundle" to the new "AbstractBundle" system following the best practices outlines in https://symfony.com/doc/current/bundles/best_practices.html

---

Nothing to see here yet...

see - https://github.com/symfony/symfony-docs/pull/19793

I'm not sure if we _should_ do this yet... Maker has a "leg up" on the other `symfony/*` bundles because we're a bit bleeding edge. Our minimum `symfony/*` is `6.4` - but by merging this PR, we would be breaking away from say, `FrameworkBundle` that still uses the legacy `Bundle` class... 

On the flip side, if we do press forward with this PR - we could reference this change somehow in the docs to show others how we "converted" to the new simplification strategy.

Pinging @nicolas-grekas @javiereguiluz & @yceruto This isn't review ready yet, but feedback on if we should press forward with this would be awesome. 

## Merge blockers:
- [x] `src/DependencyInjection/CompilerPass/*` figure out if these should live here still - or if there is a better way...
- [x] ~`src/Resources/bin` - This is our "bundled" php-cs-fixer, it needs a new home (not intended to be directly called by the user...)~ the bundled php-cs-fixer used by the template linter was replaced with a shim
- [x] `src/Resources/config/php-cs-fixer.config.php` - used by our internal php-cs-fixer when running `make:*` - make sure this path isnt hard coded.. If it is, fix it to use the new `config/php-cs-fixer.config.php` path.
- [x] `src/Resources/doc` -> `docs/`
- [x] #1605  ~`src/Resources/help` - figure out a home for these files - the `AbstractMaker`? (or another parent class) needs to be updated to point the command to the new help location~ This will done in a separate PR as it will create a messy diff...
- [x] #1606 ~`src/Resources/Skeleton` - our "templates" need a new home. These feel like they should live in `src` maybe something like `src/Skeleton`? hmm....~ This will done in a separate PR as it will create a messy diff...
- [x] Test, test in ci, test in linked apps, test some more, run some tests....


yceruto:
> `src/DependencyInjection/CompilerPass/*` figure out if these should live here still - or if there is a better way...

It's still the correct place for them in my opinion.

>`src/Resources/bin` - This is our "bundled" php-cs-fixer, it needs a new home (not intended to be directly called by the user...)

It should be `<root>/bin/` instead (according best practices document)

> `src/Resources/config/php-cs-fixer.config.php` - used by our internal php-cs-fixer when running make:* - make sure this path isnt hard coded.. If it is, fix it to use the new config/php-cs-fixer.config.php path.

I usually put this config file at the root of the project, but it'd be okay under `<root>/config/` as well

> src/Resources/doc -> doc/

`<root>/docs/` in plural (according to best practices doc) 

> src/Resources/help - figure out a home for these files - the AbstractMaker? (or another parent class) needs to be updated to point the command to the new help location

They look like configuration files to me, as such they could be moved to `<root>/config/help/` dir

> src/Resources/Skeleton - our "templates" need a new home. These feel like they should live in src maybe something like src/Skeleton? hmm....

Kind of templates! `<root>/skeleton/` lower case 👍

PR #1520, opened 2024-04-19
[make:entity] Simplify repository PHPDoc removes method annotations

lyrixx:
I removed theses from project project, and my IDE (vscode) and PHPStan are okay with that.
It works, because doctrine has already templated the methods

### DIFF
--- a/src/Resources/skeleton/doctrine/Repository.tpl.php
+++ b/src/Resources/skeleton/doctrine/Repository.tpl.php
@@ -6,11 +6,6 @@ namespace <?= $namespace; ?>;
 
 /**
  * @extends ServiceEntityRepository<<?= $entity_class_name; ?>>
- *
- * @method <?= $entity_class_name; ?>|null find($id, $lockMode = null, $lockVersion = null)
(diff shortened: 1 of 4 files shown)


jrushlow (approved):
Awesome! I'm about 99% sure that we are safe to remove the annotations now. Some SymfonyCasts folks and I tried out code completion in PHPStorm and it appears to be working just fine without them...

I'm sure the folks out there with the pitch forks will let us know pretty quick if this _does_ break something after cutting a release. :smile:

stof (approved):

simondaigre:
On my end this PR broke autocompletion on PHPStorm on projects with ORM 2 (didn't tried with 3)

lyrixx:
This is really weird, because we tested it with ORM2 and it was okay.
However, what is the version of `doctrine/persistence`?

simondaigre:
`3.3.2`. Do you want a separate issue for this ?

jrushlow:
@simondaigre - Yes if you could create a new issue - I'll get a patch out in the next day or so.

PR #1244, opened 2022-11-29
[make:security:form-login] new maker to use built in FormLogin

jrushlow:
Introduces a new series of makers that use the built-in `SecurityBundle` authenticators. The first - `make:security:form-login` with the next probably being `make:security:json-login` ? which will abstract any common functionality that `make:security:*` can re-use...

This PR does not deprecate `make:auth` yet. Once we have a `make:security:custom-authenticator` in place - we can sunset `make:auth`.

- [ ] Update docs to show usage and examples

fixes #1049 

---

@wouterj - Im going to steal some of the concepts/code from https://github.com/symfony/maker-bundle/pull/1058 so we can implement that PR in baby steps - starting with FormLogin

PR #1523, opened 2024-04-21
[make:auth] deprecate command

jrushlow:
The `make:auth` command has been superseded by `make:security:form-login` (#1244) and `make:security:custom` (#1522).

Originally written to create login form authentication back in the days of "Guard Authenticators" - this command has reached it's EOL with the existence of the new Symfony Security System that was introduced in Symfony 5.x.

Although the command class itself is `@internal` - we'll:
- deprecate the command in `v1.59.0` maintaining its existing functionality
- remove the existing functionality _after_ the `v1.59.0` release
- remove the command completely sometime after that. 

I think this will give plenty of time to get the docs & book updated and "gently" push users to `make:security:*`

- [x] Symfony Book 6.4 - Published

cc @fabpot - Symfony Book

### DIFF
--- a/src/Maker/MakeAuthenticator.php
+++ b/src/Maker/MakeAuthenticator.php
@@ -54,7 +54,11 @@ use Symfony\Component\Security\Http\SecurityRequestAttributes;
 use Symfony\Component\Security\Http\Util\TargetPathTrait;
 use Symfony\Component\Yaml\Yaml;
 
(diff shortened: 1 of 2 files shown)


jrushlow:
Note: We're going to hold off on merging this until the Symfony 6.4 Book is published. Else, we "break" the user experience for those following along in the book.

PR #1524, opened 2024-04-24
[make:security] Add throttling support to authenticator maker

Spomky:
Reopening #1406
In the MakeAuthenticator class, a new optional argument 'support-throttling' was added. This allows the user to specify if they want to enable throttling protection during the authentication generation process. If 'support-throttling' is enabled, the LimiterInterface dependency is added and the 'throttling' node is set in the firewall configuration.

### DIFF
--- a/src/Maker/MakeAuthenticator.php
+++ b/src/Maker/MakeAuthenticator.php
@@ -38,6 +38,7 @@ use Symfony\Component\Console\Question\Question;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\RateLimiter\LimiterInterface;
 use Symfony\Component\Routing\Attribute\Route;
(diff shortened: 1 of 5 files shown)


src/Maker/MakeAuthenticator.php
```diff
@@ -209,6 +210,15 @@ function ($answer) {
                     $supportRememberMeValues[$supportRememberMeType]
                 );
             }
+
+            $command->addArgument('support-throttling', InputArgument::OPTIONAL);
```
jrushlow, line 214: `MakeAuthenticator` is being deprecated - we are not adding any new features to it.

jrushlow (changes requested):
Howdy @Spomky

jrushlow (changes requested):
Howdy @Spomky - Sorry for the double review... Coffee has not kicked in yet...

Anywho, as I said in https://github.com/symfony/maker-bundle/pull/1406#issuecomment-2074144535 `make:auth` is being deprecated in favor of the `make:security:form-login` command. We are not adding any new features to `make:auth` at this time.

PR #1527, opened 2024-04-28
WIP - [make:reset-password] handle multiple entities

### LINKED ISSUE #653: make:reset-password : Multiple entities
Hi 😄,

It would be great if we could handle multiple entities with this Reset Password.

@jrushlow said :

> [...] But off the top of my head, the `reset_password.yaml` file wouldn't change if you wanted to use multiple user entities. It just needs to know what repository to use for the `ResetPasswordRequest` entity. Which there should only be 1 of those.
> 
> What would change would be the `Entity\ResetPasswordRequest::user` Doctrine Annotation. `make:reset-password` does something like `@ORM\ManyToOne(targetEntity="App\Entity\User")` for that property.
> 
> If you have multiple entities, you couldn't use a `ManyToOne` relation. Knee jerk thinking would be the request entity would need a new property(ies) to track which user "type" (entity) and the id of the user. Then in the request repository, you would use the new properties to fetch and set `Entity\ResetPasswordRequest::user` with the correct user entity before returning the `Entity\ResetPasswordRequest::class` to the consumer.

View : https://github.com/SymfonyCasts/reset-password-bundle/issues/110#issuecomment-626244231

jrushlow:
fixes #653 
refs: 
- https://github.com/SymfonyCasts/reset-password-bundle/pull/318
- https://github.com/SymfonyCasts/reset-password-bundle/issues/110

### DIFF
--- a/src/Maker/MakeResetPassword.php
+++ b/src/Maker/MakeResetPassword.php
@@ -166,6 +166,8 @@ class MakeResetPassword extends AbstractMaker
             'What is the User entity that should be used with the "forgotten password" feature? (e.g. <fg=yellow>App\\Entity\\User</>)'
         );
 
+        // @TODO - if more than 1 $providerData - ask if we should handle more than $this->userClass
+
         $this->emailPropertyName = $interactiveSecurityHelper->guessEmailField($io, $this->userClass);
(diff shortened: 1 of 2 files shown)


ludovik-dev:
Have you been able to make any progress on managing several entities (e.g. User entity and Admin entity)? Thx

PR #1532, opened 2024-05-03
[make:auth] only fire deprecation when command is called

### LINKED ISSUE #1530: [make:auth] V1.59 - Deprecated classes causes deprecated message on each page
the trigger_deprecation function before MakeAuthenticator Class make a deprecation message on each called page not just when command make:auth is called : 

Since symfony/maker-bundle v1.59.0: The "Symfony\Bundle\MakerBundle\Maker\MakeAuthenticator" class is deprecated, use any of the Security\Make* commands instead.

can pass the deprecated in interact function ? 

should I make a PR ?

eltharin:
fixes https://github.com/symfony/maker-bundle/issues/1530


jrushlow (approved):
Thank you @eltharin!

PR #1534, opened 2024-05-03
[make:entity] add enum support

### LINKED ISSUE #1233: `make:entitiy`: Suggestion: Add `enumType`
I'm suggesting to add this possible outcome to the wizard:
```php
#[ORM\Column(nullable:true, enumType: Suit::class)]
private ?Suit $suit = null;
```
Maybe via a dedicated `enum` entry in the types list, then by inferring the actual type (i.e. `string` or `int`) from the enum itself.

### LINKED ISSUE #1440: `make:entity` and enums
It seems entity generator is unable to handle enums.
On an existing entity class 
```
#[ORM\Entity]
class MyEntity
{
    // ...
    #[ORM\Column(nullable: false)]
    protected MyEntityType $type;
    // ...
```
`make:entity --regenerate` outputs enum's backed type (or is it getting it from mapped field type actually?):
```
    public function getType(): ?string
    {
        return $this->type;
    }

    public function setType(string $type): static
    {
        $this->type = $type;

        return $this;
    }
```
Also, when running `make:entity` as a command-line wizard, while creating a property there is no option for enum type.

Quick browse of maker-bundle's code showed [almost nothing about enums](https://github.com/search?q=repo%3Asymfony%2Fmaker-bundle%20enum&type=code). 
Is it really unsupported yet?
Can it be fixed to generate a proper param/return type hint at least?

loicsapone:
Hi 👋 

The purpose of this PR is to add support for enumerations when creating an entity. 
Instead of introducing a new field type, I've opted to add a prompt to supplement a text field with the enumType option.

---
Edit by @jrushlow 
fixes #1440 fixes #1233

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -434,8 +434,8 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
             // ask for valid backed enum class
             $classProperty->enumType = $io->ask('Enum class', null, Validator::classIsBackedEnum(...));
 
-            // reset type to "string" because enum is not a valid DBAL type
-            $classProperty->type = 'string';
+            // set type according to user decision
+            $classProperty->type = $io->confirm('Can this field store multiple enum values', false) ? 'simple_array' : 'string';
         }
 
         if ($io->confirm('Can this field be null in the database (nullable)', false)) {
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -393,7 +393,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         $allValidTypes = array_merge(
             array_keys($types),
             EntityRelation::getValidRelationTypes(),
-            ['relation']
+            ['relation', 'enum']
         );
         while (null === $type) {
             $question = new Question('Field type (enter <comment>?</comment> to see all types)', $defaultType);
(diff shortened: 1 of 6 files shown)


src/Maker/MakeEntity.php
```diff
@@ -424,6 +424,10 @@ private function askForNextField(ConsoleStyle $io, array $fields, string $entity
         if ('string' === $type) {
             // default to 255, avoid the question
             $classProperty->length = $io->ask('Field length', '255', Validator::validateLength(...));
+
+            if ($io->confirm('Is this field an enum?', false)) {
+                $classProperty->enumType = $io->ask('Enum class', null, Validator::classIsBackedEnum(...));
+            }
```
jrushlow, line 430: To be consistent with the other available types, this should be a field type rather than a question. e.g. when a user does: 
image

Even though there is not a DBAL enum type* - there is a field enum type declaration - so we should advertise enums as a _type_ rather than treat as something special. 

As an aside, we also want to limit the number of questions we ask, and this is a perfect case to not ask an additional question when we already ask the user for the `Field Type`. This gives us wiggle room in the future to ask an additional question for feature xyz if we need to.

jrushlow (changes requested):
Howdy @loicsapone - Thank you for getting this going! I do have one change that we need to address before we can move forward with the review / merge.

Side note: this fixes #1440

src/Maker/MakeEntity.php
loicsapone: Hey @jrushlow,

I hesitated as well, and for the same reasons you mentioned, I didn't choose to create a new type. I'll revisit that decision to incorporate it into the type selection. 
Should we add it under the "Main Types" category?

jrushlow: Hmm... I'm thinking that it should probably be an "Other" type - just because enums are not "real" types. But having it as a main type would be perfectly acceptable too.
image

jrushlow (approved):
Look's good, let's run with it! Thank you @loicsapone

PR #1535, opened 2024-05-04
fix: Update ux-turbo package name to have proper suggestion in MakeEntity command

nfragnet:
Hi 👋

I tried to create an entity using `bin/console make:entity --broadcast` and got an error message suggesting a command to install the missing dependency. I think the command isn't working because the flex alias does not exist.

Capture d’écran 2024-05-04 à 12 48 04

I assumed that the correct dependency was `symfony/ux-turbo` and updated the MakeEntity class to declare the correct package name.

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -333,7 +333,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
         if (null !== $input && $input->getOption('broadcast')) {
             $dependencies->addClassDependency(
                 Broadcast::class,
-                'ux-turbo'
+                'symfony/ux-turbo'
             );
         }
 


jrushlow (approved):
@nfragnet - Thank you!

I'm not sure why the flex alias isn't working, but regardless, we're shifting to using the entire package name here in maker bundle rather than rely on flex aliases - so this is perfect. 

Note to future self, we should utilize https://github.com/symfony/maker-bundle/blob/b87b1b25c607a8a50832395bc751c784946a0350/src/Maker/Common/InstallDependencyTrait.php#L27 for `ux-turbo` and `api` bundles instead of throwing an error. This would behave like: https://github.com/symfony/maker-bundle/blob/b87b1b25c607a8a50832395bc751c784946a0350/src/Maker/Security/MakeCustomAuthenticator.php#L77

If you're up for that @nfragnet task, a PR would be much appreciated.

PR #1538, opened 2024-05-06
[release] v1.59.1

### LINKED ISSUE #1537: Deprecation : symfony/maker-bundle v1.59.0
Symfony 7.0.7
PHP 8.3.6
Symfony/maker-bundle v1.59.0
Dev. env.

Message retourned by profiler:

`Since symfony/maker-bundle v1.59.0: The "Symfony\Bundle\MakerBundle\Maker\MakeAuthenticator" class is deprecated, use any of the Security\Make* commands instead.`
Can i just uninstall the `Symfony\Bundle\MakerBundle\Maker\MakeAuthenticator` ? because is now a part of Security ? or wait for fix ? what is the good solution to avoid deprecation ?

Thanks by advance

jrushlow:
# Changelog

## [v1.59.1](https://github.com/symfony/maker-bundle/releases/tag/v1.59.1)

*May 5th, 2024*

### Bug

- [#1532](https://github.com/symfony/maker-bundle/pull/1532) [make:auth] only fire deprecation when command is called - *@eltharin*

---

fixes #1537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [v1.59.1](https://github.com/symfony/maker-bundle/releases/tag/v1.59.1)
+
+*May 5th, 2024*
+
+### Bug
+
+- [#1532](https://github.com/symfony/maker-bundle/pull/1532) [make:auth] only fire deprecation when command is called - *@eltharin*
+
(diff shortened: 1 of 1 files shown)


pierrealexaline:
after `composer update` , i tested and the deprecation has disappeared; as you say, it's only fired when you use the command ! good job, many thanks !

PR #1539, opened 2024-05-06
[make:crud|voter] generate classes with final keyword

### LINKED ISSUE #1536: Use final classes
Hi,
I'm referring to this issue https://github.com/symfony/demo/pull/1409.
Would it be possible to use `final` in classes? For example for controllers, repositories, validators, etc.

jrushlow:
- introduces 2 new configuration values:
    - `generate_final_classes`: Defaults to `true` - When true, all non-entity classes generated by MakerBundle will be final.
    - `generate_final_entities`: Defaults to `false` - When true, all entity classes generated by MakerBundle will be final.

- introduces an internal `ClassData` object that represents all of the meta data needed to generate a class from a template. It combines parts of `ClassNameDetails` & `UseStatementGenerator, while adding logic to generate a class declaration conditionally including the `final` keyword & `extends` class.

- `ClassData` is configured by `TemplateComponentGenerator` to handle `root_namespace`, `generate_final_classes`, & `generate_final_entities` if passed as a template `class_data` variable to `Generator::createClass()`. The idea being we don't have to modify the constructor for the "non-internal" `Generator::class`

- reduces the complexity of our `.tpl.php` templates. The `ClassData` object is accessible from within the templates (Crud & Voter for now). `$class_data->getClassDeclaration()` generates `final class SomeClass extends XyzClass`

- To keep the PR's "reviewable" this PR only adds the new functionality for `make:crud` & `make:voter`. To be implemented in additional makers in subsequent PR's.

- [x] fix test fixtures to use `final`
- [x] update _this_ description to include all of the proposed changes

fixes #1536 

---

I think instead of one massive PR, we will merge this PR first to include `make:voter` & `make:crud` -> create subsequent PR's for makers that touch entities (1 pr) and the rest of the makers (1 pr). I say this because _this_ PR is actually adding a handful of new features (mostly internal) in order to generate `final` classes.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -24,6 +24,8 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->children()
                 ->scalarNode('root_namespace')->defaultValue('App')->end()
+                ->booleanNode('generate_final_classes')->defaultTrue()->end()
+                ->booleanNode('generate_final_entities')->defaultFalse()->end()
             ->end()
         ;
 
--- a/src/DependencyInjection/MakerExtension.php
+++ b/src/DependencyInjection/MakerExtension.php
@@ -49,6 +49,7 @@ class MakerExtension extends Extension
         $componentGeneratorDefinition
             ->replaceArgument(0, $config['generate_final_classes'])
             ->replaceArgument(1, $config['generate_final_entities'])
+            ->replaceArgument(2, $rootNamespace)
         ;
 
         $container->registerForAutoconfiguration(MakerInterface::class)
--- a/src/DependencyInjection/MakerExtension.php
+++ b/src/DependencyInjection/MakerExtension.php
@@ -45,6 +45,12 @@ class MakerExtension extends Extension
         $doctrineHelperDefinition = $container->getDefinition('maker.doctrine_helper');
         $doctrineHelperDefinition->replaceArgument(0, $rootNamespace.'\\Entity');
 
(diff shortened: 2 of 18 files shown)


seb-jean:
Thanks Kevin

jrushlow:
Hmm. So I don't _think_ we can `final` classes by default without causing confusion amongst the consumers of MakerBundle. I'm not 100% on my thought process here, but doctrine orm is using lazy ghosts by default - atleast that's what we're doing in the `DoctrineBundleRecipe` -> https://github.com/symfony/recipes/blob/8cba886fb716a49e654845980cba33e211288a9b/doctrine/doctrine-bundle/2.9/config/packages/doctrine.yaml#L11

Like the failed tests for this PR, wouldn't we trigger logic exceptions everywhere in userland code similar to:
```
Symfony\Component\VarExporter\Exception\LogicException: Cannot generate lazy ghost: class "Symfony\Bundle\MakerBundle\Tests\tmp\current_project\src\Entity\User" is final.
```

if all the generated entities were `final`, we'd end up forcing the user to either a) remove the `final` keyword everytime they `make:user|entity|security` etc... || b) set `orm.enable_lazy_ghost_objects: false`. Both of those options are not viable.

I think the only way we can implement the `final` keyword for generated _entities_ would be on a opt-in basis with a `--is-final` flag. We _should_ be able to implement the `final` keyword for _non-entity_ classes by default. 

If I'm missing something obvious - please let me know.. Thoughts @symfony/mergers ?

Issue #1540, opened 2024-05-06
PHPStan - User::getPassword()


seb-jean:
Hi,

With PHPStan, I have the following error:
`Method App\Entity\User::getPassword() should return string but returns string|null.`

The generated code is as follows:
```php

// src/Entity/User.php

public function getPassword(): string
{
    return $this->password;
}
```

Instead we should have the code below:

```php

// src/Entity/User.php

public function getPassword(): ?string
{
    return $this->password;
}
```

PR #1542, opened 2024-05-06
[make:*] add types for PHPStan

seb-jean:
Hi,

I have add types for PHPStan for:

-  `make:webhook`
- `make:crud`
- `make:message`
- `make:registration-form`
- `make:validator`
- `make:reset-password`

What do you think ?

### DIFF
--- a/src/Resources/skeleton/validator/Validator.tpl.php
+++ b/src/Resources/skeleton/validator/Validator.tpl.php
@@ -7,7 +7,7 @@ use Symfony\Component\Validator\ConstraintValidator;
 
 class <?= $class_name ?> extends ConstraintValidator
 {
-    public function validate($value, Constraint $constraint)
+    public function validate(mixed $value, Constraint $constraint): void
     {
         /** @var <?= $constraint_class_name ?> $constraint */
 
--- a/src/Security/UserClassBuilder.php
+++ b/src/Security/UserClassBuilder.php
@@ -179,7 +179,7 @@ final class UserClassBuilder
             // add an empty method only
             $builder = $manipulator->createMethodBuilder(
                 'getPassword',
-                'string',
+                '?string',
                 true,
                 [
                     'This method can be removed in Symfony 6.0 - is not needed for apps that do not check user passwords.',
@@ -216,7 +216,7 @@ final class UserClassBuilder
 
             $manipulator->addGetter(
                 'password',
-                'string',
+                '?string',
                 true
             );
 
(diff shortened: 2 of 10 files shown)


seb-jean:
A commit is related to the issue: https://github.com/symfony/maker-bundle/issues/1540

src/Resources/skeleton/validator/Validator.tpl.php
```diff
@@ -7,7 +7,7 @@
 
 class <?= $class_name ?> extends ConstraintValidator
 {
-    public function validate($value, Constraint $constraint)
+    public function validate(mixed $value, Constraint $constraint): void
```
jrushlow, line 10: I'm not sure if this is possible (e.g. maker has enough knowledge about `$value`) but can we do `string|array<int, mixed>|int|etc.....` here instead of `mixed`? Thinking about phpstan level 9?  where it complains if something is `mixed` or if an `array` does not define its shape.

I'm not suggesting we do all potential types as a union, but if we know `$value` could be say, either a `string` or a `int` - we should use `string|int` as the type instead of mixed.

else `mixed` is perfectly fine here :wink:

src/Security/UserClassBuilder.php
```diff
@@ -179,7 +179,7 @@ private function addGetPassword(ClassSourceManipulator $manipulator, UserClassCo
             // add an empty method only
             $builder = $manipulator->createMethodBuilder(
                 'getPassword',
-                'string',
+                '?string',
                 true,
                 [
                     'This method can be removed in Symfony 6.0 - is not needed for apps that do not check user passwords.',
```
jrushlow, line 185: Do we even need to add this method anymore? I'm just looking at the comment that we add - maker only supports symfony `6.4`+

If we do not need the method - lets leave this alone and remove the method in a separate PR.

jrushlow (changes requested):
Hey @seb-jean - awesome! A couple minor comments below and one request:

I'm working on v2 of SymfonyCasts `ResetPasswordBundle` && `VerifyEmailBundle`, I'm planning on introducing generics either in the last `1.x` feature release or in the first `2.x` release.

 Tl;dr, can you split the changes to code that touches `make:reset-password` and `make:registration` from this PR into their own PR's here in maker bundle? e.g. `[make:reset-password] phpstan types` & `[make:registration] a good title for the changelog` Reasoning: it'll help me over in the symfonycasts repo's and I'll be able to tack on the changes made over there to your 2 new PR's. 

Thank you in advance!

seb-jean:
Closed in favor of other PRs I just created.

PR #1549, opened 2024-05-06
[make:registration-form] improve generated types for phpstan

seb-jean:
---
edited by: @jrushlow 

Improves type hints in generated code to reduce userland PHPStan squawks.


jrushlow (approved):
Thank you @seb-jean

PR #1561, opened 2024-05-15
[make:*] generate test classes properly & fix class name details suffix

jrushlow:
- When passing a `suffix` to `Generator::createClassNameDetails()`, the suffix is _not_ added to an absolute class name. Our docs say that passing a suffix "...guarantee is on the end of the class". This PR fixes that.

- Generates tests in the proper `App\Tests\` namespace instead of `App\Test\`

---

### Internals

When generating tests (`--with-tests`), I was lazy and used `generateFile()` to create the test class (in part because of the suffix bug) instead of using the correct `generateClass()` method. This PR fixes that. It's also needed to accomplish #1539 with test classes.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -153,6 +153,10 @@ class Generator
             }
         }
 
+        if (!str_ends_with($className, $suffix)) {
+            $className = sprintf('%s%s', $className, $suffix);
+        }
+
         Validator::validateClassName($className, $validationErrorMessage);
 
         // if this is a custom class, we may be completely different than the namespace prefix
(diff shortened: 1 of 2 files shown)


src/Generator.php
```diff
@@ -153,6 +153,10 @@ public function createClassNameDetails(string $name, string $namespacePrefix, st
             }
         }
 
+        if (!str_ends_with($className, $suffix)) {
+            $className = sprintf('%s%s', $className, $suffix);
+        }
```
jrushlow, line 158: Hmm... I don't think we should "fix" this here. One could argue that because you're passing an absolute class name into this method, it should not be altered. Under that assumption - this "fix" would actually be a BC break as it's changing the intended functionality from https://github.com/symfony/maker-bundle/commit/9e90cbc6d7d557187d3f097fdd0a7e1a883c7fd1#diff-4ee5eaad222ef0782fccd29473a5f1860988b7cf91cdcd0a12628b2377972b36

 Chewing on this for a better solution

PR #1562, opened 2024-05-16
[docs] document new config values

jrushlow:
- [x] depends on #1539 

---

- adds the ability to generate / preview our single page doc locally (https://symfony.com/bundles/SymfonyMakerBundle/current/index.html)

Yup - totally stole the code from https://github.com/symfony/symfony-docs/tree/7.1/_build to preview the build

Open to ideas on how we can preview the docs that ultimately get published on symfony.com without doing the whole "make change -> merge PR" rinse and repeat cycle...


javiereguiluz (approved):
I think this is correct. Thanks Jesse!

If we assume that the doc of this bundle will always be a single page (which I think it's safe to assume) then you could simplify some things if you want:

* You could remove the `disable-cache` option because when having 1 doc page, it doesn't make any noticeable difference
* All the logic related to this --> _"fix assets URLs to make them absolute (otherwise, they don't work in subdirectories)"_ could be removed because you don't have images and, if you include images in the future, you won't have subdirectories for contents because all the docs are in a single page

A final comment: something very useful for us is to run a RST linter to docs. It detects so many issues! We use DOCtor-RST (created by @OskarStark and actively developed by folks like @alamirault). We use it like this:

https://github.com/symfony/symfony-docs/blob/9bfc9fad274711f88b77364b7d36a7db86c79f7f/.github/workflows/ci.yaml#L53-L78

.doctor-rst.yaml
```diff
@@ -0,0 +1,113 @@
...
+#    - '.. versionadded:: 2.7.1' # Doctrine
+#    - '123,' # assertion for var_dumper - components/var_dumper.rst
+#    - '"foo",' # assertion for var_dumper - components/var_dumper.rst
+#    - '$var .= "Because of this `\xE9` octet (\\xE9),\n";'
+#    - '.. versionadded:: 0.2' # MercureBundle
+#    - '.. versionadded:: 3.6' # MonologBundle
+#    - '.. versionadded:: 3.8' # MonologBundle
+#    - '.. versionadded:: 3.5' # Monolog
+#    - '.. _`a feature to test applications using Mercure`: https://github.com/symfony/panther#creating-isolated-browsers-to-test-apps-using-mercure-or-websocket'
+#    - 'End to End Tests (E2E)'
```
OskarStark, line 113: new line at end of file missing

.doctor-rst.yaml
```diff
@@ -0,0 +1,113 @@
...
+#exclude_rule_for_file:
+#  - path: configuration/multiple_kernels.rst
+#    rule_name: replacement
+#  - path: page_creation.rst
+#    rule_name: no_php_open_tag_in_code_block_php_directive
+#  - path: frontend/create_ux_bundle.rst
+#    rule_name: argument_variable_must_match_type
+
+# do not report as violation
+whitelist:
```
OskarStark, line 91: I think there is no whitelist needed, right?

src/Resources/doc/index.rst
```diff
@@ -65,18 +65,74 @@ file by their respective environment variables:
...
+See https://www.php.net/manual/en/language.oop5.final.php
+
+.. code-block:: php
+
+    final class MyVoter
+    {
+        ...
+    }
+
+.. versionadded:: 1.60.0
```
OskarStark, line 108: I would propose to only show the minor versions, as patch versions does not introduce new features

src/Resources/doc/index.rst
```diff
@@ -65,18 +65,74 @@ file by their respective environment variables:
...
+classes with the ``final`` PHP keyword. Set this to ``true``
+to override this behavior for all maker commands that create
+entities.
+
+See https://www.php.net/manual/en/language.oop5.final.php
+
+.. code-block:: php
+
+    #[ORM\Entity(repositoryClass: TaskRepository::class)]
+    class Task extends AbstractEntity
```
OskarStark, line 128:
```suggestion
    final class Task extends AbstractEntity
```
?

src/Resources/doc/index.rst
```diff
@@ -65,18 +65,74 @@ file by their respective environment variables:
...
+    {
+        ...
+    }
+
+.. versionadded:: 1.60.0
+
+    ``generate_final_entities`` was introduced in MakerBundle v1.60
```
OskarStark, line 135:
```suggestion
    ``generate_final_entities`` was introduced in MakerBundle 1.60.
```

should always end with a dot and I am not sure the `v` is needed

jrushlow: note to self: when updating this, `1.60` needs to be bumped to `61` or higher - we havent merged the feature yet.

PR #1571, opened 2024-06-14
[make:twig-component] Improve `make:twig-component` by reading the configuration file

shadowc:
Small change in the way `make:twig-component` works when determining the namespace of the class to create.

In before, the namespace was hard coded. This is still good, but in cases where your Twig Component class namespace may have changed (i.e. your website is divided into `src/Admin` and `src/Site` and you've modified your `twig_component.yaml` so that your TwigComponents live under `src/Site/Twig/Components`), the component class would be still created under `src/Twig/Components`.

I've improved the command to read this config and automatically create the component under the first entry in this file. If no config is found, then we can use the old default.

TODOS:
- [x] Can a test be made with a special config file?

### DIFF
--- a/src/Maker/MakeTwigComponent.php
+++ b/src/Maker/MakeTwigComponent.php
@@ -20,7 +20,6 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Yaml\Exception\ParseException;
 use Symfony\Component\Yaml\Yaml;
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
--- a/src/Maker/MakeTwigComponent.php
+++ b/src/Maker/MakeTwigComponent.php
@@ -20,7 +20,6 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Yaml\Yaml;
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
--- a/src/Maker/MakeTwigComponent.php
+++ b/src/Maker/MakeTwigComponent.php
@@ -107,7 +107,7 @@ final class MakeTwigComponent extends AbstractMaker
         if ($this->fileManager->fileExists($path)) {
             $value = Yaml::parse($this->fileManager->getFileContents($path));
 
-            $this->namespace = \substr(\array_key_first($value['twig_component']['defaults']), 4);
+            $this->namespace = substr(array_key_first($value['twig_component']['defaults']), 4);
         }
     }
 }
--- a/src/Maker/MakeTwigComponent.php
+++ b/src/Maker/MakeTwigComponent.php
@@ -72,7 +72,7 @@ final class MakeTwigComponent extends AbstractMaker
         if ($this->fileManager->fileExists($path)) {
             $value = Yaml::parse($this->fileManager->getFileContents($path));
 
-            $nameSpacePrefix = \array_key_first($value['twig_component']['defaults']);
+            $nameSpacePrefix = \substr(\array_key_first($value['twig_component']['defaults']), 4);
         }
 
         $factory = $generator->createClassNameDetails(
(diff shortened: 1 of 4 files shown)


jrushlow (changes requested):
Awesome Work! We do need a test case for this. we can probably just copy & paste this one -> https://github.com/symfony/maker-bundle/blob/c305a02a22974670f359d4274c9431e1a191f559/tests/Maker/MakeTwigComponentTest.php#L22

For the custom configuration file needed in the test ("fixture"), we would create that in `maker-bundle/tests/fixtures/make-twig-components/custom_twig_component.yaml` 

Then in our test case, point to it using the `$runner->copy()` command. Similar to what we do here

https://github.com/symfony/maker-bundle/blob/c305a02a22974670f359d4274c9431e1a191f559/tests/Maker/MakeTwigComponentTest.php#L31

So it would look something like:
```php
$runner->copy(
    'make-twig-component/tests/custom_twig_component.yaml',    // path to fixture (omit the tests/fixture prefix)
    'config/packages/twig_compont.yaml`.  // destination relative to the "test app's root"
); 
```

If any changes need to be made to the existing tests in the `fixtures/make-twig-components` to run in our _new_ test, just copy the existing fixture file -> rename it to say "GeneratedComponentTwigTest_customNamespace.php" and modify the `$runner->copy()` appropriately`.. 

Hope this helps, if ya get stuck, slack or on here :)

src/Maker/MakeTwigComponent.php
```diff
@@ -93,5 +102,23 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+            return;
+        }
+
+        try {
+            $value = Yaml::parse($this->fileManager->getFileContents($path));
+        } catch (ParseException $e) {
+            return;
+        }
+
+        if (!array_key_exists('twig_component', $value) || !array_key_exists('defaults', $value['twig_component'])) {
```
jrushlow, line 118: I believe we would need to either wrap this in the try/catch or set `$value` to an empty array. Won't this fail if `$value` is not defined?

jrushlow: Doh! never mind, we're already returning if `parse` fails, so $value` would always exist at this point..

shadowc: Yes, on the other hand, if the Twig configuration file is corrupted, wouldn't the container just fail to boot and would we ever reach this `catch`?

shadowc: Also, by looking at `Yaml::parse` it can return `null` some times, so would a better option be:

```php
try {
    $value = Yaml::parse($this->fileManager->getFileContents($path));
    $this->namespace = substr(array_key_first($value['twig_component']['defaults']), 4);
} catch (\Throwable) {
}
```

shadowc:
@jrushlow Hey, I created the test as you told me, but it's failing to load the config file! Let me know if I'm doing something wrong!

tests/Maker/MakeTwigComponentTest.php
```diff
@@ -37,6 +37,28 @@ public function getTestDetails(): \Generator
...
+            ->addExtraDependencies('symfony/ux-twig-component', 'symfony/twig-bundle')
+            ->run(function (MakerTestRunner $runner) {
+                $output = $runner->runMaker(['Alert']);
+
+                $this->assertStringContainsString('src/Site/Twig/Components/Alert.php', $output);
+                $this->assertStringContainsString('templates/components/Alert.html.twig', $output);
+                $this->assertStringContainsString('To render the component, use <twig:Alert />.', $output);
+
+                $runner->copy(
+                    'make-twig-component/tests/custom_twig_component.yaml',
```
jrushlow, line 50:
```suggestion
                    'make-twig-component/custom_twig_component.yaml',
```

That should do the trick. Looks like the yaml file is in `tests/fixtures/make-twig-component/custom_twig_component.yaml`

src/Maker/MakeTwigComponent.php
jrushlow: Ya I think that would work. Then catch the throwable and throw a `RuntimeCommandException` https://github.com/symfony/maker-bundle/blob/main/src/Exception/RuntimeCommandException.php Which will ultimately be shown in the terminal. 

Something like:
```php
throw new RuntimeCommandException(message: 'Unable to parse twig_components.yaml', previous: $throwable);
```

shadowc:
I think this didn't work either...

src/Maker/MakeTwigComponent.php
shadowc: Should we throw though? Or just let the command continue with default values?

shadowc:
Ok, heads up: Thanks for all the help!! Looks like tests are passing and I've made the last changes of what as discussed! I think this is ready to go!

jrushlow (approved):
Awesome work @shadowc! Thank you!

DaveData:
This breaks if config is in php files.

Maybe go with default if yaml file can't be read? Do we even need to open/parse the yaml file to get access to the value?

shadowc:
@jrushlow hey! Let me know if I can fix this somehow? It seems we should be able to read php configs AND also default to the default value.

PR #1572, opened 2024-06-15
[make:twig-component] docs: Adding twig component examples and references to sister documentation

techSolstice:
Addresses [symfony/ux#480](https://github.com/symfony/ux/issues/480)

### DIFF
--- a/src/Resources/doc/index.rst
+++ b/src/Resources/doc/index.rst
@@ -41,6 +41,55 @@ optional arguments and options. Check them out with the ``--help`` option:
 
     $ php bin/console make:controller --help
 
+Creating a Twig Component
+~~~~~~~
+
+Run this command to install the `Symfony UX package`_:
+
+.. code-block:: terminal
+
+    $ composer require symfony/ux-twig-component
+
+Make a basic `twig component`_:
+
+.. code-block:: terminal
+    $ php bin/console make:twig-component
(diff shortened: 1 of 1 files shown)


jrushlow (changes requested):
Howdy @techSolstice - This is awesome but I think @kbond meant for references to be added to https://github.com/symfony/ux/blob/2.x/src/TwigComponent/doc/index.rst

I'm not opposed to adding new documentation to MakerBundles docs, but typically we only do that for things that affect MakerBundle itself. E.g. global configuration. For now, I'm not sure that we should add this here seeing as we don't have doc's for other specific maker commands. 

I'll keep this open for the time being and chew on it for a bit - this might be something we should be doing, albeit in a more limited fashion. E.g.

```
make:twig-component - Generates a TwigComponent(link to docs)
    Arguments:
        "name" - the name of the twig component, e.g. MyTwigComponent
   Options:
        -q   Quiets the output
        -v   Verbose
```

_If_ we end up doing this, I'd want to automate this process so we can just pull each commands name / description & the arguments / options from the code base directly. Otherwise these docs would probably be forgotten about. 

cc @kbond @javiereguiluz

techSolstice:
@jrushlow Created https://github.com/symfony/ux/pull/1923 for now

PR #1573, opened 2024-06-19
[make:twig-component] Fix config file in error messages

smnandre:
(minor)


jrushlow (approved):
Thank you for catching this @smnandre

PR #1574, opened 2024-06-20
[make:event] Add command for making Event class

Ippey:
I think adding "make:event" that makes `Event` class is good, although that is a small feature.
So, I wrote a simple make command.

```
php bin/console make:event


 The name of the event class (e.g. OrderPlacedEvent):
 > ItemCreatedEvent

 Field name (press enter to stop adding fields):
 > id

 Field type (e.g. string) [string]:
 > int

 Field visibility (public, protected, private) [public]:
  [0] public
  [1] private
  [2] protected
 > public

 Can this field be null (nullable) (yes/no) [no]:
 > no

 Field name (press enter to stop adding fields):
 > 
```

It makes bellow

```php
<?php

namespace App\Event;

use Symfony\Contracts\EventDispatcher\Event;

final class ItemCreatedEvent extends Event
{
    public function __construct(
        public readonly int $id,
    ) {
    }
}
```

### DIFF
--- a/src/Maker/MakeEvent.php
+++ b/src/Maker/MakeEvent.php
@@ -11,8 +11,6 @@
 
 namespace Symfony\Bundle\MakerBundle\Maker;
 
-use DateTime;
-use DateTimeImmutable;
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
 use Symfony\Bundle\MakerBundle\Doctrine\DoctrineHelper;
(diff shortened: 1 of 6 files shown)


94noni:
i thinks its a good idea, would definitely use it if it lands in the bundle

perhaps we can imagine a step forward with adding the command asking the Event "data" in an asking interactive question way? like some other existing makers commands

> bin/console make:event
> ItemCreatedEvent
> would you like to add some event's properties? yes (continue) / no (exit and generate code)
> Property name?
> identifier
> Property type?
> integer
> Property visibility?
> public

```
<?php

namespace App\Event;

use Symfony\Contracts\EventDispatcher\Event;

final class ItemCreatedEvent extends Event
{
    public function __construct(public int $identifier)
    {
    }
}
```

wdyt?

Ippey:
@94noni Thank you. That's great!
I'll try it.

Ippey:
Updated. It can make an event class with properties interactively.
We can set bellow:

- name
- type
- visibility
- nullable

94noni (approved):

PR #1575, opened 2024-06-25
[make:*] Use a PHP-CS-Fixer shim rather than an external PHAR

theofidry:
This PR proposes to leverage the PHP-CS-Fixer shim package, which is the Composer package shipping the scoped PHAR, rather than managing the PHAR directly.

IMO it is better because on the maintainer side it is a package like usual and you do not have the downside of polluting your project.

There is however the downside still that a project could use a shim with an incompatible version... And this may not be desired. In which case simply ignore this PR! :)


jrushlow:
Thank you @theofidry - I didn't realize phpcsfixer had created a shim since we introduced this feature. I'm definitely all for it as it will save me from having to maintain the `phar` versions periodically. 

I'll need to do some toying around with this locally before we merge to make sure we are on the up and up.

jrushlow (approved):
Thank you @theofidry! This will give the user the ability to use an up to date version of PHP-CS-Fixer without having to wait around on maintainers to update the phar manually. Win Win...

composer.json
```diff
@@ -16,6 +16,7 @@
         "php": ">=8.1",
         "doctrine/inflector": "^2.0",
         "nikic/php-parser": "^4.18|^5.0",
+        "php-cs-fixer/shim": "^v3.64",
```
Seb33300, line 19: This should be in `require-dev`

mmarton: yes, and it has a recipie and created/update a bunch of files in my project

HypeMC: #1648

Chris53897: https://github.com/symfony/maker-bundle/issues/1644

mbaschnitzi: It should not be in `require-dev` as php-cs-fixer is used at runtime to format the generated files. (See the changes in `\Symfony\Bundle\MakerBundle\Util\TemplateLinter`) I think `symfony/maker-bundle` should instead require `friendsofphp/php-cs-fixer`, so the decision to use the shim package can be made by the user.

pr #1651 contains this change.

PR #1576, opened 2024-06-25
[ci] allow template linter tests on OSX (Darwin)

theofidry:
See https://github.com/symfony/maker-bundle/pull/1429/files#r1653262959.

---
Edited by @jrushlow

Uses the `PHP_WINDOWS_VERSION_MAJOR` constant to determine if the OS is Windows based. Rather than doing a string `win` comparison with `\PHP_OS`. The `PHP_WINDOWS_VERSION_*` constants are only available in the Windows environment. https://www.php.net/manual/en/info.constants.php#constant.php-windows-version-major

Linter tests are still skipped on windows based CI.

### DIFF
--- a/src/Util/TemplateLinter.php
+++ b/src/Util/TemplateLinter.php
@@ -59,7 +59,7 @@ final class TemplateLinter
             $templateFilePath = [$templateFilePath];
         }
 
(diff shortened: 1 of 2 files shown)


jrushlow (approved):
Thank you @theofidry

PR #1579, opened 2024-07-03
[make:listener] Match event name against active events class/id

### LINKED ISSUE #1410: [Make:Listener] Improve event name matching
# Description
When using a class's FQCN as event name, you have to enter the FQCN in the command to make it work.

If the event name given by the user doesn't match a registered event name nor an FQCN, we could try to 'guess' if the input matches a class short name from the `App\\` namespace (we could even sneak some levenshtein in there to catch typos) and ask if the user meant one of the matching results.

Do you think that would be doable without ruining the performance (I imagine that on large projects with a lot of classes that could be a rather heavy process) ?

# Example

```bash
Choose a class name for your event listener or subscriber (e.g. ExceptionListener or ExceptionSubscriber):
 > CustomEventListener

What event do you want to listen to?:
 > MyCustomEvent

Did you mean ... [MyCustomEvent]
[0] MyCustomEvent
[1] App\Event\MyCustomEvent
 > App\Event\MyCustomEvent  # <- Here we would have auto-completion from the suggestions above
```

maelanleborgne:
Closes #1410 

## Description

The goal of this PR is to simplify the usage of the `make:listener` command by adding the following features to the event input :

- Matching of the event name to existing events FQCN.
- Typo detection (using levenshtein distance).

## Example

Matching on class short name :

image

Matching on class short name when multiple events match : 

image

Typo detection : 

image

### DIFF
--- a/src/Maker/MakeListener.php
+++ b/src/Maker/MakeListener.php
@@ -23,6 +23,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Question\Question;
 use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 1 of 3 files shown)


94noni:
I like the idea and will indeed simplify such maker, and behavior can be duplicated on others as well I think
but have no opinion on the implementation

src/Maker/MakeListener.php
```diff
@@ -102,6 +103,33 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
             $input->setArgument('event', $event);
         }
+
+        $event = $input->getArgument('event');
+        if (null === $this->getEventConstant($event) && null === $this->eventRegistry->getEventClassName($event)) {
+            $eventList = $this->eventRegistry->getAllActiveEvents();
+            $eventFQCNList = array_filter(array_map($this->eventRegistry->getEventClassName(...), $eventList), fn ($eventFQCN) => \is_string($eventFQCN));
+            $eventIdAndFQCNList = array_unique(array_merge($eventList, $eventFQCNList));
+            $suggestionList = [];
+            foreach ($eventIdAndFQCNList as $eventSuggestion) {
```
jrushlow, line 113: Add spacing for readability between logic statements (e.g. `if`'s) to be consistent with other code in the bundle.

```suggestion
            $suggestionList = [];
            
            foreach ($eventIdAndFQCNList as $eventSuggestion) {
```

We need to do this for the other `if` statements below as well.

src/Maker/MakeListener.php
```diff
@@ -102,6 +103,33 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+                    $suggestionList[] = $eventSuggestion;
+                }
+            }
+            if (!$suggestionList) {
+                return;
+            }
+            if (1 === \count($suggestionList)) {
+                $question = new ConfirmationQuestion(sprintf('<fg=green>Did you mean</> <fg=yellow>"%s"</> <fg=green>?</>', $suggestionList[0]), false);
+                $event = $io->askQuestion($question) ? $suggestionList[0] : $event;
+            } else {
```
jrushlow, line 124: We can eliminate this `else` condition by returning early inside the `if` conditional. Then just have the statements within the `else` called after the `if`.

jrushlow (changes requested):
Thank you for working on this @maelanleborgne - a couple of minor comments below...

jrushlow (approved):
Thank you @maelanleborgne

PR #1583, opened 2024-07-30
[make:crud] Remove `/` from from index action URL

seb-jean:
I propose to remove the `/` from the index action URL because all other URLs in the [src/Resources/skeleton/crud/controller/Controller.tpl.php](https://github.com/seb-jean/maker-bundle/blob/main/src/Resources/skeleton/crud/controller/Controller.tpl.php) file do not end with a `/`. This is more consistent.

### DIFF
--- a/src/Resources/skeleton/crud/controller/Controller.tpl.php
+++ b/src/Resources/skeleton/crud/controller/Controller.tpl.php
@@ -7,7 +7,7 @@ namespace <?= $class_data->getNamespace() ?>;
 #[Route('<?= $route_path ?>')]
 <?= $class_data->getClassDeclaration() ?>
 {
-<?= $generator->generateRouteForControllerMethod('/', sprintf('%s_index', $route_name), ['GET']) ?>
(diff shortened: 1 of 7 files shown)


jrushlow (changes requested):
Howdy @seb-jean - I cannot see any downsides to this change. We would need to fix the test fixtures by changing the assertion to match the path without the trailing `/`.

 e.g. https://github.com/symfony/maker-bundle/blob/bd3b4d2d0a5067ee3587db03e5da6ca4818b2f29/tests/fixtures/make-crud/tests/it_generates_basic_crud.php#L12

jrushlow:
Note to self: I vaguely remember needing to have a trailing `/` in the "root" path(s) avoid phpunit test errors - or something to that affect. 

e.g. `/path/to/resource/` instead of `/path/to/resource` 

But I do the latter in my projects without issue, so this may have been resolved upstream.

jrushlow (approved):
Thank you @seb-jean!

PR #1584, opened 2024-07-31
Fix 1580 - Mulitiple and Nullable enum

### LINKED ISSUE #1580: MakerBundle "enum" entity type prompts "Can this field store multiple enum values" but doesn't handle the case

7.1.2

### Description

When creating a field of type "enum" for an entity with Symfony MakerBundle, after setting the enum, the CLI prompts you to choose whether the field can store multiple enum values or not. Choosing yes correctly generates the code by passing the enum to the field's attribute and setting the type to "SIMPLE_ARRAY" but then throws the following error when trying to save the form.
"Expected argument of type "your Enum", "array" given at property path "your field".

### How to reproduce

Create a valid string backed enum and run `php bin/console make:entity`. Enter whatever name, choose `enum` as the field type and specify your enum's full class name (e.g.: App\Enum\MyEnum). There is the most important step. When prompted "Can this field store multiple enum values", choose yes. Then create a form for this entity, make the field an `EnumType`, specify your Enum class name and set "multiple" to true. When sending the form, the error will be displayed. (while you would expect it to work out of the box as any MakerBundle generated code)

### Possible Solution

_No response_

### Additional Context

Setup and errors (entity code was generated by MakerBundle and is untouched)
image
image
image
image
image

Fan2Shrek:
Fixes #1580
Fix: No use statement when enum are multiple

The problem was due to the first condition `'array' === $typeHint && !$nullable`. 
If both "multiple" and "nullable" were selected, `$typeHint` was overridden by the second condition.


src/Util/ClassSourceManipulator.php
```diff
@@ -121,13 +121,23 @@ public function addEntityField(ClassProperty $mapping): void
...
         $defaultValue = null;
         $commentLines = [];
-        if ('array' === $typeHint && !$nullable) {
-            $defaultValue = new Node\Expr\Array_([], ['kind' => Node\Expr\Array_::KIND_SHORT]);
-            if (null !== $mapping->enumType) {
+
+        if (null !== $mapping->enumType) {
+            if ('array' === $typeHint) {
+                // still need to add the use statement
+                $this->addUseStatementIfNecessary($mapping->enumType);
```
jrushlow, line 128: I think we can eliminate the `else` conditional in this `if` statement by moving `addUserStatement...()` outside and after `if ('array' === $typeHint) is called. 

We're calling the add use statement method regardless if `array` === `$typeHint` correct?

src/Util/ClassSourceManipulator.php
```diff
@@ -121,13 +121,23 @@ public function addEntityField(ClassProperty $mapping): void
...
+
                 $commentLines = [\sprintf('@return %s[]', Str::getShortClassName($mapping->enumType))];
+                if ($nullable) {
+                    $commentLines[0] .= '|null';
```
jrushlow, line 132: To be consistent in the bundle, lets use `sprintf()`

```suggestion
                    $commentLines[0] = \sprintf('%s|null', $commentLines[0]);
```

src/Util/ClassSourceManipulator.php
```diff
@@ -121,13 +121,23 @@ public function addEntityField(ClassProperty $mapping): void
...
+                } else {
+                    $defaultValue = new Node\Expr\Array_([], ['kind' => Node\Expr\Array_::KIND_SHORT]);
+                }
+            } else {
+                $typeHint = $this->addUseStatementIfNecessary($mapping->enumType);
             }
```
jrushlow, line 138: For clarity to my suggestion above, removing that use statement call and moving this one outside of the `if` conditional would accomplish the same thing with less code if I'm not mistaken.
```suggestion
            }
            
            $typeHint = $this->addUseStatementIfNecessary($mapping->enumType);
```

jrushlow (changes requested):
2 minor details and we should be all set!

src/Util/ClassSourceManipulator.php
Fan2Shrek: This will override `$typeHint` with the enum type even if we use an array, which was the initial problem. 
We don't want to use the returned type of `addUseStatementIfNecessary` if we specify an array.

src/Util/ClassSourceManipulator.php
Fan2Shrek: > We're calling the add use statement method regardless if `array` === `$typeHint` correct?

We're calling it in both cases, but if array === $typeHint, we don't want the returned string; it's just to automate the use statement.

tests/Maker/MakeEntityTest.php
```diff
@@ -735,6 +735,29 @@ public function getTestDetails(): \Generator
...
+                'y',
+                // nullable
+                'y',
+                // finish adding fields
+                '',
+            ]);
+
+            $this->runEntityTest($runner);
+        }),
+        ];
```
Fan2Shrek, line 760: I don't know if this test is relevant

PR #1595, opened 2024-09-17
[make:security:form-login] no interaction for make:security:form-login

eltharin:
allow command make:security:form-login with name in argument to no interaction

as help says : make:security:form-login \[\<controllerName\>\]

### DIFF
--- a/src/Maker/Security/MakeFormLogin.php
+++ b/src/Maker/Security/MakeFormLogin.php
@@ -59,11 +59,11 @@ final class MakeFormLogin extends AbstractMaker
 
     private const SECURITY_CONFIG_PATH = 'config/packages/security.yaml';
     private YamlSourceManipulator $ysm;
+    private string $controllerName;
     private string $firewallToUpdate;
     private string $userClass;
     private string $userNameField;
-    /** @var ?array<string, mixed> */
-    private ?array $securityData = null;
+    private bool $willLogout;
 
     public function __construct(
         private FileManager $fileManager,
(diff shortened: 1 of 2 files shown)


src/Maker/Security/MakeFormLogin.php
```diff
@@ -57,11 +59,11 @@ final class MakeFormLogin extends AbstractMaker
 
     private const SECURITY_CONFIG_PATH = 'config/packages/security.yaml';
     private YamlSourceManipulator $ysm;
-    private string $controllerName;
     private string $firewallToUpdate;
     private string $userClass;
     private string $userNameField;
-    private bool $willLogout;
```
jrushlow, line 64: We need to keep both of these properties. MakerBundle is getting away from "abusing" `setArgument` and `getArgument` for the purposes of passing user input from one method to the next within a maker command. e.g. separation of concerns. 

tl;dr `interact()` configures and sets any values onto their respective properties that will later be needed in `generate()`

`interact()` should get, validate, and set input.
`generate()` should only use values that are set via class properties when possible.

src/Maker/Security/MakeFormLogin.php
```diff
@@ -111,38 +116,44 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
...
+                    'SecurityController',
+                    Validator::validateClassName(...)
+                ));
+        }
 
-        $securityHelper = new InteractiveSecurityHelper();
-        $this->firewallToUpdate = $securityHelper->guessFirewallName($io, $securityData);
-        $this->userClass = $securityHelper->guessUserClass($io, $securityData['security']['providers']);
-        $this->userNameField = $securityHelper->guessUserNameField($io, $this->userClass, $securityData['security']['providers']);
-        $this->willLogout = $io->confirm('Do you want to generate a \'/logout\' URL?');
```
jrushlow, line 131: Following on from the above, these all need to stay in interact. They should not be set in the `generate` method.

jrushlow (changes requested):
A couple of comments and we can get this moving...

src/Maker/Security/MakeFormLogin.php
eltharin: ok so we don't have to use --no-interaction option for all makers ?

eltharin:
@jrushlow Is this as you want now ?

PR #1597, opened 2024-09-23
[make:crud] name repository variable with entity prefix when generating tests

dr-matt-smith:
Use lowercase-first version of entity to name repository variable in the test class

e.g. if entity named “Product” then test class repository variable will be “productRepository” rather than just “repository”

this makes copy-paste of setup code from other test classes simpler, when a related entity needs to be created for test cases


jrushlow (approved):
Thank you @dr-matt-smith I appreciate the PR and the work that you do!

dr-matt-smith:
thanks for fixing the missing “-“ Jesse!
I saw the build failed but am only getting to look at it today

.. matt ..

PR #1611, opened 2024-10-28
[doc] make:entity requires Doctrine

alexislefebvre:
Someone on Stack Overflow asked (question is not published yet) if `make:entity` can be used without Doctrine.

It looks like MakerBundle can only manage Doctrine entities, so I added it in the docs, hopefully that is true. Otherwise we'll have to change or close this PR.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -41,6 +41,9 @@ optional arguments and options. Check them out with the ``--help`` option:
 
     $ php bin/console make:controller --help
 
+.. caution::
+
+    ``make:entity`` requires ``doctrine/orm`` to be installed and configured. This maker support only ORM, not ODM.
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -41,6 +41,9 @@ optional arguments and options. Check them out with the ``--help`` option:
 
     $ php bin/console make:controller --help
 
+.. caution::
+
+    ``make:entity`` requires Doctrine to be installed and configured.
```
WedgeSama, line 46:
```suggestion
    ``make:entity`` requires ``doctrine/orm`` to be installed and configured. This maker support only ORM, not ODM.
```

kbond:
Thank you @alexislefebvre.

Issue #1401, opened 2023-11-23
Add maker for decorate services


alexander-schranz:
Most people go over inheritance instead of composition when overwriting services. But the best way would be always go with composition when decorate the service. A maker `make:service-decorator` could automatically detect the interfaces which a decorated service implements and use so the composition when decorating services and call the inner service e.g.:

```php
class SomeService implements SomeInterface
{
    public function someMethod(): string
    {
        return 'test';
    }

    public function some2Method(): string
    {
    }
}

interface SomeInterface {
    public function someMethod(): string;
    public function some2Method(): void;
}
```

Calling `bin/console make:service-decorator` does:

```php
#[AsDecorator(decorates: SomeService::class)]
class DecoratorService implements SomeInterface
{
    public function __construct(private #[MapDecorated] SomeInterface $inner)
    {}

    public function someMethod(): string
    {
         return $this->inner->someMethod();
    }

    public function someMethod(): void
    {
        $this->inner->some2Method();
    }
}
```

An idea i just came to my mind while helping somebody decorate a service correctly, one common issue is also that `AsDecorator` requires the service id and people are used to write a class in it, there a maker could help with `Did you mean service ...`. In this specific case it was a twig extra extension which only has a service id and not a public class alias in the container.

PR #1613, opened 2024-11-10
[make:decorator] Add new maker to create decorator

WedgeSama:
Add new maker to create decorator.
https://github.com/symfony/maker-bundle/issues/1401

Allow to create new decorator for existing services.

It try first to decorate by implements:
- The service ID used is an interface => implements this interface
- The service class implements one (or more) interface(s) and it do not declare more public method => implements those interfaces

If cannot implements, it will fallback to extends:
- Simply extends the service's class
- Do not work for final class

e.g.

```shell
bin/console make:decorator Symfony\Component\Routing\Generator\UrlGeneratorInterface MyUrlGenerator
```

```php
<?php

namespace App;

use Symfony\Component\DependencyInjection\Attribute\AsDecorator;
use Symfony\Component\DependencyInjection\Attribute\AutowireDecorated;
use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
use Symfony\Component\Routing\RequestContext;

#[AsDecorator(UrlGeneratorInterface::class)]
final class MyUrlGenerator implements UrlGeneratorInterface
{
    public function __construct(
        #[AutowireDecorated]
        private readonly UrlGeneratorInterface $inner,
    ) {
    }

    public function generate(string $name, array $parameters = [], int $referenceType = self::ABSOLUTE_PATH): string
    {
        return $this->inner->generate($name, $parameters, $referenceType);
    }

    public function setContext(RequestContext $context): void
    {
        $this->inner->setContext($context);
    }

    public function getContext(): RequestContext
    {
        return $this->inner->getContext();
    }
}
```

### DIFF
--- a/src/Maker/MakeDecorator.php
+++ b/src/Maker/MakeDecorator.php
@@ -1,0 +1,141 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Maker;
+
+use Psr\Container\ContainerInterface;
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Bundle\MakerBundle\InputConfiguration;
+use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Bundle\MakerBundle\Util\DecoratorInfo;
+use Symfony\Bundle\MakerBundle\Validator;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Question\Question;
+use Symfony\Component\DependencyInjection\Attribute\AsDecorator;
+
+/**
+ * @author Benjamin Georgeault <git@wedgesama.fr>
+ */
+final class MakeDecorator extends AbstractMaker
+{
+    /**
+     * @param array<string> $ids
+     */
+    public function __construct(
+        private readonly ContainerInterface $container,
+        private readonly array $ids,
+    ) {
+    }
+
+    public static function getCommandName(): string
+    {
+        return 'make:decorator';
+    }
+
+    public static function getCommandDescription(): string
+    {
+        return 'Create CRUD for Doctrine entity class';
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig): void
+    {
+        $command
+            ->addArgument('id', InputArgument::OPTIONAL, 'The ID of the service to decorate.')
+            ->addArgument('decorator-class', InputArgument::OPTIONAL, \sprintf('The class name of the service to create (e.g. <fg=yellow>%sDecorator</>)', Str::asClassName(Str::getRandomTerm())))
+            ->setHelp($this->getHelpFileContents('MakeDecorator.txt'))
+        ;
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies): void
+    {
+        $dependencies->addClassDependency(
+            AsDecorator::class,
+            'dependency-injection',
+        );
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
+    {
+        // Ask for service id.
+        if (null === $input->getArgument('id')) {
+            $argument = $command->getDefinition()->getArgument('id');
+
+            ($question = new Question($argument->getDescription()))
+                ->setAutocompleterValues($this->ids)
+                ->setValidator(fn ($answer) => Validator::serviceExists($answer, $this->ids))
+                ->setMaxAttempts(3);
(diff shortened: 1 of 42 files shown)


src/Maker/MakeDecorator.php
```diff
@@ -0,0 +1,141 @@
...
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
+    {
+        // Ask for service id.
+        if (null === $input->getArgument('id')) {
+            $argument = $command->getDefinition()->getArgument('id');
+
+            ($question = new Question($argument->getDescription()))
+                ->setAutocompleterValues($this->ids)
+                ->setValidator(fn ($answer) => Validator::serviceExists($answer, $this->ids))
```
Neirda24, line 77: based on the issue I'd suggest that the command accept not only the id but a classname and tries to guess the id based on it (if the service does not exists for example)

Neirda24: 3 yes with fault tolerance maybe ? `serializer` vs `serialiser` There are example in symfony's code about this.

tests/DependencyInjection/Fixtures/ServiceA.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Symfony\Bundle\MakerBundle\Tests\DependencyInjection\Fixtures;
+
+class ServiceA implements ServiceInterface
+{
+    public function getName(): string
+    {
+        return 'service_a';
+    }
+
+    public function getDefault(string $mode = self::MODE_FOO): ?string
```
Jean-Beru, line 21: Maybe we can use this test to check different argument representations. Ex: without type, without default value, PHP8 attributes, variadic arguments, etc.)

src/Maker/MakeDecorator.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public static function getCommandName(): string
+    {
+        return 'make:decorator';
+    }
+
+    public static function getCommandDescription(): string
+    {
+        return 'Create CRUD for Doctrine entity class';
```
PierreCapel, line 49:
```suggestion
        return 'Create a decorator of a service class';
```

src/Maker/MakeDecorator.php
WedgeSama: Done. I also did guess from class/interface short name.

src/Maker/MakeDecorator.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public function configureCommand(Command $command, InputConfiguration $inputConfig): void
+    {
+        $command
+            ->addArgument('id', InputArgument::OPTIONAL, 'The ID of the service to decorate.')
+            ->addArgument('decorator-class', InputArgument::OPTIONAL, \sprintf('The class name of the service to create (e.g. <fg=yellow>%sDecorator</>)', Str::asClassName(Str::getRandomTerm())))
+            ->setHelp($this->getHelpFileContents('MakeDecorator.txt'))
```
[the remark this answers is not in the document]
WedgeSama: Added

src/Maker/MakeDecorator.php
WedgeSama: Done

tests/DependencyInjection/Fixtures/ServiceA.php
WedgeSama: I test it here [tests/Util/ClassSource/ClassMethodTest.php#L62](https://github.com/WedgeSama/maker-bundle/blob/feature/make-decorator/tests/Util/ClassSource/ClassMethodTest.php#L62)
Is it enough?

src/Util/DecoratorInfo.php
```diff
@@ -0,0 +1,270 @@
...
+    {
+        foreach ($method->getParameters() as $parameter) {
+            $default = null;
+            if ($parameter->isOptional()) {
+                if ($parameter->isDefaultValueConstant()) {
+                    $default = $parameter->getDefaultValueConstantName();
+                } elseif ($parameter->isDefaultValueAvailable()) {
+                    $defaultValue = $parameter->getDefaultValue();
+
+                    if (\is_string($defaultValue)) {
```
Neirda24, line 181: is variadic also supported ?

src/Util/DecoratorInfo.php
```diff
@@ -0,0 +1,270 @@
...
+
+        if ($type instanceof \ReflectionUnionType) {
+            return implode('|', array_map($this->parseType(...), $type->getTypes()));
+        }
+
+        if ($type instanceof \ReflectionIntersectionType) {
+            return implode('&', array_map($this->parseType(...), $type->getTypes()));
+        }
+
+        throw new RuntimeCommandException('Should never be reach.');
```
Neirda24, line 225:
```suggestion
        throw new RuntimeCommandException('Should never be reached.');
```

GromNaN:
Thank you for this maker, it fits well within the project's scope for a feature that's genuinely tricky to get right by hand.

Since this branch conflicts with the config changes merged into `1.x` since (XML config moved to PHP, help files restructured), I rebased it and applied the remaining review feedback from @Neirda24 (strict `in_array`, explicit short name map initialization, choice prompt wording, static map for `onInvalid` constants, variadic parameter support, exception message typo) in #1807.

Feel free to have a look, and let me know if you'd rather continue on this branch instead.

WedgeSama:
Sorry, I was in holiday :sweat_smile: 
No problem @GromNaN  :wink:

PR #1618, opened 2024-12-05
[make:twig-component] read processed twig_component config instead of parsing the yaml file

IndraGunawan:
changes from #1571 breaks if config is in php files even `$namespace` has a default value

this PR read the processed twig_component config no matter how the config was created

### DIFF
--- a/config/makers.xml
+++ b/config/makers.xml
@@ -22,7 +22,6 @@
 
             <service id="maker.maker.make_twig_component" class="Symfony\Bundle\MakerBundle\Maker\MakeTwigComponent">
                 <tag name="maker.command" />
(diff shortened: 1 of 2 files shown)


smnandre:
This will need to be updated if we change anything in the extension, add a pass, etc etc.. 

Seems a bit risky, but in the same time in understand the need 🤷

IndraGunawan:
@smnandre thanks for your feedback

if this change is a bit risky, i would revert the changes and instead make the command not throwing an error if the yaml config is not exist but it makes the feature only be available on yaml config

in current condition it is totally broken for php config

wdyt?

PR #1619, opened 2024-12-07
[make:entity] Issue with detected mixed type set to be nullable

### LINKED ISSUE #1612: [make:entity] Crash when return type from DoctrineHelper is mixed
Hello,

I'm able to reliably crash the `make:entity` command on 1.61 because of the following line:
https://github.com/symfony/maker-bundle/blob/ce60831a9d6c41c032117ddf59f303f56d004c87/src/Doctrine/DoctrineHelper.php#L291

As this helper relies on the `convertToPHPValue` return type (very clever btw) it will crash if the return value is set to `mixed`. This can be replicated with any custom doctrine type that is set in the application, with the following stack trace.

```php
in BuilderHelpers.php line 208:
                                 
  [LogicException]               
  mixed type cannot be nullable  
                                 

Exception trace:
  at /var/www/vendor/nikic/php-parser/lib/PhpParser/BuilderHelpers.php:208
 PhpParser\BuilderHelpers::normalizeType() at /var/www/vendor/nikic/php-parser/lib/PhpParser/Builder/Property.php:127
 PhpParser\Builder\Property->setType() at /var/www/vendor/symfony/maker-bundle/src/Util/ClassSourceManipulator.php:402
 Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator->addProperty() at /var/www/vendor/symfony/maker-bundle/src/Util/ClassSourceManipulator.php:151
 Symfony\Bundle\MakerBundle\Util\ClassSourceManipulator->addEntityField() at /var/www/vendor/symfony/maker-bundle/src/Maker/MakeEntity.php:243
 Symfony\Bundle\MakerBundle\Maker\MakeEntity->generate() at /var/www/vendor/symfony/maker-bundle/src/Command/MakerCommand.php:102
 Symfony\Bundle\MakerBundle\Command\MakerCommand->execute() at /var/www/vendor/symfony/console/Command/Command.php:326
```

It appears that the builder tries to make a field with type `mixed|null` or `?mixed` but `mixed` already implicitly includes null, which causes the error.

After updating the return type for our custom type from `mixed` to a more sensible `string|null` the error has stopped. Maybe in such a case a note should be printed to cli?

sadikoff:
Hi there, pretty strange edge case found in #1612 when `mixed` type is detected from a custom doctrine type and was set to be nullable. Maker tried to set is as `?mixed` which is not allowed.

That should fix this behavior.

closes #1612

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -143,7 +143,7 @@ final class ClassSourceManipulator
         }
 
         $propertyType = $typeHint;
-        if ($propertyType && !$defaultValue) {
+        if ($propertyType && !$defaultValue && $propertyType !== 'mixed') {
             // all property types
             $propertyType = '?'.$propertyType;
         }
@@ -162,13 +162,13 @@ final class ClassSourceManipulator
             // getter methods always have nullable return values
             // because even though these are required in the db, they may not be set yet
             // unless there is a default value
-            null === $defaultValue,
+            null === $defaultValue && $propertyType !== 'mixed',
             $commentLines
         );
 
         // don't generate setters for id fields
         if (!($mapping->id ?? false)) {
-            $this->addSetter($mapping->propertyName, $typeHint, $nullable);
+            $this->addSetter($mapping->propertyName, $typeHint, $nullable && $propertyType !== 'mixed');
         }
     }
 


nicolas-grekas:
Can you please check for adding a test case?

PR #1621, opened 2024-12-07
[make-entity]Keep the 'is' prefixes for booleans properties setters

octoseth:
When generating a new entity, if we name a boolean property with the prefix "is", this prefix is removed to construct the setter.

example:
 ```
bool $isInternational;

public function isInternational(): ?bool
{
    return $this->isInternational;
}

public function setInternational(bool $isInternational): static
{
    $this->isInternational = $isInternational;

    return $this;
}
```
This is breaking when we want to access the property via a form :

> [NoSuchPropertyException]
> HTTP 500 Internal Server Error
> The method "isInternational" in class "App\Entity\Conference" requires 0 arguments, but should accept only 1.. Make the property public, add a setter, or set the "mapped" field option in the form type to be false.

This PR revert this change introduce in [PR 1493](https://github.com/symfony/maker-bundle/pull/1493) to keep the 'is' prefix on the setter.

example :

```
public function setIsInternational(bool $isInternational): static 
{
    $this->isInternational = $isInternational;

    return $this;
}

```

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -485,10 +485,6 @@ final class ClassSourceManipulator
 
     private function getSetterName(string $propertyName, $type): string
     {
-        if ('bool' === $type && 0 === strncasecmp($propertyName, 'is', 2)) {
-            return 'set'.Str::asCamelCase(substr($propertyName, 2));
-        }
-
         return 'set'.Str::asCamelCase($propertyName);
     }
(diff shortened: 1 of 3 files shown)


nicolas-grekas (approved):
My bad for https://github.com/symfony/maker-bundle/pull/1493#pullrequestreview-1962866946

kbond:
Thank you @octoseth.

PR #1633, opened 2024-12-25
[minor] Fix "twig" case (-> "Twig") in make:twig-component & various places

smnandre:
Same thing for 
* `HTML` in component template
* `Live component` in maker help


kbond (approved):

kbond:
Thanks Simon.

PR #1634, opened 2024-12-26
[Docs] Remove Symfony version in introduction

smnandre:
At first, wanted to update the doc to write "Symfony 7 directory structure" instead of "6.4".. but i  think the best choice here (leading to the minimal amount of frustrations or doubts for the reader) would probably be to remove the version precision.

wdyt ?


MrYamous:
Looks good to me, I think "Symfony 7 directory structure" does not give much value as there is no new specific structure for this version, even for 6.x

MrYamous (approved):

kbond (approved):

kbond:
Thanks Simon.

Issue #1626, opened 2024-12-16
Allow make:entity specify the table name


omgslinux:
It happens that though the default is to name the table to the lowercase entity name (i.e: Task-> task), it's not unusual that the table name could be "tasks" (i.e. in plural).

So, I can think of two ways for this:
1) Just add an extra parameter where one can specifiy the table name

2) Explicitily prompt for the table name, and in case it's the same as the entity name, you can skip using the ORM\Table attribute in the entity file. You can default to the same name or try to guess the plural, and offer it in the prompt.

There could be a third option, changing the current idea, and it's to directly insert the plural in the entity file, and maybe encourage to review if necessary.

Thanks in advance.

PR #1635, opened 2024-12-30
Allow `make:entity` specify the table name

MrYamous:
Give a try on #1626

### DIFF
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -40,7 +40,7 @@ final class EntityClassGenerator
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade, bool $generateRepositoryClass, bool $broadcast, EntityIdTypeEnum $useUuidIdentifier, ?string $tableName): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, ?string $tableName): string
     {
         $repoClassDetails = $this->generator->createClassNameDetails(
             $entityClassDetails->getRelativeName(),
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -40,7 +40,7 @@ final class EntityClassGenerator
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, ?string $tableName): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, string $tableName = null): string
     {
         $repoClassDetails = $this->generator->createClassNameDetails(
             $entityClassDetails->getRelativeName(),
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -40,7 +40,7 @@ final class EntityClassGenerator
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade, bool $generateRepositoryClass, bool $broadcast, EntityIdTypeEnum $useUuidIdentifier, ?string $tableName): string
     {
         $repoClassDetails = $this->generator->createClassNameDetails(
             $entityClassDetails->getRelativeName(),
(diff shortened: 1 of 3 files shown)


MrYamous:
Struggling with tools in local but i'm going to have a look and those errors

src/Doctrine/EntityClassGenerator.php
```diff
@@ -40,15 +40,17 @@ public function __construct(
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, ?string $tableName): string
```
smnandre, line 43:
```suggestion
    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, ?string $tableName = null): string
```

src/Maker/MakeEntity.php
```diff
@@ -97,6 +97,7 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
             ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Add the ability to broadcast entity updates using Symfony UX Turbo?')
             ->addOption('regenerate', null, InputOption::VALUE_NONE, 'Instead of adding new fields, simply generate the methods (e.g. getter/setter) for existing fields')
             ->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite any existing getter/setter methods')
+            ->addOption('table-name', null, InputOption::VALUE_NONE, 'Overwrite default table name')
```
Menelion, line 100:
```suggestion
            ->addOption('table-name', 't', InputOption::VALUE_NONE, 'Overwrite default table name')
```

src/Doctrine/EntityClassGenerator.php
```diff
@@ -40,15 +40,17 @@ public function __construct(
...
     {
         $repoClassDetails = $this->generator->createClassNameDetails(
             $entityClassDetails->getRelativeName(),
             'Repository\\',
             'Repository'
         );
 
-        $tableName = $this->doctrineHelper->getPotentialTableName($entityClassDetails->getFullName());
+        if (null === $tableName) {
```
Menelion, line 51: Wouldn't you consider handling the most common case of pluralizing the class name somehow? Maybe via another option or something else? You can use [this](https://symfony.com/doc/current/string.html#inflector) Symfony inflector capability.

Menelion (approved):

src/Doctrine/EntityClassGenerator.php
MrYamous: This is a good point. But I think it's outside the scope of this PR, my purpose is to be able to explicitly choose the table name, not to change the default behavior. Both can also coexist, I think.

MrYamous: I keep the idea to open a new PR in the next few days, or maybe you want to do it

src/Doctrine/EntityClassGenerator.php
```diff
@@ -40,15 +40,17 @@ public function __construct(
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, string $tableName = null): string
```
smnandre, line 43:
```suggestion
    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, ?string $tableName = null): string
```

Missed the `?` here :)

src/Maker/MakeEntity.php
```diff
@@ -97,6 +97,7 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
             ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Add the ability to broadcast entity updates using Symfony UX Turbo?')
             ->addOption('regenerate', null, InputOption::VALUE_NONE, 'Instead of adding new fields, simply generate the methods (e.g. getter/setter) for existing fields')
             ->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite any existing getter/setter methods')
+            ->addOption('table-name', 't', InputOption::VALUE_NONE, 'Overwrite default table name')
```
smnandre, line 100: Overwrite seems more something to change existing things no ? 

I'd say this will be also (mainly?) used during first Entitiy creation ? 

Maybe something more like "Set custom table name" ? 

Feel free to ignore if you disagree, just a comment passing by ;)

MrYamous: To be honest I hadn't thought that far ahead. But I agree with you for this change, sounds better

src/Maker/MakeEntity.php
```diff
@@ -97,6 +97,7 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
             ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Add the ability to broadcast entity updates using Symfony UX Turbo?')
             ->addOption('regenerate', null, InputOption::VALUE_NONE, 'Instead of adding new fields, simply generate the methods (e.g. getter/setter) for existing fields')
             ->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite any existing getter/setter methods')
+            ->addOption('table-name', 't', InputOption::VALUE_NONE, 'Set custom table name')
```
smnandre, line 100:
```suggestion
            ->addOption('table-name', 't', InputOption::VALUE_REQUIRED, 'Set custom table name')
```

This is the reason of the test failures I think.

But i'm wondering here... should we add a test to avoid using an existing table name ?

omgslinux:
Sorry, I didn't see this. I created a PR with a very similar approach, but currently is failing the test (CI/CD is fine). I don't know what the next steps are, beyond just waiting...

Thanks in advance.

PR #1637, opened 2025-01-01
[make:webhook] generate final classes

MrYamous:
Inspired from #1589


GromNaN (approved):
Thank you! I rebased the branch on 1.x and pushed a few changes:

- Fixed the new test: the data provider is static, so `$this->createMakerTest()` was replaced with `self::buildMakerTest()` and the closure and assertions were made static.
- Removed a stray semicolon in the expected `extends AbstractRequestParser` string.
- Fixed a bug in `MakeWebhook::interact()`: the webhook name entered interactively was only stored in `$this->name`, while `generate()` reads `$input->getArgument('name')` to build the class names. The argument stayed empty, so the generated files were `RequestParser.php` and `WebhookConsumer.php` instead of `RemoteServiceRequestParser.php` and `RemoteServiceWebhookConsumer.php`. The argument is now set at the end of `interact()`, which keeps the non-interactive mode working. All tests in `MakeWebhookTest` pass again.

GromNaN (approved):

GromNaN:
Thanks for your work on this new feature!

PR #1638, opened 2025-01-01
Allow `ClassData` to define an implemented class

MrYamous:
A new `ClassData` has been introduced in #1539 and allows to define a class to extend
While working on #1637 I also wanted to be able to choose a class to implement to generate `WebhookConsumer` class (and maybe others will use this, I didn't check)

Implementation is fully inspired from extends as both lead to a similar result

### DIFF
--- a/src/Util/ClassSource/Model/ClassData.php
+++ b/src/Util/ClassSource/Model/ClassData.php
@@ -30,13 +30,14 @@ final class ClassData
         private bool $isFinal = true,
         private string $rootNamespace = 'App',
         private ?string $classSuffix = null,
+        public readonly ?string $implements = null,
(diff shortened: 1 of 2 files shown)


GromNaN:
Closing as out of scope for now.

The `implements` parameter added to `ClassData` currently has no concrete consumer: `#1637` (make:webhook), which motivated this change, was merged using `extendsClass` instead, and no other maker in the codebase uses `implements` on `ClassData::create()`. Per the project's scope, we avoid adding broadly extensible or pluggable infrastructure ahead of an actual, identified need.

Happy to reconsider if a maker in this repo needs to generate a class implementing an interface.

PR #1640, opened 2025-01-02
[make:twig-component] do not throw an error if config file does not exist

IndraGunawan:
alternative of #1618

changes from https://github.com/symfony/maker-bundle/pull/1571 break if config is in php files even $namespace has a default value


IndraGunawan:
@kbond any chance to review this?

kbond (approved):
This seems like a good call. But yeah, there definitely should be a more robust solution in the future.

PR #1642, opened 2025-01-05
Custom root namespace make:twig-component

### LINKED ISSUE #1641: `make:twig-component` doesn't work well when using a custom root namespace
Hi!

Our app has a custom root namespace. Let's say `Foobar\` instead of `App\`. We have this configured in `maker.yaml`:

```php
maker:
  root_namespace: 'Foobar'
```

Now when we run `make:twig-component` we get:

```
bin/console make:twig-component

 The name of your twig component (ie Notification):
 > Button

 Make this a live component? (yes/no) [no]:
 > no

 created: src/ar/Twig/Components/Button.php
 created: templates/components/Button.html.twig

           
  Success! 
           


 To render the component, use <twig:Button />.
```

The first 4 characters of `Foobar` are removed and a file is created in `src/ar/Twig/...` instead of `src/Twig`.

marcriemer:
Fixed #1641

### DIFF
--- a/src/Maker/MakeTwigComponent.php
+++ b/src/Maker/MakeTwigComponent.php
@@ -71,7 +71,7 @@ final class MakeTwigComponent extends AbstractMaker
 
         $factory = $generator->createClassNameDetails(
             $name,
-            $this->namespace,
+            str_replace($generator->getRootNamespace().'\\', '', $this->namespace),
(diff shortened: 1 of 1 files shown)


kbond:
Thanks for fixing this bug Marc.

PR #1646, opened 2025-01-14
Missmatch between PHPDoc and the reality on User::getRoles()

Curryed:
See description in #1652

### DIFF
--- a/src/Security/UserClassBuilder.php
+++ b/src/Security/UserClassBuilder.php
@@ -132,7 +132,7 @@ final class UserClassBuilder
             'getRoles',
             'array',
             false,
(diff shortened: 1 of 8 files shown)


kbond:
Thanks for digging into this @Curryed.

What about removing the `@return` docblock and just letting the `UserInterface` dictate? I typically never override docblocks from interfaces/abstract classes (unless required) but maybe I'm in the minority?

Curryed:
Sure! If that's the way you usually go for 😁

I updated the it removing the `@return` :)

kbond (approved):
👍🏻

kbond:
Thank you @Curryed.

PR #1648, opened 2025-01-14
Move PHP CS fixer to require dev

HypeMC:
I'm pretty sure it was added to the `require` section by accident.


kbond:
I've reverted #1575 and released v1.62.1. It was causing a lot of problems. I think we need to rethink the php-cs-fixer integration.

Issue #1644, opened 2025-01-14
php-cs-fixer/shim is userland


Chris53897:
is it intended that php-cs-fixer/shim is in `require` and not in `require-dev`? 
https://github.com/symfony/maker-bundle/commit/4735ff12ce4350b16c15735de1ed6e1126a3471d#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R19

Now `friendsofphp/php-cs-fixer` will be uninstalled and `php-cs-fixer/shim` will be installed in userland code.

@theofidry @jrushlow

theofidry:
>Now friendsofphp/php-cs-fixer will be uninstalled and php-cs-fixer/shim will be installed in userland code.

You'll need to expend on why this is the case, it is not the expected behaviour for sure.

From this (`symfony/maker-bundle`) perspective, the shim _is_ a required dependency: it relies on it to execute it against generated code (so it uses it as part of its source code). However, you likely consume this library as a dev dependency so the shim would remain a dev dependency from your application.

Can you develop as to why?

Chris53897:
I hope this infos will help.

my composer.json

```
"friendsofphp/php-cs-fixer": "^3.68",
"symfony/maker-bundle": "^1.62",
```

`symfony composer update`

`Symfony operations: 4 recipes (e3eb63c1ad767ffc8e3fff791285e9d0)
  - Configuring php-cs-fixer/shim (>=3.0): From github.com/symfony/recipes:main
  - Unconfiguring friendsofphp/php-cs-fixer (>=3.0): From github.com/symfony/recipes:main
  - Unconfiguring composer/xdebug-handler (>=3.0.5): From auto-generated recipe
  - Unconfiguring composer/pcre (>=3.3.2): From auto-generated recipe
`

kbond:
I've reverted #1575 and released v1.62.1. It was causing a lot of problems. I think we need to rethink the php-cs-fixer integration.

PR #1651, opened 2025-01-14
Replace php-cs-fixer/shim with friendsofphp/php-cs-fixer

mbaschnitzi:
In #1575 the dependency `php-cs-fixer/shim` was added to replace a self packaged phar of php-cs-fixer.

This change causes `php-cs-fixer/shim` to be installed in all projects using `symfony/maker-bundle`, even if the non-shim package was used before. In symfony flex projects this also causes further unexpected changes since `php-cs-fixer/shim` also executes an additional recipe modifying additional files. (#1644 #1648)

I have changed the dependency to `friendsofphp/php-cs-fixer` instead, so the decision on which package to use can be made by the user instead of forcing the use of the shim package. The execution in `\Symfony\Bundle\MakerBundle\Util\TemplateLinter` remains the same for both packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
     "require": {
         "php": ">=8.1",
         "doctrine/inflector": "^2.0",
+        "friendsofphp/php-cs-fixer": "^v3.64",
         "nikic/php-parser": "^4.18|^5.0",
-        "php-cs-fixer/shim": "^v3.64",
         "symfony/config": "^6.4|^7.0",
         "symfony/console": "^6.4|^7.0",
         "symfony/dependency-injection": "^6.4|^7.0",


Seb33300:
`friendsofphp/php-cs-fixer` also triggers a symfony flex recipe

mbaschnitzi:
> `friendsofphp/php-cs-fixer` also triggers a symfony flex recipe

Then maybe the change should be reverted and/or php-cs-fixer should be added as an optional dependency?

If php-cs-fixer is installed globally neither `php-cs-fixer/shim` or `friendsofphp/php-cs-fixer` are really needed.

kbond:
I've reverted #1575 and released v1.62.1. It was causing a lot of problems. I think we need to rethink the php-cs-fixer integration.

Issue #1653, opened 2025-01-14
projects using maker-bundle are enforced to switch from friendsofphp/php-cs-fixer to php-cs-fixer/shim


madmajestro:
@jrushlow @theofidry
The composer.json of `php-cs-fixer/shim` contains a replace section which replaces `friendsofphp/php-cs-fixer` by `php-cs-fixer/shim` in all projects which are using the maker-bundle.

I suspect that this invasive effect was not intended. I therefore propose to revert #1575 quickly and make a new release. Then you can work on a better solution later.

kbond:
I've reverted #1575 and released v1.62.1. It was causing a lot of problems. I think we need to rethink the php-cs-fixer integration.

PR #1654, opened 2025-01-15
Revert #1575 (Use a PHP-CS-Fixer shim rather than an external PHAR)

kbond:
This reverts #1575. This PR has caused a lot of trouble as it made php-cs-fixer a required dep which installs a recipe.

See #1644, #1653, #1651, #1648.

We can't have php-cs-fixer (or the shim) as a required dependency. I think we need to have this bundle only use php-cs-fixer if available/configured (or drop it entirely). I'm not sure why it is used at all - feels like a lot of added complexity...


alexander-schranz:
The why is mostly because php parser putput has no styling.
But I would like go the way rector does it by just document it that you should run your favorite code style fixer (phpcsfixer, ecs, ..) after you generated code. I would do nothing automaticall in my opinion.

nicolas-grekas:
> I would do nothing automaticall in my opinion.

I agree with that: make-bundle should stick to one "Symfony CS" for output and any other CS is user's problem to solve.

PR #1656, opened 2025-01-15
Fix putting the class-opening bracket on its own line

nicolas-grekas:
Without this change, eg make:controller generates the following. A PHP closing tag eats its CR in some cases.

```php
final class TotoController extends AbstractController{
```


smnandre (approved):

kbond:
Thanks Nicolas.

PR #1657, opened 2025-01-15
Fix indentation of options_code

nicolas-grekas:
Similar to #1469 but doesn't rely on any CS fixer.

Note that using PHP_EOL was wrong: this constant is meant to be used ONLY when outputting something on a terminal. NOT when dealing with new lines in files.


kbond:
Thanks Nicolas.

PR #1659, opened 2025-02-05
Possibility to add custom templates when [make] command

ghost:
Possibility to use custom templates when using make command.

Update on Generator.php to check if a template exists in `templates/bundles/MakerBundle/' . $templateName`

Tree inside template's folder needs to follow the bundle convention.

`make:crud` example to customize the show.html.twig:
create a custom copy of show.tpl.php in $PROJECT_DIR/templates/bundles/MakerBundle/crud/templates

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "composer/semver": "^3.0",
         "doctrine/doctrine-bundle": "^2.5.0",
         "doctrine/orm": "^2.15|^3",
+        "doctrine/persistence": "^2.5|^3",
         "symfony/http-client": "^6.4|^7.0",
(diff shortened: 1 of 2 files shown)


eltharin:
Idea is good but allow templates bundle folder can be better, everyone can make his own maker in their own bundles and have it in many projects.
Maybe have a configuration to add makerbundle custom template folder ?

grichards:
Yes please. In my case, I don't want to create a custom maker... I really just want to modify the template make:entity uses to set some sensible default for #[ApiResource].

The simplest way would be if this bundle would follow the convention that allows me to add template overrides in my templates directory.

Also... why are these classes all "final"... I mean, why shouldn't I be able to extend MakeEntity?

PR #1660, opened 2025-02-05
[make:entity] Use stricter DateTime type instead of DateTimeInterface for mutable datetime fields

rosier:
The current DateTimeInterface type hint suggest that you can put a DateTimeImmutable in a mutable datetime field, but that is no longer the case.

It results in the following error:
```
Could not convert PHP value of type DateTimeImmutable to type Doctrine\DBAL\Types\DateTimeType. 
Expected one of the following types: null, DateTime.
```

Follows:
- https://github.com/doctrine/dbal/pull/6017
- https://github.com/doctrine/dbal/pull/6161


kbond (approved):
Makes sense!

kbond:
Thank you @rosier.

PR #1661, opened 2025-02-07
Make:entity: Add question about unsigned integer

Menelion:
It is my first contribution, so please bear with me.  
So often, when adding entities, I wanted integers to be unsigned, so I decided to add this question if the field type is integer. Most probably did I miss something, so I'm all ears for your comments.


Menelion:
Dear core contributors, is it OK to merge now? All checks passed. Thanks!

PR #1662, opened 2025-02-09
Prevent the installation of doctrine/persistence ^4

rosier:
The `DoctrineHelper` uses `Doctrine\Persistence\Mapping\StaticReflectionService` which was removed in doctrine/persistence ^4

Fixes Error: Class "Doctrine\Persistence\Mapping\StaticReflectionService" not found


kbond:
This has been fixed in #1677.

PR #1674, opened 2025-03-18
Fix default param key in validator

PierreBaud:
Default param key not matching with default validator.

Failure it's seem not related to my changes

### DIFF
--- a/templates/validator/Constraint.tpl.php
+++ b/templates/validator/Constraint.tpl.php
@@ -8,7 +8,7 @@ namespace <?= $class_data->getNamespace(); ?>;
 <?= $class_data->getClassDeclaration(); ?>
 
 {
-    public string $message = 'The string "{{ string }}" contains an illegal character: it can only contain letters or numbers.';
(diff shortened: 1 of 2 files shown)


templates/validator/Constraint.tpl.php
```diff
@@ -7,7 +7,7 @@
 #[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
 <?= $class_data->getClassDeclaration(); ?>
 {
-    public string $message = 'The string "{{ string }}" contains an illegal character: it can only contain letters or numbers.';
+    public string $message = 'The string "{{ value }}" contains an illegal character: it can only contain letters or numbers.';
```
lyrixx, line 10: See https://github.com/symfony/maker-bundle/blob/main/templates/validator/Validator.tpl.php#L19

lyrixx (approved):
Good Catch. Thanks 👍🏼

kbond (approved):

kbond:
Thank you @PierreBaud.

PR #1482, opened 2024-03-16
WIP - failing generator test with global ns collision

jrushlow:
refs #1479 #1480

PR #1676, opened 2025-03-24
Allow creation of classes with the same name as PHP native classes

### LINKED ISSUE #1675: Cannot create an entity named 'Directory'
**API Platform version(s) affected**: 4.1.1

**Description**  
I started a project using api-platform one year ago. Inside this project i created an entity called Directory.
I had no issue using this name at the time. Nowadays i cannot use the command :
```
php bin/console make:entity Directory
```
The result is :

In FileManager.php line 111:
                                                                                       
  Notice: file_get_contents(): Read of 8928 bytes failed with errno=21 Is a directory  

**How to reproduce**  
Clone latest api-platform, run :
```
docker compose build --no-cache
```
and then 
```
docker compose up --wait
````
Go inside app-php container and run :
```
php bin/console m:e Directory
```
To me, it seems like a regression, but I cannot be certain. I hope there is a workaround. Thanks a lot for your time.

**Possible Solution**  
None.

**Additional Context**  

Image

Fan2Shrek:
Fix #1675 

When trying to create a class with a native PHP name (e.g., Directory), `Validator::classDoesNotExist` incorrectly throws an error stating that the class already exists. To fix this, we need to prepend the namespace before performing the check.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -188,8 +188,8 @@ class Generator
             $className = Str::asClassName($name, $suffix);
 
             try {
-                Validator::classDoesNotExist($className);
                 $className = rtrim($fullNamespacePrefix, '\\').'\\'.$className;
+                Validator::classDoesNotExist($className);
             } catch (RuntimeCommandException) {
             }
         }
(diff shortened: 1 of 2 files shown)


Fan2Shrek:
The test failures are related, but I'm not entirely sure how `Generator::createClassNameDetails` is supposed to behave. In my view, it should assert that the FQCN does not exist, not just the class itself. However, related tests are failing.

GPHemsley:
See also:
* #1480
* #1482
* #1681

GromNaN:

PR #1677, opened 2025-03-26
[make:entity] `doctine/persistence` v4 support, drop xml entity config support

kbond:
- drops support for xml configured entities
- `doctrine/persistence` dropped StaticReflectionService - added backport


weaverryan:
We should drop ``regenerate`: that's an antiquated flag to go from XML -> entity class. We should be 100% all-in on attribute mapping. I don't consider this a BC break as the `make:` commands can't realistically be used in automated scripts.

That being said a 2.0 major is overdue

PR #1681, opened 2025-04-07
Prevent vendor class name overlap

### LINKED ISSUE #1566: [make:entity] Maker modifies vendor files if FQCN covers up local alias
Using 1.59.1 the following interaction creates a funny edit in the vendor directory.

I earlier created entities under our regular doctrine mapping, with MangoPay as the preceding namespace.
I then named the entity Wallet, so a class under `App/Common/Entity/MangoPay/Wallet` exists.

The doctrine mapping for this section is as following:

```yaml
        mappings:
          App:
            is_bundle: false
            type: attribute
            dir: '%kernel.project_dir%/src/Common/Entity'
            prefix: 'App\Common\Entity'
            alias: App
```

The following interaction then occurs.

```
# bin/console make:entity

 Class name of the entity to create or update (e.g. VictoriousElephant):
 > MangoPay\Wallet

 > stuff

 > 

 Field length [255]:
 > 

 Can this field be null in the database (nullable) (yes/no) [no]:
 > 

 updated: vendor/mangopay/php-sdk-v2/MangoPay/Wallet.php

 Add another property? Enter the property name (or press <return> to stop adding fields):
 > 


           
  Success! 
           

 Next: When you're ready, create a migration with php bin/console make:migration
```

Obviously, I am expecting it to modify my entity, which is within the doctrine mapping. 
Using the full namespace the edit is performed as expected.

In this scenario it should be at least partially checked, and then possibly asked which class you're exactly referring to. The same might be the case when a FQCN name might cover up a local one through relations, as I believe I've seen something like this already.

pkly:
Fixes #1566 

I've pasted this over the currently installed version in the project in which I had these issues, seems to work fine. Would be good if it could be re-checked and generally looked over, since I'm not 100% sure about it, but this is the place which causes the issue mentioned.

It should prevent situations where a known, working class name is skipped over and a vendor class is used instead, as there is no instance in which modifying vendor classes is required.

Actually this also seems to not work when it comes to classes with relations, so the target entity class (one we're adding relation to) may still be pulled from the vendor.


smnandre:
cf
- https://github.com/symfony/maker-bundle/issues/1479
- https://github.com/symfony/maker-bundle/issues/1422

pkly:
Any chance one of the maintainers could look at this? It's horribly annoying.

pkly:
@javiereguiluz sorry to bother you, but is there a chance for this to get merged? I've been using this patched in for... well months really without problems and it solves few issues

GPHemsley:
See also:
* #1480
* #1482
* #1676

GromNaN (dismissed):
Thanks for tackling this! The logic is sound: keeping the app-namespaced candidate as a fallback and preferring it when it exists is exactly what avoids the silent vendor-file write reported in #1566. I reproduced the original issue and confirmed the fix behaves as expected.

pkly:
Yeah, feel free to open up an alternate as this PR is extremely stale. I'd appreciate some sort of mention though.

Once it's ready, I'll close this.

GromNaN:

PR #1683, opened 2025-04-25
Tighten up CI matrix

kbond:


bocharsky-bw (commented):
Looks good to me 👍 

How about bumping PHP version as well? I see the latest 8.4 isn't used at all

weaverryan (approved):

weaverryan:
> How about bumping PHP version as well? I see the latest 8.4 isn't used at all

Agreed, but maybe separate PR? And maybe we do one more release first?

PR #1684, opened 2025-04-25
[make:schedule] deprecate in favor of recipe

kbond:
The recipe for `symfony/scheduler` now creates a default schedule. I think it's super rare that you'd create multiple schedules.


bocharsky-bw (commented):
Makes sense to me 👍

weaverryan (approved):

PR #1685, opened 2025-04-25
[make:form] suffix with "Form" instead of "Type"

kbond:
This may be controversial but I've always found the term "Type" confusing for your forms. For the purposes of what this maker creates, these are "Forms".

If there is agreement on this change, I'll fix up the tests.


weaverryan:
💯

bocharsky-bw:
The same for me 👍

tests/fixtures/make-registration-form/tests/it_generates_registration_with_entity_and_authenticator.php
```diff
@@ -26,6 +26,7 @@ public function testRegistrationSuccessful()
         $form['registration_form[agreeTerms]'] = true;
         $client->submit($form);
 
+        dump($client->getResponse());
```
nebkam, line 29: Was this intentional? :thinking:

kbond: It was not, good catch!

lyrixx:
See https://github.com/symfony/maker-bundle/issues/1711

PR #1687, opened 2025-04-25
[make:listener] use FQCN for kernel events, fix adding unnecessary `#[AsEventListener]` arg

kbond:
- adding the `event` argument to `#[AsEventListener]` isn't required when it's a FQCN
- use FQCN for kernel events - I believe this is the best practice

### DIFF
--- a/tests/fixtures/make-listener/tests/EventListener/BarListener.php
+++ b/tests/fixtures/make-listener/tests/EventListener/BarListener.php
@@ -7,7 +7,7 @@ use Symfony\Component\HttpKernel\Event\RequestEvent;
 
 final class BarListener
 {
-    #[AsEventListener(event: RequestEvent::class)]
+    #[AsEventListener]
     public function onRequestEvent(RequestEvent $event): void
     {
         // ...
--- a/src/Maker/MakeListener.php
+++ b/src/Maker/MakeListener.php
@@ -153,12 +153,11 @@ final class MakeListener extends AbstractMaker
         $eventFullClassName = $this->eventRegistry->getEventClassName($event);
         $eventClassName = $eventFullClassName ? Str::getShortClassName($eventFullClassName) : null;
 
-        if (null !== ($eventConstant = $this->getEventConstant($event))) {
-            $useStatements->addUseStatement(KernelEvents::class);
-            $eventName = $eventConstant;
-        } else {
-            $eventName = class_exists($event) ? \sprintf('%s::class', $eventClassName) : \sprintf('\'%s\'', $event);
+        if ($this->getEventConstant($event)) {
+            $event = $eventFullClassName;
         }
+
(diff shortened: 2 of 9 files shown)


tests/fixtures/make-listener/tests/EventListener/BarListener.php
```diff
@@ -7,7 +7,7 @@
 
 final class BarListener
 {
-    #[AsEventListener(event: RequestEvent::class)]
+    #[AsEventListener]
     public function onRequestEvent(RequestEvent $event): void
```
weaverryan, line 11: Feature creep: can these be invokable?

weaverryan (approved):

tests/fixtures/make-listener/tests/EventListener/BarListener.php
kbond: Yep!

My only hesitation on this: I sometimes have multiple related listeners in the same file (think security or request/response). I don't know this is a good enough reason to not have invokable listeners though.

weaverryan: Let's skip

weaverryan: Ideas for better method names:

* `whenKernelRequest()`
* `handleKernelRequest()`

or descriptive when possible
* `beforeRequest()
* `afterException()`
* `afterResponse()`
* `beforeCommand()`
* `afterCommand()`
* `beforeLogin()` (`AUTHENTICATION_SUCCESS`)

I quite like the descriptive version

kbond: Do you think this could be generated dynamically? Or some kind of common event map?

MolloKhan: descriptive is better.`whenKernelRequest` does not tell me at what point of the workflow is executed

kbond: I'm thinking we should go with `__invoke()` - this would match `make:message`

weaverryan: > I'm thinking we should go with __invoke() - this would match make:message

I'm not convinced by `__invoke` here or even in controllers: I don't want a whole new class for a 2nd listener. Message handlers, like commands, really never have a multi-method use-case

>  you think this could be generated dynamically? Or some kind of common event map?

event map

kbond: Here's my argument for `__invoke()` (this applies to `make:controller` also):
1. Saves you from thinking of two names: class & method
2. Since these makers don't allow adding additional methods, you have to manually add them if desired - renaming `__invoke()` to something more descriptive at the same time seems easy.
That being said, I don't feel super strongly. It's easy enough to rename the generated method to invoke and that's what I usually do.

PR #1690, opened 2025-04-25
[make:message] use `AsMessage` for routing if available

kbond:
The less files a maker touches, the better, IMO.


src/Maker/MakeMessage.php
```diff
@@ -97,9 +100,19 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             'Handler'
         );
 
+        $useStatements = new UseStatementGenerator([]);
+
+        if ($chosenTransport && class_exists(AsMessage::class)) {
+            $useStatements->addUseStatement(AsMessage::class);
+        }
```
weaverryan, line 108: Jesse & I got in the habit of adding something like `@legacy to support < symfony/messenger 7.2`

weaverryan (approved):

PR #1692, opened 2025-04-25
Fix docs branch following the renaming of main to 1.x

chalasr:
So the docs on [symfony.com/bundles/SymfonyMakerBundle](https://symfony.com/bundles/SymfonyMakerBundle) are back


kbond:
Thanks Robin.

PR #1703, opened 2025-05-13
[make:user] Hash passwords using crc32c and deprecate eraseCredentials()

### LINKED ISSUE #1700: [make:user] `UserInterface::eraseCredentials` is deprecated
Since https://github.com/symfony/symfony-docs/issues/20616, `UserInterface::eraseCredentials()` is deprecated.

So I think `make:user` shoud not add this method when symfony version is >= 7.3

nicolas-grekas:
Fix #1700
Replace ##1701


kbond:
Thanks Nicolas.

PR #1704, opened 2025-05-18
[make:doctrine:listener] Add maker for Doctrine event/entity listeners

HypeMC:
Similar to `make:listener`, this maker helps create Doctrine event or entity listeners.

### DIFF
--- a/tests/Doctrine/DoctrineEventRegistryTest.php
+++ b/tests/Doctrine/DoctrineEventRegistryTest.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MakerBundle\Tests\Doctrine;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\MakerBundle\Doctrine\DoctrineEventRegistry;
+
+class DoctrineEventRegistryTest extends TestCase
+{
+    private static ?DoctrineEventRegistry $doctrineEventRegistry = null;
+
+    public static function setUpBeforeClass(): void
+    {
+        self::$doctrineEventRegistry = new DoctrineEventRegistry();
+    }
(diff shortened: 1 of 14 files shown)


tests/Doctrine/DoctrineEventRegistryTest.php
```diff
@@ -0,0 +1,69 @@
...
+        self::$doctrineEventRegistry = new DoctrineEventRegistry();
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        self::$doctrineEventRegistry = null;
+    }
+
+    /**
+     * @testWith ["prePersist", true]
```
GromNaN: Note that using a conventional data provider avoids having to rewrite this sample data when migrating to attributes. It also allows the static analyzer to identify to class names.

tests/Maker/MakeDoctrineListenerTest.php
```diff
@@ -0,0 +1,183 @@
...
+namespace Symfony\Bundle\MakerBundle\Tests\Maker;
+
+use Symfony\Bundle\MakerBundle\Maker\MakeDoctrineListener;
+use Symfony\Bundle\MakerBundle\Test\MakerTestCase;
+use Symfony\Bundle\MakerBundle\Test\MakerTestDetails;
+use Symfony\Bundle\MakerBundle\Test\MakerTestRunner;
+
+class MakeDoctrineListenerTest extends MakerTestCase
+{
+    private const EXPECTED_LISTENER_PATH = __DIR__.'/../../tests/fixtures/make-doctrine-listener/tests/EventListener/';
```
GromNaN: This file is already in the `tests` dir. 
```suggestion
    private const EXPECTED_LISTENER_PATH = __DIR__.'/../fixtures/make-doctrine-listener/tests/EventListener/';
```

GromNaN (commented):
Could you add tests using the `--non-interactive` option. The code looks corrected but it's worst testing so that we ensure it works with coding agents.

HypeMC:
@GromNaN Thanks for the feedback! All comments have been addressed in https://github.com/symfony/maker-bundle/compare/c76f3c9c62bff933fe25e0480fa8609d0ad846e7..fcaec08253c34d46b0188b8d47da4fdb50306a97

PR #1709, opened 2025-06-04
Update Controller.tpl.php to use createView() method by default

DocFX:
Hi, just stumble recurringly on this while using `make:crud`.

There are several ways to address that, like using render() or else, but I think the most concensual way of doing so would be to just add `createView()` to all Controller methods that pass forms to templates.

I might be wrong on that, though, but it's one of the first things I do in those files right after generating them.

### DIFF
--- a/templates/crud/controller/Controller.tpl.php
+++ b/templates/crud/controller/Controller.tpl.php
@@ -45,7 +45,7 @@ namespace <?= $class_data->getNamespace() ?>;
 
         return $this->render('<?= $templates_path ?>/new.html.twig', [
             '<?= $entity_twig_var_singular ?>' => $<?= $entity_var_singular ?>,
-            'form' => $form,
+            'form' => $form->createView(),
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


templates/crud/controller/Controller.tpl.php
```diff
@@ -45,7 +45,7 @@ public function new(Request $request, EntityManagerInterface $entityManager): Re
 
         return $this->render('<?= $templates_path ?>/new.html.twig', [
             '<?= $entity_twig_var_singular ?>' => $<?= $entity_var_singular ?>,
-            'form' => $form,
+            'form' => $form->createView(),
```
ker0x, line 48: `render()` will automatically call `createView()` on all parameters that are an instance of `FormInterface`!

DocFX: I know that it does... If not yet done. The question is: shouldn't this be done?

Pros: more explicit.
Cons: doesn't handle the 422 code, for instance (like Turbo doesn't like this).

I'm split on that. Hence the questioning! :)

DocFX:
Annnd... The test runner crashed :(

stof:
To me, this PR should be rejected. Passing the `FormInterface` to `render()` is the preferred usage, to let FrameworkBundle handle automatically the 422 status code for invalid form submissions (otherwise, you have to implement that yourselves).

*Removing* the `->createView()` call in the MakerBundle template was an intended change when we introduced this status code handling logic in `AbstractController::render()`

GromNaN:
Closing for the reason given by stof.

Issue #1135, opened 2022-06-04
configurable template paths


tacman:
If the absolute path to the template is not passed to the Generator class, it defaults to looking in makerBundle's skeleton directory.

```php
        if (!file_exists($templatePath)) {
            $templatePath = __DIR__.'/Resources/skeleton/'.$templateName;
```

What I'd like to do is set a list of paths to use when looking for a  template in the config, e.g.

```yaml
maker:
    root_namespace: Foo
    template_dirs: 
          - my/templates
          - vendor/symfony/maker-bundle/src/Resources/skeleton/twig
```

This would also allow my own makers to not require hardcoding the path or overriding the generator class.  That is, if I have a make:something (MakeSomething.php), I'd like to be able to call

```php
        $generator->generateClass(
            $extensionClassNameDetails->getFullName(),
            'Something.tpl.php',
            ['use_statements' => $useStatements]
        );
```

without having to prefix the template path myself with __DIR__ . '../../skeleton/'.

Happy to do the PR if this approach seems worthwhile.  I love the maker bundle, and use my own makers all the time.

PR #1710, opened 2025-06-12
add customizable templates path

eltharin:
allow user to customize their templates by adding a config parameter to list templates folders.
generator will search in all folders listed in the configurations before to select its own template.
With that, user can modify one or more templates to get what he want without copy all maker classes :).

If no config is made => nothing move, all stay as before,
config can be set in a bundle (and the templates can be there too) to maximize reusability.

by example config can be : 
```
maker:
    templates_folders:
        - "makertemplates"
        - "vendor/eltharin/myBundle/makerTemplates
```

### DIFF
--- a/config/services.xml
+++ b/config/services.xml
@@ -57,6 +57,7 @@
                 <argument /> <!-- root namespace -->
                 <argument>null</argument> <!-- PhpCompatUtil -->
                 <argument type="service" id="maker.template_component_generator" />
+                <argument /> <!-- templates_folders -->
             </service>
 
             <service id="maker.entity_class_generator" class="Symfony\Bundle\MakerBundle\Doctrine\EntityClassGenerator">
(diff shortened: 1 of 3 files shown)


pkly:
Any chance this could be merged sometime? I'd also like to see this feature, and it'd resolve #1135

GromNaN (commented):
👎🏻 Making the template directory configurable means we have to maintain backward compatibility for filenames and the parameters passed to templates. This makes it much harder to make changes without breaking your usage. I am therefore opposed to this change.

pkly:
Wouldn't it be better to simply... place the templates in the template directory, and have symfony resolve it like it does with twig? That way you wouldn't have to worry about backwards compatibility and anyone who'd like to override could simply do so using the built-in bundle override feature for templates

eltharin:
> 👎🏻 Making the template directory configurable means we have to maintain backward compatibility for filenames and the parameters passed to templates. This makes it much harder to make changes without breaking your usage. I am therefore opposed to this change.

Yes, this requires some procedures and avoids making random changes. Now, do the filenames change with each version? Isn't this a "Symfony" project, which is supposed to be a well-structured project? Your answer is surprising in this context.

eltharin:
The ability to extend templates is requested by a large majority of users; perhaps they should be listened to.

PR #1715, opened 2025-06-23
CI updates/fixes

kbond:


GromNaN:
Looking at the successful jobs following the latest fixes, this PR no longer seems necessary.

PR #1725, opened 2025-08-05
Fixed undefined method issue

sape-petrov:
This is a small fix `PHP Fatal error:  Uncaught Error: Call to undefined method SymfonyDocsBuilder\BuildConfig::diableBuildCache()` issue for `_docs_build`


javiereguiluz:
Roman, thanks a lot for fixing this and sorry it took us so long to merge this.

PR #1727, opened 2025-08-18
Remove obsolete check for 'version' in docker compose file

kasperh90:
Running ``symfony console make:docker:database`` fails with the error "[ERROR] compose.yaml file version is not set." if the version: field is missing.

Since Docker Compose v2, the version field is optional and considered obsolete. Users with a modern compose.yaml may see this error unnecessarily.

See: https://docs.docker.com/reference/compose-file/version-and-name/

This PR removes the version check and updates the tests accordingly.

### DIFF
--- a/src/Util/ComposeFileManipulator.php
+++ b/src/Util/ComposeFileManipulator.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\Bundle\MakerBundle\Util;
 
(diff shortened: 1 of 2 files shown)


javiereguiluz (approved):

GromNaN:
I rebased and fixed the CS.

GromNaN:
Thank you @kasperh90.

PR #1728, opened 2025-08-21
PHP 8.4 Deprecation Fix

natewiebe13:
Resolves the following deprecation: https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


javiereguiluz (approved):

javiereguiluz:
Thanks for fixing this bug Nate.

PR #1724, opened 2025-08-05
Add support for Symfony 8

Kocal:
The same tests on Symfony <8.0 & 8.0.x-dev jobs are failing, that's fine but we should take some time to fix them

PR #1729, opened 2025-08-25
Convert config from XML to PHP

nicolas-grekas:
PHP is nicer than XML and https://github.com/symfony/symfony/pull/60568 might make this required at some point.


javiereguiluz:
Thanks Nicolas. This change is indeed mandatory to make this bundle work with Symfony 8.

However, given that in #1724 Hugo is working to make this bundle work in Symfony 8, let's close this PR in favor of the other one, which includes this and other changes. Thanks!

PR #1730, opened 2025-08-26
[make:entity] Allow specify the table name

omgslinux:
Related to #1626, here's a PR to fit that need. I've checked it's backwards compatible, so if you want to keep the current behaviour, you just have to confirm the default guessing. If you want to have a different table name from the default guessing, just enter when prompted. And I've tested that if you enter a keyword, it's properly quoted.

### DIFF
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -40,7 +40,7 @@ final class EntityClassGenerator
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, array $params = []): string
     {
         $repoClassDetails = $this->generator->createClassNameDetails(
             $entityClassDetails->getRelativeName(),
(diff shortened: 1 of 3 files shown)


src/Maker/MakeEntity.php
```diff
@@ -186,14 +187,29 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+
             $broadcast = $input->getOption('broadcast');
             $entityPath = $this->entityClassGenerator->generateEntityClass(
                 entityClassDetails: $entityClassDetails,
                 apiResource: $input->getOption('api-resource'),
                 broadcast: $broadcast,
                 useUuidIdentifier: $this->getIdType(),
+                params: [
+                    'tableName' => $tableName,
+                ]
```
GromNaN, line 212: I understand the intention, which is to stop adding parameters to the function and use an array instead. But in that case, we'd have to revise the entire method signature; this is just a stopgap solution.

Since the class is internal, I'm convinced that the best thing to do is to add a new parameter to the method. That way, it's clearly documented. Otherwise, we'd have to use an array-shape in the phpdoc, which is less strict.


```suggestion
                $tableName,
```

src/Doctrine/EntityClassGenerator.php
```diff
@@ -40,15 +40,15 @@ public function __construct(
     ) {
     }
 
-    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT): string
+    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, array $params = []): string
```
GromNaN, line 43:
```suggestion
    public function generateEntityClass(ClassNameDetails $entityClassDetails, bool $apiResource, bool $withPasswordUpgrade = false, bool $generateRepositoryClass = true, bool $broadcast = false, EntityIdTypeEnum $useUuidIdentifier = EntityIdTypeEnum::INT, string $tableName = ''): string
```

GromNaN (changes requested):
Aside from my preference regarding the signature of the `generateEntityClass` method, I think it's useful. It adds a question to the `make:entity` command, but the default answer makes it easy to skip.

PR #1733, opened 2025-09-24
Fix flex `SYMFONY_REQUIRE` configuration in CI

GromNaN:
Flex uses the `SYMFONY_REQUIRE` variable for version constraint https://github.com/symfony/flex/blob/f356aa35f3cf3d2f46c31d344c1098eb2d260426/src/Flex.php#L129

### DIFF
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -35,17 +35,15 @@ jobs:
                 ports:
                     - 1337:1337
 
-        env:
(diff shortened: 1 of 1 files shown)


GromNaN:
I made the wrong assumption about the purpose of `SYMFONY_VERSION`; it's used for testing.
https://github.com/symfony/maker-bundle/blob/c86da84640b0586e92aee2b276ee3638ef2f425a/src/Test/MakerTestEnvironment.php#L405-L408

PR #1740, opened 2025-11-03
Removed assertion on status code after redirect

rensieeee:
The unit test for the created LoginController does an assertion on the http status code after redirecting. Since it's not required to have a root route, this assertion could fail if there is no root configured.

The functionality works as intended if the redirect works, there's no real benefit to further check the status of the page its redirected to. This functionality should be tested in the unit test of the root controller.


rensieeee:
This solution was discussed in [https://github.com/symfony/symfony/pull/62245](https://github.com/symfony/symfony/pull/62245)

javiereguiluz (approved):

javiereguiluz:
Thank you Rens.

PR #1743, opened 2025-11-15
do not implement `__serialize()` for users without passwords

xabbuh:


src/Test/MakerTestEnvironment.php
```diff
@@ -260,7 +260,7 @@ private function buildFlexSkeleton(): void
         }
 
         // fetch a few packages needed for testing
-        MakerTestProcess::create('composer require phpunit browser-kit symfony/css-selector --prefer-dist --no-progress --no-suggest', $this->flexPath)
+        MakerTestProcess::create('composer require phpunit:1.1.* browser-kit symfony/css-selector --prefer-dist --no-progress --no-suggest', $this->flexPath)
```
xabbuh, line 263: `symfony/test-pack` 1.2 brings us PHPUnit 12 which the tests don't run with

src/Test/MakerTestEnvironment.php
```diff
@@ -246,7 +246,7 @@ private function buildFlexSkeleton(): void
         $flexProjectDir = \sprintf('flex_project%s', $targetVersion);
 
         MakerTestProcess::create(
-            \sprintf('composer create-project symfony/skeleton%s %s --prefer-dist --no-progress', $versionString, $flexProjectDir),
+            \sprintf('composer create-project symfony/skeleton%s %s --prefer-dist --no-progress --keep-vcs', $versionString, $flexProjectDir),
```
xabbuh, line 249: not sure when this broke but we need the `.git` directory as the created project serves as the base repository for the following tests

alexandre-daubois (approved):

HypeMC (approved):

Kocal:
👀

tests/Security/fixtures/expected/UserEntityWithoutPassword.php
```diff
@@ -73,7 +73,6 @@ public function setRoles(array $roles): static
     public function __serialize(): array
     {
         $data = (array) $this;
-        $data["\0" . self::class . "\0password"] = hash('crc32c', $this->password);
```
nicolas-grekas, line 76: then we can remove the whole method (same below of course)

xabbuh: indeed, done

nicolas-grekas (approved):

src/Test/MakerTestEnvironment.php
stof: `1.1.*` is weird. Don't you mean `11.*` instead ?

src/Test/MakerTestEnvironment.php
stof: IIRC, it relied on Flex running `git init` to create a *new* git repository (free from the history of the skeleton)

src/Test/MakerTestEnvironment.php
xabbuh: no, `phpunit` is an alias for `symfony/test-pack` (see https://github.com/symfony/recipes/blob/main/symfony/test-pack/1.0/manifest.json#L2 and https://packagist.org/packages/symfony/test-pack#v1.1.0)

src/Test/MakerTestEnvironment.php
xabbuh: ah, that might indeed explain it

nextpageblog: Thank you @xabbuh, this change solved the problem I faced while trying to run the repository tests.

Now, I can complete my feature.

PR #1746, opened 2025-11-27
[make:command] Create invokable command by default

seizan8:
Updated make:command to generate a command with __invoke method.

Also added an option to use the old structure.

### DIFF
--- a/src/Maker/MakeCommand.php
+++ b/src/Maker/MakeCommand.php
@@ -18,13 +18,12 @@ use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
 use Symfony\Bundle\MakerBundle\Util\UseStatementGenerator;
+use Symfony\Component\Console\Attribute\Argument;
 use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Attribute\Option;
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Command\LazyCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
@@ -76,12 +75,10 @@ final class MakeCommand extends AbstractMaker
 
         $useStatements = new UseStatementGenerator([
             Command::class,
-            InputArgument::class,
-            InputInterface::class,
-            InputOption::class,
-            OutputInterface::class,
             SymfonyStyle::class,
             AsCommand::class,
+            Argument::class,
+            Option::class,
         ]);
 
         $generator->generateClass(
@@ -90,7 +87,6 @@ final class MakeCommand extends AbstractMaker
             [
                 'use_statements' => $useStatements,
                 'command_name' => $commandName,
-                'set_description' => !class_exists(LazyCommand::class),
             ]
         );
 
--- a/src/Maker/MakeCommand.php
+++ b/src/Maker/MakeCommand.php
@@ -18,13 +18,16 @@ use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Str;
 use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
 use Symfony\Bundle\MakerBundle\Util\UseStatementGenerator;
+use Symfony\Component\Console\Attribute\Argument;
 use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Attribute\Option;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Command\LazyCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
@@ -58,8 +61,21 @@ final class MakeCommand extends AbstractMaker
     {
         $command
             ->addArgument('name', InputArgument::OPTIONAL, \sprintf('Choose a command name (e.g. <fg=yellow>app:%s</>)', Str::asCommand(Str::getRandomTerm())))
+            ->addOption('old-structure', null, InputOption::VALUE_NONE, 'Generate the command with the old structure (configure and execute method) instead of an invokable command?')
             ->setHelp($this->getHelpFileContents('MakeCommand.txt'))
         ;
+    }
+
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
+    {
+        if (!$input->getOption('old-structure'))
+        {
+            $description = $command->getDefinition()->getOption('old-structure')->getDescription();
+            $question = new ConfirmationQuestion($description, false);
+            $isUsingOldStructure = $io->askQuestion($question);
+
+            $input->setOption('old-structure', $isUsingOldStructure);
+        }
     }
 
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator): void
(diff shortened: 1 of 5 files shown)


seizan8:
I do not understand why the tests are failing. They should be fine from what I understand.

src/Maker/MakeCommand.php
```diff
@@ -58,10 +61,22 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
     {
         $command
             ->addArgument('name', InputArgument::OPTIONAL, \sprintf('Choose a command name (e.g. <fg=yellow>app:%s</>)', Str::asCommand(Str::getRandomTerm())))
+            ->addOption('old-structure', null, InputOption::VALUE_NONE, 'Generate the command with the old structure (configure and execute method) instead of an invokable command?')
```
Kocal: I think the opposite way is the good way to go, an option `invokable` would create less frictions for actual users, and we could make it opt-out  (or single behavior) in next major version.

templates/command/InvokableCommand.tpl.php
```diff
@@ -0,0 +1,32 @@
...
+)]
+class <?= $class_name; ?>
+{
+    public function __invoke(
+        SymfonyStyle                                                  $symfonyStyle,
+        #[Argument('Required argument description')] string           $requiredArg,
+        #[Argument('Optional nullable argument description')] ?string $optionalNullableArg,
+        #[Argument('Optional non-nullable argument description')] int              $optionalNonNullableArg = 1,
+        #[Option('Option description')] string                        $option = "defaultOptionValue",
+        #[Option('Nullable option description')] ?string              $nullableOption = null,
```
Kocal:
```suggestion
        SymfonyStyle $io,
        #[Argument('Required argument description')] string $requiredArg,
        #[Argument('Optional nullable argument description')] ?string $optionalNullableArg,
        #[Argument('Optional non-nullable argument description')] int $optionalNonNullableArg = 1,
        #[Option('Option description')] string $option = "defaultOptionValue",
        #[Option('Nullable option description')] ?string $nullableOption = null,
```

Kocal: Anyway, I don't think we need so much examples, let's use something lighter instead: 
```suggestion
        SymfonyStyle $io,
        #[Argument] string $name,
        #[Option] bool $enable = false,
```

Kocal (changes requested):

tests/Maker/MakeCommandTest.php
```diff
@@ -25,11 +25,26 @@ protected function getMakerClass(): string
...
+                $this->runCommandTest($runner, 'it_makes_a_command.php');
+            }),
+        ];
+
+        yield 'it_makes_an_old_structured_command_no_attributes' => [$this->createMakerTest()
+            ->run(function (MakerTestRunner $runner) {
+                $runner->runMaker([
+                    // command name
+                    'app:foo',
+                    // create invokable command by default
```
Kocal: the comment is actually wrong

src/Maker/MakeCommand.php
seizan8: @Kocal I made the PR based on @chalasr talk at the symfony con. As far as I understand it, the invokable command should be better and how people create commands moving forward. So I wanted it to make it the default to encourage using this structure from now on. 
We can switch it around. If this would be prefered.

templates/command/InvokableCommand.tpl.php
seizan8: @Kocal I agree. I did this, because I noticed the documentation is not complete yet. So I added more examples to make it clearer what different options there are. For example the optional nullable argument is not intuitive at all if you ask me. It is nullable but does not need a default value. Usually I would expect an error because I did not pass the argument. In this case, Symfony simply passes null if the argument is not passed in the console. Which does not seem obvious to me.

src/Maker/MakeCommand.php
Kocal: Symfony 6.4 is still supported by Symfony and by the Maker Bundle, generating invokable commands by default would break for people on Symfony 6.4. But yes, invokable commands are the future. 

I think the best we can do here, is to:
- drop `old-structure` option
- add `invokable` boolean option, defaults to `Kernel::VERSION_ID >= 70300` (we already have a maker that relies on Kernel version ID) and with flag `InputOption::VALUE_NEGATABLE`

So, running:
- `make:command` on Symfony <7.3 will create a _legacy-structured_ command
- `make:command` on Symfony >=7.3 will create an invokable command
- `make:command --no-invokable` will force creating a _legacy-structured_ command
- `make:command --invokable` will force creating an invokable command

src/Maker/MakeCommand.php
```diff
@@ -58,10 +63,22 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
     {
         $command
             ->addArgument('name', InputArgument::OPTIONAL, \sprintf('Choose a command name (e.g. <fg=yellow>app:%s</>)', Str::asCommand(Str::getRandomTerm())))
+            ->addOption('invokable', null, InputOption::VALUE_NEGATABLE, 'Generate an invokable command instead of a command with the old structure (configure and execute method)?')
```
Kocal: Could be enough I think
```suggestion
            ->addOption('invokable', null, InputOption::VALUE_NEGATABLE, 'Generate an invokable command (using PHP attributes) starting from Symfony 7.3?')
```

src/Maker/MakeCommand.php
```diff
@@ -58,10 +63,22 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
     }
 
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
+    {
+        if ($input->getOption('invokable') === null) {
+            $description = $command->getDefinition()->getOption('invokable')->getDescription();
+            $question = new ConfirmationQuestion($description, Kernel::VERSION_ID >= 70300);
+            $isUsingOldStructure = $io->askQuestion($question);
+
+            $input->setOption('invokable', $isUsingOldStructure);
```
Kocal:
```suggestion
            $invokable = $io->askQuestion($question);

            $input->setOption('invokable', $invokable);
```

templates/command/InvokableCommand.tpl.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public function __invoke(
+        SymfonyStyle $io,
+        #[Argument('Required argument description')] string $requiredArg,
+        #[Argument('Optional nullable argument description')] ?string $optionalNullableArg,
+        #[Argument('Optional non-nullable argument description')] int $optionalNonNullableArg = 1,
+        #[Option('Option description')] string $option = "defaultOptionValue",
+        #[Option('Nullable option description')] ?string $nullableOption = null,
+    ): int
+    {
```
Kocal:
```suggestion
    ): int {
```

templates/command/InvokableCommand.tpl.php
Kocal: I see what you mean, but then, maybe this is more a documentation issue that.. should be fixed on the documentation itself? ;)

src/Maker/MakeCommand.php
seizan8: @Kocal Great idea! I just updated the code.

templates/command/InvokableCommand.tpl.php
seizan8: @Kocal fair point. I reduced the number of parameters in the template 👍

chalasr:
Noted myself to look into this asap. Thanks for the PR @seizan8!

seizan8:
This PR might be obsolete since PR [#1750 ](https://github.com/symfony/maker-bundle/pull/1750) seems to cover more features

src/Maker/MakeCommand.php
```diff
@@ -58,10 +63,22 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
...
+            ->addOption('invokable', null, InputOption::VALUE_NEGATABLE, 'Generate an invokable command (using PHP attributes) starting from Symfony 7.3?')
             ->setHelp($this->getHelpFileContents('MakeCommand.txt'))
         ;
     }
 
+    public function interact(InputInterface $input, ConsoleStyle $io, Command $command): void
+    {
+        if ($input->getOption('invokable') === null) {
+            $description = $command->getDefinition()->getOption('invokable')->getDescription();
+            $question = new ConfirmationQuestion($description, Kernel::VERSION_ID >= 70300);
```
GromNaN, line 75: This check should use [`Composer\InstalledVersion::satisfies()`](https://getcomposer.org/doc/07-runtime.md#knowing-whether-package-x-is-installed-in-version-y) with `symfony/console`. Distinct versions of the kernel and console might be installed.

GromNaN:
Rebased the branch on top of the latest `1.x`. Since `1.x` already requires Symfony 7.4 and PHP 8.2, invokable commands are always available, so `make:command` now generates an invokable command in every case. Following the decision in #1800, the generated output always follows the current best practices, so the choice between the invokable and the old structure has been dropped.

Changes:

- `templates/command/Command.tpl.php`: replaced by the invokable template. The command logic lives in a single `__invoke()` method with `#[Argument]` and `#[Option]` parameters. Fixed a typo on the `$enable` parameter and gave `$arg` a default value so the generated command runs out of the box.
- `src/Maker/MakeCommand.php`: updated the use statements for the invokable template (added `Argument` and `Option`, removed `InputOption`, `OutputInterface`, `LazyCommand` and the `set_description` variable). There was no option or `interact()` method left to remove, `1.x` had already dropped them.
- `tests/fixtures/make-command/tests/it_makes_a_command_in_custom_namespace.php`: adapted, because for an invokable command `Application::find()` returns a `Command` wrapper instead of the user class. The test now checks that the generated class exists in the `Custom` namespace directly.

`MakeCommandTest` passes (3 tests, 5 assertions).

Follow-up idea for later: we could let the user add arguments and options interactively during generation, the same way `make:entity` builds fields step by step, so the generated `__invoke()` signature matches the command needs from the start.

GromNaN (approved):

GromNaN:
It took time, but here we go, this is in now. Thank you very much @seizan8.

PR #1749, opened 2025-11-28
Updated MigrationDiffFilteredOutput to implement all methods from OutputInterface

joshmcrae:
This PR adds an implementation for `OutputInterface::isSilent()` which is currently missing from the `MigrationDiffFilteredOutput` class.

Currently, the `./bin/console make:migration` command is failing when the latest version of `symfony/console` is installed because the class is incomplete.


Ishadijcks:
I just had this happen on a fresh Symfony 8 project, thanks for the fix!

Kocal (approved):

javiereguiluz:
Thanks for fixing this bug Josh.

PR #1747, opened 2025-11-27
Update CI after release of Symfony 7.4 and 8.0

alexisLefebvre:
It's very likely that composer used the stable versions after their release but let’s clean this up for consistency.

Follow-up of:

- GH-1724

PR #1755, opened 2025-12-03
fix: `Authenticator` must implement `\Stringable`

alexislefebvre:
This will fix the PHP-CS-Fixer job that is failing on main branch: https://github.com/symfony/maker-bundle/actions?query=branch%3A1.x example from scheduled job: https://github.com/symfony/maker-bundle/actions/runs/19877750206

I don’t know how to fix the PHPunit tests:

- https://github.com/symfony/maker-bundle/pull/1747#issuecomment-3608114026

So this PR is only about PHP-CS-Fixer.


GromNaN:
That was also fixed in #1747

Issue #1738, opened 2025-10-28
make:controller error "An option with shortcut "i" already exists"


ctapyxin:
Conflict between options _[-i|--invokable]_ and _[-i| --id=ID]_ [The App ID](https://symfony.com/doc/current/configuration/multiple_kernels.html#step-4-update-the-front-controllers)

PR #1761, opened 2025-12-17
[maker:controller] Fix invokable `-i` conflict with APP_ID  `-i`

ayyoub-afwallah:
Fixs #1738

Unability to generate an app-specific controller

`php bin/console make:controller test -iapi`

> [critical] Error thrown while running command "make:controller test -iapi". Message: "The "-a" option does not exist."

` php bin/console make:controller test -i -iapi`

> [critical] Error thrown while running command "make:controller test -i -iapi". Message: "The "-a" option does not exist."

I think we are better off without this short param


GromNaN (approved):

GromNaN:
Thank you @ayyoub-afwallah.

PR #1766, opened 2026-01-07
PHPUnit 11 compatibility

GromNaN:
- Run tests with PHPUnit 11
- Add PHPUnit attributes, in addition to annotations

Making all the data providers static require to make many other methods static in the `tests` directory.

The main point is the `MakerTestCase` that is part of the API and is extended in other projects. Creating a new class is the only way to change abstract methods from non-static to static in a BC way.

### DIFF
--- a/src/Test/MakerTestCase.php
+++ b/src/Test/MakerTestCase.php
@@ -12,11 +12,15 @@
 namespace Symfony\Bundle\MakerBundle\Test;
 
 use Composer\Semver\Semver;
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bundle\MakerBundle\MakerInterface;
 use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\Process\Process;
 
+/**
+ * @method static iterable<array{0: MakerTestDetails}> getTestDetails()
+ */
 abstract class MakerTestCase extends TestCase
 {
     private ?KernelInterface $kernel = null;
@@ -26,18 +30,27 @@ abstract class MakerTestCase extends TestCase
      *
      * @return void
      */
+    #[DataProvider('getTestDetails')]
     public function testExecute(MakerTestDetails $makerTestDetails)
     {
         $this->executeMakerCommand($makerTestDetails);
     }
 
-    abstract public function getTestDetails();
-
     abstract protected function getMakerClass(): string;
 
+    /**
+     * @deprecated Since 1.66.0, use static::buildMakerTest() instead
+     */
     protected function createMakerTest(): MakerTestDetails
     {
+        trigger_deprecation('symfony/maker-bundle', '1.66.0', 'The "%s()" method is deprecated. Use "self::buildMakerTest()" instead.', __METHOD__, self::class);
+
         return new MakerTestDetails($this->getMakerInstance($this->getMakerClass()));
+    }
+
+    protected static function buildMakerTest(): MakerTestDetails
+    {
+        return new MakerTestDetails();
     }
 
     /**
@@ -53,6 +66,7 @@ abstract class MakerTestCase extends TestCase
             $this->markTestSkipped($testDetails->getSkippedTestMessage());
         }
 
+        $testDetails->setMaker($this->getMakerInstance($this->getMakerClass()));
         $testEnv = MakerTestEnvironment::create($testDetails);
 
         // prepare environment to test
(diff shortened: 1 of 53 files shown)


src/Test/AbstractMakerTestCase.php
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+    }
+
+    protected static function assertContainsCount(string $needle, string $haystack, int $count): void
+    {
+        self::assertEquals(1, substr_count($haystack, $needle), \sprintf('Found more than %d occurrences of "%s" in "%s"', $count, $needle, $haystack));
+    }
+
+    private static function getMakerInstance(string $makerClass): MakerInterface
```
stof: this should not be static. The maker instance should be created inside the test, so that it is coming from the containing booted for that test.

Data providers are executed *before* all tests, so all your cases are currently using the same container instance, breaking test isolation.

I think the retrieval of the maker instance from the service locator (introduced in #1771) should be moved to the test itself.

src/Test/MakerTestCase.php
```diff
@@ -12,11 +12,15 @@
...
 
 use Composer\Semver\Semver;
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bundle\MakerBundle\MakerInterface;
 use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\Process\Process;
 
+/**
+ * @method static iterable<array{0: MakerTestDetails}> getTestDetails()
```
GromNaN, line 22: By removing the abstract method declaration, moved to a PHPDoc, I avoid the breaking change.
We have to be careful to not add the abstract static method declaration in 1.x, but only in 2.x.

src/Test/MakerTestCase.php
```diff
@@ -26,20 +30,29 @@ abstract class MakerTestCase extends TestCase
...
-    abstract public function getTestDetails();
-
     abstract protected function getMakerClass(): string;
 
+    /**
+     * @deprecated Since 1.66.0, use static::newMakerTest() instead
+     */
     protected function createMakerTest(): MakerTestDetails
     {
+        trigger_deprecation('symfony/maker-bundle', '1.66.0', 'The "%s()" method is deprecated. Use "self::newMakerTest()" instead.', __METHOD__, self::class);
```
GromNaN: This method is renamed because it needs to be static. We don't want to introduce a breaking change in case this method is overridden.

src/Test/MakerTestCase.php
```diff
@@ -87,9 +101,9 @@ protected function executeMakerCommand(MakerTestDetails $testDetails)
     /**
      * @return void
      */
-    protected function assertContainsCount(string $needle, string $haystack, int $count)
+    protected static function assertContainsCount(string $needle, string $haystack, int $count)
```
GromNaN: This is a breaking change, but there is no reason to override this method. And all assertion methods are usually static.

src/Test/AbstractMakerTestCase.php
GromNaN: That's an excellent point. By delaying the retrieval of the maker in the test body, there are far fewer methods that need to be made static. It's no longer necessary to create a new class for `MakerTestCase`.

src/Test/MakerTestCase.php
stof: Do we actually *need* to make this BC break ? AFAICT, we don't *need* this method to be static to support PHPUnit 11 (making it static is more about theoretical purity here, and custom assertions methods can also be non-static if they need to access some state of the test instance)

src/Test/MakerTestCase.php
stof: shouldn't this mention `buildMakerTest` ?

src/Test/MakerTestCase.php
```diff
@@ -26,20 +30,29 @@ abstract class MakerTestCase extends TestCase
...
     {
         $this->executeMakerCommand($makerTestDetails);
     }
 
-    abstract public function getTestDetails();
-
     abstract protected function getMakerClass(): string;
 
+    /**
+     * @deprecated Since 1.66.0, use static::newMakerTest() instead
```
stof: isn't this `buildMakerTest` instead ?

src/Test/MakerTestCase.php
GromNaN: We can also deprecate this method and create a new static one with a new name.
It needs to be static because it's used inside a closure created in the static data provider method.

PR #1768, opened 2026-01-08
Upgrade to PHPStan v2

GromNaN:


src/Maker/MakeStimulusController.php
```diff
@@ -340,6 +340,7 @@ private function generateUsageExample(string $name, array $targets, array $value
     public function configureDependencies(DependencyBuilder $dependencies): void
     {
         // lower than 8.1, allow WebpackEncoreBundle
+        // @phpstan-ignore smaller.alwaysFalse
```
stof: phpstan is actually right. This is dead code due to the PHP requirement in composer.json.

GromNaN: Good point, I removed the dependency on webpack encore.

PR #1770, opened 2026-01-08
Make the root package name and path dynamically discovered in `MakerTestCase`

GromNaN:
An issue when using the `MakerTestCase` to test maker classes provided by other packages, is that it depends on the `tests` directory of the `symfony/maker-bundle` package.

In this PR, we use `Composer\InstalledVersions::getRootPackage()` to retrieve the path and name of the root package under test. This way, that's the tested package that is linked into the temporary test projects.

This assumes a certain structure of the maker package (`src`, `tests` and `tests/fixtures` dirs, `tests/tmp` ignored, `tools/twigcs` subproject).

### DIFF
--- a/src/Test/MakerTestEnvironment.php
+++ b/src/Test/MakerTestEnvironment.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MakerBundle\Test;
 
+use Composer\InstalledVersions;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Process\InputStream;
 
@@ -28,9 +29,11 @@ final class MakerTestEnvironment
     public const GENERATED_FILES_REGEX = '#(?:created|updated):\s(?:.*\\\\)*(.*\.[a-z]{3,4}).*(?:\\\\n)?#ui';
 
     private Filesystem $fs;
-    private bool|string $rootPath;
+    private string $packageName;
+    private string $rootPath;
     private string $cachePath;
     private string $flexPath;
+    private string $fixturesPath;
     private string $path;
     private MakerTestProcess $runnedMakerProcess;
     private bool $isWindows;
@@ -41,7 +44,9 @@ final class MakerTestEnvironment
         $this->isWindows = '\\' === \DIRECTORY_SEPARATOR;
 
         $this->fs = new Filesystem();
-        $this->rootPath = realpath(__DIR__.'/../../');
+        $composerPackage = InstalledVersions::getRootPackage();
+        $this->packageName = $composerPackage['name'];
+        $this->rootPath = realpath($composerPackage['install_path']);
         $cachePath = $this->rootPath.'/tests/tmp/cache';
 
         if (!$this->fs->exists($cachePath)) {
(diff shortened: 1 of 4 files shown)


src/Test/MakerTestEnvironment.php
```diff
@@ -441,12 +445,12 @@ private function addMakerBundleRepoToComposer(string $composerJsonPath): void
         // Require-dev is empty and composer complains about this being an array when we encode it again.
         unset($composerJson['require-dev']);
 
-        $composerJson['repositories']['symfony/maker-bundle'] = [
+        $composerJson['repositories'][$this->packageName] = [
```
stof: Composer 2.9 is deprecating the usage of object-based structure for `repositories` (as JSON technically does not guarantee an order for object keys). It might make sense to avoid relying on it, using the array-based structure instead (maybe using the `composer repo` command)

src/Test/MakerTestEnvironment.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MakerBundle\Test;
 
+use Composer\InstalledVersions;
```
stof, line 14: this should declare a dependency on `composer-runtime-api` with a constraint matching the version of the runtime API providing the info you rely on (probably `2.0` as lowest bound for that case, unless some of the feature you rely on was added later).

GromNaN: The `install_path` was added in Composer 2.1 by https://github.com/composer/composer/pull/9699

src/Test/MakerTestEnvironment.php
```diff
@@ -431,26 +434,24 @@ private function composerRequireMakerBundle(string $projectDirectory): void
...
                 'versions' => [
-                    'symfony/maker-bundle' => '9999.99', // Arbitrary version to avoid stability conflicts
+                    $this->packageName => '9999.99', // Arbitrary version to avoid stability conflicts
                 ],
             ],
         ];
 
-        file_put_contents($composerJsonPath, json_encode($composerJson, \JSON_THROW_ON_ERROR | \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES));
+        MakerTestProcess::create(\sprintf(
+            'composer repo add %s \'%s\' ',
```
stof: this won't work on Windows (where you cannot use single-quote to wrap arguments).

I think the best solution would be to make MakerTestProcess support the array-based signature to define a Process (which is a lot safer than building a command-line that is parsed by a shell as it solves the need for shell escaping)

GromNaN: I didn't realize that was already possible. I added the param type to the relevant methods (all internal or final).

src/Test/MakerTestEnvironment.php
```diff
@@ -190,7 +186,7 @@ public function prepareDirectory(): void
         }
     }
 
-    public function runCommand(string $command): MakerTestProcess
+    public function runCommand(string|array $command): MakerTestProcess
```
GromNaN, line 205: Not a breaking change because the class is internal.

src/Test/MakerTestProcess.php
```diff
@@ -31,7 +31,7 @@ private function __construct($commandLine, $cwd, array $envVars, $timeout)
         $this->process->setEnv($envVars);
     }
 
-    public static function create($commandLine, $cwd, array $envVars = [], $timeout = null): self
+    public static function create(string|array $commandLine, string $cwd, array $envVars = [], ?float $timeout = null): self
```
GromNaN, line 40: Not a breaking change because the class is final and internal

src/Test/MakerTestEnvironment.php
stof, line 205: I suggest adding a phpdoc `@param string|list<string> $command` to document the type of the array.

stof: same in MakerTestProcess btw

PR #1773, opened 2026-01-10
[make:test] Remove trailing slash in assertResponseRedirects tests

tcoch:
Out of the box, the generated tests do not work.
I found out that removing the trailing slash is actually enough to make them work !
This PR aims to fix that.


GromNaN (approved):

GromNaN:
Thank you @tcoch.

PR #1774, opened 2026-01-11
[make:crud] Ensure the generated functional tests run in MakerBundle's CI

GromNaN:
In order to ensure there is no regression in the generated test case (https://github.com/symfony/maker-bundle/pull/1773), they must be run in MakerBundle's CI.

The `markTestIncomplete` call stops the execution by throwing and exception. Skipping all the test body. By moving this call at the end of the method, the tests are fully run.

I even wonder if marking tests as incomplete serves any purpose.

For reference, this was added in https://github.com/symfony/maker-bundle/pull/307, this point was not discussed.

### DIFF
--- a/templates/crud/test/Test.EntityManager.tpl.php
+++ b/templates/crud/test/Test.EntityManager.tpl.php
@@ -40,7 +40,6 @@ namespace <?= $namespace ?>;
 
     public function testNew(): void
     {
-        $this->markTestIncomplete();
         $this->client->request('GET', sprintf('%snew', $this->path));
 
         self::assertResponseStatusCodeSame(200);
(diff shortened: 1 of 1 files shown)


tcoch:
I believe it's good for now to mark tests as incomplete, for two reasons:
- check L75 : `// Use assertions to check that the properties are properly displayed.`. Wit hthat you state that the generated tests are incomplete
- When using the command, there's a disclaimer about this being an experimental feature. Marking test as incomplete forces the developer to check what was generated, or at least to remove theses lines.

But yes, moving them at the end of the method is good for the CI!

tcoch (approved):

PR #1786, opened 2026-02-17
Remove redundant PHP CS Fixer job

HypeMC:
PHP CS Fixer is already run by Fabbot, so there's no need to run it again in this workflow.


Kocal (approved):

GromNaN (approved):

GromNaN:
Thanks @HypeMC.

PR #1785, opened 2026-02-13
[make:entity] Autocompletion for enum classes

tcoch:
This PR aims to provide autocompletion for `enum` classes, just like it exists for doctrine relations.

Two tests on generating entity with enum types already exist (https://github.com/symfony/maker-bundle/blob/1.x/tests/Maker/MakeEntityTest.php#L697-L741) to cover that this new feature won't break anything.

PR #1790, opened 2026-03-17
Update Yaml indentation in tests after symfony/symfony#62967

tcoch:
https://github.com/symfony/symfony/pull/62967 introduced a change in how YAML files are generated.
For this reason, some tests are broken here (e.g. https://github.com/symfony/maker-bundle/pull/1785).
This PR aims to resolve that.

The modifications are:
- Test files used have been adapted towards the new format
- Special case has been added to `YamlSourceManipulatorTest.php` to address specify of tests in this CI with PHP8.1 and SF6.4.* lowest deps

### DIFF
--- a/tests/Util/YamlSourceManipulatorTest.php
+++ b/tests/Util/YamlSourceManipulatorTest.php
@@ -11,11 +11,14 @@
 
 namespace Symfony\Bundle\MakerBundle\Tests\Util;
 
+use Composer\InstalledVersions;
+use Composer\Semver\VersionParser;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Psr\Log\LogLevel;
 use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\HttpKernel\Kernel;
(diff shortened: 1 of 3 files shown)


tests/Util/YamlSourceManipulatorTest.php
```diff
@@ -48,7 +49,11 @@ public function testSetData(string $startingSource, array $newData, string $expe
          */
         $actualContents = $manipulator->getContents();
         $actualContents = str_replace("\r\n", "\n", $actualContents);
-        $this->assertSame($expectedSource, $actualContents);
+        if (Kernel::VERSION_ID < 81000) {
```
stof: this should detect the version of `symfony/yaml`, not of `symfony/http-kernel` (they might differ)

tcoch: I agree with you but I couldn't figure out a clean way of doing so. Got any ideas ?

stof: I suggest using the composer runtime API: https://getcomposer.org/doc/07-runtime.md#installed-versions

tcoch: Done, thanks for the tip!

GromNaN (approved):

GromNaN:
Thank you @tcoch.

Issue #1344, opened 2023-08-11
Specific namespace for different objects


samihsoylu:
Imagine that you want to create a controller in different directory, and entities in another. How could you achieve this through the config?

```yaml
maker:
  root_namespace: 'App'
  controller:
    namespace: App\Application\Controller
    dir: src/Application/Controller
  entity:
    namespace: App\Domain\Entity
    dir: src/Domain/Entity
  repository:
    namespace: App\Core\Repository\Doctrine
    dir: src/Core/Repository/Doctrine
```

The keywords controller, entity, and repository are not recognised. The only allowed option is `root_namespace`.

tg-iserv:
I'm also interested in this feature, as our project also do not use the default directory structure

GromNaN:
The `dir` can be found from the namespace with the psr-4 autoloading rule. Therefore, only the namespace should be specified.
Would an option when launching the command be sufficient? I don't think adding configuration is in line with the philosophy of maker-bundle, which is intended to be "plug-and-play".

tg-iserv:
I agree that the `dir` option is not necessary.
Having to specify the whole namespace each time i want to make something would be annoying therefore I would prefer a configuration option. 
You can use the current value as default and the maker-bundle would still be "plug-and-play".
The same way as the `root_namespace` which must only be set if someone does not follow the Symfony conventions.

Still I see advantages in the command option:
- it allows for using multiple directories for controllers/entities/repositories
- it can be used to create sub namespace of entities.

Amoifr:
Hey @GromNaN! I put together a proof-of-concept in #1794 to move the discussion forward — **not meant to be merged as-is**, just a concrete base to iterate on.

Here's the approach:

- **Config-based** with sensible defaults (nothing changes if you don't configure anything):
```yaml
maker:
    namespaces:
        entity: Domain\Entity
        controller: Application\Controller
```

- **No `dir` option** — paths are derived from PSR-4 autoloading, exactly as you suggested
- **`NamespaceType` enum** to avoid magic strings in the codebase
- **Open map** (`scalarPrototype`) — users only override what they need, defaults come from the enum
- **Validation** via `Validator::validateClassName()` on each namespace value
- **10 makers updated** — all makers using the 5 main types (controller, command, entity, form, repository)

A CLI option (`--namespace`) for one-off overrides could complement this as a follow-up — would love your thoughts on whether that's worth pursuing too.

Happy to adjust the approach based on your feedback! 🙂

PR #1794, opened 2026-04-10
Add configurable namespaces for generated classes

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Issues        | #1344
| License       | MIT

This PR is a **proof of concept** to support the discussion in #1344 — it is not intended to be merged as-is, but to serve as a concrete base for the ongoing reflection.

### What it does

Adds an optional `namespaces` configuration to override the default namespace for generated classes:

```yaml
maker:
    namespaces:
        entity: Domain\Entity
        controller: Application\Controller
        repository: Infrastructure\Repository
```

The directory is derived automatically from PSR-4 autoloading rules (no `dir` option needed, as @GromNaN suggested).

### Design choices

- **`NamespaceType` enum** — replaces magic strings with a typed enum for all 5 configurable types (controller, command, entity, form, repository). Provides IDE autocompletion and prevents typos.
- **Defaults centralized in the enum** — each case knows its default via `defaultNamespace()`. The `Generator` merges user config on top of these defaults.
- **Open map via `scalarPrototype`** — the config accepts any key, not a fixed list. Users can override only what they need.
- **Validation** — each namespace value is validated via `Validator::validateClassName()`.
- **10 makers updated** — all makers using the 5 configurable types now go through `$generator->getNamespace(NamespaceType::Entity)` instead of hardcoded `'Entity\\'`.
- **No constructor injection needed** — `Generator` is already available in `generate()`. The `MakeController` logic was moved from `interact()` to `generate()` to avoid adding a constructor dependency.

### What's NOT in scope (possible follow-ups)

- CLI option (`--namespace`) for one-off overrides (mentioned by @GromNaN)
- Additional types beyond the initial 5 (listener, message, voter, etc.)

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -27,19 +27,35 @@ class Generator
     private GeneratorTwigHelper $twigHelper;
     private array $pendingOperations = [];
     private array $generatedFiles = [];
+    private array $configuredNamespaces;
 
     public function __construct(
         private FileManager $fileManager,
         private string $namespacePrefix,
         ?PhpCompatUtil $phpCompatUtil = null,
         private ?TemplateComponentGenerator $templateComponentGenerator = null,
+        array $configuredNamespaces = [],
     ) {
         $this->twigHelper = new GeneratorTwigHelper($fileManager);
         $this->namespacePrefix = trim($namespacePrefix, '\\');
 
+        $defaults = [];
+        foreach (NamespaceType::cases() as $case) {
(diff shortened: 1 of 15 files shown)


Amoifr:
Closing this proof of concept since the discussion it was supporting (#1344) has been closed. Happy to revive it if the topic ever comes back on the table. Thanks for all the work on the maker!

PR #1795, opened 2026-05-12
[make:entity] Fix `--with-uuid` and `--with-ulid` being ignored when name is passed as argument

### LINKED ISSUE #1789: `--with-uuid` and `--with-ulid` options ignored when entity name provided as argument
When running `make:entity` command with `--with-uuid` or `--with-ulid` options AND providing the entity name as an argument, the options are silently ignored and the entity is generated with a default integer ID instead.

**Steps to Reproduce:**

```shell script
php bin/console make:entity MyEntity --with-uuid
```

**Expected Behavior:**

Entity should be generated with UUID as the ID type.

**Actual Behavior:**

Entity is generated with integer ID type. The UUID/ULID options are completely ignored.

**Root Cause:**

In `MakeEntity::interact()` method (lines 110-112), there is an early return when the entity name is provided as an argument. This happens **before** the `checkIsUsingUid($input)` call on line 126, which means the `$usesUuid` and `$usesUlid` flags are never set.

**Proposed Solution:**

Move the `checkIsUsingUid($input)` call before the early return condition on line 110, so it's always executed regardless of whether the entity name is provided as an argument.

**Impact:**

This bug affects:
- Non-interactive usage (CI/CD, automated tests, scripts)
- Any scenario where entity name is provided as argument
- Third-party bundles that extend `make:entity`

**Environment:**

- Symfony MakerBundle: 1.x
- Symfony: 7.4+
- PHP: 8.1+

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

When the entity name is provided as a command argument, `MakeEntity::interact()` returns early before reaching `$this->checkIsUsingUid($input)`. The `$usesUuid` / `$usesUlid` flags are therefore never set, and the entity is always generated with an integer `id`, silently ignoring `--with-uuid` / `--with-ulid`.

```bash
# Before this PR: id is int, options silently ignored
php bin/console make:entity MyEntity --with-uuid
```

The fix moves the call to the top of `interact()`, which:

- Matches the pattern already used in `MakeUser::interact()` and `MakeResetPassword::interact()`.
- Matches the docblock in `UidTrait::checkIsUsingUid()`: *"Call this as early as possible in a maker's interact()."*
- Also surfaces the existing validation (mutually-exclusive `--with-uuid` / `--with-ulid`, missing `symfony/uid`) earlier in non-interactive flows.

Fixes #1789

### DIFF
--- a/src/Maker/Common/UidTrait.php
+++ b/src/Maker/Common/UidTrait.php
@@ -25,9 +25,6 @@ use Symfony\Component\Uid\Uuid;
  */
 trait UidTrait
 {
-    private bool $usesUuid = false;
-    private bool $usesUlid = false;
-
     /**
      * Call this in a maker's configure() to consistently allow entity's with UUID's.
      * This should be called after you calling "setHelp()" in the maker.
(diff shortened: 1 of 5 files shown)


GromNaN:
Rebased onto 1.x

The branch was rebased on the current `1.x` head and force-pushed, with no conflicts. Since the base moved a lot (56 commits), the commit that validates the fix still sits on top.

### Also works under `--no-interaction`

While verifying the fix, I found that `--with-uuid` and `--with-ulid` were still ignored when `make:entity` runs with `--no-interaction`. `UidTrait` kept the choice in properties set only during `interact()`, and Symfony skips `interact()` entirely in that mode.

The follow-up commit `cf2d5053` makes `getIdType()` read the options directly from the input, so the flags are honoured by `make:entity`, `make:user` and `make:reset-password` in interactive and non-interactive modes alike.

Regression tests covering both paths:

- `it_creates_a_new_class_with_uuid_when_name_is_passed_as_argument`
- `it_creates_a_new_class_with_uuid_under_no_interaction`

GromNaN:
Thank you @Amoifr.

Issue #198, opened 2018-06-10
make:entity wrong setter for non nullable fields


sbruev:
Hi!
It's a minor syntax mistake, but right now generated setters allow to pass the null value to non nullable fields, if they are ManyToOne relationship.
```php
    /**
     * @ORM\ManyToOne(targetEntity="App\Entity\AnotherEntity")
     * @ORM\JoinColumn(nullable=false)
     */
    private $anotherEntity;

    //...

    public function setAnotherEntity(?AnotherEntity $anotherEntity): self
    {
        $this->anotherEntity = $anotherEntity;

        return $this;
    }
```
The arg should be `AnotherEntity` not `?AnotherEntity`

Cheers!

lenar:
Are you sure this is a bug? This would allow to use validations, otherwise the failure occurs too early.
I would say that currently there is a bug with ordinary fields which don't have ? before type if nullable=false. Failure occurs too early, makes it hard to validate. Maybe it should be configurable if one wants to allow nulls or not for setters.

weaverryan:
Hmm, interesting. Are you using the form submit, and it’s submitting the values as null?

lenar:
No, not using form submit. "Legacy" code, build with older Symfony releases. Manual validations with Symfony validation/constraint classes.

Anyway it should be selectable (by argument to console command) if you want very strict definitions (nulls not allowed if nullable=false and thus fail with PHP error) or you want to handle things yourself later with validations (and this seems to be the default with older releases and doctrine:generate:entitites).

And I really prefer handling nulls with validations and not manually modifying string => ?string in setters of generated entities.

lenar:
Looking at older (doctrine:generate) generated entities, the setters are `setSomething($something = null)` irrelevant of `nullable` of column annotation. They were generated automatically. And it makes sense IMHO.

javiereguiluz:
Even if this is correct in theory ... I'm afraid that in practice, when using Symfony Forms, is not correct. You must make everything nullable or the app won't work. That's why we did that in the Symfony Demo app for example: https://github.com/symfony/demo/blob/master/src/Entity/Post.php

mesolaries:
+1 for this to be non-nullable. 
One thing to consider is the generated `remove*` methods for bidirectional associations, which set the owning side to `null`.

Also, when using typed properties, phpstan/phpstan-doctrine reports a `doctrine.associationType` error:

```php
    /**
[...]

PR #1797, opened 2026-06-18
[make:entity] Generate a strict setter for non-nullable unidirectional ManyToOne relations

Amoifr:
## Summary

Concrete design proposal for #198, which reports that generated `ManyToOne` setters accept `null` even when the relation is non-nullable (`setCategory(?Category $category)` for a `nullable: false` association).

The discussion stalled on a real constraint: the setter must stay nullable for **bidirectional** relations, because the collection `remove*()` generated on the inverse side sets the owning side back to null (`$item->setOwner(null)`, orphanRemoval) — making every `ManyToOne` setter strict would break that generated code.

This PR tightens only the **provably-safe subset**: a non-nullable, **unidirectional** `ManyToOne`, which has no such `remove*()` null path:

```php
public function setCategory(Category $category): static  // was: ?Category
```

Bidirectional and explicitly-nullable relations are unchanged. This mirrors what already happens for non-nullable **OneToOne** setters (`$relation->isNullable()`), just extended to the `ManyToOne` case that is provably safe.

## Scope / open questions

- Deliberately **setter-only** (the original report). The backing property stays `?Category $category = null` — consistent with the current OneToOne behavior, and it avoids uninitialized-typed-property pitfalls with Forms.
- Also making the **property** non-nullable (raised again recently re: phpstan-doctrine `doctrine.associationType`) is a bigger discussion — it reopens @javiereguiluz's Forms concern about uninitialized properties — so it's left out here. Happy to follow up if you'd like to go there.

No existing fixtures change (no current test case is non-nullable + unidirectional); a new `many_to_one_not_nullable_no_inverse` case + fixture covers it.

### DIFF
--- a/src/Util/ClassSourceManipulator.php
+++ b/src/Util/ClassSourceManipulator.php
@@ -531,12 +531,14 @@ final class ClassSourceManipulator
         $setterNodeBuilder = $this->createSetterNodeBuilder(
             $relation->getPropertyName(),
             $typeHint,
-            // make the type-hint nullable always for ManyToOne to allow the owning
-            // side to be set to null, which is needed for orphanRemoval
-            // (specifically: when you set the inverse side, the generated
-            // code will *also* set the owning side to null - so it needs to be allowed)
-            // e.g. $userAvatarPhoto->setUser(null);
-            $relation instanceof RelationOneToOne ? $relation->isNullable() : true
+            // A ManyToOne setter must stay nullable as soon as the relation maps an
+            // inverse side: the collection remove*() generated on that side sets the
+            // owning side back to null (orphanRemoval), e.g. $userAvatarPhoto->setUser(null);
+            // so null must be allowed there. A non-nullable, *unidirectional* ManyToOne
+            // has no such code path, so its setter can use a strict, non-nullable type-hint.
+            $relation instanceof RelationOneToOne
+                ? $relation->isNullable()
+                : ($relation->isNullable() || $relation->getMapInverseRelation())
         );
 
         // set the *owning* side of the relation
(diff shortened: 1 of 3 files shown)


Amoifr:
A heads up on the `Fabbot / Checks` job, because two of its steps ask for opposite things and I do not think a pull request touching a test file can satisfy both.

`tests/Util/ClassSourceManipulatorTest.php` has test methods without a return type. Since this pull request touches that file, the job checks it:

* the **Run PHP-CS-Fixer** step fails and hands me a patch that adds `: void` to nine test methods. The rule is `void_return`, pulled in by `@PHPUnit84Migration:risky` in `.php-cs-fixer.dist.php`;
* the **Test method names should not have a return type** step fails as soon as they are there.

I tried both states today. With the return types, the second step failed with `Test case methods should not have a return type. Please apply the patch below.` Without them, the first one fails with `PHP-CS-Fixer found style issues.` The file is untouched by me in that respect: running the repository's own fixer on `1.x` flags it too, so this is not something this pull request introduced.

I left the file as it is on `1.x`, so the diff stays minimal, and applied the other thing the job asked for, the full stop on the `Unknown relation type` exception message.

Nothing urgent here for me, and I am not asking for anything on this pull request. It just seemed worth reporting, since anyone touching a test file with the same shape will hit the same wall. Excluding `void_return` for `tests/` or dropping that fabbot step would both settle it, and you know better than I do which one fits.

For the rest of the CI on this branch: I rebased it on the current `1.x`, which turned `PHPStan` and `Testing on Windows` green again. The four remaining matrix jobs fail on `1.x` as well, with the same names.

PR #1798, opened 2026-06-28
[MakeEntity] Remove dead is_string() branch in generate()

### LINKED ISSUE #1509: [make:entity] `generate()` call to `is_string()` appears to always evaluate to `false`
https://github.com/symfony/maker-bundle/blob/f7ee2503fbea634f2e8c30170a8ad8bd0e3d378f/src/Maker/MakeEntity.php#L309

I believe `$manipulatorOrMessage` is always an instance of `ClassSourceManipulator` but I may be wrong. This could be an artifact leftover from previous versions where we did display a message. We need to confirm what use case would cause this _not_ to be an instance of `CSM`. 

For debugging:

- change the conditional to `$manipulatorOrMessage instanceOf ClassSourceManipulator` -> `getSourceCode()` && exit early. 
- then attach breakpoint || `dump($manipulatorOrMessage)` in the else statement.

Amoifr:
## Summary

Fixes #1509.

As suspected in the issue, the `is_string($manipulatorOrMessage)` check in `MakeEntity::generate()` is always `false`: every value stored in `$fileManagerOperations` comes from `createClassManipulator()` and therefore is a `ClassSourceManipulator`. The string branch (`$io->comment(...)`) is dead code left over from older versions that returned a message — which is exactly why it carried a `@phpstan-ignore-line` pointing to this issue.

This PR drops the dead branch (and the `@phpstan-ignore-line`), and dumps the files directly:

```php
foreach ($fileManagerOperations as $path => $manipulator) {
    $this->fileManager->dumpFile($path, $manipulator->getSourceCode());
}
```

No behavior change — only the unreachable branch is removed. `MakeEntityTest` produces identical results before/after.

### DIFF
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -305,12 +305,8 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
                 throw new \Exception('Invalid value.');
             }
 
-            foreach ($fileManagerOperations as $path => $manipulatorOrMessage) {
-                if (\is_string($manipulatorOrMessage)) {     /* @phpstan-ignore-line - https://github.com/symfony/maker-bundle/issues/1509 */
-                    $io->comment($manipulatorOrMessage);
-                } else {
-                    $this->fileManager->dumpFile($path, $manipulatorOrMessage->getSourceCode());
-                }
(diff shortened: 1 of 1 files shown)


GromNaN (approved):

GromNaN:
Thank you @Amoifr.

PR #1799, opened 2026-06-30
[MakeEntity] Apply the generate_final_entities config to generated entities

### LINKED ISSUE #1731: generate_final_entities configuration does not work
I have the config for final entities enabled

```yaml
maker:
    generate_final_entities: true
```

But when using the `make:entity` command the class is still created without the final keyword. It seems that the option is not passed/used in the process of generating an entity.

https://github.com/symfony/maker-bundle/blob/c86da84640b0586e92aee2b276ee3638ef2f425a/src/Maker/MakeEntity.php#L192-L197

https://github.com/symfony/maker-bundle/blob/c86da84640b0586e92aee2b276ee3638ef2f425a/src/Doctrine/EntityClassGenerator.php#L80-L92

https://github.com/symfony/maker-bundle/blob/c86da84640b0586e92aee2b276ee3638ef2f425a/templates/doctrine/Entity.tpl.php#L21

Amoifr:
## Summary

Fixes #1731.

The `maker.generate_final_entities` option was wired into the `TemplateComponentGenerator` (used by makers based on `ClassData`), but **not** into the `EntityClassGenerator`, which generates entities from the legacy `doctrine/Entity.tpl.php` template. As a result, `make:entity` (and `make:user` / `make:reset-password`) always produced non-`final` entities, regardless of the configuration.

This PR passes the configured value to the `EntityClassGenerator` and renders the `final` keyword in the template:

```yaml
maker:
    generate_final_entities: true
```

```php
// before: class User { ... }
// after:  final class User { ... }
```

The option still defaults to `false`, so the default behavior is unchanged. Added a functional test covering the `generate_final_entities: true` case.

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -89,6 +89,7 @@ return static function (ContainerConfigurator $container) {
         ->args([
             service('maker.generator'),
             service('maker.doctrine_helper'),
+            false, // overridden by the "generate_final_entities" config in MakerBundle::loadExtension()
         ]);
 
     $services->set('maker.user_class_builder', UserClassBuilder::class);
--- a/src/Doctrine/EntityClassGenerator.php
+++ b/src/Doctrine/EntityClassGenerator.php
@@ -37,6 +37,7 @@ final class EntityClassGenerator
     public function __construct(
         private Generator $generator,
         private DoctrineHelper $doctrineHelper,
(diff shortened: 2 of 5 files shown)


GromNaN (approved):

GromNaN:
Thank you @Amoifr.

Issue #1787, opened 2026-03-01
Unable to extend existing makers because they are declared final


sovetski:
Currently, each maker class is declared as final, which prevents extending an existing maker to adapt it to custom needs.

Here: https://symfony.com/bundles/SymfonyMakerBundle/current/index.html#creating-your-own-makers, it is explained that we can create our own makers. However, the problem is that this requires rebuilding everything from scratch every time.

My personal use case is to extend the MakeEntity class (https://github.com/symfony/maker-bundle/blob/1.x/src/Maker/MakeEntity.php) in order to change the default directory.
At the moment, when running `make:entity`, it always looks for or creates entities in the `src/Entity` directory, whereas in some projects entities are located in different folders, for example `src/FolderA/TestEntity.php`.

Related issues:
https://github.com/symfony/maker-bundle/issues/51
https://github.com/symfony/symfony/issues/28790
https://github.com/symfony/maker-bundle/issues/247

sachingk:
I do have the same concern. I am not able to extend make:entity and add my own logic into it... 

I still don't understand what is the necessary to make the them final ?

GromNaN:
> My personal use case is to extend the MakeEntity class (https://github.com/symfony/maker-bundle/blob/1.x/src/Maker/MakeEntity.php) in order to change the default directory. At the moment, when running `make:entity`, it always looks for or creates entities in the `src/Entity` directory, whereas in some projects entities are located in different folders, for example `src/FolderA/TestEntity.php`.

You may update the [`maker.doctrine_helper` service definition](https://github.com/symfony/maker-bundle/blob/ebfdbd164e62400a4f1a97391de96f6be66d3489/src/MakerBundle.php#L60-L61) to change the entity namespace. This is not covered by our BC policy, but the codebase is quite stable.

> I still don't understand what is the necessary to make the them final ?

This is a maintainer concern: when a class is final, we can modify protected methods, add optional arguments to public methods, and other changes without breaking backward compatibility.

We strongly recommend using composition instead of inheritance: create a new **decorator** class that implements `InputAwareMakerInterface` and delegates all method calls to the methods of these interfaces. Then [configure this decorator in the DI container](https://symfony.com/doc/current/service_container/service_decoration.html).

```php
use Symfony\Bundle\MakerBundle\Maker\MakeEntity;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputInterface;

[...]

Issue #1679, opened 2025-03-29
make:entity to generate non-nullable fields


valx76:
Hey there!

Thinking about the issue #1175.

How about an option in the maker command to define the nullable behavior ?
Some examples:
- Everything (mostly) nullable like the current behavior
- Set the types in PHP as they are specified by the user (and warns that accessing an unassigned non-nullable value will result in errors)
- Same as above but also defines a default value when not nullable (0, empty string, etc.) to avoid unassigned errors

What do you guys think?

Issue #1399, opened 2023-11-22
Support target directories other than the first one matching a namespace


janopae:
The Symfony Maker bundle decides the path where it creates files by transforming namespaces (given by the individual makers) to file paths, given the composer autloading configuration. This transformation is not always possible in a unique way.

The code that transforms namespaces to directories currently resides [here](https://github.com/symfony/maker-bundle/blob/112f9466c94a46ca33dc441eee59a12cd1790757/src/Util/AutoloaderUtil.php#L33).

In cases where a namespace can't be resolved to a path in a unique way, it just picks the "first" possible namespace (whatever this means, as JSON object keys, such as the `"autoload"` key inside the `composer.json`, are by definition unordered).

This means that it is not possible to generate files in directories outside this "first match" autoload path.

#### For example

If you configure your test cases to live in the same namespaces as the classes they test (useful to test classes that are `@psalm-internal`, and generally to reduce `use` statements), but in a separate directory (useful for tools like the `composer-require-checker`), you might have an autoload configuration like the following:

```json
    "autoload": {
        "psr-4": {
            "": "src/"
        },
    },
    "autoload-dev": {
        "psr-4": {
            "": "tests/"
        }
    },
```

In this case, it is not possible to create files in the `tests` directory using MakerBundle.

janopae:
I could think of the following ways to solve this:

### Interactive selection

PHPStorm provides an interactive selection when moving a class to a namespace and the path is ambigious. A maker command can't spawn a window, but it could use an interactive CLI:

```
Path for /App/AcmeFactory:
  [0] src/App/AcmeFactory.php
  [1] tests/App/AcmeFactory.php
>
```

It seems like this needs to be explicitly supported by the make commands – either by calling a function, which gets an Output passed and checks the namespace for ambiguosity and does the prompt in case, or by calling that function when handling an exception that `AutoloaderUtil` would throw, or by passing the Output down to the `FileManager`/`AutoloaderUtil`.

### Provide a way to hard-overwrite the path

Calling a maker command with `--target-directory=tests/App/Acme` could ignore the namespace and hardly create the file in `tests/App/Acme`(maybe throw in error if this doesn't make sense given the namespace).

This would be much harder to use, but easier to implement, as we wouldn't need to pass an Output to the `AutoloaderUtil` or a similar function.

However, it still needs explicit support by the make commands, as they need to catch that parameter and add its content to the `ClassNameDetails`.

PR #1723, opened 2025-08-04
[make:registration-form][make:reset-password] Refactor constraints to use named arguments

octoseth:
When generating registration or reset password forms using the MakerBundle, some validation constraints are still passed using arrays. This triggers deprecation warnings in Symfony 7.3+:

> Passing an array of options to configure the "%s" constraint is deprecated, use named arguments instead.

This PR updates the generated code to use named arguments for these constraints, removing the deprecations and ensuring compatibility with future Symfony versions.

PR #1800, opened 2026-08-03
Redefine the scope of MakerBundle

GromNaN:
Following a discussion among Symfony core team members about the future of MakerBundle and its relationship with Symfony Mate (part of `symfony/ai`), this PR documents a clearer, narrower scope for the project.

## Context

Key points that emerged from the discussion:

- MakerBundle has become exceptionally hard to maintain. The test suite is slow and complex to debug (each test spins up a temporary project and runs commands).
- Every new option adds combinatorial complexity, both for users and for maintainers, sometimes turning generators into meta-programming through the CLI.
- Most of the templates are simple boilerplate that require no AI at all. The complexity lives in advanced generators (notably `make:entity`).
- MakerBundle still has real value for newcomers, for training sessions, and for offline / AI-free workflows. Dropping it entirely is not on the table right now.
- Coding agents with dedicated skills (e.g. Symfony Mate, Laravel Boost style) are a better fit for customization, edge cases, and opinionated variations.

## Proposal

Rather than dropping MakerBundle, we narrow its scope:

- Target audience: newcomers starting a Symfony project without writing boilerplate.
- Generate code that reflects our current recommended best practices, with sensible defaults and minimal configuration.
- Reject options and switches that exist only to cover edge cases or personal preferences.
- Redirect requests for extensibility, custom templates, alternative target directories, or opinionated code shapes to coding agents and skills-based tools.

Examples of requests that fall outside this new scope:

- Extending the makers (#1787)
- `make:entity` non-nullable option (#1679)
- Configurable template paths (#1135)
- Changing target directories (#1399)

Example of requests that could be accepted, changing the generated code:
- Invokable command generation (#1746)
- `$form->createView()` in controllers (#1709)
- specify the table name of an entity (#1730)

The README already warns that generated code can change in any version; this PR makes the underlying philosophy explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,36 @@ The MakerBundle is the fastest way to generate the most common code you'll
 need in a Symfony app: commands, controllers, form classes, event subscribers
 and more!
 
+Scope
+-----
+
+This project is aimed at developers who are new to a given Symfony feature and
+want a working starting point without writing boilerplate code. That includes
+absolute Symfony beginners, but also experienced developers picking up an
+unfamiliar component. It generates code that follows the current recommended
+best practices, with sensible defaults and minimal configuration.
+
+A maker is kept even if a project only runs it once, because that single run
+is precisely when the developer benefits the most from a working skeleton. On
+the other hand, makers whose output is fully covered by a Flex recipe, or that
+duplicate another maker, are out of scope.
+
+For any requirement not covered by these defaults (custom templates, alternative
+target directories, extended generators, opinionated code shapes, etc.), we
+recommend using coding agents with dedicated skills, or other code-generation
+tools. See [Symfony Mate][3] for an AI-assisted, skills-based approach.
+
+As a consequence:
+
+- Options and switches that exist only to cover edge cases or personal
+  preferences will not be added. The generated code always reflects our current
+  recommended defaults and may evolve between releases.
+- Requests to make the generators broadly extensible, configurable, or
+  pluggable are out of scope.
+- Only the latest Symfony LTS is supported. The generated code targets that
+  LTS. New makers for post-LTS features may still be added, with a runtime
+  check on the required dependency versions.
+
 Documentation
 -------------
 
@@ -25,6 +55,7 @@ B) The generated code itself may change between minor releases. This
 
 [1]: https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html
 [2]: https://symfony.com/doc/current/contributing/code/bc.html
+[3]: https://symfony.com/doc/current/ai/components/mate.html
 
 ---
 


kbond:
Within this new scope, if we have `make:command` generate invokable commands, would it still fallback to standard command if invokable commands aren't supported? I ask because these permutations can add up too.

GromNaN:
Here are the open PRs that fall **outside** the new scope (to be closed):                                                
                                                                                         
Custom templates / paths / namespaces                                                      
- https://github.com/symfony/maker-bundle/pull/1710 add customizable templates path
- https://github.com/symfony/maker-bundle/pull/1672 Customize template inherited           
- https://github.com/symfony/maker-bundle/pull/562 function to change template name path 
- https://github.com/symfony/maker-bundle/pull/378 Configuration for custom namespaces     
                                                                                           
Extensibility / scaffolding / pluggability                                                 
- https://github.com/symfony/maker-bundle/pull/1085 [RFC] scaffolding system make:scaffold 
- https://github.com/symfony/maker-bundle/pull/1466 interactively install composer         
dependencies                                                                               
- https://github.com/symfony/maker-bundle/pull/436 Fluent setters bundle configuration     
option                                                                                     
                                                                                           
Edge-case options / personal preferences on existing makers                              
- https://github.com/symfony/maker-bundle/pull/1730 make:entity specify table name         
- https://github.com/symfony/maker-bundle/pull/1635 make:entity specify table name         
(duplicate of #1730)                                                              
- https://github.com/symfony/maker-bundle/pull/1661 make:entity unsigned integer question  
- https://github.com/symfony/maker-bundle/pull/1558 make:entity default value option     
- https://github.com/symfony/maker-bundle/pull/1119 make:entity multiple entity managers   
- https://github.com/symfony/maker-bundle/pull/1555 make:registration-form add           
hash_property_path option                                                                  
- https://github.com/symfony/maker-bundle/pull/1595 make:security:form-login no-interaction
 option                                                                                    
- https://github.com/symfony/maker-bundle/pull/1594 make:security:custom no-interaction    
option                                                                                   
- https://github.com/symfony/maker-bundle/pull/544 make:crud change field name appearance  
- https://github.com/symfony/maker-bundle/pull/1527 make:reset-password handle multiple  
entities                                                                                   
- https://github.com/symfony/maker-bundle/pull/798 ResetPassword API Platform support      
                                                                                           
New niche makers                                                                           
- https://github.com/symfony/maker-bundle/pull/1704 make:doctrine:listener                 
- https://github.com/symfony/maker-bundle/pull/1613 make:decorator        
- https://github.com/symfony/maker-bundle/pull/1574 make:event                             
- https://github.com/symfony/maker-bundle/pull/1218 make:form-extension                    
- https://github.com/symfony/maker-bundle/pull/303 make:dto                                
- https://github.com/symfony/maker-bundle/pull/1246 make:security:json-login (WIP)         
                                                                                           
Opinionated code shape changes without clear best-practice consensus                       
- https://github.com/symfony/maker-bundle/pull/1637 make:webhook generate final classes                                                     
                                                                                           
Borderline (could go either way depending on whether the maintainers consider the change a
"new recommended default" rather than an option):                                          
- https://github.com/symfony/maker-bundle/pull/1750 make:command invokable + interactive 
args                                                                                       
- https://github.com/symfony/maker-bundle/pull/1744 make:validator --target option       
- https://github.com/symfony/maker-bundle/pull/1722 make:user UniqueConstraint class name  
                                                                                         
Note: #1746 (make:command invokable by default) is in scope, it changes the default rather 
than adding an option.

GromNaN:
> Within this new scope, if we have `make:command` generate invokable commands, would it still fallback to standard command if invokable commands aren't supported? I ask because these permutations can add up too.

I would choose to support only the latest LTS version (currently 7.4), to avoid putting too much pressure on upgrading PHP and the installed packages for newcomers, while making it easier to adopt the latest features in the generated code.
For the `make:command` example, that means we remove the `extends Command` version of the generated code; but we can't use [Symfony 8.1 Improved Console Input](https://symfony.com/blog/new-in-symfony-8-1-improved-console-input)

stof:
I agree with @GromNaN here. Changing the default code of a maker should happen only once that new code is available in LTS. And dropping support for the old LTS allows to simplify the maintenance (projects stuck on the old LTS can stick to the last release of the bundle supporting it).

A new maker for a whole new feature might rely on non-LTS feature (i.e. we can add new makers for new released features even if they are not yet part of the LTS), with a check for the compatible version of dependencies.

GromNaN:
# MakerBundle: existing makers status vs. new scope

This report reviews every maker currently shipped by MakerBundle against the new scope (target newcomers, generate best-practice code, minimal configuration, latest LTS only, no extensibility, no edge-case options).

Status legend:

- **in-scope** the maker fits the new scope as is, only marginal cleanup needed
- **needs-trim** in scope but options or code branches should be reduced
- **reconsider** may not fit the new scope; needs a decision
- **drop-candidate** likely to be removed or moved elsewhere

| Maker | Status | Rationale | Action to fit scope / drop |
|-------|--------|-----------|---------------------------|
| MakeCommand | in-scope | Simple template, useful for newcomers. | Switch default to invokable command (#1746). Drop the non-invokable fallback once LTS supports it. |
| MakeController | in-scope | Core scaffold. | Keep single recommended shape. Reject options that only tune style. |
| MakeForm | in-scope | Core scaffold. | Keep; align with current form best practices. |
| MakeVoter | in-scope | Small and stable. | Keep as is. |
| MakeSubscriber | in-scope | Straightforward event subscriber. | Keep as is. |
| MakeListener | in-scope | Companion to Subscriber, event autodiscovery adds value. | Keep, verify all branches still needed. |
| MakeMessage | in-scope | Generates message + handler, aligned with Messenger docs. | Keep. |
| MakeMessengerMiddleware | in-scope | Tiny, correct. | Keep. |
| MakeSchedule | in-scope | Matches current Scheduler component. | Keep. |
| MakeWebhook | in-scope | Follows Webhook component. | Consider making generated classes final by default (see #1637) as a recommended shape, not an option. |
| MakeTwigExtension | in-scope | Minimal. | Keep. |
| MakeTwigComponent | in-scope | Aligned with ux-twig-component. | Keep. |
| MakeStimulusController | in-scope | UX Stimulus scaffold, non trivial but well contained. | Keep, watch for option creep. |
| MakeSerializerEncoder | in-scope | Minimal template. | Keep. |
| MakeSerializerNormalizer | in-scope | Minimal template. | Keep. |
| MakeValidator | in-scope | Minimal. | Adopt `--target` (#1744) only if we consider it a recommended default; otherwise reject. |
| MakeUnitTest | in-scope | Minimal. | Keep. |
| MakeFunctionalTest | needs-trim | Overlaps with MakeTest. | Merge into MakeTest or delegate to it and remove the separate command. |
| MakeTest | needs-trim | Umbrella command that already covers unit / integration / functional. | Consolidate MakeUnitTest and MakeFunctionalTest here to reduce duplication. |
| MakeFixtures | in-scope | Minimal, common newcomer need. | Keep. |
| MakeDockerDatabase | reconsider | Docker Compose scaffolding is opinionated and outside the core Symfony concern. | Decide: keep as convenience for newcomers, or drop and defer to Symfony recipes and Docker Compose docs. |
| MakeMigration | in-scope | Thin wrapper over doctrine-migrations. | Keep. |
| MakeEntity | needs-trim | Largest and most complex maker. Combinatorial options are the root of the maintenance pain called out in the scope discussion. | Freeze the option surface. Reject new options (#1730, #1635, #1661, #1558). Consider extracting advanced relationship editing to a coding agent skill and keeping only the create/update happy path. |
| MakeCrud | needs-trim | Second most complex, tied to MakeEntity output. | Keep the single recommended layout, reject field-appearance options (#544) and route conflict options (#922). |
| MakeUser | in-scope | Core security setup. | Keep. Reject niche options (#1722) unless they become defaults. |
| MakeAuthenticator | needs-trim | Large legacy command superseded by Security/MakeFormLogin and Security/MakeCustomAuthenticator. | Decide whether to deprecate in favour of the two Security/* makers, or clearly document the split. |
| Security/MakeFormLogin | in-scope | Matches current recommended auth flow. | Keep. Reject #1595 (option to skip interaction). |
| Security/MakeCustomAuthenticator | in-scope | Small, focused. | Keep. Reject #1594. |
| MakeRegistrationForm | needs-trim | Very large; depends on MakeEntity + MakeUser output. | Reduce configurability. Freeze on the recommended shape from the latest LTS. Reject #1555, #1723 unless adopted as default. |
| MakeResetPassword | needs-trim | Similar profile to MakeRegistrationForm. | Same treatment. Reject #1527 (multi entity) and #798 (API Platform). |

javiereguiluz:
Jérôme, I love this proposal. I agree 100% on it. Thanks a lot for your work here and I hope others agree so we can start closing all those issues/PRs out of the scope. Thanks!

kbond:
I love this!

I think this new scope should have the following rules also:
- Exclude makers that likely aren't used by beginners (ie `make:authenticator`)
- Exclude makers that have a flex recipe that is enough (ie `make:schedule)`
- Exclude makers that realistically are only run once (ie `make:reset-password`)

Here are my opinions on the current list of makers:

* `MakeSubscriber`: isn't the best practice to always use listeners now? Maybe we should drop in favor of MakeListener?
* `MakeMessengerMiddleware`: imo this is super rare to need
* `MakeSchedule`: flex provides a default recipe, rare to need multiple schedules
* `MakeTwigExtension`: should upgrade to the attribute method?
* `MakeXTest`: I believe the idea was to consolidate into the single `make:test`
* `MakeAuthenticator`/`MakeCustomAuthenticator`: rare/advanced need
* `MakeFormLogin`: single use
* `MakeResetPassword`: single use
* `MakeRegistrationForm`: single use

README.md
```diff
@@ -5,6 +5,29 @@ The MakerBundle is the fastest way to generate the most common code you'll
...
+As a consequence:
+
+- Options and switches that exist only to cover edge cases or personal
+  preferences will not be added. The generated code always reflects our current
+  recommended defaults and may evolve between releases.
+- Requests to make the generators broadly extensible, configurable, or
+  pluggable are out of scope.
+- Only the latest Symfony LTS is supported. The generated code targets that
+  LTS. New makers for post-LTS features may still be added, with a runtime
+  check on the required dependency versions.
```
GromNaN, line 36: - See https://github.com/symfony/maker-bundle/pull/1803

GromNaN:
Thanks for the extra rules. I would refine the target audience a bit: the bundle is for developers who are new to *a given Symfony feature*, not only for absolute Symfony beginners. From that angle, "single-use per project" is not the same as "low value": that single time is exactly when the developer is learning the feature and benefits the most from a working skeleton.

Applied to the list:

* Keep `make:messenger-middleware`, `make:schedule`, `make:authenticator`, `make:security:custom`, `make:form-login`, `make:reset-password`, `make:registration-form`. All of them are entry points into an advanced feature. The Flex recipe for Scheduler, for instance, only wires the config; it does not give a task handler to learn from.
* Drop `make:subscriber` in favor of `make:listener`. Two ways to do the same thing, and `#[AsEventListener]` is the current recommended shape.
* Migrate `make:twig-extension` to the `#[AsTwigFilter]` / `#[AsTwigFunction]` attributes rather than dropping it.
* Consolidate `make:unit-test` and `make:functional-test` into `make:test` (already in the `needs-trim` row).

So the exclusion rule I would adopt is narrower: drop makers whose output is fully covered by the Flex recipe or that duplicate another maker, but keep the ones that scaffold a working example of a feature, even if a project only runs them once.

This scope is not set in stone: if a better way to scaffold projects emerges, either through richer Flex recipes or through another mechanism, we can revisit it and shift more responsibility away from MakerBundle.

I updated the README scope section accordingly.

kbond:
Missed this comment, I agree with you.

Dropping the single-use maker's is too big a leap currently as we don't have an alternative yet.

PR #1802, opened 2026-08-04
(WIP) Revamp the functional test suite

javiereguiluz:
The functional suite (176 end-to-end cases) was the main contributor pain point: 20–40 minutes per CI leg, ~45 Symfony apps built per run with test-time `composer require` calls, opaque positional answer arrays, and dozens of cases that only asserted generated files existed without ever running them.

This PR **rewrites it on a new internal harness** (`tests/Harness/`, not shipped). The shipped `src/Test/*` classes are untouched and now documented as frozen for third-party bundles.

## How

* **Fixed app profiles instead of per-test apps**: 16 pristine app templates (`base`, `doctrine`, `webapp`, `panther`, …) built once and cached in CI; each (worker, profile) app is materialized and cache-warmed once at its final path, then reset between tests in ~0.3 s. Zero network at test time (even Panther's browser driver downloads at profile build).
* **Everything executes**: an enforced per-artifact contract: every generated PHP class, Twig template, migration, test file and Stimulus controller must be declared (`covers:`) *and* observed at runtime (include recorder, tracing Twig loader, real migrations, esbuild + jsdom). ~84 formerly snapshot-only cases now really run, including full registration, reset-password and login flows.
* **Prompt-matched answers**: `->answer('Class name of the entity…', 'User')` replaces positional arrays; a new/reordered maker prompt fails loudly with a transcript instead of silently accepting defaults.
* **CI overhaul**: matrix unchanged in coverage (PHP 8.1–8.5 × SF 6.4/7.4/^8, lowest/highest, Windows), plus: makers under PHPUnit 13 on one leg, a real PostgreSQL service exercising non-sqlite isolation, and profile caching keyed on all build inputs + an ISO-week epoch.

Equivalence with the old suite is 1:1: every migrated method carries a `#[LegacyCase]` attribute; a coverage script (deleted in the final commit) proved all 176 original case IDs accounted for before `tests/Maker/` was removed.

## Real bugs found by executing generated code

* `ClassSourceManipulator` emitted a fatal `use` statement when relating two entities with the same short name (e.g. `User` ↔ `Friend\User`) **fixed** (+ the runtime-installer trait now fails loudly on composer errors instead of printing "successfully installed!").
* Empty `make:schedule` generated a fatal `->add()` with zero args: **fixed**.
* Recorded for follow-up: same-short-name ManyToMany join columns silently collapse; generated registration mailer assertions break when messenger is installed; `hashPassword()` generated for passwordless users; Panther's `APP_ENV=panther` is rejected by SF 8 skeletons.

## Notes for reviewers

* `CONTRIBUTING.md` documents the new workflow (`composer test:makers -- --filter=MakeCrud` iterates on one maker in seconds).
* The first Windows run validates the robocopy resetter + junction fallback: that path can only be exercised on real Windows CI.


src/Util/DependencyInstaller.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Bundle\MakerBundle\ConsoleStyle;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Component\Process\Process;
+
+/**
+ * Installs composer packages a maker needs at runtime. Single abstraction for
+ * every maker that auto-installs dependencies, so the behavior (and its
+ * failure mode) is defined — and testable — in exactly one place.
+ *
+ * @author Jesse Rushlow <jr@rushlow.dev>
```
GromNaN, line 23: That's very nice of you to credit another author for a new class.

tests/fixtures/make-entity/entities/attributes/Friend/User-sub-namespace.php
```diff
@@ -5,6 +5,8 @@
 use Doctrine\ORM\Mapping as ORM;
 
 #[ORM\Entity]
+// without an explicit name, both User entities would map to the same "user" table
+#[ORM\Table(name: 'friend_user')]
```
GromNaN, line 9: I didn't know we could specify the table name. This was requested in https://github.com/symfony/maker-bundle/pull/1730
Or is it a fixture to setup a test project?

GromNaN (commented):
It's great to see the CI running so quickly on this project. 

Question: Isn't there a risk of cross-contamination between builds if we change the way cached projects are generated?

GromNaN:
This PR is still titled `(WIP)`. Is it ready?
It have to be rebased since I removed support for older versions in https://github.com/symfony/maker-bundle/pull/1803.

javiereguiluz:
It's still WIP because even if it makes tests much faster and some parts of the new tests are better (like the fluent interface to define the answers when running each maker) there ar other parts that look a bit complex to me. I'd like to improve this much more.

PR #1803, opened 2026-08-05
Bump minimum Symfony version to 7.4 and PHP 8.2

GromNaN:
Raise the minimum supported Symfony version to 7.4 (LTS) and PHP to 8.2, then drop the compatibility guards that are no longer reachable.

## Changes

- `composer.json`: `php: >=8.2`, all `symfony/*` constraints tightened to `^7.4|^8.0`.
- `.github/workflows/ci-linux.yaml`: removed `6.4.*` from the matrix, updated the "lowest" job to PHP 8.2 + Symfony 7.4.
- `MakeSerializerEncoder`: dropped the `Kernel::VERSION_ID >= 70000` branch; the generated `decode()` method now always declares `: mixed`.
- `MakeMessage`: the `#[AsMessage]` attribute (available since Symfony 7.3) is always emitted for the chosen transport; the `messenger.yaml` routing fallback and the `updateMessengerConfig()` helper are removed.
- Test assertions covering the removed fallback are cleaned up.
- `tools/phpstan/includes/composer.json`: bumped `api-platform/core` to `^4.0` (3.x versions are blocked by security advisories) and aligned Symfony constraints on `^7.4|^8.1`.

Legacy code unrelated to the Symfony version bump (deprecated makers, ORM 2.x guards) is intentionally left out and can be handled in follow-up PRs.


src/Maker/MakeCommand.php
```diff
@@ -33,7 +33,7 @@
  */
 final class MakeCommand extends AbstractMaker
 {
-    public function __construct(private ?PhpCompatUtil $phpCompatUtil = null)
+    public function __construct(?PhpCompatUtil $phpCompatUtil = null)
```
GromNaN, line 36: Fix unused property.

PR #1804, opened 2026-08-17
Remove the PHP CS Fixer integration

### LINKED ISSUE #1801: Remove PHP CS Fixer integration (out of scope)
Following the scope discussion in #1800, PHP CS Fixer integration in `src/Util/TemplateLinter.php` should be removed.

Quoting @wouterj:

> Another good example is the integration of PHP CS Fixer, that looks quite complex and is way out of scope of MakerBundle imho. Devs should run their CS fixer tool manually if they don't agree with the Symfony CS used by the makers.

Rationale:

- The linter integration adds significant complexity to the bundle, including a downloader, executable resolution, and PHAR handling.
- The generated code already follows the Symfony code style. Developers who use a different style can run their own CS tool after generation.
- This aligns with the new scope: minimal configuration, no options for personal preferences, no extensibility that only serves edge cases.

Proposed action:

- Remove `TemplateLinter` and its wiring from the bundle.
- Drop related configuration and dependencies.
- Update the docs accordingly.

Target: next major.

Amoifr:
## Summary

Following the scope discussion in #1800, this removes the PHP CS Fixer integration:

- `TemplateLinter`, the bundled 3.5 MB `php-cs-fixer.phar`, the fallback config file and the `MAKER_PHP_CS_FIXER_BINARY_PATH` / `MAKER_PHP_CS_FIXER_CONFIG_PATH` env vars are gone;
- `MakerCommand` no longer post-processes the generated files (and no longer prints the linter message in verbose mode).

About @kbond's question on PHP-Parser prettyprint: generated code already looks right without the linter. The skeleton templates are hand-formatted, and the `ClassSourceManipulator` / PHP-Parser output is exactly what the fixtures in `tests/Util/fixtures/` pin down, since the linter only ran at the very end of `MakerCommand` and never in those tests. Devs who want a stricter style can run their own fixer on `src/`, as @wouterj suggested.

Fix #1801


GromNaN:
I see many formatting changes in the failing tests, is it possible to fix them in the code generator?
```
7) Symfony\Bundle\MakerBundle\Tests\Maker\MakeScheduleTest::testExecute with data set "it_generates_a_schedule_select_empty" (Symfony\Bundle\MakerBundle\Test\MakerTestDetails Object (...))
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
     {\n
         return (new Schedule())\n
             ->add(\n
-                // @TODO - Create a Message to schedule\n
-                // RecurringMessage::every('1 hour', new App\Message\Message()),\n
+            // @TODO - Create a Message to schedule\n
+            // RecurringMessage::every('1 hour', new App\Message\Message()),\n
             )\n
             ->stateful($this->cache)\n
         ;\n
```

kbond:
So to clarify, the php-cs-fixer was just *fluff*? The out of the box PHP-Parser generated code looks ok? I kind of figured it was like all one line with no indentation or something.

```php
public function getUser(int $id): ?User { if (!$id) { return null; } $user = $this->userRepository->find($id); if (!$user) { return null; } return $user; }
```

Amoifr:
Good call, and thanks for the failing list. The linter was doing two real jobs after all, both now moved into the generators:

- Template glitches the fixer used to paper over: indented `<?php` control tags in the webhook `RequestParser` (they leaked their leading whitespace into the generated code), the schedule placeholder indent, missing blank lines in the Voter template, docblock/comment indentation and a trailing comma in the custom Authenticator.
- One functional leftover: `make:serializer:normalizer` always added `use App\Entity\X;` even when the entity does not exist, and relied on `no_unused_imports` to clean it up. The use statement is now only added when the entity exists.

@kbond so mostly fluff, with those exceptions: the raw PHP-Parser output was already properly indented (the `ClassSourceManipulator` fixtures in `tests/Util/fixtures/` never went through the fixer), what it really covered was template typos and that unused import. No one-liner methods, promise.

The remaining CI red matches what already fails on the `1.x` nightlies (MakeController invoke, MakeForm relations, serializer-encoder legacy, Panther). Schedule, Voter, Webhook, Normalizer and Authenticator suites pass locally now.

templates/scheduler/Schedule.tpl.php
```diff
@@ -17,11 +17,11 @@ public function getSchedule(): Schedule
         return (new Schedule())
             ->add(
 <?php if ($has_custom_message): ?>
-            // @TODO - Modify the frequency to suite your needs
-            RecurringMessage::every('1 hour', new <?= $message_class_name; ?>()),
+                // @TODO - Modify the frequency to suite your needs
+                RecurringMessage::every('1 hour', new <?= $message_class_name; ?>()),
```
GromNaN, line 21: Should we keep php-cs-fixer running in tests, only for detecting such output issue? Ideally, we should call it programmatically instead of using the phar command-line tool, for testing performance. So in each test, we would have a list of files to validate. Or, run it just once after generating all the files.

GromNaN (commented):
Thanks for the update. It's definitely better to generate the correct format directly.

There is some Yaml formatting issues reported:
```
6) Symfony\Bundle\MakerBundle\Tests\Util\YamlSourceManipulatorTest::testSetData with data set "unix_simple_sequence_add_array_with_comments.test" ('# https://symfony.com/doc/cur...USER }', array(array(array(array('^/login$', 'IS_AUTHENTICATED_ANONYMOUSLY')))), '# https://symfony.com/doc/cur...USER }')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 '# https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
 security:
     access_control:
-        - path: ^/login$
-          roles: IS_AUTHENTICATED_ANONYMOUSLY
+        -
+            path: ^/login$
         # - { path: ^/admin, roles: ROLE_ADMIN }
         # - { path: ^/profile, roles: ROLE_USER }'
```

Amoifr:
The Yaml failures are pre-existing: the same four data sets fail on the `1.x` nightlies (e.g. run 32084125012, same job), they come from a symfony/yaml formatting change for sequences of maps. I can send a separate PR for those if you want.

On keeping php-cs-fixer in the tests: I like it as a safety net against template drift, now that nothing repairs the output at runtime. Concrete shape I would propose:

- keep the userland ruleset as a test fixture (basically the removed `config/php-cs-fixer.config.php`);
- after a maker runs in a functional test, run the fixer once in dry-run over the generated files (the harness knows them), through `tools/php-cs-fixer` which CI already installs, and fail with the diff;
- once per generated app keeps the cost to one process per test instead of one per file; going fully programmatic through the Runner API can come later if that is still too slow.

Want it in this PR or as a follow-up? Happy either way.

GromNaN:
Thank you @Amoifr.

GromNaN (approved):

PR #1805, opened 2026-08-18
Dump sequences of mappings compactly regardless of the symfony/yaml version

### LINKED ISSUE #1806: YamlSourceManipulatorTest fails since symfony/yaml restored the expanded dump of sequences of mappings
The `1.x` nightlies (and every PR) have four failing `YamlSourceManipulatorTest` data sets since August 15: `unix_simple_sequence_add_array.test`, `unix_simple_sequence_add_array_with_comments.test` and their `windows_` variants.

```diff
 security:
     access_control:
-        - path: ^/login$
-          roles: IS_AUTHENTICATED_ANONYMOUSLY
+        -
+            path: ^/login$
+            roles: IS_AUTHENTICATED_ANONYMOUSLY
```

Timeline:

1. symfony/symfony#62967 (February, symfony/yaml 8.1) made the Dumper produce the compact `- key: value` style by default for sequences of mappings;
2. ce97fb46 adapted the fixtures to that new default (with a whitespace-insensitive comparison for `symfony/yaml < 8.1`);
3. symfony/symfony#65365 (August 15, bug fix) restored the expanded default, making the compact style opt-in through `Yaml::DUMP_COMPACT_NESTED_MAPPING` — and the fixtures have been failing since.

Instead of following the default back and forth, `YamlSourceManipulator` can pass `Yaml::DUMP_COMPACT_NESTED_MAPPING` explicitly: the flag exists on every symfony/yaml version supported by the bundle, the generated YAML keeps the nicer compact style, and the output becomes deterministic across versions (which also removes the need for the version-dependent comparison in the test).

Amoifr:
Fix #1806

The `1.x` nightlies have been red since August 15: symfony/symfony#65365 restored the expanded default dump for sequences of mappings, reverting symfony/symfony#62967 that ce97fb46 had adapted the fixtures to.

Instead of following the default back and forth, `YamlSourceManipulator` now passes `Yaml::DUMP_COMPACT_NESTED_MAPPING`, which produces the compact `- key: value` style on every supported symfony/yaml version (verified against 7.4 and 8.2). The output becomes deterministic, so the version-dependent whitespace-insensitive comparison added to the test can go away too.

Spotted while working on #1804 (the four `YamlSourceManipulatorTest` failures there are these).

### DIFF
--- a/src/Util/YamlSourceManipulator.php
+++ b/src/Util/YamlSourceManipulator.php
@@ -709,7 +709,11 @@ class YamlSourceManipulator
             ? intdiv($this->indentationForDepths[$this->depth], $this->depth)
             : 4;
 
-        $newDataString = Yaml::dump($data, 4, $indent);
+        // DUMP_COMPACT_NESTED_MAPPING keeps sequences of mappings in the compact
+        // "- key: value" style on every supported symfony/yaml version: it was
+        // the default for a while (symfony/symfony#62967) before being reverted
+        // (symfony/symfony#65365), and the fixtures rely on it
+        $newDataString = Yaml::dump($data, 4, $indent, Yaml::DUMP_COMPACT_NESTED_MAPPING);
         // new line is appended: remove it
         $newDataString = rtrim($newDataString, "\n");
 
--- a/tests/Util/YamlSourceManipulatorTest.php
+++ b/tests/Util/YamlSourceManipulatorTest.php
@@ -11,14 +11,11 @@
 
 namespace Symfony\Bundle\MakerBundle\Tests\Util;
 
-use Composer\InstalledVersions;
-use Composer\Semver\VersionParser;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Psr\Log\LogLevel;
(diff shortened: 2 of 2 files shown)


Amoifr:
@GromNaN here is the fix for the Yaml failures you spotted on #1804, with the full diagnosis in #1806 (short version: symfony/symfony#65365 restored the expanded dump default that the fixtures had been adapted away from in March). Once this is merged, rebasing #1804 will clear them from its CI too.

src/Util/YamlSourceManipulator.php
```diff
@@ -55,7 +55,7 @@ public function __construct(
         }
     }
 
-    public function setLogger(LoggerInterface $logger)
+    public function setLogger(LoggerInterface $logger): void
```
GromNaN: Adding return types to a non-final, non-internal class is technically a breaking change. That should be reverted.

GromNaN (changes requested):

Amoifr:
Reverted, the CS fixer had slipped those two `: void` in while I formatted the file. The manipulator diff is back to the single `Yaml::dump()` line (plus the explanatory comment).

GromNaN (approved):

GromNaN:
Thank you @Amoifr to help maintain this bundle

PR #1807, opened 2026-08-21
[make:decorator] Add new maker to create decorator

### LINKED ISSUE #1401: Add maker for decorate services

GromNaN:
Rebase and continuation of #1613 by @WedgeSama, updated for the current `1.x` codebase (config moved from XML to PHP, help files restructured) and with the remaining review feedback from @Neirda24 applied:

- Fixed the `getRealId()` strict comparison (`in_array` with strict mode).
- Initialized the short name map entries explicitly instead of relying on implicit array creation, and replaced the array_map/array_unique callable-first-class syntax with a plain foreach.
- Fixed the choice prompt wording ("choose which one you want to decorate:").
- Replaced the `ReflectionClass` lookup of `ContainerInterface` constants with a static map for the `onInvalid` option.
- Added support for variadic parameters when parsing methods to decorate (previously produced an incorrect signature and a broken forwarding call).
- Fixed a typo in an internal exception message.

Add a `make:decorator` command that generates a decorator class for an existing service. The maker asks for the service to decorate and the class name of the decorator, then generates a class that either implements the decorated service interfaces or extends its class.

```shell
bin/console make:decorator Symfony\Component\Routing\Generator\UrlGeneratorInterface MyUrlGenerator
```

```php
<?php

namespace App;

use Symfony\Component\DependencyInjection\Attribute\AsDecorator;
use Symfony\Component\DependencyInjection\Attribute\AutowireDecorated;
use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
use Symfony\Component\Routing\RequestContext;

#[AsDecorator(UrlGeneratorInterface::class)]
final class MyUrlGenerator implements UrlGeneratorInterface
{
    public function __construct(
        #[AutowireDecorated]
        private readonly UrlGeneratorInterface $inner,
    ) {
    }

    public function generate(string $name, array $parameters = [], int $referenceType = self::ABSOLUTE_PATH): string
    {
        return $this->inner->generate($name, $parameters, $referenceType);
    }

    public function setContext(RequestContext $context): void
    {
        $this->inner->setContext($context);
    }

    public function getContext(): RequestContext
    {
        return $this->inner->getContext();
    }
}
```

Closes #1401

### DIFF
--- a/config/help/MakeDecorator.txt
+++ b/config/help/MakeDecorator.txt
@@ -1,0 +1,10 @@
+The <info>%command.name%</info> command generates a new service decorator class.
+
+<info>php %command.full_name%</info>
+<info>php %command.full_name% My\Decorated\Service\Class</info>
+<info>php %command.full_name% My\Decorated\Service\Class MyServiceDecorator</info>
+<info>php %command.full_name% My\Decorated\Service\Class Service\MyServiceDecorator</info>
+<info>php %command.full_name% my_decorated.service.id MyServiceDecorator</info>
+<info>php %command.full_name% my_decorated.service.id MyServiceDecorator</info>
+
+If one argument is missing, the command will ask for it interactively.
--- a/config/makers.php
+++ b/config/makers.php
@@ -15,6 +15,7 @@ use Symfony\Bundle\MakerBundle\Maker\MakeAuthenticator;
 use Symfony\Bundle\MakerBundle\Maker\MakeCommand;
 use Symfony\Bundle\MakerBundle\Maker\MakeController;
 use Symfony\Bundle\MakerBundle\Maker\MakeCrud;
+use Symfony\Bundle\MakerBundle\Maker\MakeDecorator;
 use Symfony\Bundle\MakerBundle\Maker\MakeDockerDatabase;
 use Symfony\Bundle\MakerBundle\Maker\MakeEntity;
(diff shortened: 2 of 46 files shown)


GromNaN:
Thank you @WedgeSama for the initial work and @Neirda24 for the review.

PR #1808, opened 2026-08-23
[make:constant] add new command to add a constant to an existing class

karozadev:
Adds a `make:constant` command that inserts a class constant into any existing PHP class under src/ (entity, service, controller, DTO, ...) using the AST manipulator, preserving the file's existing formatting.

- Interactive prompts for class (with autocompletion), name and value when not passed as arguments.
- --type/-t, --visibility, --comment/-c and --final options.
- Type is guessed from the value when --type is not provided.
- ClassSourceManipulator gains addConstant()/constantExists() and places the new constant after existing constants, otherwise before the first property, otherwise after traits, otherwise at the top.
- Validator gains validateConstantName() (UPPER_SNAKE_CASE).

### DIFF
--- a/config/help/MakeConstant.txt
+++ b/config/help/MakeConstant.txt
@@ -1,0 +1,23 @@
+The <info>%command.name%</info> command adds a constant to an existing PHP class in your <comment>src/</comment>
+directory (an entity, a service, a controller, a DTO, ...) using PHP's AST parser, so your
+existing code formatting is preserved.
+
+<info>php %command.full_name% Order STATUS_PENDING pending</info>
+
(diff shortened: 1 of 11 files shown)


GromNaN:
Thanks for the contribution! Adding a constant to a class is a simple PHP code manipulation that does not justify a dedicated maker command. The maker bundle focuses on leveraging Symfony features to generate boilerplate code, see the [Scope section](https://github.com/symfony/maker-bundle#scope) of the README. Closing this PR, but feel free to open new proposals for makers that generate Symfony-related boilerplate.

PR #1809, opened 2026-08-23
[make:entity] generate an association entity for ManyToMany relations…

karozadev:
… with additional properties

When adding a ManyToMany relationship, ask whether it needs additional properties (e.g. a "role" or "joinedAt" field). If so, generate a dedicated association entity (e.g. User + Group -> UserGroup) instead of a direct ManyToMany mapping, following the Doctrine-recommended pattern:

- the association entity gets a simple auto-generated id, a ManyToOne to each side of the relationship, and a unique constraint on the pair;
- both original entities get the OneToMany inverse side automatically;
- additional properties are collected by reusing the existing field-adding loop, so nested relations on the association entity work for free;
- re-running the command against an already-generated association entity reuses it (existing properties/relations/methods are preserved, and the unique constraint is not duplicated);
- self-referencing relations and vendor targets are not offered this workflow and keep the existing ManyToMany behavior unchanged.

Also fixes a latent bug where the loop dumping generated files reused the $manipulator variable name, corrupting it for any relation added afterwards in the same make:entity run.

### DIFF
--- a/config/help/MakeEntity.txt
+++ b/config/help/MakeEntity.txt
@@ -22,3 +22,8 @@ for the properties of existing entities:
 You can also *overwrite* any existing methods:
 
 <info>php %command.full_name% --regenerate --overwrite</info>
+
+When adding a ManyToMany relationship, you'll be asked whether it needs additional
+properties (e.g. a "role" or "joinedAt" field). If so, an association entity is
+generated automatically to hold the relationship and its extra properties, instead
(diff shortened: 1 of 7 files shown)


karozadev:
Hi @GromNaN , gentle ping on this PR when you have a moment. Let me know if you need any clarification or if I should break it down into smaller parts!

PR #1810, opened 2026-08-28
[make:entity] Add a --field option to add fields without prompting

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

`make:entity` is the only maker with no non-interactive path. The entity name can be passed
as an argument, but every field is asked for in a `while (true)` loop, so the command can't
be scripted, can't run in a non-interactive shell, and can't be used by anything that isn't
a person at a terminal.

This adds a repeatable `--field` option:

```console
$ php bin/console make:entity BlogPost \
    --field=title:string:100 \
    --field=body:text \
    --field=price:decimal:10:2 \
    --field=publishedAt \
    --field=status:enum:Status \
    --no-interaction
```

The format is `name[:type[:length]]` with a trailing `?` for nullable. Each option adds one
field, exactly as one pass through the interactive loop does.

A few details worth pointing out:

- **The type is optional.** When it's left out, the same guess the interactive mode makes
  applies, so `--field=publishedAt` yields a `datetime_immutable` and `--field=isActive` a
  `boolean`. That heuristic was inline in `askForNextField()`; it's now a small private
  method both paths call, which is the only change to existing behaviour.
- **Enums take their class as the third part**, and a short name is enough while it is
  unique within `src/` — `--field=status:enum:Status` — which spares the caller from
  escaping backslashes in the shell. `--field=labels:enum:Label,multiple` stores several
  values. One subtlety worth flagging for review: the interactive mode settles the column
  type *inside* its enum branch, which runs after the branch that assigns a length, so an
  enum never carries a `length`. The parser mirrors that rather than falling through the
  `string` branch, and a test asserts it.
- **Relations are rejected** with a message pointing back to the interactive mode. They
  need follow-up questions (target entity, direction, inverse side, orphan removal) that
  don't compress into a single string without inventing a syntax nobody asked for. Scoping
  them out keeps this addition reviewable; a follow-up PR adds a `--relation` option.
- **Definitions are parsed before anything is written**, so an invalid one fails without
  leaving a half-created entity behind. That includes a definition colliding with the `id`
  the entity is about to be generated with.
- **Nullable is only set when asked for**, matching the interactive path — otherwise the
  generated attribute picks up a redundant `nullable: false`.
- **Errors stay errors.** `Validator::validateDoctrineFieldName()` throws an
  `\InvalidArgumentException`, which the interactive question validator catches to re-ask.
  With no one to ask, it is wrapped in a `RuntimeCommandException` so a reserved word or an
  invalid property name prints a clean error box instead of a stack trace.

Tests cover the happy path (including the type guess and the absence of `nullable: false`),
enum fields resolved both by short name and by full class name, and the rejected
definitions — relation, missing or unknown enum class, unknown type, `id` collision,
invalid length, a surplus segment, an unknown modifier and an invalid property name.

### Why

Coding agents are the immediate motivation. An agent asked to add an entity today either
drives an interactive prompt it can't see, or gives up and hand-writes the class — which is
how projects end up with entities that don't look like the ones `make:entity` produces. The
same applies to any script, Makefile or fixture setup that wants to scaffold without a
human present.

### DIFF
--- a/config/help/MakeEntity.txt
+++ b/config/help/MakeEntity.txt
@@ -22,3 +22,22 @@ for the properties of existing entities:
 You can also *overwrite* any existing methods:
 
 <info>php %command.full_name% --regenerate --overwrite</info>
+
+Fields can be passed on the command line instead of being asked for, which is
+what you want in a script or when the terminal is not interactive:
+
+<info>php %command.full_name% BlogPost --field=title:string:100 --field=body:text --field=publishedAt:datetime_immutable?</info>
+
+The format is <comment>name[:type[:length]]</comment>, with a trailing <comment>?</comment> to make the field nullable.
+The type defaults to the same guess the interactive mode makes, so <comment>--field=createdAt</comment>
+becomes a <comment>datetime_immutable</comment>. For <comment>decimal</comment>, the third and fourth parts are the
+precision and the scale: <comment>--field=price:decimal:10:2</comment>.
+
+For <comment>enum</comment>, the third part is the backed enum class. Its short name is enough as
+long as it is unique within <comment>src/</comment>, which saves escaping backslashes in your shell.
+Append <comment>,multiple</comment> to store several values:
+
+<info>php %command.full_name% BlogPost --field=status:enum:Status --field=labels:enum:Label,multiple</info>
+
+Relations are not supported by this option - run the command without <comment>--field</comment>
+to be guided through them.
--- a/src/Maker/MakeEntity.php
+++ b/src/Maker/MakeEntity.php
@@ -98,6 +98,7 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
             ->addOption('broadcast', 'b', InputOption::VALUE_NONE, 'Add the ability to broadcast entity updates using Symfony UX Turbo?')
             ->addOption('regenerate', null, InputOption::VALUE_NONE, 'Instead of adding new fields, simply generate the methods (e.g. getter/setter) for existing fields')
             ->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite any existing getter/setter methods')
+            ->addOption('field', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Add a field without being asked for it: <fg=yellow>name[:type[:length]][?]</> (e.g. <fg=yellow>title:string:100</>). Repeat the option for each field')
             ->setHelp($this->getHelpFileContents('MakeEntity.txt'))
         ;
 
@@ -173,9 +174,14 @@ final class MakeEntity extends AbstractMaker implements InputAwareMakerInterface
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator): void
     {
         $overwrite = $input->getOption('overwrite');
+        $fieldOptions = $input->getOption('field');
 
         // the regenerate option has entirely custom behavior
         if ($input->getOption('regenerate')) {
+            if ($fieldOptions) {
+                throw new RuntimeCommandException('The "--field" option cannot be combined with "--regenerate", which only generates methods for existing fields.');
+            }
+
             $this->regenerateEntities($input->getArgument('name'), $overwrite, $generator);
             $this->writeSuccessMessage($io);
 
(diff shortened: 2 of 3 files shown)


src/Maker/MakeEntity.php
```diff
@@ -173,9 +174,14 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
     public function generate(InputInterface $input, ConsoleStyle $io, Generator $generator): void
     {
         $overwrite = $input->getOption('overwrite');
+        $fieldOptions = $input->getOption('field');
 
         // the regenerate option has entirely custom behavior
         if ($input->getOption('regenerate')) {
+            if ([] !== $fieldOptions) {
```
GromNaN: Nit: Our CS is to use falsy/truthy values in conditions.
```suggestion
            if ($fieldOptions) {
```

src/Maker/MakeEntity.php
```diff
@@ -336,6 +351,191 @@ public function configureDependencies(DependencyBuilder $dependencies, ?InputInt
...
+        $head = array_shift($modifiers);
+
+        if ($nullable = str_ends_with($head, '?')) {
+            $head = substr($head, 0, -1);
+        }
+
+        $parsedModifiers = [];
+
+        foreach ($modifiers as $modifier) {
+            if ('' === $modifier) {
```
GromNaN:
```suggestion
            if (!$modifier) {
```

src/Maker/MakeEntity.php
```diff
@@ -336,6 +351,191 @@ public function configureDependencies(DependencyBuilder $dependencies, ?InputInt
...
+            null => false,
+            true, 'true' => true,
+            'false' => false,
+            default => throw new RuntimeCommandException(\sprintf('The "%s" modifier of the field "%s" takes no value, or "true" or "false".', $name, $definition)),
+        };
+    }
+
+    private function resolveEnumClass(string $enumClass): string
+    {
+        if ('' === $enumClass) {
```
GromNaN:
```suggestion
        if (!$enumClass) {
```

GromNaN (changes requested):
Can you fix the conditions please.
 
Examples: https://github.com/symfony/symfony/pull/65327

wachterjohannes:
Done. You flagged three, there were eight in this PR and four more in #1813, so I converted all of them to keep the two consistent.

GromNaN (approved):

config/help/MakeEntity.txt
```diff
@@ -22,3 +22,22 @@ for the properties of existing entities:
...
+
+Fields can be passed on the command line instead of being asked for, which is
+what you want in a script or when the terminal is not interactive:
+
+<info>php %command.full_name% BlogPost --field=title:string:100 --field=body:text --field=publishedAt:datetime_immutable?</info>
+
+The format is <comment>name[:type[:length]]</comment>, with a trailing <comment>?</comment> to make the field nullable.
+The type defaults to the same guess the interactive mode makes, so <comment>--field=createdAt</comment>
+becomes a <comment>datetime_immutable</comment>. For <comment>decimal</comment>, the third and fourth parts are the
+precision and the scale: <comment>--field=price:decimal:10:2</comment>.
```
GromNaN, line 34: I’d like a third opinion on this choice of syntax for field types (which reminds me of Eloquent, but that’s incidental).
/cc @kbond

kbond: You mean like property hooks vs getters/setters?

GromNaN: I mean, the option syntax `--field name[:type[:length]]`. This is a syntax that only exists here, using colons as separators.

kbond: Got it. I don't see a problem, not really something a user would use directly, right?

wachterjohannes: @kbond no not really for users the interactive way is the more intuitive one, for agents the parameter is the way to go

src/Maker/MakeEntity.php
```diff
@@ -336,6 +351,191 @@ public function configureDependencies(DependencyBuilder $dependencies, ?InputInt
...
+        if ('_id' === $suffix) {
+            return 'integer';
+        }
+
+        if (str_starts_with($snakeCasedField, 'is_') || str_starts_with($snakeCasedField, 'has_')) {
+            return 'boolean';
+        }
+
+        if ('uuid' === $snakeCasedField) {
+            return Type::hasType('uuid') ? 'uuid' : 'guid';
```
GromNaN: The [UuiType](https://github.com/symfony/symfony/commits/8.2/src/Symfony/Bridge/Doctrine/Types/UuidType.php) is always available. I removed this condition.
```suggestion
            return 'uuid';
```

GromNaN:
Thank you @wachterjohannes.

PR #1811, opened 2026-08-28
Fix the failing tests on 1.x

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

The `CI Linux` workflow has been red on `1.x` for a while — three of its four jobs fail,
for three unrelated reasons. None of them is a bug in a maker; all three are the test
suite (or its dependency constraints) not having caught up with Symfony 8.

### `MakeFormTest`, three relation cases

These copy a fixture test into the generated project and run PHPUnit *in* it. The inner
run reports `OK (1 test, 4 assertions)` and still exits non-zero:

```
Remaining self deprecation notices (1)

  1x: The "App\Tests\GeneratedFormTest::getExtensions()" method will require a new
      "ViolationMapperInterface|null $violationMapper" argument in the next major version
      of its parent class "Symfony\Component\Form\Test\TypeTestCase", not defining it is
      deprecated.
```

`TypeTestCase::getExtensions()` announces the argument with a commented-out signature and
a `@param`, `DebugClassLoader` turns the missing argument into a *self* deprecation, and
`SYMFONY_DEPRECATIONS_HELPER=max[self]=0` makes it fatal. The three fixtures now declare
the argument and forward it. Adding an optional parameter to an override is legal against
the 7.4 parent too, and passing it on to a parent that does not take it is a no-op, so
this stays compatible with both branches.

### `MakeTestTest::it_makes_PantherTestCase_type`

```
Failed asserting that 'Symfony Exception' contains "Hello World".
```

The page Panther actually got is:

> The environment "panther" is not registered as allowed by `App\Kernel::getAllowedEnvs()`.

Panther serves the app under `PANTHER_APP_ENV`, which its recipe sets to `panther`, and
the Symfony 8 skeleton kernel now rejects any environment missing from `getAllowedEnvs()`.
The test adds `panther` to that list, which is what an application would have to do. The
replacement is `allowNotFound`, since the 7.4 skeleton has no such method.

### `PHP 8.2 + @7.4.* lowest deps`

This job dies with a compile error partway through the suite, taking the whole process
down with exit code 255:

```
Declaration of Symfony\Bridge\Doctrine\CacheWarmer\ProxyCacheWarmer::warmUp(string $cacheDir): array
must be compatible with WarmableInterface::warmUp(string $cacheDir, ?string $buildDir = null): array
```

`symfony/doctrine-bridge` is only pulled in transitively through DoctrineBundle, so nothing
constrains it and `--prefer-lowest` resolves it to `6.4.0` — predating the `$buildDir`
argument `WarmableInterface` has carried since 7.1 — while `symfony/http-kernel` is held at
`7.4` by this package's own constraint. Requiring it in the same range as the other Symfony
components makes `--prefer-lowest` pick `7.4.0`:

```
- Downgrading symfony/doctrine-bridge (8.2.x-dev => v6.4.0)   # before
- Downgrading symfony/doctrine-bridge (8.2.x-dev => v7.4.0)   # after
```

### Verification

Each fix was reproduced locally before and after, against both skeletons
(`SYMFONY_VERSION=8.*` and `SYMFONY_VERSION=7.4.*`):

- the form fixtures: inner PHPUnit exit code `1` → `0`, deprecation gone
- Panther: reproduced the error page, then green on 7.4 and 8
- lowest deps: `composer update --prefer-lowest` now resolves a compatible pair, and
  `ProxyCacheWarmer::warmUp()` and `WarmableInterface::warmUp()` agree by reflection

### DIFF
--- a/tests/Maker/MakeTestTest.php
+++ b/tests/Maker/MakeTestTest.php
@@ -88,6 +88,15 @@ class MakeTestTest extends MakerTestCase
                     ''
                 );
 
+                // Panther serves the app under PANTHER_APP_ENV, which its recipe sets to "panther",
+                // and the Symfony 8 skeleton refuses any environment its kernel does not list
+                $runner->replaceInFile(
+                    'src/Kernel.php',
+                    "return ['prod', 'dev', 'test'];",
+                    "return ['prod', 'dev', 'test', 'panther'];",
+                    allowNotFound: true
+                );
+
                 $runner->runMaker(
                     [
                         // type
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "doctrine/doctrine-bundle": "^2.10|^3.0",
         "doctrine/orm": "^2.15|^3",
         "doctrine/persistence": "^3.1|^4.0",
+        "symfony/doctrine-bridge": "^7.4|^8.1",
         "symfony/http-client": "^7.4|^8.1",
         "symfony/phpunit-bridge": "^7.4|^8.1",
         "symfony/security-core": "^7.4|^8.1",
--- a/tests/fixtures/make-form/tests/it_generates_form_with_many_to_many_relation.php
+++ b/tests/fixtures/make-form/tests/it_generates_form_with_many_to_many_relation.php
@@ -14,6 +14,7 @@ use Doctrine\ORM\QueryBuilder;
 use Doctrine\Persistence\ManagerRegistry;
 use PHPUnit\Framework\Attributes\DataProvider;
 use Symfony\Bridge\Doctrine\Form\DoctrineOrmExtension;
+use Symfony\Component\Form\Extension\Validator\ViolationMapper\ViolationMapperInterface;
 use Symfony\Component\Form\Test\TypeTestCase;
 
 class GeneratedFormTest extends TypeTestCase
@@ -74,7 +75,7 @@ class GeneratedFormTest extends TypeTestCase
         ];
     }
 
-    protected function getExtensions(): array
+    protected function getExtensions(?ViolationMapperInterface $violationMapper = null): array
     {
         $mockEntityManager = $this->createMock(EntityManager::class);
         $mockEntityManager->method('getClassMetadata')
@@ -107,6 +108,6 @@ class GeneratedFormTest extends TypeTestCase
         $mockRegistry->method('getManagers')
             ->willReturn([$mockEntityManager]);
 
-        return array_merge(parent::getExtensions(), [new DoctrineOrmExtension($mockRegistry)]);
+        return array_merge(parent::getExtensions($violationMapper), [new DoctrineOrmExtension($mockRegistry)]);
     }
 }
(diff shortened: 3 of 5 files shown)


tests/Maker/MakeTestTest.php
```diff
@@ -88,6 +88,15 @@ public static function getTestDetails(): \Generator
                     ''
                 );
 
+                // Panther serves the app under PANTHER_APP_ENV, which its recipe sets to "panther",
```
GromNaN, line 91: I approve this change as hotfix.

Is it something that flex users have to fix themselve? If some, can we fix it for all the users?

GromNaN (approved):
Thank you @wachterjohannes. I really appreciate you taking the time to resolve these long-standing issues.

wachterjohannes:
**Disclaimer up front:** I don't know Panther. I have never used it. This fix, the PR description, and everything below were produced entirely by an AI agent (Claude Code) working in my checkout. I have not independently verified the Panther-specific reasoning, so please read it with that in mind and push back where it's wrong.

With that said, here is what the agent found, and it did reproduce each step locally against both a Symfony 7.4 and a Symfony 8 skeleton.

**What happens.** Unlike `WebTestCase`, Panther starts a real web server (`php -S`) in a separate process and drives a real browser against it. That server process needs an `APP_ENV`, and the `symfony/panther` recipe gives it a dedicated one by writing `PANTHER_APP_ENV=panther` into `.env.test`. The Symfony 8 skeleton meanwhile generates a `getAllowedEnvs()` into the user's own `src/Kernel.php` returning `['prod', 'dev', 'test']`, and `KernelTrait` throws for anything else:

```
The environment "panther" is not registered as allowed by "App\Kernel::getAllowedEnvs()".
```

So the server returns Symfony's error page and the assertion sees `Symfony Exception` instead of the page content. Symfony 7.4 has no such list, which is why only the `@^8` jobs failed.

**Do Flex users have to fix it themselves?** As things stand, yes: anyone installing `symfony/panther` into a Symfony 8 skeleton app should hit this on their first Panther test. The fix is one line in their own `src/Kernel.php`: add `'panther'` to `getAllowedEnvs()`, or return `[]`, which disables the check entirely since it only applies when the list is non-empty. That is exactly what this test now does, rather than working around the symptom.

**Can we fix it for all of them?** Not from MakerBundle: `getAllowedEnvs()` lives in the user's own `src/Kernel.php` by design, and a recipe cannot rewrite an existing PHP file. The durable fix would belong in `symfony/recipes`. Worth noting that Panther itself looks fine here: when `PANTHER_APP_ENV` is unset, `WebServerManager` passes no env and the server simply inherits `APP_ENV=test` from the test run. It is only the recipe that introduces the extra environment, so changing it to `PANTHER_APP_ENV=test` would fix every new installation, at the cost of the dedicated `panther` env it deliberately creates. Otherwise it needs a note in the Panther docs for Symfony 8.

This is not Panther-specific either: any package whose recipe introduces its own `APP_ENV` runs into the same wall on Symfony 8. Happy to open an issue on `symfony/recipes` if you think that is the right place.

PR #1813, opened 2026-08-28
[make:entity] Add a --relation option to add relations without prompting

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

> [!NOTE]
> Built on #1810 (`--field`) and #1812 (a relation-writing bug that makes this unusable
> otherwise). Both appear in the diff until they are merged — review them first.

#1810 gave `make:entity` a non-interactive path for columns. Relations still need a person
at a prompt: `askRelationDetails()` asks for the target entity, the direction, nullability,
whether to map the inverse side and under which name, and orphan removal.

```console
$ php bin/console make:entity BlogPost \
    --relation=author:ManyToOne:User \
    --relation=comments:OneToMany:Comment \
    --relation=tags:ManyToMany:Tag \
    --no-interaction
```

The format is `name:type:target`, repeatable. The target may be a short name — it is
resolved against the entity namespace first, exactly as the prompt does — which spares
escaping backslashes in a shell. A trailing `?` makes the relation nullable, and two
modifiers cover the rest:

| Modifier | Meaning | Applies to | Default |
| --- | --- | --- | --- |
| `target-field=<prop>` | name of the property added to the **other** class | all | named after this entity |
| `target-field=none` | leave the other class alone | MTO, MTM, OTO | — |
| `orphan-removal` | activate orphanRemoval | MTO, OTM | off |

```console
--relation=author:ManyToOne:User?,target-field=posts
--relation=cover:OneToOne:Media,target-field=owner
--relation=comments:OneToMany:Comment,orphan-removal
```

### Decisions worth reviewing

- **Nullability is the one deliberate difference from the prompt.** `SymfonyStyle::confirm()`
  defaults to `true`, so the interactive mode preselects *nullable* for a relation while
  `--field` asks with `false` for a column. On the command line both follow the same rule
  instead: no `?` means NOT NULL. A prompt default is a suggestion, not a semantic — there
  is no single "what a human would have produced", since that depends on whether they hit
  enter. One rule that reads the same for a column and a relation seemed worth more, and it
  is called out in the help text.
- **`target-field`, not `inverse`.** For a `OneToMany` the property created on the target is
  the *owning* side, so calling it `inverse` would be wrong Doctrine terminology.
  `target-field` describes the same thing correctly for all four types.
- **A `OneToOne` does not map its inverse side by default**, following the interactive
  recommendation — that one encodes a real trade-off, since Doctrine cannot lazy load an
  inverse `OneToOne`.
- **Errors instead of silent no-ops.** The prompt simply does not ask about orphan removal
  when it cannot apply, and silently forces the inverse side off for a class in `vendor/`.
  In a script both would be traps, so they are rejected: an unknown target, a nullable
  `ManyToMany`, `orphan-removal` on a relation that cannot carry it, a `target-field` on a
  vendor class, and two relations whose properties on a shared target would collide
  (`createdBy` and `updatedBy` both defaulting to `blogPosts` on `User`).

### One thing the interactive mode gets away with

Resolving the target has to cope with the entity not existing yet, so a self reference is
matched against the class about to be generated instead of being looked up.

The same problem bit `generate()` one level down. It resolves the class on the other side
of a relation through reflection, which fails for an entity created moments earlier: the
`class_exists()` at the top of `generate()` poisons Composer's `missingClasses` cache before
the file is written. The interactive mode never notices, because asking for the target
entity calls `getEntitiesForAutocomplete()` → `getMetadata()`, which loads every entity
class as a side effect. Nothing does that when the relation comes from an option.

That lookup is discarded for a `OneToMany` anyway — the branch below recomputes both the
filename and the manipulator from the owning class — so it is now skipped when the class on
the other side is the entity being edited.

### Tests

One case generates all four relation types plus a self reference in a single run, asserts
the attributes on *both* classes, and finishes with `doctrine:schema:update` so Doctrine
itself validates every mapping. A second walks the rejected definitions listed above.

### DIFF
--- a/config/help/MakeEntity.txt
+++ b/config/help/MakeEntity.txt
@@ -39,5 +39,21 @@ Append <comment>,multiple</comment> to store several values:
 
 <info>php %command.full_name% BlogPost --field=status:enum:Status --field=labels:enum:Label,multiple</info>
 
-Relations are not supported by this option - run the command without <comment>--field</comment>
-to be guided through them.
+Relations use their own option, <comment>name:type:target</comment>, where the target is an entity
+class - its short name is enough:
+
+<info>php %command.full_name% BlogPost --relation=author:ManyToOne:User --relation=tags:ManyToMany:Tag</info>
+
+A trailing <comment>?</comment> makes the relation nullable here too. Note that this differs from the
+interactive mode, which offers <comment>nullable</comment> as its default answer: without a <comment>?</comment> the
+relation is NOT NULL.
+
+By default the property added to the other class is named after this entity, and a
+<comment>OneToOne</comment> does not add one at all - the interactive mode advises against it, because
+Doctrine cannot lazy load an inverse <comment>OneToOne</comment>. Both are controlled with
+<comment>target-field</comment>, and <comment>orphan-removal</comment> activates orphanRemoval on the relations that
+support it:
+
+<info>php %command.full_name% BlogPost --relation=cover:OneToOne:Media,target-field=post</info>
+<info>php %command.full_name% BlogPost --relation=author:ManyToOne:User,target-field=none</info>
+<info>php %command.full_name% BlogPost --relation=comments:OneToMany:Comment,orphan-removal</info>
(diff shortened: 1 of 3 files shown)


wachterjohannes:
@GromNaN rebased :)

GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1814, opened 2026-08-28
[make:crud] Fix the crash under --no-interaction and add --controller-class

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| License       | MIT

`make:crud` takes its entity as an argument and marks it non-interactive, so it looks
scriptable. It is not: it crashes before generating anything.

```console
$ php bin/console make:crud SweetFood --no-interaction

 [WARNING] "make:crud" is not meant to be run in non-interactive mode.

In MakeCrud.php line 135:
  Typed property Symfony\Bundle\MakerBundle\Maker\MakeCrud::$controllerClassName
```

### Cause

The controller name is the command's only other input, and it was asked for in `interact()`
and stored on the maker:

```php
private string $controllerClassName;   // typed, no default

// interact()
$this->controllerClassName = $io->ask(...);

// generate(), twice
$generator->createClassNameDetails($this->controllerClassName, ...)
```

`Command::run()` skips `interact()` under `--no-interaction`, so the property is never
assigned and the first read in `generate()` throws. There is no way to supply the value from
the command line, so the command simply cannot run without a human at the terminal.

### Fix

The name now travels on the input instead of on the maker, which leaves one source of truth:

- `--controller-class` carries it.
- `interact()` writes the answer into that option rather than into a property, and skips the
  question when the option was already passed.
- `generate()` reads the option and falls back to the same default the prompt offers,
  derived by the one method both paths now call, so the two cannot drift.
- The property is gone.

```console
$ php bin/console make:crud SweetFood --controller-class=SweetFoodAdminController --no-interaction
$ php bin/console make:crud SweetFood --no-interaction   # SweetFoodController
```

Nothing changes for anyone typing at a terminal: the same question, the same default.

I marked this as both a bug fix and a feature. The crash is the bug; `--controller-class` is
the smallest thing that makes the command usable rather than merely non-crashing, since
without it the controller name could never be chosen non-interactively.

### One thing left alone

`--with-tests` has the same shape of problem. `CanGenerateTestsTrait::interactSetGenerateTests()`
is the only place that reads the option, and it is called from `interact()`, so under
`--no-interaction` the flag is silently ignored rather than generating tests. The trait is
shared by five makers (`crud`, `controller`, `reset-password`, `registration-form`,
`form-login`), so fixing it belongs in its own PR. Happy to open one.

### Tests

Two cases in `MakeCrudTest`, both non-interactive: one passing `--controller-class` and
asserting the controller lands at the requested class, one without it asserting the derived
default is used. Both fail on `1.x` with the initialisation error above. The full
`MakeCrudTest` passes.

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -33,6 +33,7 @@ use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\HttpFoundation\Request;
@@ -49,7 +50,6 @@ final class MakeCrud extends AbstractMaker
     use CanGenerateTestsTrait;
 
     private Inflector $inflector;
-    private string $controllerClassName;
     private bool $generateTests = false;
 
     public function __construct(private DoctrineHelper $doctrineHelper, private FormTypeRenderer $formTypeRenderer)
@@ -71,6 +71,7 @@ final class MakeCrud extends AbstractMaker
     {
         $command
             ->addArgument('entity-class', InputArgument::OPTIONAL, \sprintf('The class name of the entity to create CRUD (e.g. <fg=yellow>%s</>)', Str::asClassName(Str::getRandomTerm())))
+            ->addOption('controller-class', null, InputOption::VALUE_REQUIRED, 'The class name of the controller to create (e.g. <fg=yellow>SweetFoodAdminController</>), defaults to the entity name plus "Controller"')
             ->setHelp($this->getHelpFileContents('MakeCrud.txt'))
         ;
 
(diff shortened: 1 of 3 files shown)


src/Maker/MakeCrud.php
```diff
@@ -110,6 +113,10 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
             'Entity\\'
         );
 
+        // interact() is skipped under --no-interaction, so the option is the only place the
```
GromNaN: This code comment isn't necessary to understand the code. It explains the fixed but, which is more a PR/commit description than a code comment.

GromNaN (approved):
SGTM.
I believe the generated code comment should be removed.

src/Maker/MakeCrud.php
wachterjohannes: Removed, thanks. You're right that it described the bug rather than the code, and it is already in the commit message where it belongs.

The same criticism applies to a few comments in my other open PRs, so I will go through those with this in mind rather than leave it at the one you pointed at.

wachterjohannes:
@GromNaN done, and I applied the same reading to my other open PRs rather than only the line you pointed at.

Removed there: a comment in #1810 and #1813 explaining that parsing happens up front so a bad definition cannot leave a half-created entity behind, and a docblock in #1815 explaining why the answer moved onto the input. Both described the fix rather than the code, same as this one.

Trimmed rather than removed in #1813: a comment above a branch in `generate()` had grown into six lines of archaeology about Composer's `missingClasses` cache. It is down to the two lines that say what the branch does.

Kept the ones that carry something the code cannot say on its own, for example that the interactive mode settles an enum's column type only after the branch that assigns a length, so an enum never gets one. Happy to cut further if you read any of those the same way.

GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1815, opened 2026-08-28
[make:controller] Honour --with-tests under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

`--with-tests` is accepted by five makers and silently ignored by all of them when the command runs non-interactively. `make:controller` additionally crashes before it gets that far. Both have the same cause: state that only `interact()` ever produces, read back later from the maker.

### `--with-tests` is dropped

`CanGenerateTestsTrait` read the option in `interactSetGenerateTests()`, stored the answer on the maker, and handed it back through `shouldGenerateTests()`:

```php
private bool $generateTests = false;

public function interactSetGenerateTests(InputInterface $input, ConsoleStyle $io): void
{
    $this->generateTests = $input->getOption('with-tests');

    if (!$this->generateTests) {
        $this->generateTests = $io->confirm('Do you want to generate PHPUnit tests? [Experimental]', false);
    }
}
```

`Command::run()` skips `interact()` under `--no-interaction`, so the property keeps its `false` default and the flag never reaches `generate()`. No error, no tests.

The answer now lives on the input instead: `interact()` writes the confirmation into the option rather than into a property, and `shouldGenerateTests(InputInterface $input)` reads the option, so both paths end up in the same place. The property is gone, along with the copy of it `MakeCrud` carried alongside the trait's.

### `make:controller` never reached it

```console
$ php bin/console make:controller FooBar --with-tests --no-interaction

 [WARNING] "make:controller" is not meant to be run in non-interactive mode.

  Typed property Symfony\Bundle\MakerBundle\Maker\MakeController::$controllerClassData
```

Its `interact()` asked nothing at all. It only derived the class data, the invokable flag and the Twig template path from the argument and the options, and `generate()` read them back off the maker. That derivation is now in `generate()`, where its inputs already are, and the four properties are gone. Nothing changes for anyone typing at a terminal, because none of it was ever a question.

### Deliberately left alone

The three remaining makers using the trait (`registration-form`, `reset-password`, `form-login`) still cannot run non-interactively. They ask for a user class, field names, email addresses and a firewall, and keep the answers on the maker in the same way. Giving them a non-interactive path is a feature rather than a fix, so it belongs in its own PR.

`make:crud` has the same crash, fixed separately in #1814. The two do not depend on each other.

### Note on the signature

`shouldGenerateTests()` gained an `InputInterface` argument. The trait is `@internal` and all five call sites are in this repository, each already inside `generate()` with the input in scope.

### Tests

`MakeControllerTest` gains one case running `FooBar --with-tests --no-interaction` and asserting the test class is generated. It fails on `1.x` with the initialisation error above. The suites of all five affected makers pass.

### DIFF
--- a/src/Maker/Common/CanGenerateTestsTrait.php
+++ b/src/Maker/Common/CanGenerateTestsTrait.php
@@ -24,8 +24,6 @@ use Symfony\Component\Console\Input\InputOption;
  */
 trait CanGenerateTestsTrait
 {
-    private bool $generateTests = false;
-
     public function configureCommandWithTestsOption(Command $command): Command
     {
         $testsHelp = file_get_contents(\dirname(__DIR__, 3).'/config/help/_WithTests.txt');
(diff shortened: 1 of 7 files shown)


GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1816, opened 2026-08-31
[make:security:custom] Fix the crash under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

`make:security:custom` cannot run without a terminal, in either of two ways:

```console
$ php bin/console make:security:custom CustomAuthenticator --no-interaction
No arguments expected for "make:security:custom" command, got "CustomAuthenticator"

$ php bin/console make:security:custom --no-interaction
Typed property Symfony\Bundle\MakerBundle\Maker\Security\MakeCustomAuthenticator::$authenticatorClassName
```

### Cause

The authenticator name is the command's only input. It was asked for in `interact()` and kept on the maker, and `configureCommand()` declared nothing but help text, so there was no way to supply it from the command line. `Command::run()` skips `interact()` under `--no-interaction`, so the property was never assigned and the first read in `generate()` threw.

### Fix

The name travels on the input as an optional `name` argument, marked non-interactive so `MakerCommand`'s generic prompt does not duplicate the maker's own question. `interact()` asks only when the argument is absent and writes the answer there, `generate()` builds the `ClassNameDetails` from it, and the property is gone.

Two guards lived in `interact()` and therefore disappeared without a terminal:

- The `security.yaml` existence check moved into a method both paths call.
- The `symfony/security-bundle` install turned out to be load-bearing rather than convenience, because its recipe is what writes `security.yaml` in the first place. `generate()` performs it when the file is missing, which also keeps it from issuing a second `composer require` after `interact()` already installed the bundle.

### Why

### DIFF
--- a/src/Maker/Security/MakeCustomAuthenticator.php
+++ b/src/Maker/Security/MakeCustomAuthenticator.php
@@ -19,11 +19,10 @@ use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Maker\AbstractMaker;
 use Symfony\Bundle\MakerBundle\Maker\Common\InstallDependencyTrait;
-use Symfony\Bundle\MakerBundle\Util\ClassNameDetails;
 use Symfony\Bundle\MakerBundle\Util\UseStatementGenerator;
 use Symfony\Bundle\MakerBundle\Util\YamlSourceManipulator;
-use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
@@ -47,8 +46,6 @@ final class MakeCustomAuthenticator extends AbstractMaker
 
     private const SECURITY_CONFIG_PATH = 'config/packages/security.yaml';
 
-    private ClassNameDetails $authenticatorClassName;
-
     public function __construct(
         private FileManager $fileManager,
         private Generator $generator,
@@ -68,6 +65,7 @@ final class MakeCustomAuthenticator extends AbstractMaker
     public function configureCommand(Command $command, InputConfiguration $inputConfig): void
     {
         $command
+            ->addArgument('name', InputArgument::OPTIONAL, 'What is the class name of the authenticator (e.g. <fg=yellow>CustomAuthenticator</>)')
             ->setHelp($this->getHelpFileContents('security/MakeCustom.txt'))
         ;
     }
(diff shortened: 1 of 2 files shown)


GromNaN:
Thanks for working on this, the non-interactive fix is welcome.

One thing I would change: `$inputConfig->setArgumentAsNonInteractive('name')` should not be needed here.

`MakerCommand::interact()` already loops over the command arguments before calling `$this->maker->interact()`, and for every empty argument that is not marked as non-interactive it does:

```php
$value = $this->io->ask($argument->getDescription(), $argument->getDefault(), Validator::notBlank(...));
$input->setArgument($argument->getName(), $value);
```

That is exactly what the maker now does by hand, with the same `Validator::notBlank` and a question that only differs in wording. So the default behaviour already covers this case.

The "duplicated question" concern does not apply either: the generic loop runs first and fills the argument, so `$this->maker->interact()` would return early on `if ($input->getArgument('name')) { return; }`. There would only ever be one prompt.

I would suggest dropping both `setArgumentAsNonInteractive('name')` and the manual `$io->ask()`, and moving the question text into the argument description. The makers that do use `setArgumentAsNonInteractive` need a specific prompt (autocompletion, domain validator, choice list), which is not the case here.

The one trade-off is ordering. Today `interact()` installs `symfony/security-bundle` and checks `security.yaml` before asking. With the default behaviour the user types the class name first and gets the "security.yaml does not exist" error afterwards. That is a minor UX change and consistent with the rest of the bundle, so I think it is acceptable. `initialize()` would be the natural place for such a check, but `MakerCommand` is final and `MakerInterface` exposes no hook there, so the private method called from both paths is the right approach in the meantime.

Two smaller points:

1. The guard in `generate()` is indirect:

```php
if (!$this->fileManager->fileExists(self::SECURITY_CONFIG_PATH)) {
    $this->installDependencyIfNeeded(...);
}
```

`installDependencyIfNeeded()` already starts with `if (class_exists($expectedClassToExist)) { return $io; }`, so it is safe to call unconditionally. Calling it once in `generate()` and removing it from `interact()` would let the three-line comment about the second composer call go away.

2. `throw new RuntimeCommandException('The "name" argument is required when the command runs without asking.')` is only reachable in non-interactive mode without the argument. Saying the `name` argument is required in non-interactive mode reads a bit more directly.

Unrelated to this PR but relevant to the series: `MakerCommand::initialize()` prints `"%s" is not meant to be run in non-interactive mode.` as soon as the input is not interactive. If the goal is to make the makers usable without a terminal, that generic warning becomes contradictory and may deserve its own change, probably conditioned on whether the maker can actually run without prompts.

wachterjohannes:
Applied both simplifications, thanks. Kept the `fileExists()` guard on `installDependencyIfNeeded()` in `generate()`: calling it unconditionally broke the interactive test, `class_exists()` misses a package installed earlier in the same process, so it reinstalls and corrupts the container cache, same bug as #1817.

Agreed on the `initialize()` warning too, worth its own PR once more of the series lands.

src/Maker/Security/MakeCustomAuthenticator.php
```diff
@@ -141,4 +147,11 @@ className: $this->authenticatorClassName->getFullName(),
     public function configureDependencies(DependencyBuilder $dependencies): void
     {
     }
+
+    private function checkSecurityConfigExists(): void
+    {
+        if (!$this->fileManager->fileExists(self::SECURITY_CONFIG_PATH)) {
+            throw new RuntimeCommandException(\sprintf('The file "%s" does not exist. PHP & XML configuration formats are currently not supported.', self::SECURITY_CONFIG_PATH));
```
GromNaN, line 154: PHP stands for 2 things: the legacy PHP-DSL and the new PHP-array formats. Both are not supported. Only Yaml is supported by Flex. 
This error message make me glich.

GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1817, opened 2026-08-31
[make:webhook] Fix the crash under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

`make:webhook` already declares its `name` argument, and already marks it non-interactive, so it looks scriptable. It is not:

```console
$ php bin/console make:webhook remote_service --no-interaction
Typed property Symfony\Bundle\MakerBundle\Maker\MakeWebhook::$name must not be accessed
before initialization
```

### Cause

`interact()` read the argument and copied it onto the maker, and `generate()` read it back from there. `Command::run()` skips `interact()` under `--no-interaction`, so the copy never happened. `$io` was stashed on the maker in the same way.

### Fix

`generate()` reads the argument it was given. The name check that `interact()` applied moved next to that read, so a name with illegal characters is rejected on both paths rather than only when someone is watching. `$io` is passed to the one private method that asks anything, and both properties are gone.

`symfony/webhook` was installed from `interact()`, which is the one step that does not always run, and the generated request parser extends `AbstractRequestParser` from that package. The install moved to `generate()`, which always runs, rather than being called from both: a second `composer require` in the same process rebuilds the container underneath the running command, and the run then dies on a stale cache file. The one visible change interactively is that the install now happens after the questions instead of before.

### Why

### DIFF
--- a/src/Maker/MakeWebhook.php
+++ b/src/Maker/MakeWebhook.php
@@ -62,10 +62,7 @@ final class MakeWebhook extends AbstractMaker implements InputAwareMakerInterfac
     public const WEBHOOK_NAME_PATTERN = '/^[a-zA-Z_.\-\x80-\xff][a-zA-Z0-9_.\-\x80-\xff]*$/u';
     private const WEBHOOK_CONFIG_PATH = 'config/packages/webhook.yaml';
 
-    private ConsoleStyle $io;
-
     private YamlSourceManipulator $ysm;
-    private string $name;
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1818, opened 2026-08-31
[make:security:form-login] Fix the crash under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

```console
$ php bin/console make:security:form-login --no-interaction
Typed property Symfony\Bundle\MakerBundle\Maker\Security\MakeFormLogin::$controllerName
```

### Cause

Every answer was kept on the maker, and the command declared no options beyond `--with-tests`. `Command::run()` skips `interact()` under `--no-interaction`, so nothing was ever assigned and `generate()` read six uninitialised properties.

### Fix

Five options carry the answers instead: `--controller-name`, `--firewall-name`, `--user-class`, `--username-field` and `--logout`. `interact()` asks only for what was not passed and writes each answer into its option, `generate()` reads them, and the properties are gone. The `YamlSourceManipulator` was never input, so it is built where it is used.

The interactive dialogue is unchanged: the same questions in the same order with the same defaults, except that passing an option skips its question.

**`--logout` is the one deliberate difference.** The prompt offers yes as its default; the option is opt-in. A flag that has to be spelled out is worth more in a script than matching which answer a prompt happens to preselect. Same reasoning as the nullability decision in #1813.

### The guesses

`--user-class` and `--username-field` are normally not asked at all. `InteractiveSecurityHelper::guessUserClass()` and `guessUserNameField()` return silently when `security.yaml` is unambiguous and prompt otherwise, and the prompting branch cannot be reached without a terminal: those are `choice()` calls with no default, which resolve to `null` and violate the `string` return type.

So the silent halves are extracted as `findUserClass()`, `findUserNameField()` and `findFirewallName()`, returning `null` exactly where a person would have to answer. The `guess*` methods call them first, so their behaviour is unchanged, and `generate()` uses them to fail with a message naming the option to pass rather than reaching for a prompt.

### Why

Part of a series making the makers usable by coding agents, which cannot answer prompts. This is also where the work `make:auth` would have needed goes: that command is deprecated in favour of the `make:security:*` commands, so fixing it there would have been effort spent on a dead entry point.

### DIFF
--- a/src/Maker/Security/MakeFormLogin.php
+++ b/src/Maker/Security/MakeFormLogin.php
@@ -36,6 +36,7 @@ use Symfony\Bundle\SecurityBundle\SecurityBundle;
 use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface;
 use Symfony\Component\Routing\Attribute\Route;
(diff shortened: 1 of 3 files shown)


GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1819, opened 2026-08-31
[make:reset-password] Fix the crash under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

> [!NOTE]
> Built on #1818, which extracts the silent halves of the security guesses this one also needs. That commit appears in the diff until #1818 is merged.

```console
$ php bin/console make:reset-password --no-interaction
Typed property Symfony\Bundle\MakerBundle\Maker\MakeResetPassword::$userClass must not be
accessed before initialization
```

### Cause

Seven answers were kept on the maker, and the command exposed only `--with-uuid` and `--with-tests`. `Command::run()` skips `interact()` under `--no-interaction`, so nothing was assigned and the first read in `generate()` threw.

### Fix

Seven options carry them instead:

| Option | Interactive equivalent |
| --- | --- |
| `--user-class` | `guessUserClass()`, asked only when `security.yaml` is ambiguous |
| `--email-field` | `guessEmailField()`, asked only when the class has no `email` property |
| `--email-getter` | `guessEmailGetter()`, asked only when the obvious getter is missing |
| `--password-setter` | `guessPasswordSetter()`, asked only when `setPassword()` is missing |
| `--success-redirect-route` | asked, default `app_home` |
| `--from-email-address` | asked, validated, no default |
| `--from-email-name` | asked, validated, no default |

`interact()` asks only for what was not passed and writes each answer into its option, `generate()` resolves them, and the properties are gone. The same questions are asked in the same order with the same defaults, except that passing an option skips its question.

### The guesses

Four of the seven are usually not asked. The guess helpers return silently when `security.yaml` and the User class are unambiguous, and otherwise prompt with a `choice()` that has no default, which without a terminal resolves to `null` and violates their `string` return type. Their silent halves are `findUserClass()`, `findEmailField()`, `findEmailGetter()` and `findPasswordSetter()` on `InteractiveSecurityHelper`, returning `null` where a person would have to answer. `generate()` uses them and fails with a message naming the option to pass.

The two email values have nothing to derive from, so they are required without a terminal, reported through the same validators the prompt uses.

### Why

### DIFF
--- a/src/Maker/MakeResetPassword.php
+++ b/src/Maker/MakeResetPassword.php
@@ -40,6 +40,7 @@ use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Bundle\SecurityBundle\SecurityBundle;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\EmailType;
 use Symfony\Component\Form\Extension\Core\Type\PasswordType;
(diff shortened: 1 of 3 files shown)


GromNaN:
Thank you @wachterjohannes.

PR #1820, opened 2026-08-31
[make:registration-form] Fix the crash under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

> [!NOTE]
> Built on #1818 and #1819, which extract the silent halves of the security guesses this one also needs. Those commits appear in the diff until they are merged. This is the largest of the series and the last of it.

```console
$ php bin/console make:registration-form --no-interaction
Typed property Symfony\Bundle\MakerBundle\Maker\MakeRegistrationForm::$userClass must not
be accessed before initialization
```

### Cause

Eight answers were kept on the maker across twelve decision points, and the command exposed only `--with-tests`. `Command::run()` skips `interact()` under `--no-interaction`, so nothing was assigned and the first read in `generate()` threw.

### Fix

Thirteen options carry them: `--user-class`, `--username-field`, `--password-field`, `--unique-entity`, `--verify-email`, `--verify-anonymously`, `--id-getter`, `--email-getter`, `--from-email-address`, `--from-email-name`, `--auto-login`, `--authenticator` and `--redirect-route`.

`interact()` asks only for what was not passed and writes each answer into its option, `generate()` resolves them, and the properties are gone. The same questions are asked in the same order with the same defaults, except that passing an option skips its question.

### Three deliberate differences

`--unique-entity`, `--verify-email` and `--auto-login` are opt-in, where the prompts offer **yes** as their default.

This is worth arguing rather than assuming. Inheriting those defaults would mean a scripted run silently pulls in `symfonycasts/verify-email-bundle` and `symfony/mailer`, adds an `isVerified` property to the User entity, writes a `#[UniqueEntity]` attribute onto it and wires up authentication after registration. None of that is something a caller who passed no flags asked for. A flag that has to be spelled out is worth more here than matching which answer a prompt preselects, the same trade-off taken in #1813.

### The guesses

`guessUserClass()`, `guessUserNameField()`, `guessPasswordField()`, `guessIdGetter()` and `guessEmailGetter()` return silently when the config and the class are unambiguous and prompt otherwise, and the prompting branch cannot be reached without a terminal. `generate()` uses their silent halves on `InteractiveSecurityHelper` and fails with a message naming the option when a person would have to answer. `findPasswordField()` and `findIdGetter()` are added there for this.

The two email values and the redirect route have nothing to derive from and are required when the feature that needs them is switched on.

### Auto-login

`--auto-login` had the same gap. The single-authenticator match lived only in the interactive question, so `generate()` read `--authenticator` but never resolved it on its own. `resolveAuthenticator()` now does what the question did: picks the one configured authenticator when there is exactly one, fails naming `--authenticator` when there is more than one, and prints the same note the prompt shows when there is none.

### Why

### DIFF
--- a/src/Maker/MakeRegistrationForm.php
+++ b/src/Maker/MakeRegistrationForm.php
@@ -45,6 +45,7 @@ use Symfony\Bundle\SecurityBundle\SecurityBundle;
 use Symfony\Bundle\TwigBundle\TwigBundle;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\PasswordType;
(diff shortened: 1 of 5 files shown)


GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1821, opened 2026-09-07
[make:subscriber] Remove the deprecated maker

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no (removes an existing one)
| Issues        | -
| License       | MIT

Following the scope defined in #1800: makers that duplicate another maker are out of scope.

`make:subscriber` has been deprecated since 1.51 in favor of `make:listener`, which already generates event subscribers. This removes the maker, its service, its help file and its tests, along with the alias handling in `MakeListener` and `MakeCommandRegistrationPass` — both explicitly annotated `@deprecated remove ... when removing make:subscriber`.

`make:listener` keeps generating subscribers: template, fixtures and test cases are untouched.


GromNaN:
Removing a deprecated maker should not be done in a minor release. I've tagged this PR to the milestone 2.0.0, but we have no plan to create work on this major release yet.

PR #1822, opened 2026-09-07
[make:twig-extension] Generate attribute-based Twig extensions

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Since Twig 3.21 and TwigBundle 7.3, a filter or function is declared with the `#[AsTwigFilter]` / `#[AsTwigFunction]` attribute on any service method, and such extensions are lazy-loaded out of the box — which is what the separate runtime class used to buy us. This is now the approach recommended by the docs.

`make:twig-extension` therefore generates a single `src/Twig/<Name>Extension.php` class carrying both attributes, instead of an `AbstractExtension` subclass plus its `RuntimeExtensionInterface` runtime. It also switches to `ClassData`, so the generated class follows `generate_final_classes` like the other makers.

`Twig\Attribute\AsTwigFilter` becomes the declared class dependency, which doubles as the version check: projects on an older Twig are told to upgrade rather than getting code that silently does nothing.

Only the generated code changes (README, point B).


GromNaN (approved):
Thanks. The next step will be to specify the filters and function methods interactively (or using command options).

GromNaN:
Thank you @seb-jean.

PR #1824, opened 2026-09-07
[docs] Add AGENTS.md with guidelines for non-interactive makers

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

Requested by @GromNaN on #1816: write down what the non-interactive maker series learned,
so the next contributor doesn't have to rediscover it PR by PR.

`AGENTS.md` covers the property-vs-input rule, the silent half of a guesser, why errors
must be `RuntimeCommandException`, why booleans need `VALUE_NONE` options, when
`setArgumentAsNonInteractive()` is actually needed, the double `composer require` pitfall,
and the testing convention this series used. `CONTRIBUTING.md` gets one line pointing at
it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,73 @@
+# Agents.md
+
+Every maker must also work with `--no-interaction`. `Command::run()` skips `interact()`
+entirely in that mode, so anything only set there is missing in `generate()`.
+
+## The one rule
+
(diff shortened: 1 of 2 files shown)


GromNaN:
Thank you @wachterjohannes.

PR #1825, opened 2026-09-10
[make:command] Ask for the arguments and the options

### LINKED ISSUE #1699: Upgrade `make:command` to create an invokable command, and ask arguments and options
The Invokable command feature is landing in Symfony 7.3, this is improving the DX with the resolution of the callable parameters. For new applications, the maker should generate the command class using this new syntax.

Currently, [`MakeCommand`](https://github.com/symfony/maker-bundle/blob/1.x/src/Maker/MakeCommand.php) generates the command class from its name only. We can make the command more interactive and ask for the parameter and options, to generate the full command definition: the signature of the `__invoke` method with `#[Argument]` and `#[Option]` attributes).

@tacman wrote an [implementation](https://github.com/survos/SurvosCodeBundle/blob/main/src/Command/CodeCommand.php) that is presented in this article: [Generating Symfony 7.3 Console Commands](https://medium.com/@tacman1123/generating-symfony-7-3-console-commands-a2b4da5d670d)

Amoifr:
Closes #1699.

The first half of that issue landed with #1746: `make:command` generates an `__invoke()` method. This is the second half, asking for the arguments and the options so the signature comes out ready rather than as a sample to edit.

```
$ php bin/console make:command app:send-report

 New argument name (press <return> to stop adding arguments):
 > recipient
 Type [string]:
 Description [Recipient]: Who to send the report to
 Is it required? (yes/no) [no]: yes

 New argument name (press <return> to stop adding arguments):
 >

 New option name (press <return> to stop adding options):
 > dry-run
 Type [bool]:
 Description [Dry Run]: Do not send anything
```

```php
public function __invoke(
    SymfonyStyle $io,
    #[Argument('Who to send the report to')] string $recipient,
    #[Option('Do not send anything')] bool $dryRun = false,
): int {
```

The name is typed the way it reads on the command line, so `dry-run` becomes `$dryRun`. Besides the scalar types and `array`, the type can be a backed enum.

They can also be passed on the command line (thanks @GromNaN). The type defaults to `string` for an argument and `bool` for an option, the short name of an enum is enough when it is unique within `src/`, a `?` after the name makes an argument optional as in a PHP array shape, and passing any of them skips the questions:

```
$ php bin/console make:command app:send-report --argument=recipient --argument=cc?:array --option=dry-run --option=format:Format
```

A few decisions worth flagging:

* **Answering nothing keeps the sample argument and option**, so `--no-interaction`, where `interact()` never runs, and the person who just presses return through the questions still get a command that shows both attributes at work.
* **What the console component would refuse is refused up front.** A required argument after an optional one, any argument after an array one, a parameter named `io` (taken by `SymfonyStyle $io`), an argument and an option sharing a name, and an enum that is not backed all produce a command that fails to load. The questions skip or ask again, and the command line gets an error before anything is written.
* **An enum named like an import is aliased**: an `App\Enum\Option` comes in as `EnumOption`, since `Option` is the attribute.
* **The notes use `var_export()`**, so every type, enums included, is shown the same way.

I saw the run of `--no-interaction` fixes that just landed (#1816, #1817, #1818), and this maker does collect state in `interact()` and read it in `generate()`, which is the shape those removed. The properties are initialised, so there is nothing to crash on, but `it_makes_a_command_non_interactively` pins that down rather than leaving it to reading.

#### Testing

Seven functional tests, up from three: the questions, the command line (enums included), and the definitions it rejects. `MakeCommandTest` green, PHPStan green with the CI tooling.

### DIFF
--- a/src/Maker/MakeCommand.php
+++ b/src/Maker/MakeCommand.php
@@ -13,17 +13,23 @@ namespace Symfony\Bundle\MakerBundle\Maker;
 
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
 use Symfony\Bundle\MakerBundle\Str;
+use Symfony\Bundle\MakerBundle\Util\EnumHelper;
 use Symfony\Bundle\MakerBundle\Util\PhpCompatUtil;
 use Symfony\Bundle\MakerBundle\Util\UseStatementGenerator;
+use Symfony\Bundle\MakerBundle\Validator;
 use Symfony\Component\Console\Attribute\Argument;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Attribute\Option;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
@@ -32,8 +38,30 @@ use Symfony\Component\Console\Style\SymfonyStyle;
  */
 final class MakeCommand extends AbstractMaker
 {
-    public function __construct(?PhpCompatUtil $phpCompatUtil = null)
-    {
+    private const PARAMETER_TYPES = ['string', 'bool', 'int', 'float', 'array'];
+
+    private const EMPTY_VALUES = [
+        'string' => "''",
+        'bool' => 'false',
+        'int' => '0',
+        'float' => '0.0',
+        'array' => '[]',
+    ];
+
+    /**
+     * @var list<array{name: string, type: string, description: string, required: bool}>
+     */
+    private array $arguments = [];
+
+    /**
+     * @var list<array{name: string, type: string, description: string, required: bool}>
+     */
+    private array $options = [];
+
+    public function __construct(
+        ?PhpCompatUtil $phpCompatUtil = null,
+        private ?FileManager $fileManager = null,
+    ) {
         if (null !== $phpCompatUtil) {
             @trigger_deprecation(
                 'symfony/maker-bundle',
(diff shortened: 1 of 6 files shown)


GromNaN (commented):
Could you add options to `make:command` to add options non-interactively.

```php
bin/console make:command 
    --argument 'name:string'
    --option 'dry-run:bool'
```

It should use the same syntax for the `name:type` as the `make:entity`
https://github.com/symfony/maker-bundle/blob/fba278064d5b9a8deb8d6356339b0617d2e0e30e/src/Maker/MakeEntity.php#L101

Amoifr:
Thanks for the nudge, good call! Pushed in dd635b78: `--argument` and `--option`, repeatable, with the `name[:type]` syntax of `--field`. The type defaults to `string` for an argument and `bool` for an option, a trailing `?` makes an argument optional (the way it makes a field nullable), and passing any of them skips the questions:

```
bin/console make:command app:send-report --argument=recipient --argument=cc:array? --option=dry-run --option=retries:int
```

While at it, I closed three holes the interactive mode had: an argument after an array one, a parameter named `io`, and an argument and an option sharing a name. Each now gets a clear error instead of a generated command that fails to load.

The 3 red "highest deps" jobs are unrelated: `TranslationBundle` not found in framework-bundle, same as last night's 1.x run.

tests/Maker/MakeCommandTest.php
```diff
@@ -63,11 +71,113 @@ public static function getTestDetails(): \Generator
...
+                self::assertStringContainsString("#[Option('Do not send anything')] bool \$dryRun = false,", $commandFileContents);
+                self::assertSame(1, substr_count($commandFileContents, '$times = '));
+                self::assertStringNotContainsString('$arg', $commandFileContents);
+            }),
+        ];
+
+        yield 'it_makes_a_command_from_the_command_line' => [self::buildMakerTest()
+            ->run(static function (MakerTestRunner $runner) {
+                // no answer is given: passing any of them on the command line skips the questions
+                $runner->runMaker([], 'app:foo --argument=recipient --argument=times:int? --argument=tags:array? --option=dry-run --option=retries:int');
```
GromNaN: I would move the question mark after the option/argument name. Because the type is optional and this mimic the PHP array-shape syntax for optional properties.

```suggestion
                $runner->runMaker([], 'app:foo --argument=recipient --argument=times?:int --argument=tags?:array --option=dry-run --option=retries:int');
```

src/Maker/MakeCommand.php
```diff
@@ -73,20 +110,48 @@ public function generate(InputInterface $input, ConsoleStyle $io, Generator $gen
...
+        $optionDefinitions = $input->getOption('option');
+
+        if ($argumentDefinitions || $optionDefinitions) {
+            [$arguments, $options] = self::parseDefinitions($argumentDefinitions, $optionDefinitions);
+        } else {
+            $arguments = $this->arguments;
+            $options = $this->options;
+        }
+
+        if ([] === $arguments && [] === $options) {
```
GromNaN: We use falsy check for bool/object/array.

```suggestion
        if (!$arguments && !$options) {
```

templates/command/Command.tpl.php
```diff
@@ -12,11 +12,13 @@ class <?= $class_name; ?>
...
     public function __invoke(
         SymfonyStyle $io,
-        #[Argument('Argument description')] string $arg = '',
-        #[Option('Option description')] bool $enable = false,
+<?php foreach ($command_parameters as $parameter) { ?>
+        <?= $parameter; ?>,
+<?php } ?>
     ): int {
-        $io->note(sprintf('The value of $arg is: %s', $arg));
-        $io->note(sprintf('The value of $enable is: %s', $enable ? 'true' : 'false'));
```
GromNaN, line 19: This is lost. Could you restore a dynamic list of input values? Maybe use `var_export` to show the value.

GromNaN:
You can also add support for enum (by enum name) in input types.

tests/Maker/MakeCommandTest.php
Amoifr: Good point, done: `name?:type`, like an optional key of an array shape. The old trailing form now fails with an "Invalid type" error instead of meaning something else.

src/Maker/MakeCommand.php
Amoifr: Done, here and on the two use-statement checks just below.

templates/command/Command.tpl.php
Amoifr: They were still there, generated one per parameter by `buildNote()`, which the template diff hides well. But `var_export()` is nicer: it replaces the per-type formatting and covers enums for free, so it's in.

Amoifr:
> You can also add support for enum (by enum name) in input types.

Done! The type can be the short name of a backed enum from `src/` (or its FQCN), in both modes: `--option=format:Format` gives `#[Option('Format')] ?Format $format = null`. An enum named like an import (`Option`, anyone?) is aliased, and a non-backed one gets an error, since the console needs the backing value.

I implemented it in the only form that made sense to me, but I can't quite picture a case where one would need it. Could you give me an example, just out of curiosity?

Thanks for all the ideas, this maker is much better for them. Though at this pace, I'm starting to fear the next review will ask for my credit card number and a kidney for a transplant. 😄

GromNaN (approved):

GromNaN:
Thank you @Amoifr.

PR #1826, opened 2026-09-11
[make:schedule] Fix the crash under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

`make:schedule --no-interaction` crashes:

```
In MakeSchedule.php line 113:

  [Error]
  Typed property Symfony\Bundle\MakerBundle\Maker\MakeSchedule::$scheduleName
```

### Cause

`Command::run()` skips `interact()` entirely under `--no-interaction`.
`$scheduleName`, `$message` and `$transportName` were only ever set there,
and `generate()` reads them unconditionally.

### Fix

Same pattern as the rest of this series: the three answers move onto the
input, as `--schedule-name`, `--message` and `--transport-name`. `interact()`
asks only when an option is missing, same questions, same order, same
defaults. `generate()` reads the options and no longer touches a property.

A `--message` value is checked against the files in `src/Message` and
rejected by name when it does not match one, the same guarantee the
interactive choice list gives.

The `composer require symfony/scheduler` call moves from `interact()` to
`generate()`, since `interact()` never runs under `--no-interaction` and the
package still needs to be there.

### Tests

Added non-interactive cases: the default empty schedule, `--transport-name`,
an existing `--message`, and a rejection case for an unknown `--message`
name. The existing interactive cases are unchanged and still pass.

### Why

Part of the series making the makers usable by coding agents, which cannot
answer prompts. Last remaining crash in the maker list.

### DIFF
--- a/src/Maker/MakeSchedule.php
+++ b/src/Maker/MakeSchedule.php
@@ -13,6 +13,7 @@ namespace Symfony\Bundle\MakerBundle\Maker;
 
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\FileManager;
 use Symfony\Bundle\MakerBundle\Generator;
 use Symfony\Bundle\MakerBundle\InputConfiguration;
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1827, opened 2026-09-11
[Security] Recognize "custom_authenticators" in InteractiveSecurityHelper

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

Found while testing the `--no-interaction` fixes from #1810–#1826 for real: a
custom authenticator generated by `make:security:custom` is invisible to
`InteractiveSecurityHelper`, so `--auto-login` and `--logout` never notice it
exists.

```php
$helper->getAuthenticatorsFromConfig([
    'main' => ['custom_authenticators' => ['App\Security\CustomAuthenticator']],
]);
// => [] (should contain one Authenticator)
```

### Cause

`make:security:custom` writes `custom_authenticators` (plural) to
`security.yaml`. `AuthenticatorType::CUSTOM` only maps the singular
`custom_authenticator`, the key the deprecated `make:auth` writes. The one
existing test for this path used the singular key, so it passed without ever
exercising the shape `make:security:custom` actually produces.

### Fix

Recognize both keys when scanning a firewall's config. Kept the singular
one, since projects and code generated by `make:auth` still use it.

### Why

Part of the series making the makers usable by coding agents (#1810–#1826).
This one is not about `--no-interaction`, but it was hiding behind the same
testing gap: a fixture that matched the code instead of what the maker
actually produces.

### DIFF
--- a/src/Security/InteractiveSecurityHelper.php
+++ b/src/Security/InteractiveSecurityHelper.php
@@ -289,7 +289,7 @@ final class InteractiveSecurityHelper
      *      pattern: ^/path
      *      form_login:
      *          login_path: app_login
-     *      custom_authenticator:
+     *      custom_authenticators:
      *          - App\Security\MyAuthenticator
      *
      * @param array<string, mixed> $firewallConfig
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1828, opened 2026-09-11
[make:message] Fix the crash under --no-interaction

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

`make:message --no-interaction` crashes:

```
In Input.php line 91:

  [Symfony\Component\Console\Exception\InvalidArgumentException]
  The "chosen-transport" argument does not exist.
```

### Cause

`interact()` registered the `chosen-transport` argument itself,
`$command->addArgument('chosen-transport', ...)`, instead of declaring it in
`configureCommand()`. `Command::run()` skips `interact()` entirely under
`--no-interaction`, so the argument was never registered, and
`generate()`'s `$input->getArgument('chosen-transport')` failed.

### Fix

Same pattern as the rest of this series: a `--transport` option, declared in
`configureCommand()`, replaces the dynamically added argument. `interact()`
asks the same question, same order, same default, only when the option is
missing. `generate()` reads the option and rejects an unconfigured transport
name, naming the transports that do exist in `messenger.yaml`.

### Why

Part of the series making the makers usable by coding agents (#1810–#1826),
found while testing that series for real.

### DIFF
--- a/src/Maker/MakeMessage.php
+++ b/src/Maker/MakeMessage.php
@@ -13,6 +13,7 @@ namespace Symfony\Bundle\MakerBundle\Maker;
 
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
 use Symfony\Bundle\MakerBundle\FileManager;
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

GromNaN:
Thank you @wachterjohannes.

PR #1831, opened 2026-09-12
[make:crud] Fix crash when the entity class name is left blank

GromNaN:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

Running `make:crud` interactively and pressing Enter without typing an entity class name crashed with a TypeError:

    Symfony\Bundle\MakerBundle\Maker\MakeCrud::getDefaultControllerClass(): Argument #1 ($entityClass) must be of type string, null given

### Cause

The interactive question for the entity class name returned `null` when left blank, and `MakeCrud::interact()` passed that value to `getDefaultControllerClass()`, whose parameter is typed `string`.

### Fix

The question now validates the answer with `Validator::entityExists()`, the pattern already used in `MakeForm`. A blank or unknown entity class name is rejected with a clear message and the question is asked again until a valid entity is provided.

### DIFF
--- a/src/Maker/MakeCrud.php
+++ b/src/Maker/MakeCrud.php
@@ -86,11 +86,10 @@ final class MakeCrud extends AbstractMaker
             $entities = $this->doctrineHelper->getEntitiesForAutocomplete();
 
             $question = new Question($argument->getDescription());
+            $question->setValidator(static fn ($answer) => Validator::entityExists($answer, $entities));
             $question->setAutocompleterValues($entities);
 
-            $value = $io->askQuestion($question);
-
-            $input->setArgument('entity-class', $value);
+            $input->setArgument('entity-class', $io->askQuestion($question));
         }
 
         if (null === $input->getOption('controller-class')) {
--- a/tests/Maker/MakeCrudTest.php
+++ b/tests/Maker/MakeCrudTest.php
@@ -45,6 +45,28 @@ class MakeCrudTest extends MakerTestCase
             }),
         ];
 
+        yield 'it_asks_again_when_entity_class_is_blank' => [self::buildMakerTest()
+            ->run(static function (MakerTestRunner $runner) {
+                $runner->copy(
+                    'make-crud/SweetFood.php',
+                    'src/Entity/SweetFood.php'
+                );
+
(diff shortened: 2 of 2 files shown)


GromNaN:
Following the crash fixed in this PR, I audited every interactive question in `src/Maker` and `src/Security` for the same pattern: a blank answer returning `null`, consumed inside `interact()` (a `string`-typed helper, `Str::asClassName`, `sprintf`...) before `generate()` can validate it.

**Result: `MakeCrud` was the only actual crash.** Every other question is guarded by a validator, a default value, or an explicit blank check:

| Maker | Guard | Blank answer behaviour |
| --- | --- | --- |
| `MakeForm` bound class | `Validator::existsOrNull` | `null` accepted, unbound form |
| `MakeEntity` entity / relation / enum | `notBlank`, `classIsBackedEnum` | question is asked again |
| `MakeListener`, `MakeSubscriber`, `MakeWebhook` name | `notBlank` | question is asked again |
| `MakeAuthenticator` class | `notBlank` + `classDoesNotExist` | question is asked again |
| `MakeDecorator` id / class | `serviceExists`, `validateClassName` | question is asked again |
| `MakeCommand` parameters | `null === $name` guard | stops adding, intended |
| `MakeStimulusController` targets/values/classes | `?: null` / `if (!$valueName)` | stops adding, intended |
| `MakeEntity` fields | validator allows empty | stops adding, intended |
| `MakeWebhook` matchers | `ChoiceQuestion`, default `<skip>` | skips, intended |
| `MakeEntity`, `MakeStimulusController` field type | default + manual `in_array` loop | default is used |
| `MakeTest`, `MakeUser`, `MakeFormLogin`, `MakeResetPassword`, `MakeRegistrationForm` | default and/or `notBlank`, `validateClassName`, `classIsUserInterface` | default or asked again |

The one question left without a guard is the transport name in `MakeSchedule`. I checked the Symfony `AsSchedule` attribute: its `name` argument is optional and defaults to `'default'`, so `#[AsSchedule]` and `#[AsSchedule('default')]` are strictly equivalent. A nullable transport name is therefore legitimate, and the current `make:schedule` behaviour is correct and left unchanged.

Issue #1830, opened 2026-09-12
`make:crud` failing to generate tests when PHPUnit is not installed in the project.


GromNaN:
Testing maker-bundle 1.x-dev (before 1.68.0 release)

1. Create a `Client` entity

```
bin/console make:entity Client --field 'name:string'
```

2. Generate the crud
```
$ bin/console make:crud

 The class name of the entity to create CRUD (e.g. AgreeableKangaroo):
 > Client

 Choose a name for your controller class (e.g. ClientController) [ClientController]:
 >

 Do you want to generate PHPUnit tests? [Experimental] (yes/no) [no]:
 > yes

[critical] Error thrown while running command "make:crud". Message: "Class "PHPUnit\Framework\TestCase" not found"
[critical] Uncaught Error: Class "PHPUnit\Framework\TestCase" not found

In KernelTestCase.php line 28:

  Attempted to load class "TestCase" from namespace "PHPUnit\Framework".
  Did you forget a "use" statement for another namespace?


make:crud [--controller-class CONTROLLER-CLASS] [--with-tests] [--] [<entity-class>]
```

The exception should not be thrown because PHPUnit don't need to be installed. The `PHPUnit\Framework\TestCase` is never available in a Symfony application, even if `symfony/phpunit-bridge` or `phpunit.phar` are installed.

Maxcastel:
#1718

PR #1719, opened 2025-07-14
[make:controller] [make:crud] [make:registration-form] [make:reset-password] [make:security:form-login] Fix error while generating PHPUnit tests

Maxcastel:
Fixes: #1718 

Change 

```php
if (!class_exists(WebTestCase::class)) {
    $io->caution('You\'ll need to install the `symfony/test-pack` to execute the tests for your new controller.');
}
```

to

```php
if (!class_exists(TestCase::class)) {
    $io->caution('You\'ll need to install the `symfony/test-pack` to execute the tests for your new controller.');
}
```
- `Symfony\Bundle\FrameworkBundle\Test\WebTestCase` to `PHPUnit\Framework\TestCase`

Error is now displayed and files are created

generate-phpunit-tests-make-controller-ok2

generate-phpunit-tests-make-crud-ok2

generate-phpunit-tests-make-registration-form-ok1

generate-phpunit-tests-make-reset-password-ok2

generate-phpunit-tests-make-security-form-login-ok2

PR #1833, opened 2026-09-12
[make:*] Fix crash when PHPUnit is not installed

GromNaN:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Fixed tickets | #1830
| License       | MIT

Running `make:crud` with `--with-tests` in a project without PHPUnit installed crashed with:

    Class "PHPUnit\Framework\TestCase" not found

### Cause

`MakeCrud`, `MakeController`, `MakeRegistrationForm` and `MakeFormLogin` checked the availability of the test runner with `class_exists(WebTestCase::class)`. `WebTestCase` belongs to `symfony/framework-bundle` and is always autoloadable, but loading it requires its parent class `PHPUnit\Framework\TestCase`, which is absent when PHPUnit is not installed. The `class_exists` call then throws.

### Fix

The check now targets `PHPUnit\Framework\TestCase`, which returns `false` cleanly when PHPUnit is absent. The makers no longer crash and simply display the caution telling the user to install `symfony/test-pack` to run the generated tests.

### DIFF
--- a/src/Maker/MakeController.php
+++ b/src/Maker/MakeController.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MakerBundle\Maker;
 
+use PHPUnit\Framework\TestCase;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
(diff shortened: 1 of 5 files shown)


Maxcastel:
#1719

PR #1835, opened 2026-09-15
[make:*] Offer to install missing packages with Composer

### LINKED ISSUE #1832: Offer to run `composer install ...` in case of missing dependency
When a package is missing, the maker-bundle returns an error with the command to install the package with composer.
This is frustrating.

Instead, ask interactively to run the composer command.

Before:
```
bin/console make:auth


 [ERROR] Missing package: to use the make:auth command, run:

         composer require security

```

Expected:
```
bin/console make:auth


The make:auth command requires the following packages:
- security

Do you want to install this packages with composer (Y/n)?
>


```

If Yes: run the composer command and restart the maker command.
If No or non-interactive: same behavior as before = error with the composer command to run.

Amoifr:
When a maker is missing packages, interactive runs now list them and ask whether to install them with Composer, instead of failing with the command to run. On "yes" the packages are required (dev ones with `--dev`) and the maker is restarted in a new process, since the running kernel was built without them and `composer require` corrupts its container cache. On "no", and with `--no-interaction`, the existing error is unchanged.

The restart needs a TTY. Without one (e.g. on Windows), the command stops after the installation and asks to run the maker again.

Only the dependencies declared in `configureDependencies()` are covered. Makers that check packages later in `interact()` (e.g. `make:auth`, `make:user`) keep their current message, which could be a follow-up.

Tested in a `symfony/skeleton` 8.1 app: `make:form` installs form, validator and orm, runs the recipes, restarts and generates the form; declining keeps the error.

Fix #1832


src/Command/MakerCommand.php
```diff
@@ -128,4 +147,53 @@ public function setCheckDependencies(bool $checkDeps): void
...
+    }
+
+    private function installPackages(DependencyBuilder $dependencies, OutputInterface $output): void
+    {
+        foreach ([[$dependencies->getMissingDependencies(), []], [$dependencies->getMissingDevDependencies(), ['--dev']]] as [$packages, $flags]) {
+            if (!$packages) {
+                continue;
+            }
+
+            $process = new Process(['composer', 'require', ...$flags, ...$packages]);
```
GromNaN: We have to check `composer` is available as a global process. It can be in `./composer.phar` or `./composer`.

The [ExecutableFinder](https://symfony.com/doc/current/components/process.html#finding-an-executable) might be used before offering to install dependencies.

src/Command/MakerCommand.php
```diff
@@ -128,4 +147,53 @@ public function setCheckDependencies(bool $checkDeps): void
...
+    {
+        $argv = $_SERVER['argv'] ?? [];
+
+        if (!$argv || !Process::isTtySupported()) {
+            $this->io->success(\sprintf('The packages are installed, run the %s command again.', $this->getName()));
+
+            return 0;
+        }
+
+        $process = new Process([\PHP_BINARY, ...$argv]);
```
GromNaN: We have a class dedicated to finding the PHP executable. Would it be more reliable? https://symfony.com/doc/current/components/process.html#finding-the-executable-php-binary

src/Command/MakerCommand.php
```diff
@@ -128,4 +147,53 @@ public function setCheckDependencies(bool $checkDeps): void
...
+            }
+        }
+    }
+
+    private function restart(): int
+    {
+        $argv = $_SERVER['argv'] ?? [];
+
+        if (!$argv || !Process::isTtySupported()) {
+            $this->io->success(\sprintf('The packages are installed, run the %s command again.', $this->getName()));
```
GromNaN:
```suggestion
            $this->io->success(\sprintf('The packages are installed, run the "%s" command again.', $this->getName()));
```

GromNaN (commented):
Great work on this feature; thanks!

src/Command/MakerCommand.php
Amoifr: Done. `findComposer()` looks on the path first, then for a `composer.phar` or a `composer` at the root of the project, and when it finds neither the offer is not made at all: the command throws the old message instead, since proposing an install with nothing to run it would be worse than saying nothing.

A phar is not executable on its own, so that one runs through the PHP binary: `[$php, '/path/composer.phar', 'require', ...]`.

Checked the three paths: composer on the path gives `['/bin/composer']`, an empty path with a phar at the root gives `['/usr/bin/php8.5', '.../composer.phar']`, and neither gives `null`.

No test for it, though. Making the lookup testable means injecting the finder, which is more API than this needs, so I would rather ask than decide: tell me if you want it and I will add the seam.

src/Command/MakerCommand.php
Amoifr: Yes, and in two places: the restart uses it now, with `PHP_BINARY` kept as the fallback, and it also provides the binary that runs the phar when Composer is only there.

src/Command/MakerCommand.php
Amoifr: Applied, thanks.

PR #1836, opened 2026-09-17
Fix YamlSourceManipulator losing track of position on single-key sequence items

dbu:
### Problem

`YamlSourceManipulator` throws `YamlManipulationFailedException: Cannot find the original value "..."` when adding a new entry to a YAML file that already contains **two or more** sequence items written as a one-key hash directly on the dash line — for example, a Symfony service tag with attributes:

    services:
        app.grid.foo:
            tags:
                - some.tag.name:
                        attribute: value

### Root cause

`guessNextArrayTypeAndAdvance()` scans forward, character by character, to decide whether the next array is written inline (`{`/`[`) or multi-line. For a sequence item that is a one-key hash written on the dash line, there is nothing that consumes the key text beforehand (a sequence index is an integer and has no literal text to advance past), so the scan runs straight through the key's own name before it reaches the line break that signals "multi-line".

Once the key's text is behind the current position, later logic that needs to relocate that same key (`advanceBeyondEndOfPreviousKey()` / `advanceBeyondKey()`, both forward text searches) can match a **different, later** occurrence of the same key name elsewhere in the file instead of its own — silently corrupting position tracking. This eventually surfaces several steps later as the (rather misleading) `Cannot find the original value "..."` exception.

### Fix

Detect this situation in `guessNextArrayTypeAndAdvance()`: hitting a real character that isn't whitespace, `-`, `{`, or `[` before a line break means we're still looking at an unconsumed key, not the start of a value. Rewind to the starting position and report `MULTILINE`, letting the normal per-key logic find and consume that key itself from a clean position.

### Testing

- Added `sequence_item_single_key_hash_add_second.test`, reproducing the exact failure (two pre-existing services sharing a tag name, adding a third). Tests fail without the fix, pass with this fix.


dbu:
cs failure is not related to my changes. should i fix it?

GromNaN (approved):
Rebased onto 1.x and cleaned up:

- Simplified the inline comment in `guessNextArrayTypeAndAdvance()` (the original mixed tabs and spaces, breaking indentation) into a short comment.
- Normalized the YAML indentation in the fixture from a 6-space to a 4-space step, consistent with the rest of the file.
- Verified the new test fails without the fix (`Cannot find the original value "app_foo"`) and passes with it.

GromNaN:
Thank you @dbu.

